PR #9, opened 2016-11-22
DHE-284 skip validation for archive/unarchive

sdonk:


datahub/core/mixins.py
```diff
@@ -19,16 +19,18 @@ def __init__(self, *args, **kwargs):
         self.korben_connector = KorbenConnector(table_name=self._meta.db_table)
         super(DeferredSaveModelMixin, self).__init__(*args, **kwargs)
 
-    def save(self, as_korben=False, *args, **kwargs):
```
sdonk: remove *args

PR #14, opened 2016-11-29
Remove username column, but keep full compatibity with DJ Admin

canni:


datahub/company/models.py
```diff
@@ -275,27 +277,33 @@ def clean(self):
         super(Contact, self).clean()
 
 
-class Advisor(DeferredSaveModelMixin, AbstractUser):
```
sdonk: Should the AbstractBaseUser be the last class in the inheritance? Usually mixins go first.

datahub/company/models.py
```diff
@@ -275,27 +277,33 @@ def clean(self):
...
-    username = models.CharField(
-        max_length=MAX_LENGTH,
-        unique=True,
-        validators=[AbstractUser.username_validator],
-        blank=True,
-    )
     first_name = models.CharField(max_length=MAX_LENGTH, blank=True)
     last_name = models.CharField(max_length=MAX_LENGTH, blank=True)
-    email = models.EmailField()
+    email = models.EmailField(max_length=MAX_LENGTH, unique=True)
```
sdonk: I think we have advisor with no email address, we use the unknown one to fill the gap. @bmcorser can you confirm?

PR #23, opened 2016-12-13
Catch KeyError as well

sdonk:
Just an extra safety layer, because of issue 325 in Sentry


uktitools:

PR #24, opened 2016-12-13
DHE-370 Pingdom endpoint

sdonk:


uktitools:

datahub/korben/connector.py
```diff
@@ -16,10 +16,9 @@ class KorbenConnector:
         'Accept': 'application/json',
     }
 
-    def __init__(self, table_name):
```
canni: This change is going to brake SSO, update the `CDMSUserBackend` also ;)

sdonk: Done. Not sure why that wasn't caught in a test

PR #26, opened 2016-12-14
DHE-230 Whitelist support for adviosrs via env

canni:
Also merge two user backends almost identical in functionality


uktitools:

canni:
Testing that would involve more mocks than tests code :D not worth it

PR #33, opened 2017-01-10
Do not authenticate if the response is not 200

sdonk:


uktitools:

PR #41, opened 2017-01-17
run flower in standalone mode to avoid error exit code on normal exit

nusnewob:


uktitools:

PR #42, opened 2017-01-17
add default Github OAuth backend

nusnewob:


uktitools:

PR #46, opened 2017-01-17
Change timezone and admin for task info

sdonk:


uktitools:

PR #87, opened 2017-01-24
separate worker for db migrate

nusnewob:


uktitools:

canni:
Not needed anymore...

Singletoned:
Why isn't this needed anymore?  We want the migrations to be run on a separate worker.

canni:
It'll be useful if we could leverage it on `flynn` somehow, otherwise it'll just hang as unused `Procfile` entry for no apparent reason

PR #88, opened 2017-01-25
Initial proposal for concurrent migrations problem on flynn

canni:


uktitools:

sdonk:
Tests? If it's even possible to test this thing

canni:
@sdonk Deployment to `flynn-staging` is a test for that ;)

sdonk:
@canni can you exclude that python file from coverage then? So the % doesn't get affected.

Singletoned:
If it's not testable then it's not a good solution, particularly as (IIUC) the migrations not happening at the same time is reasonably important.  That it works once or twice on staging doesn't really tell us anything, particularly as (IIUC) the migrations sometimes pass anyway.

Singletoned:
Also, we will at the minimum need a regression test to tell us whether it is still working in 6 months time.

PR #104, opened 2017-02-01
DHE-643 service deliveries model

sdonk:


uktitools:

canni:
@sdonk you need to merge master here, new migration from #107

PR #115, opened 2017-02-08
Fix loading up companies with undef advisor in interacions

canni:


uktitools:

PR #157, opened 2017-03-23
DHE-827 service deliveries repository

canni:


sdonk:
Add more tests to cover the parser class

datahub/v2/exceptions.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework.exceptions import ValidationError
+
+
+class RepoDataValidation(ValidationError):
```
canni: Add Error to the class name

datahub/v2/repos/service_deliveries.py
```diff
@@ -0,0 +1,112 @@
...
+from . import utils
+
+
+class ServiceDeliveryDatabaseRepo:
+    """DB repo."""
+
+    def __init__(self, config=None):
+        """Initialise the repo using the config."""
+        self.model_class = ServiceDelivery
+        self.schema_class = ServiceDeliverySchema
```
canni: Shouldn't those be just a class variables instead?

datahub/v2/repos/service_deliveries.py
```diff
@@ -0,0 +1,112 @@
...
+        self.config = config
+        self.url_builder = config['url_builder']
```
canni: Add what the urlbuilder is in docs

datahub/v2/repos/service_deliveries.py
```diff
@@ -0,0 +1,112 @@
...
+            return utils.remove_null(data)
+        except colander.Invalid as e:
+            raise RepoDataValidation(
+                detail=e.asdict()
+            )
+
+    def get(self, object_id):
+        """Get and return a single object by its id."""
+        entity = self.model_class.objects.get(id=object_id)
+        data = utils.model_to_json_api_data(entity, self.schema_class(), url_builder=self.url_builder)
```
canni: I'm thinking that passing URL builder should be done passed when needed, rather than stored in repo config (I assume that this method is called from a view)

datahub/v2/repos/service_deliveries.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def filter(self, company_id=utils.DEFAULT, contact_id=utils.DEFAULT, offset=0, limit=100):
+        """Filter objects."""
+        filters = {}
+        if company_id != utils.DEFAULT:
+            filters['company__pk'] = company_id
+        if contact_id != utils.DEFAULT:
+            filters['contact__pk'] = contact_id
+        start, end = offset, offset + limit
+        entities = list(self.model_class.objects.filter(**filters).all()[start:end])
```
canni: A lot is happening in that line, split it into `queryset` variable and then do the splicing

datahub/v2/repos/utils.py
```diff
@@ -0,0 +1,182 @@
...
+DEFAULT = object()
+
+
+class RepoResponse:
+    """Encapsulate the repo response.
+
+    Data can be either a dictionary (detail view) or a list
+    """
+
+    def __init__(self, data, metadata=None, links=None, status=None):
```
canni: Looks like a hidden `namedtuple` to me, no?

datahub/v2/repos/utils.py
```diff
@@ -0,0 +1,182 @@
...
+
+def extract_id_for_relationship_from_data(data, relationship_name):
+    """Give JSON api formatted data and a relationship name return the ID."""
+    return data.get('relationships', {}).get(relationship_name, {}).get('data', {}).get('id')
+
+
+def remove_null(data):
+    """Remove fields from deserialized data with colander.null."""
+    cleaned_data = {
+        'attributes': {k: v for k, v in data['attributes'].items() if v},
```
canni: That does not feel right to me, shouldn't we target explicitly the `cplander.null` here eg: `if v is not colander.null`

datahub/v2/repos/utils.py
```diff
@@ -0,0 +1,183 @@
...
+    return a dictionary inverting keys with values
+
+    {'Foo': 'foo', 'Bar': 'bar'}
+    """
+    return dict((v, k) for (k, v) in mapping)
+
+
+def model_to_json_api_data(model_instance, schema_instance, url_builder):
+    """Convert the model instance to the JSON api format."""
+    attributes = dict()
```
reupen: `{}` is better than calling the `dict()` constructor for literals, because the `{}` form is the standard representation of a dict (and it's also faster).

datahub/v2/parsers.py
```diff
@@ -0,0 +1,54 @@
...
+
+    We extract the attributes so that DRF serializers can work as normal.
+    """
+
+    media_type = 'application/vnd.api+json'
+    renderer_class = JSONRenderer
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        """Parse the incoming bytestream as JSON and returns the resulting data."""
+        result = super(JSONParser, self).parse(
```
reupen: `super()` by itself (without `JSONParser, self`) will work in Python 3.

datahub/v2/tests/repos/test_service_deliveries_repo.py
```diff
@@ -0,0 +1,227 @@
...
+        """Test get service delivery."""
+        service_offer = factories.ServiceOfferFactory()
+        service_delivery = factories.ServiceDeliveryFactory(
+            service=service_offer.service,
+            dit_team=service_offer.dit_team
+        )
+        result = ServiceDeliveryDatabaseRepo(config=DUMMY_CONFIG).get(service_delivery.pk)
+        data = result.data
+        assert isinstance(result, RepoResponse)
+        assert data['relationships']
```
reupen: Would be nice to check the contents of the entire object.

datahub/v2/views/service_deliveries.py
```diff
@@ -0,0 +1,72 @@
...
+        """Handle the GET."""
+        params = {k: v for (k, v) in request.query_params.items() if k in self.param_keys}
+        url_builder = functools.partial(
+            reverse, viewname=self.detail_view_name, request=request)
+        repo_config = {'url_builder': url_builder}
+        service_deliveries = self.repo_class(config=repo_config).filter(**params)
+        return Response(service_deliveries)
+
+    def post(self, request):
+        """Handle the POST."""
```
reupen: Would be nice to have a more meaningful docstring.

datahub/v2/tests/views/test_service_delivery.py
```diff
@@ -24,13 +21,90 @@ def test_service_delivery_detail_view(self):
             service=service_offer.service,
             dit_team=service_offer.dit_team
         )
-        url = reverse('v2:servicedelivery-detail', kwargs={'pk': servicedelivery.pk})
+        url = reverse('v2:servicedelivery-detail', kwargs={'object_id': servicedelivery.pk})
         response = self.api_client.get(url)
+        content = json.loads(response.content.decode('utf-8'))
         assert response.status_code == status.HTTP_200_OK
-        assert response.data['id'] == str(servicedelivery.pk)
+        assert set(content.keys()) == {'data'}
```
reupen: Don't need `set()` around the return value of `dict.keys()`, it acts like a set object already.

datahub/v2/renderers.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def view_has_errors(view):
+    """Return True if view has errors."""
+    try:
+        code = str(view.response.status_code)
+    except (AttributeError, ValueError):
+        pass
+    else:
+        if code.startswith('4') or code.startswith('5'):
```
reupen: Why do we need to convert `status_code` to a string? It's not clear – I'd prefer to be dealing with an integer.

sdonk: This code is copied from DRF, I think it's converted to string to use the `startwith` function.
How would handle it with ints?  `>= 400 and <=499`?

datahub/v2/renderers.py
```diff
@@ -0,0 +1,97 @@
...
+    }]
+
+    """
+    errors = []
+    for k, v in data.items():
+        error = {}
+        key = '/data/{path}'.format(path=k.replace('.', '/'))
+        error['detail'] = v
+        error['source'] = {'pointer': key}
+        error['status'] = status_code
```
sdonk: It will get converted into a string once the `render` method is called. 
I'm happy to drop the status code for errors.

datahub/v2/renderers.py
```diff
@@ -0,0 +1,97 @@
...
+        if data.status:  # 201
+            response.status_code = data.status
+        return super(JSONRenderer, self).render(
+            render_data, accepted_media_type, renderer_context
+        )
+
+
+def view_has_errors(view):
+    """Return True if view has errors."""
+    try:
```
sdonk: Well spotted, I'll change it.

datahub/v2/repos/utils.py
```diff
@@ -0,0 +1,182 @@
...
+    {('foo', 'Foo'), ('bar', 'Bar')}
+
+    return a dictionary
+
+    {'foo': 'Foo', 'bar': 'Bar'}
+    """
+    return dict(mapping)
+
+
+def types_to_attributes(mapping):
```
sdonk: Leftover code from a previous attempt. I'll delete it

datahub/v2/repos/utils.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    return dict((v, k) for (k, v) in mapping)
+
+
+def model_to_json_api_data(model_instance, schema_instance, url_builder):
+    """Convert the model instance to the JSON api format."""
+    attributes = dict()
+    relationships = dict()
+    links = {'self': build_self_link(model_instance, url_builder)}
+    attribute_to_type_mapping = attributes_to_types(model_instance.API_MAPPING)
```
sdonk: Because it's used to build `types_to_attributes`, but I'm deleting that function so this can be converted back to a dictionary

datahub/v2/repos/utils.py
sdonk: The good old `dict()` vs `{}` fight :)

What we use is provably insignificant in this case. You're right to say that we use both the forms in our codebase. We should decided on a form, add it to our style guide and get stick to it.

datahub/v2/repos/utils.py
sdonk: Actually, we need the API_MAPPING in the format of tuples, it's used in `model_to_json_api_data()`

datahub/v2/repos/utils.py
```diff
@@ -0,0 +1,182 @@
...
+def build_relationship(model_instance, attribute, attribute_to_type_mapping):
+    """Build relationships object from models."""
+    entity_name = attribute_to_type_mapping[attribute]
+    data_dict = {'data': {'type': entity_name, 'id': str(model_instance.pk)}}
+    return data_dict
+
+
+def build_attribute(model_instance, attribute):
+    """Build attributes object from model."""
+    value = getattr(model_instance, attribute, None)
```
sdonk: When the FK is not set

canni:
I'd like to move auth code from korben to leeloo and stop korben komponent

PR #159, opened 2017-03-27
DHE-830 RO Account management tier

canni:


datahub/company/test/test_company_views.py
```diff
@@ -137,6 +137,24 @@ def test_update_company(self, mocked_save_to_korben):
...
+        company = CompanyFactory(
+            name='Foo ltd.',
+            registered_address_1='Hello st.',
+            registered_address_town='Fooland',
+            registered_address_country_id=constants.Country.united_states.value.id,
+            classification_id=constants.CompanyClassification.tier_a.value.id,
+        )
+
+        url = reverse('v1:company-detail', kwargs={'pk': company.pk})
+        self.api_client.patch(url, {
```
reupen: Would be nice to check the response status code here, since it could be failing due to some unrelated problem, and we'd never know.

canni: Sure thing

PR #165, opened 2017-03-31
Include company details for interactions in homepage endpoint

reupen:


datahub/dashboard/test/test_views.py
```diff
@@ -37,4 +37,7 @@ def test_intelligent_homepage(self):
         assert len(response.data['contacts']) == 1
         assert response.data['contacts'][0]['id'] == str(api_response.data['id'])
         assert len(response.data['interactions']) == 1
-        assert response.data['interactions'][0]['id'] == str(interaction.pk)
+        resp_interaction = response.data['interactions'][0]
+        assert resp_interaction['id'] == str(interaction.pk)
+        assert isinstance(resp_interaction['company'], dict)
+        assert isinstance(resp_interaction['company']['name'], str)
```
sdonk: LGTM. One thing, I would assert that the name is actually what we expect.

reupen: Updated.

PR #173, opened 2017-04-07
Use bulk_create to import CH companies into DB

sdonk:


datahub/company/models.py
```diff
@@ -168,7 +168,7 @@ def clean(self):
 class CompaniesHouseCompany(CompanyAbstract):
     """Representation of Companies House company."""
 
-    company_number = models.CharField(max_length=MAX_LENGTH, unique=True)
+    company_number = models.CharField(max_length=MAX_LENGTH, primary_key=True)
```
canni: I'm not happy with this :P as per conversation

datahub/company/management/commands/sync_ch.py
```diff
@@ -73,43 +76,51 @@ def open_ch_zipped_csv(fp):
...
 
 
-def is_changed(company, csv_data):
-    """Return whatever company data changed."""
-    return any(csv_data[name] != getattr(company, name) for name in csv_data)
-
-
-def sync_ch(tmp_file_creator, endpoint=None):
+def sync_ch(tmp_file_creator, endpoint=None, truncate_first=False):
     """Do the sync."""
```
reupen: Would be nice to add a comment here explaining why we're batching the records ourselves instead of using the functionality Django provides in `QuerySet.bulk_create()`.

datahub/company/test/test_sync_ch_command.py
```diff
@@ -56,11 +55,9 @@ def test_unzip_and_csv_read_on_the_fly():
 @mock.patch('datahub.company.management.commands.sync_ch.get_ch_latest_dump_file_list')
 def test_full_ch_sync(file_list_mock):
     """Test the whole process."""
-    CompaniesHouseCompanyFactory(company_number='08209948', name='Will change')
```
reupen: What batch size (`BULK_CREATE_BATCH_SIZE`) are we using during the tests? We need to test that the batching works.

PR #184, opened 2017-04-27
Drambuie release

jim68000:
There's a lot of commits in this: we will have to be a bit careful in reviewing these files...


canni:
First thing, let's add new `env` vars required for CDMS auth before releasing this, it won't affect current prod, and it'll save us from rollbacks & DB migration problems

PR #196, opened 2017-05-11
Allow null's instead of undefs

canni:


reupen:
Looks fine, are we planning to get rid of the on_delete=CASCADE?

PR #212, opened 2017-05-22
[DH-164] [DH-179] [DH-180] [DH-206] Search endpoints and sync command.

elcct:
This adds management command to sync `Company` and `Contact` models with Elasticsearch.

It also adds 3 search endpoints for basic and detailed search - specs are available in `data-hub-api-spec` repository.


datahub/search/models.py
```diff
@@ -0,0 +1,117 @@
...
+    archived = Boolean()
+    archived_on = Date()
+    archived_reason = String()
+    created_on = Date()
+    modified_on = Date()
+    id = String(index='not_analyzed')
+    name = String()
+    title = Nested(properties={'id': String(index='not_analyzed'), 'name': String(copy_to='name')})
+    first_name = String(copy_to='name')
+    first_name = String(copy_to='name')
```
marcofucci: duplicate ;-)

elcct: Nice! Fixed

datahub/search/elasticsearch.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        connections.configure(
+            default={
+                'host': settings.ES_HOST,
+                'port': settings.ES_PORT
+            }
+        )
+
+
+def get_basic_search_query(term, entities=['company'], offset=0, limit=100):
```
reupen: > entities=['company']

Should avoid using mutable values as defaults since they aren't reinitialised between calls. (Although entities isn't being mutated here, still better to avoid completely I'd say.)

Could use a tuple or None.

datahub/search/elasticsearch.py
```diff
@@ -0,0 +1,130 @@
...
+        Q('multi_match', query=term, fields=['name', '_all']),
+        Q('term', _type=entity),
+    )
+
+    query_filter = []
+    for k, v in filter(lambda f: '.' not in f[0], filters.items()):
+        query_filter.append(Q('term', **{k: v}))
+
+    # query nested fields
+    for k, v in filter(lambda f: '.' in f[0], filters.items()):
```
reupen: I think this would be clearer with a simple `if` inside the for loop instead of the double-filtering?

datahub/search/elasticsearch.py
```diff
@@ -0,0 +1,130 @@
...
+    """Performs filtered search for contact."""
+    return get_search_by_entity_query(entity='contact', **kwargs)
+
+
+def bulk(actions=None, chunk_size=None, **kwargs):
+    """Send data in bulk to Elasticsearch."""
+    return es_bulk(connections.get_connection(), actions=actions, chunk_size=chunk_size, **kwargs)
+
+
+def legacy_search_by_term_query(term, doc_type=None, offset=0, limit=100):
```
reupen: Is this function used?

datahub/search/elasticsearch.py
```diff
@@ -0,0 +1,130 @@
...
+    name_map = {
+        'sector': 'sector.id',
+        'account_manager': 'account_manager.id',
+        'export_to_country': 'export_to_countries.id',
+        'future_interest_country': 'future_interest_countries.id',
+        'uk_region': 'uk_region.id',
+        'trading_address_country': 'trading_address_country.id',
+        'advisor': 'advisor.id',
+    }
+    return dict(zip(map(lambda x: name_map.get(x, x), filter.keys()), filter.values()))
```
reupen: Can we use a dict comprehension instead? Easier to read and shorter ;)

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,179 @@
...
+logger = getLogger(__name__)
+
+ES_INDEX = settings.ES_INDEX
+
+DataSet = namedtuple('DataSet', ('queryset', 'es_model',))
+FieldsColumnFn = namedtuple('FieldsColumnFn', ('fields', 'fn'))
+
+fields_column_fn = (
+    FieldsColumnFn(fields=('companies_house_data'),
+                   fn=lambda col: _company_dict(col)),
```
reupen: Can this be just `fn=_company_dict`?

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,179 @@
...
+    FieldsColumnFn(fields=('account_manager', 'archived_by', 'one_list_account_owner',),
+                   fn=lambda col: _contact_dict(col)),
```
reupen: Can this be just `fn=_contact_dict`?

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,179 @@
...
+        'company_number': obj.company_number,
+    }
+
+
+def _model_to_dict(model, fields_column_fn=fields_column_fn):
+    """Converts model instance to a dictionary suitable for ElasticSearch."""
+    result = {}
+    for fcf in fields_column_fn:
+        result.update({field: fcf.fn(getattr(model, field))
+                       for field in fcf.fields if hasattr(model, field) and getattr(model, field)})
```
reupen: Looks like we could just pass a default to `getattr()` instead of calling `hasattr()` and `getattr()`?

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _company_dict(obj):
+    return {
+        'id': str(obj.id),
+        'company_number': obj.company_number,
+    }
+
+
+def _model_to_dict(model, fields_column_fn=fields_column_fn):
```
reupen: This shadows `fields_column_fn` from above – one of them should be renamed...

datahub/search/elasticsearch.py
```diff
@@ -0,0 +1,130 @@
...
+
+    query_filter = []
+    for k, v in filter(lambda f: '.' not in f[0], filters.items()):
+        query_filter.append(Q('term', **{k: v}))
+
+    # query nested fields
+    for k, v in filter(lambda f: '.' in f[0], filters.items()):
+        query_filter.append(
+            Q('nested', path=k.split('.')[0],
+              query=Q('term', **{k: v}))
```
reupen: Is `k` right here? Might be what elasticsearch wants, but just looks a bit odd as it's split on the previous line.

conftest.py
```diff
@@ -30,5 +31,62 @@ def api_client():
...
+    application, _ = Application.objects.get_or_create(
+        user=get_test_user(),
+        client_type=Application.CLIENT_CONFIDENTIAL,
+        authorization_grant_type=Application.GRANT_PASSWORD,
+        name='Test client'
+    )
+    return application
+
+@pytest.fixture
+def logged_in_api_client():
```
elcct: Fixed that.

datahub/search/elasticsearch.py
elcct: Fixed that.

datahub/search/elasticsearch.py
elcct: Good point. Fixed :)

datahub/search/elasticsearch.py
elcct: Yes, it is what Elasticsearch wants.

datahub/search/elasticsearch.py
elcct: It is not. Deleted.

datahub/search/elasticsearch.py
elcct: True :) Fixed.

datahub/search/management/commands/sync_es.py
elcct: Don't want to include in the result fields/attrs that don't exist in the model. It processes both Company and Contact and for example Contact doesn't have some fields that are in the Company model. Default would not work in this case.

datahub/search/management/commands/sync_es.py
elcct: Fixed

datahub/search/management/commands/sync_es.py
elcct: Fixed

datahub/search/management/commands/sync_es.py
elcct: True. Fixed.

datahub/search/management/commands/sync_es.py
reupen: Isn't `if getattr(model, field, None)` the same as `if hasattr(model, field) and getattr(model, field)`?

Anyway, can't we define a `_fields_column_fn` per model? That should avoid the need for hasattr() or setattr() + default which might mask bugs.

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,181 @@
...
+    return {
+        'id': str(obj.id),
+        'company_number': obj.company_number,
+    }
+
+
+FieldsColumnFn = namedtuple('FieldsColumnFn', ('fields', 'fn'))
+
+_fields_column_fn = (
+    FieldsColumnFn(fields=('companies_house_data'),
```
reupen: Missing comma in `('companies_house_data')`

datahub/search/test/test_management_commands.py
```diff
@@ -0,0 +1,146 @@
...
+def test_id_name_dict():
+    """Tests _id_name_dict."""
+    obj = mock.Mock()
+    obj.id = 123
+    obj.name = 'test'
+
+    res = sync_es._id_name_dict(obj)
+
+    assert res['id'] == str(obj.id)
+    assert res['name'] == obj.name
```
marcofucci: Not a big deal but I find
```
assert res == {
    'id': str(obj.id),
    'name': obj.name
}
```

Better to read as it's obvious what the expected data should be.

datahub/search/management/commands/sync_es.py
elcct: Good point with getattr, not sure what I was thinking :) - Fixed

Regarding splitting _fields_column_fn per model - problem is that Elasticsearch needs to have mappings of the same name to be the same across different DocTypes, for instance name in Company must be String, just as name in Contact. Processing columns in the same way for all models could help catch errors in case one model's column is of different type and needs extra "treatment". More at: https://www.elastic.co/guide/en/elasticsearch/guide/current/mapping.html

datahub/search/management/commands/sync_es.py
elcct: Nice! Fixed

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,181 @@
...
+                   fn=_company_dict),
+    FieldsColumnFn(fields=('account_manager', 'archived_by', 'one_list_account_owner',),
+                   fn=_contact_dict),
+    FieldsColumnFn(fields=('business_type', 'classification', 'employee_range',
+                           'headquarter_type', 'parent', 'registered_address_country',
+                           'sector', 'trading_address_country', 'turnover_range', 'uk_region',
+                           'address_country', 'company', 'title',),
+                   fn=_id_name_dict),
+    FieldsColumnFn(fields=('interactions',),
+                   fn=lambda col: list(map(_id_type_dict, col.all()))),
```
elcct: Yes, you're right :) Fixed.

datahub/search/test/test_management_commands.py
elcct: That is indeed easier to read. Fixed.

datahub/search/test/test_views.py
```diff
@@ -1,71 +1,116 @@
...
 
-class SearchViewTestCase(LeelooTestCase):
-    """Search test case."""
 
-    def test_search_missing_required_parameter(self):
-        """Required parameter missing."""
-        url = reverse('search')
-        response = self.api_client.post(url)
+@pytest.mark.usefixtures('client', 'setup_data')
+class SearchTestCase(LeelooTestCase):
```
reupen: These tests are getting skipped (in CircleCI too), how should they be run?

reupen:
Is the front end ready for these changes? How will the switchover be managed?

datahub/search/elasticsearch.py
```diff
@@ -0,0 +1,114 @@
...
+
+    query_filter = []
+
+    for k, v in filters.items():
+        if '.' not in k:
+            query_filter.append(Q('term', **{k: v}))
+        else:
+            # query nested fields
+            query_filter.append(
+                Q('nested', path=k.split('.')[0], query=Q('term', **{k: v}))
```
elcct: Good point. Fixed

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,181 @@
...
+    return {
+        'id': str(obj.id),
+        'company_number': obj.company_number,
+    }
+
+
+FieldsColumnFn = namedtuple('FieldsColumnFn', ('fields', 'fn'))
+
+_fields_column_fn = (
+    FieldsColumnFn(fields=('companies_house_data', ),
```
elcct: I had initially done it via dict field -> fn, but found that it was too much repetition. I agree this way is not too pleasing to look at, so I changed it back to dict and also split into company and contact.

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,181 @@
...
+        actions = list(_dict_to_es(doc_type, _models_to_dict(batch)))
+        bulk(actions=actions,
+             chunk_size=len(actions),
+             request_timeout=300,
+             raise_on_error=True,
+             raise_on_exception=True,
+             )
+
+
+def sync_es(batch_size, dataset=None):
```
elcct: Thanks, it is left over from the beginnings and forgot to remove. Fixed.

datahub/search/management/commands/sync_es.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def handle(self, *args, **options):
+        """Handle."""
+        try:
+            sync_es(batch_size=options['batch_size'], dataset=get_dataset())
+        except Exception as e:
+            with log_and_ignore_exceptions():
+                client.captureException()
+
+            logger.exception('Failed to sync to ES.')
```
elcct: I copied it from sync_ch, without giving much thought. I agree this is pointless, so I removed entire block to leave just the call to sync_es.

datahub/search/views.py
```diff
@@ -1,35 +1,121 @@
...
+class SearchBasicAPIView(APIView):
+    """Aggregate company and contact search view."""
 
-    http_method_names = ('post', )
+    http_method_names = ('get',)
+
+    def get(self, request, format=None):
+        """Performs basic search."""
+        term = request.query_params['term']
+        entity = request.query_params.get('entity', 'company')
```
elcct: Indeed there should be validation. Fixed.

datahub/search/views.py
```diff
@@ -1,35 +1,121 @@
...
+        response = {
+            'count': results.hits.total,
+            'aggregations': list(map(lambda x: {'count': x['doc_count'], 'entity': x['key']},
+                                     results.aggregations['count_by_type']['buckets'])),
+        }
+
+        hits = list(map(lambda x: x.to_dict(), results.hits))
+
+        if entity == 'company':
+            response.update({'companies': hits})
```
elcct: I think that's also left over from the beginnings. Didn't want to switch back and forth between the both forms when playing around. Fixed.

datahub/search/views.py
```diff
@@ -1,35 +1,127 @@
...
+        results = elasticsearch.get_basic_search_query(
+            term=term,
+            entities=entity.split(','),
+            offset=offset,
+            limit=limit
+        ).execute()
+
+        response = {
+            'count': results.hits.total,
+            'aggregations': list(map(lambda x: {'count': x['doc_count'], 'entity': x['key']},
```
marcofucci: Again, a list comprehension could be clearer here

datahub/search/views.py
```diff
@@ -1,35 +1,127 @@
...
+                                     results.aggregations['count_by_type']['buckets'])),
+        }
+
+        hits = list(map(lambda x: x.to_dict(), results.hits))
```
marcofucci: and here

datahub/search/test/test_views.py
elcct: These run fine locally. I'll have a look as indeed I can see they don't run in CircleCi.

datahub/search/views.py
elcct: Fixed.

datahub/search/views.py
elcct: Fixed.

datahub/search/test/test_views.py
elcct: Turns out `TEST_ES_SERVER` env variable in `docker-compose.yml` was not set. Test client by default connects to localhost - it is where I had instance of Elasticsearch running, that's why my tests worked. Fixed.

elcct:
@reupen As far as I know front end doesn't support it - they are supposed to implement it once it is deployed somewhere. Ideally in deployment it should point at new instance of Elasticsearch, then `sync_es` command can be run from the container to sync the data.

reupen:
@elcct I asked as the old endpoints are being removed, so merging this looks like it will break the front end. @jim68000 How do you think we should handle this?

marcofucci:
Maybe we could keep the old endpoints, merge this in, make the frontend use the new endpoints, delete the old endpoints but @jim68000 might have a better idea.

elcct:
@reupen @marcofucci @jim68000 I think that would need some extra work for them to coexist, as old endpoint would have to use the new index and transform the output to be consistent with the old schema or we would have to use two Elasticsearch endpoints, because of the difference in mappings. I think this would be unnecessary work to get it up, so it could be killed right afterwards.

jim68000:
So, I think we agreed in planning that we would start work on the new FE for search, and that @feedmypixel would be doing it. 

In general we should try and maintain old and new endpoints so we can upgrade the FE service when we want to, rather than have to keep things in sync (my last place was a microservices-based system, and you run into that requirement quite quickly). 

The question then is how long long dev will be in an unusable state. As search is fundamental to the way we use the site, changing the API will stop other paths from working - if it's < sprint we can probably just go ahead and build out the new version and not worry about backward compatibility, but if it's going to be longer than that, we probably need to fix this so we can use both indexes until front end is migrated over.

PR #213, opened 2017-05-23
Remove teams from contact model and endpoint

marcofucci:
As discussed with the team, the `Contact.teams` field doesn't seem very useful at the moment so we decided to get rid of it. We might add it in the future if we find a use for it.

Also, `Contact.address_country` and `Contact.company` had `null=True` but not `blank=True` which has been fixed.


marcofucci:
Apart from the `teams` this also fixes `Contact.address_country` and `Contact.company`. They had `null=True` but not `blank=True` which I think it's not correct. Is there anything I might be missing about those fields?

PR #214, opened 2017-05-23
Remove the selectable feature

canni:
Depends on uktrade/data-hub-fe-beta2#114


reupen:
The front end uses selectable when adding interactions (for interaction type) so we need to kill that first.

PR #222, opened 2017-05-26
Add ability to filter interactions by contact

marcofucci:
This adds the `?contact_id=...` filter to the interaction API v1.

The reason behind it is that at the moment the frontend is getting the interactions by using the contact endpoint v1 and looping over the `contact.interactions` field.
By adding this, we can refactor the frontend and then get rid of the contact API v1.

I would consider this just as a temporary solution and we should really convert the interaction endpoints to v3 as well.

### DIFF
--- a/datahub/interaction/test/test_interaction_views.py
+++ b/datahub/interaction/test/test_interaction_views.py
@@ -63,11 +63,8 @@ class InteractionTestCase(LeelooTestCase):
         InteractionFactory.create_batch(3, contact=contact1)
         interactions = InteractionFactory.create_batch(2, contact=contact2)
 
-        url = '{}?contact_id={}'.format(
-            reverse('api-v1:interaction-list'),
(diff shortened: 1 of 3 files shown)


datahub/interaction/test/test_interaction_views.py
```diff
@@ -54,3 +54,21 @@ def test_modify_interaction(self):
...
+
+    def test_list_filtered(self):
+        """List of interactions filtered by contact"""
+        contact1 = ContactFactory()
+        contact2 = ContactFactory()
+
+        InteractionFactory.create_batch(3, contact=contact1)
+        interactions = InteractionFactory.create_batch(2, contact=contact2)
+
+        url = '{}?contact_id={}'.format(
```
reupen: Should be able to pass the URL params to `self.api_client.get()` as the data param (as a dict).

PR #223, opened 2017-05-26
Remove squashed migrations and add MPTT hierarchy to Company

canni:


reupen:
The Django docs say to remove the replaces attribute from the squashed migrations – any reason for keeping them?

https://docs.djangoproject.com/en/1.11/topics/migrations/#squashing-migrations

canni:
@reupen nice catch, I forgot about it, corrected

PR #224, opened 2017-05-30
Add business leads endpoints

reupen:
Adds the business lead model and following endpoints:

- `GET v3/business-leads`
- `POST v3/business-leads`
- `GET v3/business-leads/{businessLeadId}`
- `PATCH v3/business-leads/{businessLeadId}`
- `POST v3/business-leads/{businessLeadId}/archive`
- `POST v3/business-leads/{businessLeadId}/unarchive`


datahub/leads/models.py
```diff
@@ -0,0 +1,66 @@
...
+    contactable_by_dit = models.BooleanField(default=False)
+    contactable_by_dit_partners = models.BooleanField(default=False)
+    contactable_by_email = models.BooleanField(default=False)
+    contactable_by_phone = models.BooleanField(default=False)
+    notes = models.TextField(null=True, blank=True)
+
+    @property
+    def name(self):
+        """Formatted name."""
+        return f'{self.first_name} {self.last_name}'
```
canni: Hmm should we keep this consistent? ;) Most of the code uses `"".format(...)`

datahub/leads/serializers.py
```diff
@@ -0,0 +1,86 @@
...
+        company_name = data_view.get_value('company_name')
+        trading_name = data_view.get_value('trading_name')
+        company = data_view.get_value('company')
+        first_name = data_view.get_value('first_name')
+        last_name = data_view.get_value('last_name')
+        telephone_number = data_view.get_value('telephone_number')
+        email = data_view.get_value('email')
+
+        if not any((company_name, company, trading_name)) and not (
+                first_name and last_name):
```
canni: This `if` is really terrible to read :(

datahub/leads/serializers.py
```diff
@@ -0,0 +1,86 @@
...
+            errors['company_name'] = NAME_REQUIRED_MESSAGE
+            errors['first_name'] = NAME_REQUIRED_MESSAGE
+            errors['last_name'] = NAME_REQUIRED_MESSAGE
+
+        if not email and not telephone_number:
```
canni: `not ( .. or ..)`?

datahub/leads/models.py
reupen: Well, we are using Python 3.6 now :stuck_out_tongue: I'd rather convert all the existing ones to be f-strings where they can be unless we have a reason not to use them.

datahub/leads/serializers.py
reupen: I used some variables to make it a bit nicer.

datahub/leads/serializers.py
reupen: Changed.

datahub/leads/models.py
canni: I like `f`-strings but I like consistency more, feel free to convert all others ;)

PR #227, opened 2017-05-31
Add investment project interactions

reupen:


datahub/core/validate_utils.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def __call__(self, attrs):
+        """
+        Performs validation. Called by DRF.
+
+        :param attrs:   Serializer data (post-field-validation/processing)
+        """
+        data_view = UpdatedDataView(self.serializer.instance, attrs)
+        values = (data_view.get_value(field) for field in self.fields)
+        value_present = any(value for value in values if value is not None)
```
canni: Hmm... Would the use of `set` and `in` test be nicer here?

PR #229, opened 2017-06-01
Explicitly set on_delete for all foreign keys

reupen:
This fixes the large amount of deprecation warnings that were being generated due to not explicitly specifying values for on_delete for foreign keys.


reupen:
@canni I've done that. Don't merge this though as I'll do a squash and merge.

PR #234, opened 2017-06-02
Audit log initial work

canni:


datahub/investment/serializers.py
```diff
@@ -103,6 +105,50 @@ class Meta:  # noqa: D101
...
+
+    def to_representation(self, instance):
+        """Overwrite serialization process completely to get the Versions."""
+        versions = Version.objects.get_for_object(instance)
+        version_pairs = [
+            (versions[n], versions[n + 1]) for n in range(len(versions) - 1)
+        ]
+
+        return {
+            'audit_log': self._construct_changelog(version_pairs),
```
reupen: `results` might be more consistent with the other collection endpoints.

datahub/investment/serializers.py
```diff
@@ -103,6 +105,50 @@ class Meta:  # noqa: D101
         extra_kwargs = {'nda_signed': {'required': True}}
 
 
+class IProjectAuditSerializer(serializers.Serializer):
+    """Serializer for Investment Project audit log."""
+
+    def to_representation(self, instance):
+        """Overwrite serialization process completely to get the Versions."""
+        versions = Version.objects.get_for_object(instance)
+        version_pairs = [
```
reupen: Not a big deal but could leave it as a generator expression.

datahub/investment/serializers.py
```diff
@@ -103,6 +105,50 @@ class Meta:  # noqa: D101
...
+        return {
+            'audit_log': self._construct_changelog(version_pairs),
+        }
+
+    def _construct_changelog(self, version_pairs):
+        changelog = []
+
+        for v_new, v_old in version_pairs:
+            changelog.append({
+                'user': AdvisorSerializer(v_new.revision.user).data,
```
reupen: Generally for v3 endpoints we've only been including `id`, `first_name` and `last_name` for advisers returned as related fields.

What happens if `user` is `None`?

datahub/investment/test/test_investment_serializers.py
```diff
@@ -0,0 +1,24 @@
+from datahub.investment import serializers
+
+
+def test_audit_log_diff_algo():
+    """Test simple diff algorithm."""
+    given = {
+        'old': {
+            'field1': 'val1',
+            'field2': 'val2',
+            'field3': None,
```
reupen: Not sure if `field3` was meant to be in `old` but would be nice to test a field being in new and not old.

PR #237, opened 2017-06-05
Removed unnecessary type conversion.

elcct:
Client sends POST data as JSON, so Booleans don't come as strings.

### DIFF
--- a/datahub/search/elasticsearch.py
+++ b/datahub/search/elasticsearch.py
@@ -90,7 +90,7 @@ def bulk(actions=None, chunk_size=None, **kwargs):
     return es_bulk(connections.get_connection(), actions=actions, chunk_size=chunk_size, **kwargs)
 
 
-def remap_fields(filter):
+def remap_fields(fields):
     """Replaces fields to match Elasticsearch data model."""
     name_map = {
         'sector': 'sector.id',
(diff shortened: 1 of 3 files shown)


datahub/search/test/test_views.py
```diff
@@ -114,13 +115,13 @@ def test_search_company(self):
         assert response.data['results'][0]['trading_address_country']['id'] == constants.Country.united_states.value.id
 
     @mock.patch('datahub.search.views.elasticsearch.ES_INDEX', 'test')
-    def test_search_foreign_company(self):
+    def test_search_foreign_company_json(self):
         """Tests detailed company search."""
         url = f"{reverse('api-v3:search:company')}?offset=0&limit=100"
 
-        response = self.api_client.post(url, {
+        response = self.api_client.post(url, json.dumps({
```
reupen: Should be able to use `format='json'`.

datahub/search/elasticsearch.py
```diff
@@ -101,7 +101,4 @@ def remap_fields(filter):
         'trading_address_country': 'trading_address_country.id',
         'advisor': 'advisor.id',
     }
-    value_map = {
-        'uk_based': lambda x: x.lower() == 'true'
-    }
-    return {name_map.get(k, k): value_map.get(k, str)(v) for k, v in filter.items()}
+    return {name_map.get(k, k): v for k, v in filter.items()}
```
canni: `filter` variable name shadows global func, nothing terrible, but... ;)

elcct: good spot!

PR #239, opened 2017-06-05
Rename advisor to adviser in public API

reupen:
Renames advisor to adviser across the public API (all URLs and request and response bodies) and wherever else possible.

The database model is still Advisor as that is more tricky to rename (as it is also the user model).

@elcct This will affect Elasticsearch.


canni:
Shouldn't this be synchronised with FE deploy?

reupen:
@canni Yes, @teneightfive are you planning to update your PR?

teneightfive:
@reupen 
>@teneightfive are you planning to update your PR?

I've updated the [frontend PR]() to the latest develop but haven't made the changes that will be affected by this PR yet. I can make those now though.

From what we have in frontend these will be the effected areas:

- `/advisor/` endpoint
- `referral_source_advisor` property on the investment object
- `dit_advisor` properties on interactions and service deliveries

Are there any other areas that will change that we will need to update?

@ztolley might be a good idea for you to take a look too.

reupen:
@teneightfive 
There's also an advisor/adviser property on contacts, and project_assurance_adviser on investment projects (team section). There's also the `type` for the `dit_advisor` field in the v2 service deliveries endpoint.

Basically, you should be able to completely get rid of advisor from the front end.

canni:
`s/([Aa])dvisor/$1dviser/g` ;]

teneightfive:
>Basically, you should be able to completely get rid of advisor from the front end.

@reupen Excellent.

@canni I originally did that but it obviously broke a few journeys...

PR #241, opened 2017-06-06
Auto sync Company and Contact to Elasticsearch on save.

elcct:
It binds to `post_save` signals of Company and Contact models to perform sync.


datahub/search/models.py
```diff
@@ -110,8 +212,17 @@ class Contact(DocType):
...
+        'adviser': _id_name_dict,
+        'company': _id_name_dict,
+        'archived_by': _contact_dict,
+    }
+
     class Meta:
         """Default document meta data."""
 
-        index = settings.ES_INDEX
+        index = elasticsearch.ES_INDEX
```
reupen: What's the reason for changing this?

elcct: I had a mix up with ES_INDEX during tests and used `elasticsearch.ES_INDEX` as a source of truth, which is not right. I should have cleaned it up. It is fixed now.

conftest.py
```diff
@@ -1,10 +1,17 @@
...
 
+import django
 import pytest
+from django.conf import settings
 from django.core.management import call_command
 from pytest_django.lazy_django import skip_if_no_django
 
 
+def pytest_configure():
+    settings.ES_INDEX = 'test'
```
reupen: How about defining a test settings module and changing `DJANGO_SETTINGS_MODULE` in `pytest.ini` to point at that?

elcct: This is a good idea. Fixed.

elcct:
@reupen Elasticsearch should now only be needed for `datahub.search` tests.

config/settings/test.py
```diff
@@ -0,0 +1,14 @@
+import environ
+
+environ.Env.read_env()  # reads the .env file
+env = environ.Env()
+
+from .common import *
+
+search_config_index = INSTALLED_APPS.index('datahub.search.apps.SearchConfig')
+
+INSTALLED_APPS = INSTALLED_APPS[0:search_config_index] + \
```
reupen: Can we modify INSTALLED_APPS in-place?

reupen:
@elcct I still seem to be hitting Elasticsearch and the signals when running e.g. company tests.

config/settings/test.py
elcct: Yes, we can.

elcct:
@reupen It is not the case here. I disabled my `Elasticsearch` container and issued:
```
pytest --cov -k "not liveserver" -s datahub/company
```
It worked fine, then I run:
```
pytest --cov -k "not liveserver" -s datahub/search
```
It fails, because can't connect to `Elasticsearch`.

reupen:
Ah, it's because the `DJANGO_SETTINGS_MODULE` environment variable overrides the setting in pytest.ini.

If we specify the config module in addopts instead using `--ds=config.settings.test` that will sort it.

elcct:
@reupen I didn't have `DJANGO_SETTINGS_MODULE` set, so looks like that was the issue then. I have added the flag now.

tests-auth.sh
```diff
@@ -3,4 +3,4 @@
 # For some obscure reason all the tests that run after the auth ones get stuck into using this different thread process
 # which doesn't have any access to the testing db hence there are no fixtures and the tests fail.
 # The solution is to run the auth tests independently.
-pytest -k liveserver
+pytest --ds=config.settings.test -k liveserver
```
reupen: Shouldn't need to specify `--ds` here since it's in `pytest.ini`?

elcct: That is true. I removed it.

config/settings/test.py
reupen: I reckon it would be clearer modifying it in-place.

elcct: Ah sorry, I meant that it is possible, but not necessarily useful here. What I want is to avoid a call to `ready` in `SearchConfig`, that sets up `Elasticsearch` for use in the backend. I thought easiest way to avoid dependencies is to just "replace" `datahub.search.apps.SearchConfig` with  `datahub.search` in the test settings. I could add a comment to explain why such operation is there. What do you think?

reupen: Yes, a comment would be useful here. But I don't think that's related to how `INSTALLED_APPS` is modified. I see that it's a tuple at the moment but we could just make it a list.

elcct: I see. I was reluctant to replace these with lists, but it did simplified it. I have updated the PR. Thanks!

PR #243, opened 2017-06-07
Fixes check for the type in order to perform correct count method.

elcct:


datahub/search/management/commands/sync_es.py
```diff
@@ -93,7 +93,19 @@ def _company_dict(obj):
 def get_dataset():
     """Returns dataset that will be synchronised with Elasticsearch."""
     return (
-        DataSet(Company.objects.all().order_by('pk'), ESCompany, _company_mappings),
+        DataSet(Company.objects.prefetch_related('registered_address_country',
```
reupen: Can we shift the company query to a separate line/statement? This one's a bit overloaded.

elcct: That's true, it doesn't look pleasant. I have amended it.

PR #250, opened 2017-06-12
Investment Project Documents

canni:


datahub/documents/models.py
```diff
@@ -0,0 +1,46 @@
...
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    path = models.CharField(
+        max_length=settings.CHAR_FIELD_MAX_LENGTH,
+        unique=True,
+    )
+    uploaded_on = models.DateTimeField(
+        null=True, blank=True
+    )
+    av_clean = models.BooleanField(default=False)
```
elcct: I think something like `checked_on` would be useful to see performance of the checks and for audit - for example if it turns out file is infected or if we want to check files that has not been checked in let's say two weeks.

datahub/documents/models.py
```diff
@@ -0,0 +1,46 @@
...
+        return path.basename(self.path)
+
+    @property
+    def name(self):
+        """Model name."""
+        return self.path
+
+    @property
+    def url(self):
+        """Generate pre-signed download URL."""
```
elcct: At least property should say that the url is signed?

datahub/investment/migrations/0006_add_investemnt_document.py
```diff
@@ -0,0 +1,44 @@
...
+                ('project', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='documents', to='investment.InvestmentProject')),
+            ],
+            options={
+                'verbose_name': 'Investment Project Document',
+                'verbose_name_plural': 'Investment Project Documents',
+            },
+        ),
+        migrations.AlterUniqueTogether(
+            name='iprojectdocument',
+            unique_together=set([('project', 'doc_type', 'filename')]),
```
reupen: Could use a set literal.

PR #255, opened 2017-06-13
Investment Project search endpoint and fixes.

elcct:


datahub/search/elasticsearch.py
```diff
@@ -98,5 +111,35 @@ def remap_fields(fields):
...
+    range_keys = set()
+
+    for k, v in fields.items():
+        if k.endswith('_before') or k.endswith('_after'):
+            range_keys.add(k[:k.rindex('_')])
+            continue
+
+        result.update({k: v})
+
+    for r in range_keys:
```
reupen: What is r? Can we use a more meaningful variable name?

datahub/search/elasticsearch.py
```diff
@@ -98,5 +111,35 @@ def remap_fields(fields):
...
+        ranges[r] = {}
+        before_key = f'{r}_before'
+        after_key = f'{r}_after'
+
+        if after_key in fields:
+            ranges[r]['gte'] = dateutil.parser.parse(fields.get(after_key))
```
reupen: Why `fields.get(after_key)` and not `fields[after_key]`?

What happens if parse() raises an exception due to invalid input?

datahub/search/management/commands/sync_es.py
```diff
@@ -6,8 +6,11 @@
 from django.db import models
 
 from datahub.company.models import Company, Contact
+from datahub.investment.models import InvestmentProject
 from datahub.search.elasticsearch import bulk
-from datahub.search.models import Company as ESCompany, Contact as ESContact
+from datahub.search.models import Company as ESCompany, \
```
reupen: PEP 8 style is to prefer brackets for continuation instead of `\`.

datahub/search/models.py
```diff
@@ -162,6 +163,13 @@ class Company(DocType, MapDBModelToDict):
         'future_interest_countries': lambda col: [_id_name_dict(c) for c in col.all()],
     }
 
+    IGNORED_FIELDS = (
+        'subsidiaries', 'servicedeliverys', 'investor_investment_projects',
```
reupen: `subsidiaries` was renamed to `children`, looks like that's been reverted here.

datahub/search/models.py
```diff
@@ -201,21 +209,183 @@ class Contact(DocType, MapDBModelToDict):
...
+    )
+
     class Meta:
         """Default document meta data."""
 
         index = settings.ES_INDEX
         doc_type = 'contact'
+
+
+class InvestmentProject(DocType, MapDBModelToDict):
```
reupen: Perhaps we should split the models up into multiple files (subpackage)? As we're probably going to end up with quite a bit in this file otherwise.

datahub/search/elasticsearch.py
```diff
@@ -98,5 +111,35 @@ def remap_fields(fields):
...
+    for k, v in fields.items():
+        if k.endswith('_before') or k.endswith('_after'):
+            range_keys.add(k[:k.rindex('_')])
+            continue
+
+        result.update({k: v})
+
+    for r in range_keys:
+        ranges[r] = {}
+        before_key = f'{r}_before'
```
reupen: We stripped off `_before`/`_after` above and are adding them again here, I'm sure this could be tidied up a bit.

datahub/search/elasticsearch.py
elcct: My Golang is leaking :) I am sorry. Fixed.

datahub/search/elasticsearch.py
elcct: It is indeed awful, simplified it.

datahub/search/elasticsearch.py
elcct: Re .get vs [] it was just a habit. Replaced to [] since it doesn't need default value.

I have added check for exception in the view.

datahub/search/management/commands/sync_es.py
elcct: Fixed. Thanks!

datahub/search/models.py
elcct: Fixed.

datahub/search/models.py
elcct: I was thinking about that and I think it is still small to do that, maybe when fourth comes in?

PR #257, opened 2017-06-14
Pin indirect dependencies using pip-compile

reupen:
This pins indirect dependencies using pip-compile (from pip-tools). This avoids random breakages, and also fixes the current deployment problem with Flynn.

`requirements.in` is now the user-edited file and `requirements.txt` is a generated lock file.

Whenever `requirements.txt` is edited, the lock file needs to be updated by running:

```
pip-compile --upgrade --output-file requirements.txt requirements.in
```

Additional info is in the README.


reupen:
Removed the hashes as they currently aren't portable.

PR #265, opened 2017-06-16
Improve sync_ch performance in Django debug mode

reupen:
In debug mode, django keeps a history of all SQL statements executed. This was causing the sync_ch command to exhaust the available memory. The main change here is to call `django.db.reset_queries()` to clear that history after each batch is inserted in the sync_ch command.

There's also more logging added to give more visibility of what the command is doing, and changes to the local logging config so that standard Python logging works more seamlessly.


datahub/company/management/commands/sync_ch.py
```diff
@@ -112,20 +127,22 @@ def truncate_ch_companies_table():
...
+
 class Command(BaseCommand):
     """Companies House sync command."""
 
     def handle(self, *args, **options):
         """Handle."""
         try:
             sync_ch(tmp_file_creator=tempfile.TemporaryFile, truncate_first=True)
-        except Exception as e:
-            with log_and_ignore_exceptions():
```
canni: Why remove that? If sentry is down (and this happened a couple of times), the whole process will blow up without this.

reupen: So what? It's already blown-up by this point and the exception chain will be there.

The old code was broken because `self.stderr.write(e)` takes a string and not an exception. Even if you fixed it, it would mess up the exit code because it was swallowing up any errors.

reupen:
The explicit capturing of CLI command exceptions for Sentry is no longer needed, because they fixed [the bug stopping that happening automatically in Raven 6.1](https://github.com/getsentry/raven-python/issues/884).

PR #274, opened 2017-06-23
Merge ES_HOST and ES_PORT environment variables into ES_URL

reupen:
This fixes the following warning:
```
data-hub-leeloo/datahub/search/elasticsearch.py:16: DeprecationWarning: invalid escape sequence \:
  auth = re.search('https\:\/\/(.*)\@', bonsai).group(1).split(':')
<unknown>:16: DeprecationWarning: invalid escape sequence \:
```

It should've been a raw string. but the fix here uses urlparse() to parse the URL instead of a regex.

We could also just switch to using a proper URL instead of separate ES_HOST and ES_PORT variables? The elasticsearch library supports them.


canni:
I'm in favour of using fully-qualified URL the less ENV variables to maintain the better

reupen:
OK, let's do that.

PR #275, opened 2017-06-23
AV Scan functionality

canni:


datahub/documents/av_scan.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _process_document(doc_pk: str):
+    if not settings.AV_SERVICE_URL:
+        logger.warning(f'Ignoring AV scan of file=UUID={doc_pk}; '
+                       'No AV service URL configured')
+        return
+
+    doc = Document.objects.get(pk=doc_pk)
+    doc.scan_initiated_on = now()
```
canni: I was thinking about this, we might need some kind of locking around this, for scenario of re-plying the success callback REST API call. It's a bit of DOS attack surface here

reupen: It would be good to stop a file being scanned if it is already being scanned or is in the queue to be scanned.

PR #279, opened 2017-06-26
Improve search query.

elcct:
In the previous query, results had too narrow difference in the score that resulted sometimes in having order that is not desired. Proposed query sets the scores further apart by favouring matches that have matching words closer to each other (using `slop`). This should fix tests failing at random.

### DIFF
--- a/datahub/search/elasticsearch.py
+++ b/datahub/search/elasticsearch.py
@@ -21,10 +21,10 @@ def configure_connection():
 def get_search_term_query(term):
     """Returns search term query."""
     return Q('bool', should=[
-        MatchPhrase(name={'query': term, 'boost': 2}),
-        MatchPhrase(_all={'query': term, 'boost': 1.5}),
-        Match(name={'query': term, 'boost': 1.0}),
-        Match(_all={'query': term, 'boost': 0.5}),
+        MatchPhrase(name={'query': term, 'slop': 200}),
+        Match(name={'query': term}),
+        MatchPhrase(_all={'query': term}),
+        Match(_all={'query': term}),
     ])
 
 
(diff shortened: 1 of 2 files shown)


datahub/search/elasticsearch.py
```diff
@@ -21,10 +21,10 @@ def configure_connection():
 def get_search_term_query(term):
     """Returns search term query."""
     return Q('bool', should=[
-        MatchPhrase(name={'query': term, 'boost': 2}),
-        MatchPhrase(_all={'query': term, 'boost': 1.5}),
-        Match(name={'query': term, 'boost': 1.0}),
-        Match(_all={'query': term, 'boost': 0.5}),
+        MatchPhrase(name={'query': term, 'slop': 200}),
```
reupen: Does this still prioritise the queries (without boost)?

elcct: Yes, boost actually was not making a huge difference and scores for `The Advisory` and `The Advisory Group` were similar, regardless of the boost value. Best results I got with adding `slop` parameter. For example: 

Without `slop`:
```
(‘The Advisory’, 4.6572275), (‘The Advisory Group’, 3.8401701)
```
With `slop` 200:
```
(‘The Advisory’, 5.492494), (‘The Advisory Group’, 3.8421867)
```

```
Whereas a phrase query simply excludes documents that don’t contain the exact query phrase, a proximity query—a phrase query where slop is greater than 0—incorporates the proximity of the query terms into the final relevance _score. By setting a high slop value like 50 or 100, you can exclude documents in which the words are really too far apart, but give a higher score to documents in which the words are closer together.
```

With `slop` 100 or lower the scores for `The Advisory` and `The Advisory Group` were almost identical - which could affect the order at random.

PR #281, opened 2017-07-03
Update and expand database fixtures

reupen:


elcct:
Is it possible to have a bit more records of each entity (so you could see paging etc.) or is it worth having a command that would generate arbitrary number of test records?

reupen:
> Is it possible to have a bit more records of each entity (so you could see paging etc.)
Yes, we do need more, but this is just a starting point.

> is it worth having a command that would generate arbitrary number of test records?
They're for use with the E2E tests and I don't think we have a way to plug the two together if we did that.

PR #286, opened 2017-07-05
Add validation for 'Verify win' stage of investment projects

reupen:
This also changes how the validation works so it's simple and more maintainable.


datahub/core/utils.py
```diff
@@ -11,6 +11,11 @@
 logger = getLogger(__name__)
 
 
+def is_falsey(val):
+    """Returns True if bool(val) is False"""
+    return not val
```
canni: RLY, why not a plain lambda? ;]

reupen: DRY? :wink:

canni: Then maybe run everywhere `s/if not (.+):/if is_falsey($1)/` ;]

reupen: operator.not_() did the same thing, so it's gone now.

datahub/investment/validate.py
```diff
@@ -52,11 +52,11 @@
     'non_fdi_type':
         CondValRule('investment_type', InvestmentType.non_fdi.value.id, Phase.prospect.value),
     'total_investment':
-        CondValRule('client_cannot_provide_total_investment', is_falsey, Phase.assign_pm.value),
+        CondValRule('client_cannot_provide_total_investment', not_, Phase.assign_pm.value),
```
canni: :+1:

datahub/investment/test/test_validate.py
```diff
@@ -191,8 +204,96 @@ def test_validate_team_instance_success():
...
+        'r_and_d_budget': 'This field is required.',
+        'non_fdi_r_and_d_budget': 'This field is required.',
+        'new_tech_to_uk': 'This field is required.',
+        'export_revenue': 'This field is required.',
+        'address_line_1': 'This field is required.',
+        'address_line_2': 'This field is required.',
+        'address_line_postcode': 'This field is required.',
+        'average_salary': 'This field is required.',
+        'client_cannot_provide_foreign_investment': 'This field is required.',
+        'foreign_equity_investment': 'This field is required.',
```
elcct: Do we need to repeat `This field is required.` like that?

reupen: Not sure what you mean – what is your suggestion?

elcct: That way it doesn't immediately tells that the responses are identical. I would either use constant `REQUIRED_MESSAGE` or create a variable `this_field_is_required` and assert against it. 
For example:
```
...
'address_line_postcode': REQUIRED_MESSAGE,
'average_salary': REQUIRED_MESSAGE,
...
```

reupen: It doesn't make much difference to me, but I kinda like seeing the output as is in tests.

elcct: To me it hints that maybe one of the responses is different than the others in a subtle way (for example missing dot or a typo), which you can't tell without looking more closely.

PR #289, opened 2017-07-07
Split company models

canni:


datahub/company/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .adviser import Advisor  # noqa: F401
+from .company import CompaniesHouseCompany, Company  # noqa: F401
+from .contact import Contact  # noqa: F401
```
reupen: Setting __all__ should get rid of F401 I think.

canni: Yeah, forgot about that :+1:

datahub/company/models/__init__.py
```diff
@@ -1,3 +1,5 @@
-from .adviser import Advisor  # noqa: F401
-from .company import CompaniesHouseCompany, Company  # noqa: F401
-from .contact import Contact  # noqa: F401
+from .adviser import Advisor
+from .company import CompaniesHouseCompany, Company
+from .contact import Contact
+
+__all__ = (Advisor, CompaniesHouseCompany, Company, Contact)
```
reupen: Should be strings in `__all__`.

PR #290, opened 2017-07-10
Add investment likelihood of landing and priority

reupen:


datahub/investment/test/test_views.py
```diff
@@ -804,6 +809,48 @@ def test_patch_project_conditional_failure(self):
             'fdi_type': ['This field is required.']
         }
 
+    def test_patch_likelihood_of_landing_too_low(self):
+        """Test updating a project a likelihood_of_landing above 100."""
```
elcct: I think comment should be like:
```
"""Test updating a project a likelihood_of_landing equal or above 0."""
```

datahub/investment/models.py
```diff
@@ -14,6 +14,12 @@
 
 MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
 
+PRIORITIES = Choices(
```
canni: SHouldn't that be placed on a Model class?

datahub/investment/test/test_views.py
reupen: Fixed :+1:

datahub/investment/models.py
reupen: OK, I've moved them.

datahub/investment/test/test_views.py
elcct: That should be `below 0` actually :) sorry

PR #291, opened 2017-07-10
Improved search results.

elcct:
This fixes problem with non deterministic results whilst testing by setting number of shards to 1 for the test.

It uses alias field for names, which has lowercase keyword analyser mapped to it. This field is used in the term query to boost results of exact matches.


datahub/search/models.py
```diff
@@ -307,7 +310,8 @@ class InvestmentProject(DocType, MapDBModelToDict):
         'id': String(index='not_analyzed'),
         'name': String()
     })  # InvestmentType
-    name = String()
+    name = String(copy_to='name_keyword')
+    name_keyword = String(analyzer='lowercase_keyword_analyzer')
     description = String()
```
reupen: Can we remove the duplicate description attribute?

elcct: Nice catch :) Done.

PR #294, opened 2017-07-11
Fixes serialization of two Adviser fields in Investment search.

elcct:


datahub/search/models.py
```diff
@@ -271,10 +271,16 @@ class InvestmentProject(DocType, MapDBModelToDict):
         'id': String(index='not_analyzed'),
         'name': String()}
     )  # Adviser
-    project_manager = Nested(properties={
-        'id': String(index='not_analyzed'),
-        'name': String()}
-    )  # Adviser
+    project_manager = Nested(properties={'id': String(index='not_analyzed'),
```
canni: Can we re-use properties dict here?

elcct: Perhaps it could be generalised - there would be some fun with `copy_to` parameters.

elcct: @canni @reupen done.

PR #295, opened 2017-07-11
Add investment team members endpoints

reupen:
Also removes the reverse side of the archived_by relationship.


datahub/investment/urls.py
```diff
@@ -4,7 +4,8 @@
 
 from datahub.investment.views import (
     IProjectAuditViewSet, IProjectDocumentViewSet, IProjectRequirementsViewSet,
-    IProjectTeamViewSet, IProjectUnifiedViewSet, IProjectValueViewSet, IProjectViewSet
+    IProjectTeamMembersViewSet, IProjectTeamViewSet, IProjectUnifiedViewSet, IProjectValueViewSet,
```
canni: This file is getting a bit convoluted, maybe a split?

reupen: Will be removing the old project, value, requirements and team endpoints once a new front-end release is done, so it'll be a lot tidier after that.

PR #301, opened 2017-07-14
Add OMIS Order model and create endpoint

marcofucci:
This:

1. creates an Order model with basic fields
2. adds a create endpoint with url `/v3/omis/order` and response
```
{
    'id': 'aa6998d6-ab5a-4ba8-8090-9b663813d0d2',
    'reference': 'TTU169/17',
    'company': {
        'name': 'My Corp',
        'id': 'd98da59a-84fd-4e08-8ffb-c702d4706548'
    },
    'contact': {
        'name': 'John Doe',
        'id': 'f56b4017-2132-4c98-9356-b3859f409490'
    },
    'primary_market': {
        'name': 'France',
        'id': '82756b9a-5d95-e211-a939-e4115bead28a'
    }
}
```


datahub/omis/order/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+        migrations.CreateModel(
+            name='Order',
+            fields=[
+                ('created_on', models.DateTimeField(auto_now_add=True, null=True)),
+                ('modified_on', models.DateTimeField(auto_now=True, null=True)),
+                ('id', models.UUIDField(db_index=True, default=uuid.uuid4, primary_key=True, serialize=False)),
+                ('reference', models.CharField(max_length=100)),
+                ('company', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='orders', to='company.Company')),
+                ('contact', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='orders', to='company.Contact')),
+                ('primary_market', models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='orders', to='metadata.Country')),
```
teneightfive: Are we going to have secondary markets?

If not might be better just to call it `market`.

marcofucci: Good question.
An order might have more Post Advisors collaborating so there could be many markets involved.

However, there's one primary market / Post Team somewhat responsible for the order.

teneightfive: I thought we agreed to not support this for the November replacement?

The idea was to keep orders simple and structured around 1 market with 1 post working on it.

Supporting collaboration brings a whole lot of work I don't think we have the ability to do before November.

marcofucci: The collaborative aspect is necessary in OMIS but it's not mandatory for November as you said (although the wireframes at the moment include that).

The data structure / API have to be in place for the migration, for old orders and for the ones in the future. The frontend can be very simple though and ignore most of this complexity.

datahub/omis/order/urls.py
```diff
@@ -0,0 +1,16 @@
+"""Company views URL config."""
+
+from django.conf.urls import url
+
+from .views import OrderViewSet
+
+# CONTACT
```
reupen: Looks like this comment needs updating...

datahub/omis/order/models.py
```diff
@@ -0,0 +1,74 @@
...
+            <(3) letters><(3) numbers>/<year> e.g. GEA962/16
+        or RuntimeError if no reference can be generated.
+        """
+        year_suffix = now().strftime('%y')
+        manager = self.__class__.objects
+
+        max_retries = 10
+        tries = 0
+        while tries < max_retries:
+            reference = '{letters}{numbers}/{year}'.format(
```
marcofucci: I do like f-strings and I used them as often as I can but I believe `format` sometimes makes more sense.

In this case, I don't think it much makes sense to store the values in variables just to use it in the f-string and nowhere else.
`letters` and `numbers` are really just params.

datahub/omis/order/test/factories.py
```diff
@@ -0,0 +1,20 @@
...
+import factory
+
+from datahub.company.test.factories import CompanyFactory, ContactFactory
+from datahub.core.constants import Country
+
+
+class OrderFactory(factory.django.DjangoModelFactory):
+    """Order factory."""
+
+    id = factory.Sequence(lambda _: str(uuid.uuid4()))
```
marcofucci: 👍  a lot better

datahub/omis/order/urls.py
marcofucci: 👍

datahub/omis/order/models.py
reupen: I don't know; the format() version is already five lines but it's not a big deal anyway.

PR #307, opened 2017-07-17
Only validate modified investment project fields when stage unchanged

reupen:
Simple change to avoid a project being unable to be updated via the front end if invalid data gets into the system (e.g. via migration or the validation rules changing)


reupen:
Just need to tweak something.

PR #308, opened 2017-07-17
Adds sort feature.

elcct:


datahub/search/elasticsearch.py
```diff
@@ -46,7 +46,28 @@ def get_search_term_query(term):
...
+    }
+    return name_map.get(field, field)
+
+
+def get_sort_query(qs, field_order=None):
+    """Attaches sort query."""
+    if field_order is None:
+        return qs
+
+    tokens = field_order.split(':')
```
reupen: I think str.rparitition() or str.rsplit(maxsplit=1) would be better here.

datahub/search/views.py
```diff
@@ -145,13 +151,15 @@ def post(self, request, format=None):
...
         limit = int(request.data.get('limit', 100))
 
         results = elasticsearch.get_search_investment_project_query(
             term='',
             filters=filters,
             ranges=ranges,
+            field_order=sortby,
             offset=offset,
             limit=limit,
         ).execute()
```
reupen: Should check the response status code in these tests.

datahub/search/elasticsearch.py
elcct: Good idea. Added rsplit.

datahub/search/views.py
elcct: Indeed, somehow missed that. Thanks!

elcct:
@reupen we have not defined which fields we want to sort by, writing validation for all fields seems disproportionate. I was thinking about adding validation once we get better idea what is really needed.

reupen:
It shouldn't blow up if you pass an invalid field. It needs to at least check the field is one that has been defined.

PR #316, opened 2017-07-19
Use max line length of 99

reupen:


canni:
BTW can we add that as a warning to pep8/flake8?

reupen:
> BTW can we add that as a warning to pep8/flake8?

max-line-length is set in the Flake8 config to 99 (was 119)

canni:
:+1:

PR #320, opened 2017-07-20
[DH-392] Add FDI value field to investment projects

reupen:


reupen:
@elcct @canni Any comments?

PR #322, opened 2017-07-21
Improve performance of adviser endpoint

reupen:
Add index on (first_name, last_name), select dit_team in the same query and exclude a few fields that were killing performance and probably shouldn't have been there.


datahub/company/serializers.py
```diff
@@ -49,7 +49,8 @@ class AdviserSerializer(serializers.ModelSerializer):
 
     class Meta:  # noqa: D101
         model = Advisor
-        exclude = ('is_staff', 'is_active', 'date_joined', 'password')
+        exclude = ('is_staff', 'is_active', 'is_superuser', 'date_joined', 'password', 'groups',
```
marcofucci: Isn't it safer to whitelist the fields to expose instead of blacklist the ones we don't want?
Just a thought so.

reupen: Yes. I'll change it.

PR #330, opened 2017-07-25
Split metadata fixtures

marcofucci:
This splits the fixtures into separate groups so that they are easier to manage.
It also adds a `loadmetadata` django command that imports all of them automatically.

The `datahub_businesstypes.yaml` fixture has been deleted and incorporated into the original one.


README.md
```diff
@@ -28,7 +28,7 @@ Leeloo uses Docker compose to setup and run all the necessary components. The do
 
     ```shell
     docker-compose run leeloo python manage.py migrate
-    docker-compose run leeloo python manage.py loaddata /app/fixtures/metadata.yaml
+    docker-compose run leeloo python manage.py loadmetadata
     docker-compose run leeloo python manage.py loaddata /app/fixtures/datahub_businesstypes.yaml
```
canni: Shouldn't that line be removed also?

marcofucci: 👍

fixtures/metadata/contacts.yaml
```diff
@@ -0,0 +1,51 @@
...
+  pk: 74756b9a-5d95-e211-a939-e4115bead28a
+  fields: {name: Partner}
+- model: metadata.role
+  pk: 75756b9a-5d95-e211-a939-e4115bead28a
+  fields: {name: Student}
+- model: metadata.role
+  pk: dc02a687-5da3-e211-a646-e4115bead28a
+  fields: {name: Managing Director}
+
+# SalaryRange
```
reupen: This one isn't related to contacts; at the moment it's just used for investment projects (in reference to new jobs).

marcofucci: Interesting, should I move it to the investment one?

reupen: Yes, I think it can go there for now. We can always move it later if we need to.

marcofucci: 👌

conftest.py
```diff
@@ -9,8 +9,7 @@
 def django_db_setup(django_db_setup, django_db_blocker):
     fixtures_dir = PurePath(__file__).parent / 'fixtures'
```
reupen: Can we remove `fixtures_dir` now?

datahub/core/management/commands/loadmetadata.py
```diff
@@ -0,0 +1,31 @@
...
+        It loads all metadata fixtures.
+
+        The algorithm could iterate and import all the files in the `metadata`
+        folder but some could have dependencies so it's safer to specify the
+        list manually.
+        """
+        call_command(
+            'loaddata',
+            *[
+                'fixtures/metadata/companies.yaml',
```
reupen: Better to build absolute paths here (like conftest.py was) I think (in case you want to run pytest from a different directory etc.)

marcofucci: @reupen makes sense. Have a look now and see if it looks right

conftest.py
marcofucci: 👍

marcofucci: We need to remove `conftest.py` from the ignored folders in flake8

reupen: Yep, not sure why it was put there...

PR #332, opened 2017-07-25
Corrects single filters.

elcct:


datahub/search/elasticsearch.py
```diff
@@ -110,25 +110,26 @@ def get_search_by_entity_query(term=None,
     if term != '':
         query.append(get_search_term_query(term))
 
-    query_filter = []
+    should_filter = []
```
reupen: Could a comment be added explaining why/how should and must are being used?

elcct: I have added comments.

PR #335, opened 2017-07-26
Team extended

marcofucci:
This:
- adds a new model called `TeamRole`
- adds `role`, `uk_region` and `country` fields to `Team`
- updates the fixtures to only load the teams we care about
- updates the `test_data` fixtures to use new values
- updates the model admin for Team and adds a new one for TeamRole
- updates/adds the Team/TeamRole API views

**Important**
As the team fixtures should replace the old teams, old records would have to be deleted manually using something like:
```shell
echo 'from datahub.metadata.models import Team\nTeam.objects.filter(role__isnull=True).delete()' | ./manage.py shell
```

Things that might be worth discussing:
1. why the metadata endpoints are not under `v3`?
2. should we paginate the metadata? At the moment we are not.
3. I don't know if I like the dynamic view and tests for teams as the model has now grown.


canni:
1) No reason (besides B/C)
2) Metadata isn't paginated because FE caches it, any change to metadata data requires node restart, so pagination wasn't necessary
3) Improve it :P

PR #338, opened 2017-07-27
Update fixtures from prod values

marcofucci:
This updates the fixtures from the prod env:
- sectors: new records have been added
- services: new records have been added
- interactions: the following interaction types have been deleted:
	- Fax
	- Telex
	- UKTI Website
	- SMS
	- Business Card

Please check the files as I'm not sure if this is what we want.
Also, as some interaction types have been deleted, you'll need to delete old records:

```
echo "from datahub.metadata.models import InteractionType; InteractionType.objects.filter(id__in=('71c226d7-5d95-e211-a939-e4115bead28a', '73c226d7-5d95-e211-a939-e4115bead28a', '75c226d7-5d95-e211-a939-e4115bead28a', 'a4d71fdd-5d95-e211-a939-e4115bead28a', 'a6d71fdd-5d95-e211-a939-e4115bead28a')).delete()" | ./manage.py shell
```

### DIFF
--- a/datahub/core/constants.py
+++ b/datahub/core/constants.py
@@ -310,16 +310,11 @@ class Country(Enum):
 class InteractionType(Enum):
     """Interaction types."""
 
-    business_card = Constant('Business Card', 'a6d71fdd-5d95-e211-a939-e4115bead28a')
     email_website = Constant('Email/Website', '70c226d7-5d95-e211-a939-e4115bead28a')
     face_to_face = Constant('Face to Face', 'a5d71fdd-5d95-e211-a939-e4115bead28a')
-    fax = Constant('Fax', '71c226d7-5d95-e211-a939-e4115bead28a')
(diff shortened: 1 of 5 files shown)


reupen:
Looks like a couple of the tests need updating.

marcofucci:
@jim68000 can you have a look please?

jim68000:
Data looks good

yadmaster:
The values that have been removed are used by data in the migration. These will now fail to migrate. We need some kind of change control around LOVs as they are as much a part of the target structure as the field definitions. Once we have mapped corresponding values they cannot simply be removed without an impact analysis, The fact that a value may not be required going forwards does not mean it is not required by historic data.

marcofucci:
@yadmaster I completely agree. 
This was just to align what we have in production at the moment with our code and it was not part of my work.

Could you please double check what interaction types are needed?

PR #342, opened 2017-07-28
Add endpoint to update order details

marcofucci:
This allows clients to PATCH the contact field of existing orders.
Company and primary market cannot be changed after creation.

PATCH `v3/omis/order/<order-id>` - [docs](https://documenter.getpostman.com/view/2403194/omis/6fZxPC3#63fb90aa-6499-6eaf-7e88-a8350216d4ed).

### DIFF
--- a/datahub/omis/order/serializers.py
+++ b/datahub/omis/order/serializers.py
@@ -30,10 +30,29 @@ class OrderSerializer(serializers.ModelSerializer):
         """
         Extra check that a contact works at the given company.
         """
-        if data['contact'].company != data['company']:
+        # get the data from `data` or fallback to the instance data in case of PATCH
+        data_to_validate = dict(data)
+        if self.partial:
+            for field in self.get_fields().keys() - data.keys():
+                data_to_validate[field] = getattr(self.instance, field)
+
+        # check that contact works at company
+        if data_to_validate['contact'].company != data_to_validate['company']:
             raise serializers.ValidationError({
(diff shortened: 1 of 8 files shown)


datahub/omis/order/serializers.py
```diff
@@ -30,11 +30,30 @@ def validate(self, data):
         """
         Extra check that a contact works at the given company.
         """
-        if data['contact'].company != data['company']:
+        # get the data from `data` or fallback to the instance data in case of PATCH
+        data_to_validate = dict(data)
```
reupen: How about `data_to_validate = data.copy()`? Just find it a clearer indication that it's only creating a (shallow) copy and not (intentionally) converting types.

Also could probably use `datahub.core.validate_utils.UpdatedDataView` (up to you).

marcofucci: The `datahub.core.validate_utils.UpdatedDataView` is very nice but can we change the name? It's not really a view isn't it?

reupen: Yep, go ahead :+1:

marcofucci: How about `UpdatedDataResolver` for the class and `data_resolver` for instance vars?

reupen: I don't mind. View came from the database sense, but I agree that's confusing in a Django app. Some other options could be DataCombiner, DataCoalescer.

PR #344, opened 2017-07-30
Fixes sorting by nested field.

elcct:


datahub/search/test/test_views.py
```diff
@@ -317,8 +316,47 @@ def test_search_sort_asc(self):
 
     @mock.patch('datahub.core.utils.executor.submit', synchronous_executor_submit)
     @mock.patch('django.db.transaction.on_commit', synchronous_transaction_on_commit)
+    def test_search_sort_nested_desc(self):
+        """Tests sorting by nested field."""
+        InvestmentProjectFactory(
+            name='Potato 1',
+            stage_id=constants.InvestmentProjectStage.active.value.id,
+        ).save()
```
reupen: I don't think the calls to save() are needed, factory boy normally does that by default.

datahub/search/elasticsearch.py
```diff
@@ -62,10 +62,18 @@ def get_sort_query(qs, field_order=None):
...
 
+    sort_params = {
+        'order': order,
+        'missing': '_first' if order == 'asc' else '_last'
+    }
+
+    # check if we sort by field in nested document (example: 'stage.name')
+    if '.' in tokens[0]:
+        # extract and add path to nested document (example: 'stage')
+        sort_params['nested_path'] = tokens[0].split('.')[0]
```
reupen: `maxsplit`?

datahub/search/test/test_views.py
elcct: Indeed, I have removed them now. Thanks!

datahub/search/elasticsearch.py
elcct: Good idea, added.

PR #348, opened 2017-07-31
Add created_by and modified_by to OMIS order

marcofucci:
This:
- adds an `order.created_by` field
- adds an `order.modified_by` field
- sets those fields to the logged in user during API calls
- includes those fields in all responses
- includes `created_on` and `modified_on` in all responses


reupen:
Should we just add them to all models that derive from `BaseModel` and put the logic to set them in CoreViewSetV1/CoreViewSetV3?

marcofucci:
Could do, do you think it's a good idea?

The model affected would be:
- Company
- CompaniesHouseCompany
- Document
- Interaction
- ServiceDelivery
- InvestmentProject
- IProjectDocument
- BusinessLead
- Order
- OrderSubscriber (I might take this out and only make it extend models as it doesn't need created and modified by).

reupen:
> Could do, do you think it's a good idea?

Yep, I think so, because a couple of them already track the creator so it feels like we should just make it consistent. (CompaniesHouseCompany probably doesn't need them, but maybe we just kill the inheritance from BaseModel for that as well.)

marcofucci:
Closing this because part of a different PR #353

PR #353, opened 2017-08-01
Add created_by/modified_by fields to BaseModel

marcofucci:
This:
- removes created_on and modified on from companieshousecompany
- adds created_by and modified_by to BaseModel and therefore to:
	- company
	- contact
	- document
	- interaction
	- servicedelivery
	- investmentproject
	- iprojectdocument
	- businesslead
	- order
	- ordersubscriber
- adds logic to populate created_by and updated_by automatically in CoreViewSetV1 and CoreViewSetV3
- adds created_by and modified_by to factories
- <s>adds created_by and modified_by to ES models</s> (see comments)
- tests CoreViewSetV1 and CoreViewSetV3 in a generic way
- updates the tests to check for those fields in response data

It does NOT add created_by/modified_by to the response JSON of viewsets using serializers with explicit `fields` set.

There's a bit of repetition in `CoreViewSetV1`/ `CoreViewSetV3` and in the related tests but if we start abstracting that out we would never get rid of `CoreViewSetV1` (which by the way we should try to remove sooner).

Please check if I'm missing anything as this touches many parts that I'm don't have much knowledge of.


datahub/search/test/test_views.py
```diff
@@ -256,10 +256,13 @@ def test_search_results_quality(self):
 
         assert response.status_code == status.HTTP_200_OK
         assert response.data['count'] == 4
-        assert ['The Advisory',
-                'The Advisory Group',
-                'The Risk Advisory Group',
-                'The Advisories'] == [company['name'] for company in response.data['companies']]
+        # no sortby specified so no ordering applied
```
elcct: @marcofucci If `sortby` is not specified then `_score` is being used to order the results (from most relevant to the least). This is testing that the results are returned in this particular order, so I would recommend leaving the list comparison.

marcofucci: Well the test fails, meaning it's not returned in this particular order.
I'm not sure if it depends on the fields I've added, if so it seems a bit weird.

marcofucci: @elcct (name, score):
```
[
    ('The Advisory', 5.8421426), 
    ('The Risk Advisory Group', 1.1033473),
    ('The Advisory Group', 1.1033473), 
    ('The Advisories', 0.31485617)
]
```

it looks like _The Risk Advisory Group_ and _The Advisory Group_ have the same score

marcofucci: @elcct I changed the test, have a look

elcct: I am getting following scores on develop:

```
[
    ('The Advisory', 6.7450495), 
    ('The Advisory Group', 1.3361949), 
    ('The Risk Advisory Group', 1.30279), 
    ('The Advisories', 0.38268062)
]
```

But since you have not changed the query, it looks like this test is too fragile, I agree with dropping the check for the order. This should be revisited in the next iteration of search results improvement.

reupen: It does seem weird that this change affects the ranking. Another option is to just exclude created_by and modified_by from the search index for now until we understand it better...

@elcct What do you think?

elcct: @marcofucci Sorry I didn't see your comment about changed test :) It's alright

elcct: @reupen that actually would be ideal - I don't think front end is using those fields for search, so it makes sense to just not include them by adding to `IGNORED_FIELDS`. What do you think @marcofucci ?

marcofucci: works for me

marcofucci: @elcct @reupen have another quick look. I've remove those fields from all search models.

reupen: Looks good!

PR #356, opened 2017-08-02
Fixes Integrity error when Primary is left out in the form.

elcct:


marcofucci:
@elcct @reupen can't we have default False instead of `NullBooleanField`?
Or we want to use `null` to state that the field has not been set yet?

elcct:
@reupen @marcofucci I assumed that front end would do validation of data before posting and since it was allowed to be posted without making selection I thought it is optional also someone in the meeting raised the need of null value for this field. If it is not the case, then indeed this fix is not correct.

reupen:
Generally the front end leaves the validation to the back end. The field isn't marked as optional in the front end at the moment, is it for the data migration it needs to be nullable?

elcct:
@reupen Just confirmed that the field is required, so I will work on the validation

reupen:
There was a similar fix needed here:

https://github.com/uktrade/data-hub-leeloo/blob/ee30134c7c4adc0cde1d318537776214a259cdbf/datahub/investment/serializers.py#L129-L132

It is probably affecting all (non-nullable) boolean fields though...

PR #367, opened 2017-08-03
Turn off W503 Flake8/pycodestyle rule

reupen:
It's in the default pycodestyle ignore list^, and is not the style recommended by PEP 8^^. 

Because we were overriding the ignore list, we were re-enabling all the checks ignored by default.

Unfortunately, there is no corresponding check for the inverse style.

^ https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
^^ https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator


elcct:
Only negative thing I see is that this way you can accidentally delete a line and it more likely go unnoticed with operator in front, but this could be sacrificed for being easier to read.

PR #368, opened 2017-08-04
Make company country and sector required again

reupen:
Fixes a couple of regressions when undefined was removed (https://github.com/uktrade/data-hub-leeloo/commit/04743bc8c75d6fb4643fddedadda3ddf78fc4a50).


marcofucci:
I couple of points:

1. There will be some records with null as sector in the resulting migrated db. Isn't it weird that when loading one of them and saving it without changing anything the system gives you a validation error because now you need to specify a sector?

2. Why do we still have `CompanyViewSetV1`?

elcct:
@marcofucci if sector is required, then why we are going to have null as sector in the migrated db? I mean if we require sector, then migrated data without it would be invalid, no?

marcofucci:
`Company.sector` is not mandatory so we are requiring it to be mandatory only via API.
There are currently 20K records in the CDMS db with sector == null.
Unless we know how to fix those, we will have null values in the migrated db.

reupen:
Sector is already mandatory in the CDMS UI. The organisations with a null sector in the CDMS database might be old records. Do you know if they're active or inactive organisations in CDMS?

reupen:
OK, I had a look in the CDMS UI and a good proportion of them are active. 

I'll have a look at writing a custom validator so it doesn't complain in that case. But this is a bit more of a general problem, and also there may be cases where you do want the user to go and fill in missing information.

marcofucci:
@reupen yes I agree completely. It's a more general problem which we might have ignored until now and which might be a bit hard to solve properly.
We have the same problem in OMIS where we have null values when we really want non-null ones.

PR #369, opened 2017-08-04
Add OrderAssignee and related endpoints

marcofucci:
This:
- adds a new `OrderAssignee` model
- adds 2 new endpoints:

`GET /v3/omis/order/<order-id>/assignee`

Returns the list of advisers assigned to an order.

`PATCH /v3/omis/order/<order-id>/assignee`

(Partially) updates the advisers assigned to an order.
By default it keeps the assignees not included in the request body but you can pass the query string `force-delete=1` and it will delete them.

There's a bit of noise because I split the `test_views` into different modules.

[Docs](https://documenter.getpostman.com/view/2403194/omis/6fZxPC3#454e3dd8-3bae-0e73-20b9-0742580a1515)


datahub/metadata/test/factories.py
```diff
@@ -28,7 +28,7 @@ class Meta:  # noqa: D101
 class TeamRoleFactory(factory.django.DjangoModelFactory):
     """TeamRole factory."""
 
-    id = factory.LazyFunction(lambda: str(uuid.uuid4()))
+    id = factory.LazyFunction(lambda: uuid.uuid4())
     name = factory.Sequence(lambda n: f'name {n}')
```
reupen: How about `id = factory.LazyFunction(uuid.uuid4)`?

datahub/omis/order/models.py
```diff
@@ -98,6 +98,62 @@ def __str__(self):
         return f'{self.order} – {self.adviser}'
 
     class Meta:  # noqa: D101
+        ordering = ['created_on']
```
reupen: Do we need an index on created_on?

datahub/metadata/test/factories.py
marcofucci: very nice

datahub/omis/order/models.py
marcofucci: probably

datahub/metadata/test/factories.py
marcofucci: I might do it in a different PR if you don't mind, I would like to get rid of the `str` everywhere if I manage to.

datahub/omis/order/models.py
marcofucci: @reupen I've added an index to the BaseModel and therefore all the subclasses. See if it makes sense.

datahub/metadata/test/factories.py
reupen: Yeah not everywhere, but the ones being updated already in this PR could be updated...

datahub/omis/order/models.py
reupen: Hmm, it's an interesting question about whether we want it on all subclasses of BaseModel. Could be argued either way, so I think it's fine. If we do want it on just a couple of the models though, we should be able to do it by setting `indexes` in the Meta class...

marcofucci: yeah, I think that in general you might want to order by `created_on` quite often as it's a very useful field to base your ordering logic on but that might not be always be the case.

Do you think that having it always on could cause any problems? If not, I would probably leave it everywhere.

datahub/omis/order/models.py
```diff
@@ -98,6 +98,62 @@ def __str__(self):
...
+        future-proofing reasons, that is, if an adviser moves to another team in the future
+        we don't want to change history.
+        """
+        if not self._state.adding and self.__adviser != self.adviser:
+            raise ValueError('Updating the value of adviser isn\'t allowed.')
+
+        if self._state.adding:
+            self.team = self.adviser.dit_team
+            if self.team:
+                self.country = self.team.country
```
reupen: Should it set `self.country` to None if the adviser doesn't have a team?

marcofucci: Wouldn't it be None already by default? `country` is set only when adding and when not None and it never change again.

PR #370, opened 2017-08-07
Support of partial matching, remove _all query, hyphens.

elcct:
It adds partial matching using trigrams, filters "-" to be able to match t-shirt and tshirt, removes _all field and uses _combined_fields instead, where we can include fields we want to search by. It also does exact matching of ids. It doesn't change existing tests.

### DIFF
--- a/datahub/search/elasticsearch.py
+++ b/datahub/search/elasticsearch.py
@@ -64,10 +64,7 @@ def get_search_term_query(term, fields=None):
     ]
 
     if fields:
-        for field in fields:
-            should_query.append(
-                get_match_query(field, term)
-            )
+        should_query.extend([get_match_query(field, term) for field in fields])
 
     return Q('bool', should=should_query)
 
@@ -192,10 +189,7 @@ def get_search_by_entity_query(term=None,
                 must_filter.append(get_term_query(k, v))
 
     if ranges:
-        for k, v in ranges.items():
-            must_filter.append(
-                Q('range', **{k: v})
-            )
+        must_filter.extend([Q('range', **{k: v}) for k, v in ranges.items()])
 
     s = Search(index=settings.ES_INDEX).query('bool', must=query)
     s = get_sort_query(s, field_order=field_order)
--- a/datahub/search/elasticsearch.py
+++ b/datahub/search/elasticsearch.py
@@ -47,12 +47,12 @@ def configure_index(index_name, settings=None):
         index.create()
 
 
-def get_search_term_query(term):
+def get_search_term_query(term, fields=None):
     """Returns search term query."""
     if term == '':
         return Q('match_all')
 
-    return Q('bool', should=[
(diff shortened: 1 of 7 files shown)


datahub/search/test/test_views.py
```diff
@@ -266,6 +267,89 @@ def test_search_results_quality(self):
...
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == 4
+
+        # order of relevance
+        assert [
+            't-shirt',
+            'tshirt',
+            'electronic shirt',
+            't and e and a'
```
marcofucci: Do we want these 2? Is splitting by '-' and matching the individual terms any useful?

elcct: As long as the most relevant results are on the top, then I don't see an issue. This could be improved in further iterations if the need arise.

datahub/search/elasticsearch.py
```diff
@@ -40,9 +53,16 @@ def get_search_term_query(term):
...
         MatchPhrase(name_keyword={'query': term, 'boost': 2}),
+        # Exact match by id
+        MatchPhrase(id={'query': term}),
+        # Match similar name
         Match(name={'query': term}),
-        Match(_all={'query': term}),
+        # Partial match name
+        MatchPhrase(name_trigram={'query': term}),
+        # Match in filter fields
+        Match(_combined_fields={'query': term}),
```
reupen: Could we use `MultiMatch` instead of copying things to `_combined_fields`?

elcct: You can't really use `MultiMatch` with nested documents. It would make it more complex, as you would have to build queries for each nested document. The documentation states that it is better for computing relevance if you search in individual fields. I'll give it another go and see if I could replace that. Building a query has an advantage that we don't need to rebuild the index if we want to add a new field into consideration for search.

reupen: Would using flat structures help?

elcct: I've been thinking about that, but we have some one to many relations, so I have to build a query anyway (the query for one to one and one to many is the same). I think it will be simpler without flattening at this moment.

datahub/search/elasticsearch.py
```diff
@@ -29,21 +41,35 @@ def configure_index(index_name, settings=None):
...
+        # Match similar name
         Match(name={'query': term}),
-        Match(_all={'query': term}),
-    ])
+        # Partial match name
+        MatchPhrase(name_trigram={'query': term}),
+    ]
+
+    if fields:
+        for field in fields:
```
reupen: Could use list.extend() e.g. `should_query.extend(get_match_query(field, term) for field in fields)`

elcct: That indeed doesn't look right. I have fixed it. Thanks!

PR #371, opened 2017-08-07
Fixes grouping of multiple filters.

elcct:
For query `a.x, a.y, b.x, b.y, c.x` we created filter like `c.x and (a.x or a.y or b.x or b.y)` which is not what we want. This changes it so the filter query becomes `(a.x or a.y) and (b.x or b.y) and c.x`

### DIFF
--- a/datahub/search/test/test_views.py
+++ b/datahub/search/test/test_views.py
@@ -485,8 +485,8 @@ class TestSearch(APITestMixin):
 
         # checks if we only have investment projects with investor companies we filtered
         assert {
-            investment_project1.investor_company.pk,
-            investment_project2.investor_company.pk
+            str(investment_project1.investor_company.pk),
+            str(investment_project2.investor_company.pk)
         } == {
             investment_project['investor_company']['id']
             for investment_project in response.data['results']
--- a/datahub/search/test/test_views.py
+++ b/datahub/search/test/test_views.py
@@ -430,6 +430,12 @@ class TestSearch(APITestMixin):
 
         We make sure that out of provided investment projects, we will
         receive only those that match our filter.
+
+        We are testing following filter:
+
+        investment_type = fdi
+        AND (investor_company = compA OR investor_company = compB)
+        AND (stage = won OR stage = active)
         """
         url = reverse('api-v3:search:investment_project')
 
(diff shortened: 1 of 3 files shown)


datahub/search/test/test_views.py
```diff
@@ -447,6 +447,10 @@ def test_search_investment_project_multiple_filters(self):
             stage_id=constants.InvestmentProjectStage.prospect.value.id,
         )
         InvestmentProjectFactory(
+            investor_company=CompanyFactory(),
```
marcofucci: Did you mean `company_a`? Adding These lines like this don't change much, right?

Also, would it be better to save:
```
        InvestmentProjectFactory(
            investment_type_id=constants.InvestmentType.fdi.value.id,
            investor_company=company_a,
            sector_id=constants.Sector.aerospace_assembly_aircraft.value.id,
            stage_id=constants.InvestmentProjectStage.active.value.id
        )
``` 
to a var and assert its id instead of the stages?

Also, could you describe in the docstring what the search term is and the expected result and why it's the only one returned? It would help me next time as I don't know much about the search code.

elcct: These are good points, I should have spent more time on this test. I will update soon. Thanks!

elcct: @marcofucci I have updated the PR

datahub/search/test/test_views.py
```diff
@@ -465,15 +465,16 @@ def test_search_investment_project_multiple_filters(self):
         }, format='json')
 
         assert response.status_code == status.HTTP_200_OK
-        assert response.data['count'] == 1
-        assert len(response.data['results']) == 1
+        assert response.data['count'] == 2
+        assert len(response.data['results']) == 2
 
+        # checks if we only have investment projects with stages we filtered
```
marcofucci: @elcct sorry If I wasn't clear but I meant that this bit of code is not only checking that right stages are returned. 

The query is basically 
```investment_type = fdi AND (investor_company = compA OR investor_company = compB) AND (stage = won OR stage = active)``` right?

Maybe you can just assert that the `results.ids == [investment_project1.id and investment_project2.id]` 

What do you think? Also you could add the query to the docstring so that it's super clear.

elcct: I added assertions that we only get results with values we filter by and also added the query we test in the doc string.

PR #375, opened 2017-08-08
Use RequiredUnlessAlreadyBlank for a few more fields

reupen:
Use it for a few more fields that were made nullable in https://github.com/uktrade/data-hub-leeloo/pull/196:

* Company.business_type
* Interaction.dit_team
* Interaction.interaction_type
* Interaction.service


reupen:
> Although having an null interaction_type on Interaction seems weird. Maybe we could try and fill in that field somehow during the migration process? Probably not possible anyway

Yep, not ideal. I can't imagine editing old interactions would be something frequently done, so we could probably get away with making them required in the normal way in the API. (The disadvantage of using RequiredUnlessAlreadyBlank is that it only runs if all fields pass validation...)

PR #379, opened 2017-08-08
Add order service type and extend metadata logic

marcofucci:
This redesigns how we deal with metadata.

1. metadata models can be defined in specific django apps
2. a metadata registry keeps track of all the metadata
3. you need to register a metadata using a `metadata` module with
```python
>>> from datahub.metadata.registry import registry
>>> registry.register(
        id=<metadata-id>,
        model=<model>,
        queryset=<queryset>,
        serializer=<serializer>
    )
```
4. that automatically generates the right view and tests

This also adds a new OMIS ServiceType metadata which uses the new registry logic.


datahub/metadata/test/test_registry.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_register_id_already_used():
+    """
+    Tests that if I try to register the same metadata twice, the second call fails.
+    """
+    reg = MetadataRegistery()
+
+    reg.register('sector', model=Sector)
+
+    with pytest.raises(AssertionError):
```
reupen: If the behaviour's important, I'd use a proper exception rather an assertion (although we don't use it, assertions are stripped out with `python -O`).

datahub/metadata/test/test_views.py
```diff
@@ -1,115 +1,75 @@
...
+        metafunc.parametrize(
+            'metadata_view_name',
+            view_names,
+            ids=[f'{view_name} metadata view' for view_name in view_names]
+        )
+
+    if 'ordered_mapping' in metafunc.fixturenames:
+        view_data = []
+        for view_id, mapping in registry.mappings.items():
+            with suppress(FieldDoesNotExist):
```
reupen: Would we be able to use `if issubclass(mapping.model, datahub.core.models.BaseOrderedConstantModel):` here?

datahub/metadata/test/test_registry.py
```diff
@@ -0,0 +1,54 @@
...
+    """Tests the default values."""
+    reg = MetadataRegistery()
+
+    reg.register('sector', model=Sector)
+    reg.register('uk-region', model=UKRegion)
+
+    assert set(reg.mappings) == {'sector', 'uk-region'}
+
+    mapping = reg.mappings['sector']
+    assert mapping.model == Sector
```
reupen: Shouldn't that be `mapping.model is Sector`?

datahub/metadata/test/test_views.py
```diff
@@ -118,72 +78,25 @@ def test_metadata_view_patch(view_name, api_client):
...
-
-
-@pytest.mark.parametrize('view_name,expected_names',
-                         ordered_metadata_view_params,
-                         ids=ordered_metadata_view_test_ids)
-def test_ordered_metadata_order_view(view_name, expected_names, api_client):
-    """Test that turnover and no. of employee ranges are returned in order.
-
-    The response elements should be ordered according to the order column in the model.
+    assert set(pattern.name for pattern in patterns) == set(registry.mappings.keys())
```
reupen: Could be simplified a little bit:
`assert {pattern.name for pattern in patterns} == registry.mappings.keys()`

datahub/omis/order/serializers.py
```diff
@@ -3,13 +3,19 @@
...
-from datahub.core.serializers import NestedRelatedField
+from datahub.core.serializers import ConstantModelSerializer, NestedRelatedField
 from datahub.core.validate_utils import DataCombiner
 from datahub.metadata.models import Country, Sector, Team
 
 from .models import Order, OrderAssignee, OrderSubscriber
 
 
+class ServiceTypeSerializer(ConstantModelSerializer):
+    """Service Type DRF serilizer"""
```
reupen: Typo in serilizer.

marcofucci: 👍

datahub/metadata/test/test_views.py
marcofucci: 👍

datahub/metadata/test/test_views.py
marcofucci: `issubclass` is not generally pythonic and checking if an object behaves in a certain way is usually preferable.

datahub/metadata/test/test_registry.py
marcofucci: 👍

datahub/metadata/test/test_registry.py
marcofucci: 👍

datahub/metadata/registry.py
```diff
@@ -0,0 +1,50 @@
...
+        - serializer (optional): if you want to override the default one
+
+    The data registered is currently used by metadata.views to generate views automatically.
+    """
+
+    def __init__(self):
+        """Keeps a local copy of the metadata registered."""
+        self.metadata = {}
+
+    def register(self, id, model, queryset=None, serializer=ConstantModelSerializer):
```
marcofucci: very true 👍

datahub/metadata/registry.py
```diff
@@ -0,0 +1,50 @@
+from collections import namedtuple
+
+from datahub.core.serializers import ConstantModelSerializer
+
+MetadataMapping = namedtuple('MetadataMapping', ['model', 'queryset', 'serializer'])
+
+
+class MetadataRegistery:
```
marcofucci: 👍

datahub/metadata/apps.py
```diff
@@ -0,0 +1,14 @@
...
+class MetadataConfig(AppConfig):
+    """
+    Django App Config for the Metadata app.
+    """
+
+    name = 'datahub.metadata'
+
+    def ready(self):
+        """Calls the autodiscover logic after all apps are loaded."""
+        super(MetadataConfig, self).ready()
```
marcofucci: 👍

marcofucci:
@reupen thanks for the comments, very helpful. Have another look when you can.

PR #381, opened 2017-08-09
Add Order.service_types

marcofucci:
This
- adds the many2many field Order.service_types
- adds service_types to the creation endpoint
- adds service_types to the update endpoint

As a service type can be disabled at any point, the endpoint raises a validation error if the client uses a value that was disabled at the time of creation of an order.

Disabled service types can be replaced by other service types if they were not disabled when the order was first created.


marcofucci:
@reupen @elcct maybe the `disabled_on` field would actually be useful for other metadata models as well?

datahub/omis/order/test/factories.py
```diff
@@ -19,6 +21,20 @@ class OrderFactory(factory.django.DjangoModelFactory):
...
+        if not create:
+            return  # Simple build, do nothing.
+
+        if extracted:
+            service_types = extracted
+        else:
+            service_types = ServiceType.objects.filter(disabled_on__isnull=True).order_by('?')[:2]
+
+        for service_type in service_types:
+            self.service_types.add(service_type)
```
reupen: `self.service_types.set(service_types)` should work.

There is a to_many_field decorator in datahub.core.test.factories which you could consider extending.

reupen:
> @reupen @elcct maybe the disabled_on field would actually be useful for other metadata models as well?

Yes, it might be useful (particularly with the data migration) and it looks like a good way of handling this. Not sure if we should rush into adding it everywhere though, there's the front-end aspect to consider too.

datahub/omis/order/test/factories.py
marcofucci: @reupen Have a look at my last commit and see if you like how I extended the `to_many_field`.

datahub/core/test/factories.py
```diff
@@ -19,6 +19,19 @@ def to_many_field(wrapped_func):
 
     This decorator doesn't call the decorated function; the decorated function is only used
```
reupen: Comment here needs updating.

datahub/core/test/factories.py
```diff
@@ -28,8 +41,13 @@ def wrapping_func(self, create, extracted, **kwargs):
             # should not be set
             return
 
+        field = getattr(self, wrapped_func.__name__)
         if extracted:
```
reupen: Should it be `if extracted is not None:`? So you can still pass an empty list etc.?

datahub/omis/order/test/factories.py
```diff
@@ -19,6 +22,15 @@ class OrderFactory(factory.django.DjangoModelFactory):
...
     sector_id = Sector.aerospace_assembly_aircraft.value.id
 
+    @to_many_field
+    def service_types(self):
+        """
+        Add support for setting service_types.
+        If nothing specified when instantiating the object, the value returned by
+        this method will be used by default.
+        """
+        return ServiceType.objects.filter(disabled_on__isnull=True).order_by('?')[:2]
```
elcct: I thought selecting records in random order this way is a bit inefficient or is the table too small to make a difference?

datahub/core/test/factories.py
marcofucci: Yes I was thinking the same.

datahub/omis/order/test/factories.py
marcofucci: Yeah the table is very small and it's just during tests so I'm not really worried.

marcofucci:
@reupen @elcct have another quick look

datahub/omis/order/test/factories.py
elcct: Great

PR #383, opened 2017-08-10
Fixes sorting by first_name, last_name and copying of contact's name …

elcct:
…fields.

We are unable to sort by first_name and last_name - this adds that.

I also realised that `copy_to` statements in `Contact` are wrong. All fields must be copied from original field (so you can't copy last_name to name and then from name to name_trigram) otherwise they are empty.


datahub/search/elasticsearch.py
```diff
@@ -106,15 +107,17 @@ def get_basic_search_query(term, entities=None, field_order=None, offset=0, limi
 
     Also returns number of results in other entities.
     """
-    fields = set()
-    map(fields.update, [entity.SEARCH_FIELDS for entity in entities])
+    fields = set(chain.from_iterable([entity.SEARCH_FIELDS for entity in entities]))
```
reupen: Should be no need for `[` and `]` here. Can just leave it at a generator expression.

datahub/search/elasticsearch.py
```diff
@@ -106,15 +107,17 @@ def get_basic_search_query(term, entities=None, field_order=None, offset=0, limi
...
+    # we need alphabetical order of fields preserved, so the query is deterministic
+    fields = sorted(fields)
```
reupen: Do we know why the order of fields affects the results? Would be useful to elaborate.

datahub/search/views.py
```diff
@@ -127,9 +127,9 @@ class SearchContactAPIView(APIView):
     """Filtered contact search view."""
 
     SORT_BY_FIELDS = (
-        'archived', 'archived', 'created_on',
```
reupen: Can we break these up to one per line to make it easier to see what's changed?

elcct: Indeed, it is hard to read. I changed it.

datahub/search/elasticsearch.py
elcct: Good point! I keep forgetting that :/ Thanks

datahub/search/elasticsearch.py
elcct: Yes, order of queries doesn't affect the results. I sorted the fields, so that the output is the same every time, which makes it easier to compare in tests.

datahub/search/elasticsearch.py
```diff
@@ -106,15 +107,17 @@ def get_basic_search_query(term, entities=None, field_order=None, offset=0, limi
 
     Also returns number of results in other entities.
     """
-    fields = set()
-    map(fields.update, [entity.SEARCH_FIELDS for entity in entities])
+    fields = set(chain.from_iterable((entity.SEARCH_FIELDS for entity in entities)))
```
elcct: True! :) Removed.

datahub/search/elasticsearch.py
```diff
@@ -106,15 +107,17 @@ def get_basic_search_query(term, entities=None, field_order=None, offset=0, limi
 
     Also returns number of results in other entities.
     """
-    fields = set()
-    map(fields.update, [entity.SEARCH_FIELDS for entity in entities])
+    fields = set(chain.from_iterable(entity.SEARCH_FIELDS for entity in entities))
+    # we need alphabetical order of fields preserved, so the query is deterministic
```
reupen: Not sure where my comment here went, it was:

> Do we know why the order of fields affects the results? Would be useful to elaborate.

datahub/search/elasticsearch.py
reupen: OK, I think the comment could be clearer, because it sounds like it's talking about the query itself. E.g. something like:

> Sort the fields so that this function is deterministic and the same query is always generated with the same inputs

elcct: Alright, I have updated it. Thanks

PR #384, opened 2017-08-11
Add extra descriptive fields to an OMIS Order

marcofucci:
This:
- adds the fields `description` and `contacts_not_to_approach`
- adds a few other legacy readonly fields
- includes those fields in the order API endpoints


datahub/omis/order/test/factories.py
```diff
@@ -21,6 +22,12 @@ class OrderFactory(factory.django.DjangoModelFactory):
     contact = factory.LazyAttribute(lambda o: ContactFactory(company=o.company))
     primary_market_id = Country.france.value.id
     sector_id = Sector.aerospace_assembly_aircraft.value.id
+    description = factory.fuzzy.FuzzyText()
```
reupen: The [factory_boy docs aren't very clear](http://factoryboy.readthedocs.io/en/latest/fuzzy.html), but they seem to suggest using factory.Faker instead (e.g. `factory.Faker('text')`) of the fuzzers. Might be worth thinking about.

PR #389, opened 2017-08-14
More search fields, case-insensitive keyword fields.

elcct:
At the moment to find a `Contact` by country you have to type country with correct case, for example `United Kingdom`, otherwise it will not bring any results. This PR fixes that. 

It also adds more fields for each entity to be searchable.

### DIFF
--- a/datahub/search/models.py
+++ b/datahub/search/models.py
@@ -202,7 +202,7 @@ class Company(DocType, MapDBModelToDict):
         'tree_id'
     )
 
-    SEARCH_FIELDS = (
+    SEARCH_FIELDS = [
         'classification.name',
         'export_to_countries.name',
         'future_interest_countries.name',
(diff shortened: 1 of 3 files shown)


datahub/search/models.py
```diff
@@ -48,31 +48,31 @@ def _contact_mapping(field):
     return Nested(properties={'id': String(index='not_analyzed'),
                               'first_name': String(copy_to=f'{field}.name'),
                               'last_name': String(copy_to=f'{field}.name'),
-                              'name': String(index='not_analyzed'),
+                              'name': String(analyzer='lowercase_keyword_analyzer'),
```
reupen: How about defining a CaseInsensitiveString e.g. using partial()?

elcct: Good idea, it should make it more readable. Fixed.

PR #390, opened 2017-08-14
Search apps

marcofucci:
This moves the search models into their own sub modules as we discussed.

The idea is to have
`datahub.search` as parent
`datahub.search.company` for company stuff
`datahub.search.contact` for contact stuff
`datahub.search.investment` for investment stuff

Soon we would also add an extra `datahub.search.omis` for omis stuff and we wouldn't need to touch the other ones.

`datahub.search` shouldn't contain anything subapp specific which is the case at the moment apart from 2 test files:
1. `datahub.search.test.test_management_command` which should now generate search models dynamically instead of hardcoding them
2. `datahub.search.test.test_views` which should now generate search models dynamically as well for the basic search.

There's a lot going on and maybe the best way is to go through it commit by commit to understand the different steps I worked on.

I expect lots of feedback so go ahead and I'll change things.


elcct:
> I expect lots of feedback so go ahead and I'll change things.

I don't know, it looks good to me :)

PR #391, opened 2017-08-14
Basic search query fields across all entities.

elcct:
Basic search needs to look across all entities for aggregation to work correctly. This change includes all `SEARCH_FIELDS` from all entities in the basic search.

### DIFF
--- a/datahub/search/elasticsearch.py
+++ b/datahub/search/elasticsearch.py
@@ -10,6 +10,12 @@ from elasticsearch_dsl.query import Match, MatchPhrase, Q
 
 from .models import Company, Contact, InvestmentProject
 
+ALL_MODELS = (
+    Company,
+    Contact,
+    InvestmentProject,
+)
+
 lowercase_keyword_analyzer = analysis.CustomAnalyzer(
     'lowercase_keyword_analyzer',
     tokenizer='keyword',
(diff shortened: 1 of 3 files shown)


datahub/search/elasticsearch.py
```diff
@@ -107,7 +109,11 @@ def get_basic_search_query(term, entities=None, field_order=None, offset=0, limi
 
     Also returns number of results in other entities.
```
reupen: Should the entities param be just 'entity' now?

datahub/search/elasticsearch.py
```diff
@@ -107,7 +109,11 @@ def get_basic_search_query(term, entities=None, field_order=None, offset=0, limi
...
     """
-    fields = set(chain.from_iterable(entity.SEARCH_FIELDS for entity in entities))
+    fields = set(chain.from_iterable(entity.SEARCH_FIELDS for entity in (
+        Company,
```
reupen: Might be worth defining a tuple of all models in a constant (e.g. ALL_MODELS).

elcct: It still accepts multiple entities, so you could get results from both Company and Contact, but also have aggregations of all 3 entities.

datahub/search/elasticsearch.py
elcct: It still accepts multiple entities, so you could get results from both Company and Contact, but also have aggregations of all 3 entities.

datahub/search/elasticsearch.py
elcct: I added it, should be more readable now. Thanks

PR #392, opened 2017-08-15
Define search app OMIS

marcofucci:
This:
- adds a new search app called *omis*
- defines the to_dict and mappings for ES indexing

It does not define any search views yet.

### DIFF
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -28,6 +28,7 @@ def _contact_dict(obj):
         'id': str(obj.id),
         'first_name': obj.first_name,
         'last_name': obj.last_name,
+        'name': obj.name,
     }
 
 
--- a/datahub/search/apps.py
+++ b/datahub/search/apps.py
@@ -11,6 +11,7 @@ SEARCH_APPS = [
     'datahub.search.company.CompanySearchApp',
     'datahub.search.contact.ContactSearchApp',
     'datahub.search.investment.InvestmentSearchApp',
+    'datahub.search.omis.OrderSearchApp',
 ]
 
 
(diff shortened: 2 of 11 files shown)


datahub/search/dict_utils.py
```diff
@@ -28,6 +28,7 @@ def _contact_dict(obj):
         'id': str(obj.id),
         'first_name': obj.first_name,
         'last_name': obj.last_name,
+        'name': obj.name,
```
elcct: `name` is being built by Elasticsearch by having specified `copy_to` mapping parameter for `first_name` and `last_name`, so that is redundant.  When you sync the data, you only sync `first_name` and `last_name` for contact.

marcofucci: Yes but we still have to return `name` to the API to make it consistent with the other endpoints, that's why I added it.

Actually that's something I wanted to check with you, if we add `name` can we remove the `copy_to` bit?

elcct: But neither `Adviser` nor `Contact` have `name` field. `copy_to` is being used to create "a view" of concatenated `first_name` and `last_name`. If you want resulting `name` to be returned from the search, you can add `store` parameter in the mapping. But I though `name` is being build in the frontend out of `first_name` and `last_name`. Sending back all three fields feels a bit redundant.

elcct: Ok I can see that there exists `name` property on those models that concatenates `first_name` and `last_name`. In that case you can drop `copy_to` :)

marcofucci: @elcct 👍  please have a look at my last commit as you have a broader view

elcct: It appears that `dbmodel_to_dict` doesn't return `name` property, so `name` is not getting synced nor returned from the view (in case of Contact for example).

marcofucci: Ah you mean when `name` is not nested in a nested object. 
Yeah I haven't changed that, I only changed the nested ones.

elcct: I see, so it is not really consistent. I checked that `name` in nested ones indeed works, so that's cool.

marcofucci: 👍  we should probably fix that, the `Contact` API endpoints don't return `name`...

PR #393, opened 2017-08-15
Implement basic OMIS order search

marcofucci:
This adds the ability to get all orders.
There's no way to filter or sort the results using querystring params.

### DIFF
--- a/datahub/search/omis/views.py
+++ b/datahub/search/omis/views.py
@@ -1,0 +1,32 @@
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from .models import Order
+from .. import elasticsearch
+
+
+class SearchOrderAPIView(APIView):
+    """Filtered order search view."""
+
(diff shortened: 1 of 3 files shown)


datahub/search/omis/views.py
```diff
@@ -0,0 +1,32 @@
...
+class SearchOrderAPIView(APIView):
+    """Filtered order search view."""
+
+    http_method_names = ('post',)
+    DEFAULT_ORDERING = 'created_on:desc'
+
+    def post(self, request, format=None):
+        """Perform filtered order search."""
+        offset = int(request.data.get('offset', 0))
+        limit = int(request.data.get('limit', 100))
```
reupen: Can we get the default page size from REST_FRAMEWORK['PAGE_SIZE'] in the settings?

marcofucci: yes I was thinking to work on the pagination a bit more as well.

PR #394, opened 2017-08-16
Abstract out paginated logic in search API views

marcofucci:
@reupen @elcct what do you think about abstracting out the pagination logic?

I've added it just to the OMIS search API View but could be applied to the other ones as well if you want.

This also catches the 500 error about the result window being too big that we are getting from sentry recently.

### DIFF
--- a/datahub/search/omis/views.py
+++ b/datahub/search/omis/views.py
@@ -1,3 +1,7 @@
+from django.conf import settings
+
+from rest_framework.exceptions import ValidationError
+from rest_framework.pagination import _positive_int
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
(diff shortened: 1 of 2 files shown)


marcofucci:
I wanted to use the DRF logic but it's not nicely separated unfortunately.

datahub/search/omis/views.py
```diff
@@ -1,20 +1,65 @@
...
 
 
-class SearchOrderAPIView(APIView):
+class PaginatedAPIMixin:
+    """Mixin for paginated API Views."""
+
+    default_limit = settings.REST_FRAMEWORK['PAGE_SIZE']
+    limit_query_param = 'limit'
+    offset_query_param = 'offset'
+    max_results = 10000
```
reupen: Do we limit the result count returned to the front end to this value then?

marcofucci: Well at the moment it's 10000 so if we change the ES limit we can update this easily
It's not clear to the FE that specifying a value > this raises 500 I think and it's also not nice to receive those errors in general.

marcofucci: oh sorry, I understand the problem now. I think it's a good idea to limit the number of pages to this overall value. Anything after this value is probably not very useful anyway.

reupen: Hmm, might not be ideal on the collection pages when no or few filters are applied though.

marcofucci: @reupen OK should I take out the `max_results` and the validation check and we keep it for later?

reupen: No strong opinion. As long as we remember to do something about it...

PR #400, opened 2017-08-17
Add ability to filter orders by assigned advisers

marcofucci:
This:
- adds subscribers and assignees to the indexed order
- adds two filters to the search view:
	- assigned_to_adviser: filters by assignee adviser id
	- assigned_to_team: filters by assignee adviser team id

When creating or deleting an OrderAssignee or an OrderSubscriber, we need to update the order as well as it normalises those values in nested lists.
This means that if I create n assignees or n subscribers, at approx the same time, it triggers n equal commands to sync the related order.
It doesn't seem easy to group those ES syncs and only sync once per order.

### DIFF
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -32,7 +32,10 @@ def _contact_dict(obj, include_dit_team=False):
     }
 
     if include_dit_team:
-        data['dit_team'] = _id_name_dict(obj.dit_team)
+        if obj.dit_team:
+            data['dit_team'] = _id_name_dict(obj.dit_team)
+        else:
+            data['dit_team'] = {}
     return data
 
 
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -22,14 +22,18 @@ def _id_uri_dict(obj):
     }
 
 
-def _contact_dict(obj):
+def _contact_dict(obj, include_dit_team=False):
     """Creates dictionary with selected field from supplied object."""
-    return {
+    data = {
         'id': str(obj.id),
         'first_name': obj.first_name,
         'last_name': obj.last_name,
         'name': obj.name,
     }
+
+    if include_dit_team:
+        data['dit_team'] = _id_name_dict(obj.dit_team)
+    return data
 
 
 def _company_dict(obj):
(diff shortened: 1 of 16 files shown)


datahub/search/dict_utils.py
```diff
@@ -22,15 +22,19 @@ def _id_uri_dict(obj):
     }
 
 
-def _contact_dict(obj):
+def _contact_dict(obj, include_dit_team=False):
```
reupen: I find the name of `_contact_dict()` etc. confusing since it's also used for advisers – might be worth renaming it in a separate PR (`_person_dict()`? `_adviser_or_contact_dict()`?)

datahub/search/omis/test/test_models.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
-from datahub.omis.order.test.factories import OrderFactory
+from datahub.omis.order.test.factories import OrderAssigneeFactory, OrderFactory, \
```
reupen: Can we stick to using `(` `)` for multi-line imports? PEP 8 does say to use round-/square-/curly-brackets over `\` for line continuation, there's also PEP 328, http://docs.python-guide.org/en/latest/writing/style/#line-continuations etc.

datahub/search/dict_utils.py
marcofucci: yeah I agree

datahub/search/omis/test/test_views.py
```diff
@@ -153,6 +172,38 @@ def test_incorrect_dates_raise_validation_error(self, setup_es, setup_data):
         assert response.status_code == status.HTTP_400_BAD_REQUEST
         assert response.json() == {'non_field_errors': 'Date(s) in incorrect format.'}
 
+    def test_filter_by_assigned_to_assignee_adviser(self, setup_es, setup_data):
+        """Test that results can be filtered by assignee."""
+        setup_es.indices.refresh()
```
elcct: If you don't create any model that needs to be synced, there is no need to refresh the index.

marcofucci: `setup_data` does and if I don't think there might be some models that haven't been made into the index yet.

elcct: Ah ok, I thought `setup_data` would also refresh the index, but now I see if you want to add more objects in the test method, then you would have ended up refreshing twice. :+1:

PR #402, opened 2017-08-18
[DH-483] Adds sector name to Contact.

elcct:
This adds a `sector` from related `Company` object to the `Contact`.

I have added new `COMPUTED_MAPPINGS` map that will contain fields that do not exist directly in source model.

### DIFF
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -39,6 +39,26 @@ def _contact_dict(obj, include_dit_team=False):
     return data
 
 
+def _nested_id_name_dict(nested_field):
+    """Creates dictionary with selected fields from nested_field."""
+    def id_name_dict(obj):
+        fields = nested_field.split('.', maxsplit=1)
+        if len(fields) != 2:
+            raise ValueError("nested_field must be in 'nested_object.nested_field' format.")
+
+        related_object = getattr(obj, fields[0], None)
+        if related_object is None:
+            return None
+
+        field = getattr(related_object, fields[1], None)
+        if field is None:
+            return None
+
+        return _id_name_dict(field)
+
+    return id_name_dict
+
+
 def _company_dict(obj):
     return {
         'id': str(obj.id),
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -39,24 +39,24 @@ def _contact_dict(obj, include_dit_team=False):
     return data
 
 
-def _nested_id_name_dict(nested_field):
+def _computed_nested_id_name_dict(nested_field):
     """Creates dictionary with selected fields from nested_field."""
-    def id_name_dict(obj):
+    def get_dict(obj):
         fields = nested_field.split('.', maxsplit=1)
         if len(fields) != 2:
             raise ValueError("nested_field must be in 'nested_object.nested_field' format.")
 
-        related_object = getattr(obj, fields[0], None)
+        related_object = getattr(obj, fields[0])
         if related_object is None:
             return None
 
-        field = getattr(related_object, fields[1], None)
+        field = getattr(related_object, fields[1])
         if field is None:
             return None
 
         return _id_name_dict(field)
 
-    return id_name_dict
+    return get_dict
 
 
 def _company_dict(obj):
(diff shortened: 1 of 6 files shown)


datahub/search/dict_utils.py
```diff
@@ -32,6 +32,26 @@ def _contact_dict(obj):
...
+
+        related_object = getattr(obj, fields[0], None)
+        if related_object is None:
+            return None
+
+        field = getattr(related_object, fields[1], None)
+        if field is None:
+            return None
+
+        return _id_name_dict(field)
```
reupen: A bit confusing here to use both `_id_name_dict()` and `id_name_dict()`, can we rename one?

datahub/search/dict_utils.py
```diff
@@ -32,6 +32,26 @@ def _contact_dict(obj):
...
 
 
+def _nested_id_name_dict(nested_field):
+    """Creates dictionary with selected fields from nested_field."""
+    def id_name_dict(obj):
+        fields = nested_field.split('.', maxsplit=1)
+        if len(fields) != 2:
+            raise ValueError("nested_field must be in 'nested_object.nested_field' format.")
+
+        related_object = getattr(obj, fields[0], None)
```
reupen: Why would the model instance not have the field? What are we protecting against by provided `None` as the third argument?

datahub/search/dict_utils.py
```diff
@@ -32,6 +32,26 @@ def _contact_dict(obj):
...
+        if related_object is None:
+            return None
+
+        field = getattr(related_object, fields[1], None)
```
reupen: Same question here about `getattr()` default value.

datahub/search/dict_utils.py
elcct: This is true. It should have required attribute here.

datahub/search/dict_utils.py
```diff
@@ -32,6 +32,26 @@ def _contact_dict(obj):
     }
 
 
+def _nested_id_name_dict(nested_field):
```
reupen: It would be nice to differentiate this function from the other ones in this module, as this one returns a function while as the other ones return dicts.

datahub/search/dict_utils.py
elcct: It is when, as an example, `Contact` doesn't have `Company`, then related_object will be `None` and you won't be able to get `sector` out of it.

reupen: The `if related_object is None` check just before this line will take care of that :)

datahub/search/dict_utils.py
elcct: I changed it to `get_dict`. That wasn't best choice of a name.

datahub/search/dict_utils.py
elcct: Of course :)

datahub/search/dict_utils.py
elcct: I have added `_computed` prefix to differentiate.

marcofucci:
Not really convinced by the sector being on the contact though. Why do we even need it? And if so maybe we could use a 2 level nested `contact.company.sector` instead?

elcct:
@marcofucci this is required for `Contact` collection. I added it like that to avoid having over complicated queries - if you wanted to search by sector, the resulted query would be quite big for what it would be doing, if we added another level of nesting.

marcofucci:
OK but in that case, you would have to add extra signals for managing when when a company and a sector change. In those cases the related contact has to be reindexed.
I would actually try to avoid these cases as much as possible as the signals and syncs to ES can add up quite quickly.
I have a similar `order.adviser.dit_team` scenario and I can oversea problems if we overuse this.

elcct:
We have to do something similar for Contact address if it is the same as company, With this PR I really only wanted to expose sector for FE and I would tackle signal stuff when working on address soon. Sadly I can't think of a way to avoid this this time.

marcofucci:
Sorry I meant that you need to catch post_save signal on company as somebody changes the company sector, you need to reindex the related contacts as you are denormalising the sector.
This should really be part of this PR as if you don't handle this, the indexed data might have inconsistent values.

PR #406, opened 2017-08-21
Add manager_email and OMIS flag to countries

marcofucci:
This:

- adds `manager_email` to `Country`. This is the email address of the overseas manager overseeing that market.
- adds `omis_disabled_on` to Country. This is None if the market offers OMIS, manages legacy records and makes sure that a market can always be disabled in the future.
- `omis_disabled_on` is exposed via API but not `manager_email` as the endpoint does not require authentication and that piece of info is not really needed.
- the fixtures have been updated to include the correct value for omis_disabled_on. All manager_emails are currently empty.


datahub/metadata/models.py
```diff
@@ -42,6 +42,21 @@ class UKRegion(BaseConstantModel):
 class Country(BaseConstantModel):
     """Country."""
 
+    manager_email = models.EmailField(blank=True)
```
reupen: Should this be more specific than `manager_email` (e.g. `omis_manager_email`)? Does this mean that the manager isn't an adviser?

marcofucci: It's not necessarily OMIS specific, they are also called _Head of Trade and Investment_.

I was thinking to have a foreign they to Adviser but maybe some markets might want a generic email instead? A simple email seems safer.

marcofucci:
Closing as we decided to use a different approach instead (see #410)

PR #409, opened 2017-08-22
[DH-486] Sync address from Company to Contact if "address_same_as_com…

elcct:
…pany"

If `address_same_as_company` is `True`, it gets trading address from linked Company

It updates `Contact` if related company is being updated.

### DIFF
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -130,12 +130,14 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
     class MPTTMeta:  # noqa: D101
         order_insertion_by = ['name']
 
-    @cached_property
+    @property
     def uk_based(self):
         """Whether a company is based in the UK or not."""
         if not self.registered_address_country:
             return None
-        return self.registered_address_country.name == constants.Country.united_kingdom.value.name
+
+        united_kingdom_id = uuid.UUID(constants.Country.united_kingdom.value.id)
+        return self.registered_address_country.id == united_kingdom_id
 
     @cached_property
     def companies_house_data(self):
@@ -148,8 +150,7 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
             except CompaniesHouseCompany.DoesNotExist:
                 return None
 
-    @cached_property
-    def has_all_required_trading_address_fields(self):
+    def has_valid_trading_address(self):
         """Tells if Company has all required trading address fields defined."""
         return all(
             getattr(self, field) for field in self.REQUIRED_TRADING_ADDRESS_FIELDS
@@ -169,7 +170,7 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
             self.trading_address_postcode,
             self.trading_address_country
         ))
-        if any_trading_address_fields and not self.has_all_required_trading_address_fields:
+        if any_trading_address_fields and not self.has_valid_trading_address():
             return False
         return True
 
@@ -196,21 +197,6 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
                 {'uk_region': ['UK region is required for UK companies.']}
             )
         super().clean()
-
-    def refresh_from_db(self, using=None, fields=None, **kwargs):
-        """Clears cached properties on reload from database."""
-        super(Company, self).refresh_from_db(using, fields, **kwargs)
-
-        cached_properties = (
-            'companies_house_data',
-            'has_all_required_trading_address_fields',
-            'uk_based',
-        )
-        for key in cached_properties:
-            try:
-                del self.__dict__[key]
-            except KeyError:
-                pass
 
 
 class CompaniesHouseCompany(CompanyAbstract):
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -148,6 +148,13 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
             except CompaniesHouseCompany.DoesNotExist:
                 return None
 
+    @cached_property
+    def has_all_required_trading_address_fields(self):
+        """Tells if Company has all required trading address fields defined."""
+        return all(
+            getattr(self, field) for field in self.REQUIRED_TRADING_ADDRESS_FIELDS
+        )
+
     def _validate_trading_address(self):
         """Trading address fields are not mandatory in the model definition.
 
@@ -162,10 +169,7 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
             self.trading_address_postcode,
             self.trading_address_country
         ))
-        all_required_trading_address_fields = all(
-            getattr(self, field) for field in self.REQUIRED_TRADING_ADDRESS_FIELDS
-        )
-        if any_trading_address_fields and not all_required_trading_address_fields:
+        if any_trading_address_fields and not self.has_all_required_trading_address_fields:
             return False
         return True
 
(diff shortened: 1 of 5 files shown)


datahub/search/contact/signals.py
```diff
@@ -14,9 +14,25 @@ def contact_sync_es(sender, instance, **kwargs):
...
+        contact_sync_es,
+        sender=DBContact,
+        dispatch_uid='contact_sync_es'
+    )
+
+    post_save.connect(
+        related_contact_sync_es,
+        sender=DBCompany,
+        dispatch_uid='related_contact_sync_es'
+    )
```
marcofucci: add this to the disconnect signals func as well?

datahub/search/contact/test/test_views.py
```diff
@@ -111,3 +111,49 @@ def test_search_contact_has_sector(self, setup_es, setup_data):
...
+
+        url = reverse('api-v3:search:contact')
+        response = self.api_client.post(url, {
+            'original_query': contact.id,
+        })
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == 1
+
+        assert 'Updated Street' == response.data['results'][0]['address_1']
```
marcofucci: It would be good to have an extra test similar to `test_search_contact_has_company_address_updated` but with `address_same_as_compay=False`.

datahub/search/contact/signals.py
elcct: ahh I know I forgot about something! Thanks

elcct: Ok I have updated it now.

datahub/search/contact/test/test_views.py
elcct: Good idea, I also have added that test and improved others.

datahub/search/contact/dict_utils.py
```diff
@@ -0,0 +1,18 @@
+from .. import dict_utils
+
+
+def computed_address_field(field):
+    """Gets Contact address from Company is address_same_as_company."""
+    def get_field(contact):
+        if contact.address_same_as_company:
+            return getattr(contact.company, f'trading_{field}')
```
reupen: What about if the company doesn't have a trading address defined?

elcct: I was thinking about best way of adding it and will add it soon. Thanks.

elcct: @reupen it is added now.

datahub/company/models/company.py
```diff
@@ -193,6 +197,21 @@ def clean(self):
             )
         super().clean()
 
+    def refresh_from_db(self, using=None, fields=None, **kwargs):
+        """Clears cached properties on reload from database."""
+        super(Company, self).refresh_from_db(using, fields, **kwargs)
+
+        cached_properties = (
+            'companies_house_data',
+            'has_all_required_trading_address_fields',
```
reupen: Is there any reason for `has_all_required_trading_address_fields` and `uk_based` to be cached? It looks like they can just be made normal properties, I don't see anything expensive happening in them.

datahub/company/models/company.py
```diff
@@ -148,6 +148,13 @@ def companies_house_data(self):
             except CompaniesHouseCompany.DoesNotExist:
                 return None
 
+    @cached_property
+    def has_all_required_trading_address_fields(self):
```
reupen: I think `has_all_required_trading_address_fields` sounds a bit like a method. Could maybe make it a method or rename it (`trading_address_valid` ?)

datahub/company/models/company.py
```diff
@@ -193,6 +197,21 @@ def clean(self):
...
+        super(Company, self).refresh_from_db(using, fields, **kwargs)
+
+        cached_properties = (
+            'companies_house_data',
+            'has_all_required_trading_address_fields',
+            'uk_based',
+        )
+        for key in cached_properties:
+            try:
+                del self.__dict__[key]
```
reupen: Can we do this as suggested in the Django docs i.e. using `del` or `delattr()`?

datahub/company/models/company.py
```diff
@@ -193,6 +197,21 @@ def clean(self):
             )
         super().clean()
 
+    def refresh_from_db(self, using=None, fields=None, **kwargs):
+        """Clears cached properties on reload from database."""
+        super(Company, self).refresh_from_db(using, fields, **kwargs)
```
reupen: super() arguments aren't needed here.

datahub/search/contact/dict_utils.py
```diff
@@ -0,0 +1,22 @@
+from .. import dict_utils
+
+
+def computed_address_field(field):
+    """Gets Contact address from Company is address_same_as_company."""
+    def get_field(contact):
```
reupen: I don't think these two functions need to be nested. Could just move them up a level and make them internal (underscore-prefixed).

datahub/company/models/company.py
elcct: These were originally cached, but I can removed that

datahub/company/models/company.py
elcct: I could rename it, as name is indeed too long :)

datahub/company/models/company.py
elcct: I will fix that.

datahub/company/models/company.py
marcofucci: I vote for a method

datahub/company/models/company.py
marcofucci: Also `uk_based` seems to check on `name` instead of `id`, it might be worth changing that.

datahub/search/contact/dict_utils.py
elcct: But if I move them a level up, how do I pass through field that I want to get and still have matching signature that is required? I could use functools for that, but that would seem over engineered.

datahub/company/models/company.py
elcct: True :)

datahub/company/models/company.py
elcct: @marcofucci do you mind telling why? It doesn't do anything with the fields just tells if required number of fields are set.

datahub/search/contact/dict_utils.py
reupen: Oops, sorry, I missed that. I think it's OK then.

datahub/company/models/company.py
marcofucci: I might be wrong but my perception is that a property is a kind of static piece of data whilst with this we are _computing_ if the current set of data has certain values.

E.g. I would expect:

```python
company.has_all_required_trading_address_fields()
>> False
company. trading_address_1 = ...
company. trading_address_country = ...
company. trading_address_town = ...
>> 
company.has_all_required_trading_address_fields()
>> True
```

It doesn't feel as good if you use properties as you'd expect the related value to stay the same.

It might just be me so I'm easy if you want to use a property instead.

elcct: I am not convinced. My test was "could this exist as a field in the db?" and my answer was yes, as this could be set on save to indicate that company indeed has a trading address for benefit of not having to check all the fields every time. It is similar to the `uk_based` property - I also don't think this should be a method and only thing what it does is checking if the registered country matches the UK. This could also exist as a field in the db.

marcofucci: OK, I'm easy.

reupen: I'd have gone for a method as well, it's just clearer what's going on as in Marco's example.

But it's somewhat subjective which is why I didn't say to change it (but put method first in my original comment :stuck_out_tongue: )

datahub/company/models/company.py
elcct: Ok I removed that method as it is no longer needed.

datahub/company/models/company.py
elcct: Ok 2:1 so I changed it to be a method :)

PR #410, opened 2017-08-22
OMIS markets

marcofucci:
This is another way of implementing #406 which as @reupen noticed leaks some OMIS data so this other option seems more preferable.

This: PR 
- adds a new app `market` and the related model `Market` with country, manager_email and disabled_on fields 
- adds market to the metadata registry to generate the API view 
- adds the command `load_initial_omis_markets` to load the initial markets data. As some columns are going to change via the admin, the command only loads the records if the table is empty or if you pass --override 
- moves the OMIS fixtures to the _omis_ folder and adds a new command `load_omis_metadata` to load all OMIS metadata. This is not going to be called by `loadmetadata` so they both have to be run manually  
- `DisableableModel` has been moved to `datahub.core`

The new apps in the `omis` folder `core` and `market` now use a custom app name `omis-...` so that they don't collide with potential other apps called in the same way but not part of OMIS. This also generates tables with name `omis-...` which has nice namespaced behaviour. 
The other app `order` does not have this as it's not as straightforward.


datahub/omis/fixtures/initial_markets.yaml
```diff
@@ -0,0 +1,1002 @@
...
+  fields: {disabled_on: !!timestamp '2010-01-01 00:00:00.000000', manager_email: ''}
+# name: Andorra
+- model: omis-market.market
+  pk: 975f66a0-5d95-e211-a939-e4115bead28a
+  fields: {disabled_on: !!timestamp '2010-01-01 00:00:00.000000', manager_email: ''}
+# name: Angola
+- model: omis-market.market
+  pk: 985f66a0-5d95-e211-a939-e4115bead28a
+  fields: {disabled_on: null, manager_email: ''}
+# name: BLANK
```
reupen: Is this needed?

datahub/omis/fixtures/initial_markets.yaml
```diff
@@ -0,0 +1,1002 @@
...
+  fields: {disabled_on: null, manager_email: ''}
+# name: Ecuador
+- model: omis-market.market
+  pk: 75af72a6-5d95-e211-a939-e4115bead28a
+  fields: {disabled_on: null, manager_email: ''}
+# name: Jamaica
+- model: omis-market.market
+  pk: 766a9ab2-5d95-e211-a939-e4115bead28a
+  fields: {disabled_on: null, manager_email: ''}
+# name: TEST
```
reupen: Is this needed?

datahub/omis/fixtures/initial_markets.yaml
marcofucci: I left it as it might be needed in the near future because of a possible use case in our roadmap but you're right, I'll remove it and add it later if necessary.

datahub/omis/fixtures/initial_markets.yaml
marcofucci: I left it as it might be needed in the near future because of a possible use case in our roadmap but you're right, I'll remove it and add it later if necessary.

PR #415, opened 2017-08-23
[DH-496] Allow filtering by company name - sector - country - region.

elcct:
This brings `Contact` filters in line with requirements, which state we need 4 filters:

- Company Name (with partial matching)
- Sector
- Country
- UK Region

### DIFF
--- a/datahub/search/contact/test/test_views.py
+++ b/datahub/search/contact/test/test_views.py
@@ -2,7 +2,7 @@ import pytest
 from rest_framework import status
 from rest_framework.reverse import reverse
 
-from datahub.company.test.factories import ContactFactory
+from datahub.company.test.factories import CompanyFactory, ContactFactory
 from datahub.core.constants import Country, Sector, UKRegion
 from datahub.core.test_utils import APITestMixin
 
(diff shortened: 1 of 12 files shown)


datahub/search/contact/test/test_views.py
```diff
@@ -27,15 +27,49 @@ def test_search_contact(self, setup_es, setup_data):
...
+
+    def test_filter_contact(self, setup_es, setup_data):
+        """Tests matching contact using multiple filters."""
+        contact = ContactFactory(address_same_as_company=True)
+        company = contact.company
+        company.name = 'SlothsCats'
+        company.trading_address_country_id = Country.united_kingdom.value.id
+        company.uk_region_id = UKRegion.east_of_england.value.id
+        company.sector_id = Sector.renewable_energy_wind.value.id
+        company.save()
```
reupen: Could create a company first and then pass it to ContactFactory()?

reupen:
Would be nice to cover a few more scenarios in the tests (e.g. correct handling of companies/contacts without a uk_region, testing of cases where filters don't match results).

datahub/search/contact/test/test_views.py
elcct: I see no reason why not. Changed.

elcct:
I will add some more tests.

elcct:
@reupen @marcofucci I added a little more tests for `Contact` filters and also cleaned up the search views.

reupen:
I think it would be good to split off the search view refactor to a separate PR. For the interactions search app I was looking at using DRF serialisers:
https://github.com/uktrade/data-hub-leeloo/compare/feature/DH-488-interactions-in-search?expand=1

elcct:
That's nice. I can definitely split that and then also incorporate DRF serializers.

marcofucci:
Yes maybe you can split this PR.

I really like serializers, I tried the same for filters but the REMAP_FIELDS logic doesn't play well with DRF serializers as you have to remap keys of nested props (e.g. `sector` has to become `sector.id`). It should be doable in a way or another though if we spend more time thinking about it.

@reupen I would keep the pagination logic away from the serializsers, you can use my PaginatedAPIView with the methods that I _borrowed_ from the DRF codebase which should cover all the cases. There's a bit more logic there.

reupen:
@marcofucci I get where you're coming from, but my converse argument would be that it's all in the POST body and should be validated in a consistent way. It could be split to a separate serialiser though.

I'm also not sure I agree with silently ignoring bad values – a negative offset or non-positive limit can't happen by accident, surely? Also if offset <= 10,000 but offset + limit > 10,000 it might be more desirable to limit the number of results instead of returning an error.

elcct:
Are we not going to over-engineer it? It seems like a lot of hassle to just validate a fairly simple request.

marcofucci:
@reupen that's fine, I always try to be consistent even in less desirable cases and I guess we are using DRF pagination elsewhere but I see what you mean as well. 
TBH the DRF code is getting pagination params from GET whilst we are always going to use POST so serializers should feel natural.

PR #419, opened 2017-08-29
Add interactions search endpoint

reupen:
It was a bit unfortunate that I had to modify those query tests in the other search apps – I think we could probably move those up a level and maybe consolidate them?

I've also used a DRF serialiser here to validate the POST body. I think we can look at doing this across the search apps. I've seen a few other things that we could probably tidy up as well.

### DIFF
--- a/datahub/search/interaction/apps.py
+++ b/datahub/search/interaction/apps.py
@@ -1,8 +1,10 @@
 from datahub.interaction.models import Interaction as DBInteraction
+from datahub.interaction.queryset import get_interaction_queryset_v3
 
 from datahub.search.apps import SearchApp
 from datahub.search.interaction.models import Interaction
 from datahub.search.interaction.views import SearchInteractionAPIView
+from datahub.search.models import DataSet
 
 
 class InteractionSearchApp(SearchApp):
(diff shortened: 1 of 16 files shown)


datahub/search/interaction/apps.py
```diff
@@ -0,0 +1,23 @@
...
+from datahub.interaction.queryset import get_interaction_queryset_v3
+
+from datahub.search.apps import SearchApp
+from datahub.search.interaction.models import Interaction
+from datahub.search.interaction.views import SearchInteractionAPIView
+from datahub.search.models import DataSet
+
+
+class InteractionSearchApp(SearchApp):
+    """SearchApp for interactions"""
```
elcct: Missing dot :)

reupen: I copied the other ones :stuck_out_tongue:

reupen:
> Can you just remove the [DH-488] from the commit messages? You can add it to the PR text as reference.

Then the commits won't appear in JIRA and it's a useful point of reference when looking at why something changed in a file etc. They could go in the second line but then it'd be the only thing there in most cases.

PR #420, opened 2017-08-29
EnglishAnalyzer for strings.

elcct:
This can be used for strings that contain English text to be searchable.

### DIFF
--- a/datahub/search/elasticsearch.py
+++ b/datahub/search/elasticsearch.py
@@ -10,7 +10,6 @@ from elasticsearch_dsl.query import Match, MatchPhrase, Q
 
 from .apps import get_search_apps
 
-
 MAX_RESULTS = 10000
 
 lowercase_keyword_analyzer = analysis.CustomAnalyzer(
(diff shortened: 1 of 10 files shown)


reupen:
Could probably stick it on company description as well.

datahub/search/elasticsearch.py
```diff
@@ -31,6 +30,35 @@
...
+    'english_stop',
+    type='stop',
+    stopwords='_english_'
+)
+
+english_analyzer = analysis.CustomAnalyzer(
+    'english_analyzer',
+    tokenizer='standard',
+    filter=[
+        english_possessive_stemmer,
```
reupen: Do we need the `english_possessive` stemmer if we're using the plain `english` stemmer?

elcct: The doc says: `The possessive stemmer removes 's from any words before passing them on to the english_stop, english_keywords, and english_stemmer.`.

elcct:
I added it to description fields across models as well.

PR #423, opened 2017-08-29
Fix sync when address_country is None.

elcct:


marcofucci:
I was actually thinking that it would make more sense to change all the `*_dict` functions so that they check and return None (or maybe {} ?) if the input param is None.
This is also to avoid similar problems in the future. What do you think?

elcct:
I am not sure if validation should be `*_dict` responsibility, but indeed that way we could avoid this kind of problems.

elcct:
@marcofucci I changed `*_dict` to return None if input param is None

PR #429, opened 2017-08-30
Changes plural entity name to results in the response of basic search.

elcct:


elcct:
Good point as it is no longer needed :) Thanks!

reupen:
Can we do it in a backwards compatible way? Like temporarily including both properties until the front-end change is made and released?

marcofucci:
Very good point, yes that sounds like a good idea. @elcct you would have to revert back to the previous commit, sorry about that.

elcct:
@marcofucci @reupen @tyom said that front end is already compatible with this change

reupen:
So if we deployed this to production today, nothing would break?

tyom:
@reupen Nothing should break but obviously we need to test it before deploying :)

PR #430, opened 2017-08-30
Add ability to generate a quote for an OMIS order.

marcofucci:
This:
- creates a new omis quote django app with a Quote model
- adds a create quote endpoint
- adds some validation logic to check that a quote for an order can be created

I'm re-using some DRF concepts (validator, exceptions) in different way in order to avoid adding new unnecessary layers.

The text of the quote is currently a placeholder to keep this PR small.


datahub/omis/order/models.py
```diff
@@ -91,43 +101,59 @@ def __str__(self):
...
     def get_datahub_frontend_url(self):
         """Return the url to the Data Hub frontend order page."""
         return f'{settings.DATAHUB_FRONTEND_BASE_URL}/omis/{self.pk}'
 
+    @transaction.atomic
+    def generate_quote(self, quote_data):
+        """
+        Generate a quote for this order if possible or raise
+
+        - rest_framework.exceptions.ValidationError in case of validation error
```
reupen: Uses :raises: here?

datahub/omis/order/validators.py
```diff
@@ -61,6 +64,67 @@ def __call__(self, data):
...
+        'delivery_date',
+    )
+
+    message = 'This field is required.'
+
+    def set_instance(self, instance):
+        """Set the current instance."""
+        self.instance = instance
+
+    def __call__(self, data={}):
```
reupen: I know data isn't being mutated here, but it would be nice to avoid `{}` etc. as a default value for a parameter in general due to the problems it causes when the argument is mutated.

marcofucci: True, I always forget that!

datahub/omis/quote/models.py
```diff
@@ -0,0 +1,34 @@
...
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    reference = models.CharField(max_length=100)
+    content = models.TextField()
+
+    @classmethod
+    def generate_reference(cls, order):
+        """
+        Return a random unused reference of form:
+            <order.reference>/Q-<(2) lettes>/<(1) number> e.g. GEA962/16/Q-AB1
```
elcct: Typo - `lettes`

datahub/omis/order/validators.py
```diff
@@ -61,6 +64,67 @@ def __call__(self, data):
...
+        data_combiner = DataCombiner(self.instance, data)
+
+        meta = self.instance._meta
+        errors = {}
+        for field_name in self.MANDATORY_FIELDS:
+            field = meta.get_field(field_name)
+
+            value = data_combiner.get_value(field_name)
+            if isinstance(field, models.ManyToManyField):
+                value = value.all()
```
reupen: May want to consider the case where data isn't None, it probably won't be a query set then (a pain, I know)...

For investment I did it using DataCombiner.get_value_to_many(): https://github.com/uktrade/data-hub-leeloo/blob/38510da17940967dca19ef97cd04dfe3e72acb45/datahub/investment/validate.py#L149-L153

datahub/omis/core/utils.py
```diff
@@ -0,0 +1,17 @@
+def generate_reference(model, gen, field='reference', prefix='', max_retries=10):
+    """
+    Generate a unique reference given:
+
+    :param model: the class of the django model
+    :param gen: a function without arguments that returns part or all the reference
+    :param field: reference field of the model that needs to be unique
+    :param prefix: optional prefix
+    :param max_retries: max number of retries before failing
```
elcct: Worth adding `:raises` ?

datahub/omis/order/models.py
```diff
@@ -91,43 +101,59 @@ def __str__(self):
...
+        for validator in [
+            validators.OrderDetailsFilledInValidator(),
+            validators.NoOtherActiveQuoteExistsValidator()
+        ]:
+            validator.set_instance(self)
+            validator()
+
+        quote = Quote.objects.create(
+            **quote_data,
+            reference=Quote.generate_reference(self),
```
elcct: Does it also raise `RuntimeError` via `Quote.generate_reference()` ?

datahub/omis/order/validators.py
```diff
@@ -61,6 +64,67 @@ def __call__(self, data):
...
-                    raise serializers.ValidationError({
+                    raise ValidationError({
                         field: self.message.format(field_name)
                     })
+
+
+class OrderDetailsFilledInValidator:
+    """Validator which checks that the order has all detail fields filled in."""
+
+    MANDATORY_FIELDS = (
```
reupen: Might be worth calling this `REQUIRED_FIELDS` for consistency.

marcofucci:
👍  thank you both, I've corrected / improved the code based on your feedback

PR #432, opened 2017-08-31
Fix paging params in Company search endpoint.

elcct:
It makes it consistent with other search endpoints.


reupen:
I think it'll need to handle both for now as I see code in the front end relying on the old behaviour:

https://github.com/uktrade/data-hub-frontend/blob/50c81292f494432acf719a3a2f49dc4ac863cd93/src/apps/search/services.js#L58

elcct:
Good catch. It now looks in both places for params.

elcct:
I am working on that, it will be another PR at some point

PR #436, opened 2017-09-01
Add create event endpoint

reupen:
Still a few things to clarify, so may be a few changes next week.

Decided to leave programmes in the event app for now, for review later on if it becomes a generic concept. Have still included them in the loadmetadata command as I don't think we need another command. Could perhaps add some arguments to it though.

In relation to the null=False/required=True thing, a different workaround that seems to work is specifying `default=''` on the model field. Don't have much preference really.


datahub/event/models.py
```diff
@@ -0,0 +1,59 @@
...
+        ('Other', 'other', 'Other'),
+    )
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    name = models.CharField(max_length=MAX_LENGTH)
+    event_type = models.CharField(max_length=MAX_LENGTH, choices=EVENT_TYPES)
+    start_date = models.DateField(null=True, blank=True)
+    end_date = models.DateField(null=True, blank=True)
+    location_type = models.CharField(
+        max_length=MAX_LENGTH, null=True, blank=True, choices=LOCATION_TYPES
```
marcofucci: Can't we remove `null=True`?

datahub/event/models.py
```diff
@@ -0,0 +1,59 @@
...
+    address_county = models.CharField(max_length=MAX_LENGTH, blank=True)
+    address_postcode = models.CharField(max_length=MAX_LENGTH)
+    address_country = models.ForeignKey(
+        'metadata.Country', on_delete=models.PROTECT, related_name='+'
+    )
+    notes = models.TextField(blank=True)
+    lead_team = models.ForeignKey(
+        'metadata.Team', on_delete=models.PROTECT, null=True, blank=True, related_name='+'
+    )
+    additional_teams = models.ManyToManyField('metadata.Team', blank=True, related_name='+')
```
marcofucci: wouldn't it be better to have a through model with a lead flag? What if you want to change the lead team? You would have to add the old `lead_team` to the `additional_teams`, remove the new lead from the `additional_teams` and set it to the `lead_team`.

datahub/event/models.py
reupen: Yep, I'll get rid of it.

datahub/event/models.py
```diff
@@ -0,0 +1,59 @@
...
+        ('HQ', 'hq', 'HQ'),
+        ('Post', 'post', 'Post'),
+        ('Regional network', 'regional_network', 'Regional network'),
+        ('Other', 'other', 'Other'),
+    )
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    name = models.CharField(max_length=MAX_LENGTH)
+    event_type = models.CharField(max_length=MAX_LENGTH, choices=EVENT_TYPES)
+    start_date = models.DateField(null=True, blank=True)
```
elcct: Are we going to have event without start date?

datahub/event/models.py
reupen: It reflects the API (which in turn reflects the UI). But in the other place I used through (investment project) it brought about a lot of complexity. It could equally be fixed by changing additional_teams to teams and then you only need to validate that lead_team is in teams.

Anyway I think this equally applies to the API so might be better discussed there: https://github.com/uktrade/data-hub-api-spec/pull/35

elcct: Or we can have `teams` that contain all teams and `lead_team` as a "pointer" to the team that is leading. That removes complexity @marcofucci mentioned.

datahub/event/models.py
reupen: At the moment on JIRA it says it is optional. It might also need to be nullable for the migration.

datahub/event/models.py
elcct: @reupen haha commented same time and looks like we think same thing :)

marcofucci: 👍  for `teams`

datahub/event/models.py
marcofucci: Yeah it doesn't seem to make sense. We should find out why it's nullable and fix the data during the migration if that is the case.

reupen:
@marcofucci @elcct Have made some changes (changed additional teams to teams, added organiser, changed event and location type to be foreign keys).

There might be some other changes needed, but I'll do those in separate PRs.

datahub/event/serializers.py
```diff
@@ -0,0 +1,64 @@
...
+from datahub.company.serializers import NestedAdviserField
+from datahub.core.serializers import NestedRelatedField
+from datahub.core.validate_utils import DataCombiner
+from datahub.event.models import Event
+
+
+class EventSerializer(serializers.ModelSerializer):
+    """Event serialiser."""
+
+    _lead_team_error_text = 'Lead team must be in teams array.'
```
marcofucci: You could try use the same DRF logic.
Define the message like this: 
```
default_error_messages = {
    'lead_in_teams': _('Lead team must be in teams array.'),
}
```
and then use it this way:
```self.error_messages['lead_in_teams']```

It might work ;)

File: datahub/event/constants.py [hunk 5]
```diff
@@ -0,0 +1,22 @@
...
+class LocationType(Enum):
+    """Location type constants."""
+
+    hq = Constant('HQ', 'b71fa81c-0c22-44c6-ab6f-13b9e045dc10')
+
+
+class EventType(Enum):
+    """Event type constants."""
+
+    seminar = Constant('Seminar', '771f654d-e6b1-4fad-8a89-b6ac44147830')
```
marcofucci: I'm trying to stay away from creating more constants and just get a random record in the tests in order to avoid duplications.
I know that this is more performant though so I'm easy on this, we could refactor later on.

datahub/event/serializers.py
reupen: Done!

datahub/core/management/commands/loadmetadata.py
```diff
@@ -4,7 +4,29 @@
...
+    SHARED_METADATA_FIXTURE_DIR / 'teams.yaml',
+    SHARED_METADATA_FIXTURE_DIR / 'titles.yaml',
+    SHARED_METADATA_FIXTURE_DIR / 'uk_regions.yaml',
+)
+
+EVENTS_FIXTURES = (
+    EVENTS_FIXTURE_DIR / 'programmes.yaml',
+    EVENTS_FIXTURE_DIR / 'location_types.yaml',
+    EVENTS_FIXTURE_DIR / 'event_types.yaml',
+)
```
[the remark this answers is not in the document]
reupen: Yeah, that might be better.

datahub/event/constants.py
reupen: I don't like them either, but we don't have another solution at present.

marcofucci: Sorry I wasn't clear, the other solution is to use `EventType.objects.first()` instead.
It's less performant so I'll leave it to you to decide.

datahub/event/serializers.py
```diff
@@ -0,0 +1,67 @@
...
+        }
+        fields = (
+            'address_1',
+            'address_2',
+            'address_country',
+            'address_country',
+            'address_county',
+            'address_postcode',
+            'address_town',
+            'end_date',
```
marcofucci: Do you need to check if `end_date <= start_date` ?

reupen: Yep, good point. I'll do it in another PR if that's OK.

datahub/event/test/test_views.py
```diff
@@ -0,0 +1,179 @@
...
+                'id': str(self.user.pk),
+                'first_name': str(self.user.first_name),
+                'last_name': str(self.user.last_name),
+                'name': str(self.user.name),
+            },
+            'lead_team': {
+                'id': Team.crm.value.id,
+                'name': Team.crm.value.name,
+            },
+            'teams': [{
```
marcofucci: asserting lists fails in my local env 🤔

reupen: Ah, sorry, the order would be undefined. Another PR coming up!

PR #437, opened 2017-09-01
Add ability to get the content of a quote

marcofucci:
The endpoint `/v3/omis/order/<order-id>/quote` now accepts ~~`full=1`~~ `expand=1` to indicate that the response body has to include the content of the quote as well.

By default we want to return a lighter version.

### DIFF
--- a/datahub/omis/quote/views.py
+++ b/datahub/omis/quote/views.py
@@ -5,14 +5,15 @@ from datahub.core.viewsets import CoreViewSetV3
 from datahub.omis.order.models import Order
 
 from .models import Quote
-from .serializers import QuoteSerializer
+from .serializers import BasicQuoteSerializer, ExpandedQuoteSerializer, ExpandParamSerializer
 
 
 class QuoteViewSet(CoreViewSetV3):
     """Quote ViewSet."""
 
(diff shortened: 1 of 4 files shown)


elcct:
`full=1` url feels a bit hacky. Would something like `expand` that accept boolean be better?

datahub/omis/quote/views.py
```diff
@@ -39,6 +41,26 @@ def get_object(self):
...
+    def _requires_full(self):
+        """
+        :returns: True if full response required, False otherwise
+
+        This can be implicit from the action or explicitly requested with the
+        `full` query param.
+        """
+        if self.action == 'create':
+            return True
+        return self.request.query_params.get(self.full_query_param, '').strip() == '1'
```
reupen: Could potentially use BooleanField from DRF rather than explicitly comparing to `'1'`? Would give us some consistency about what is considered truthy/falsely for boolean values in the API.

reupen: Also why `.strip()`?

marcofucci:
@reupen @elcct thanks for the feedback, all good. That was just me being lazy!
Replaced the param name and started using a serializer now :)

PR #438, opened 2017-09-04
Add ability to preview an order

marcofucci:
This adds a new endpoint `v3/order/<order-id>/quote/preview` which acts exactly like the create but without actually saving the changes.
The endpoint is used to preview the content before creating a quote.

### DIFF
--- a/datahub/omis/quote/test/test_views.py
+++ b/datahub/omis/quote/test/test_views.py
@@ -16,11 +16,61 @@ from ..models import Quote
 pytestmark = pytest.mark.django_db
 
 
-class TestCreateQuote(APITestMixin):
-    """Create quote test case."""
+class TestCreatePreviewOrder(APITestMixin):
+    """Tests for creating and previewing a quote."""
+
+    @pytest.mark.parametrize('quote_view_name', ('item', 'preview'))
+    def test_404_if_order_doesnt_exist(self, quote_view_name):
+        """Test that if the order doesn't exist, the endpoint returns 404."""
+        url = reverse(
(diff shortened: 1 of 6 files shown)


datahub/omis/quote/views.py
```diff
@@ -17,6 +20,15 @@ class QuoteViewSet(CoreViewSetV3):
 
     order_lookup_url_kwarg = 'order_pk'
 
+    def preview(self, request, *args, **kwargs):
+        """
+        Same as `create` but without actually saving the changes.
```
reupen: So what is being saved when it calls `serializer.save()`? Just a bit confusing...

datahub/omis/quote/test/test_views.py
```diff
@@ -16,36 +16,17 @@
...
 
 
-class TestCreateQuote(APITestMixin):
-    """Create quote test case."""
-
-    @freeze_time('2017-04-18 13:00:00.000000+00:00')
-    def test_success(self):
-        """Test a successful call to create a quote."""
-        order = OrderFactory()
+class CreatePreviewMixin:
```
reupen: What about using parametrize (on the class maybe)? Just think using a mixin makes it a bit harder to work out what's going on (but happy either way).

datahub/omis/quote/views.py
marcofucci: I agree, it was clean (less duplications) but maybe too magic. Replaced with explicit call.

datahub/omis/quote/test/test_views.py
marcofucci: 👍  changed

PR #439, opened 2017-09-04
[DH-541] Support Elasticsearch 5.5.

elcct:
Biggest difference is that Elasticsearch 5 has `fielddata` set to `False` by default. It means that we have to set it to `True` for the string fields we want to sort by.

### DIFF
--- a/datahub/search/dsl_utils.py
+++ b/datahub/search/dsl_utils.py
@@ -1,9 +1,7 @@
 from functools import partial
 from elasticsearch_dsl import Keyword, Nested, String
 
-
-KeywordString = Keyword
-CaseInsensitiveKeywordString = partial(
+SortableCaseInsensitiveKeywordString = partial(
     String,
     analyzer='lowercase_keyword_analyzer',
     fielddata=True
@@ -16,10 +14,10 @@ SortableString = partial(String, fielddata=True)
 def contact_or_adviser_mapping(field, include_dit_team=False):
     """Mapping for Adviser/Contact fields."""
     props = {
-        'id': KeywordString(),
-        'first_name': CaseInsensitiveKeywordString(),
-        'last_name': CaseInsensitiveKeywordString(),
-        'name': CaseInsensitiveKeywordString(),
+        'id': Keyword(),
+        'first_name': SortableCaseInsensitiveKeywordString(),
+        'last_name': SortableCaseInsensitiveKeywordString(),
+        'name': SortableCaseInsensitiveKeywordString(),
     }
 
     if include_dit_team:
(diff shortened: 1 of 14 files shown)


datahub/search/dsl_utils.py
```diff
@@ -1,11 +1,16 @@
 from functools import partial
-from elasticsearch_dsl import Nested, String
+from elasticsearch_dsl import Keyword, Nested, String
 
 
-KeywordString = partial(String, index='not_analyzed')
-CaseInsensitiveKeywordString = partial(String, analyzer='lowercase_keyword_analyzer')
+KeywordString = Keyword
```
reupen: Do we need to keep this?

datahub/search/dsl_utils.py
```diff
@@ -1,11 +1,16 @@
...
+CaseInsensitiveKeywordString = partial(
```
reupen: Should 'Sortable' be added to the name here?

reupen:
Do we have ES 5.5 instances on standby ready to switch to? Should we change the name of the ES_URL environment variable to e.g. ES5_URL, just thinking it might make deployment easier.

marcofucci:
I don't know exactly why you started to use `fielddata` but the docs say that we should really avoid them as 
> fielddata can consume a lot of heap space

As I understand, fielddata is just for rare cases. Most of the times, if you need sorting or aggregations on text, it's better to use `keywords` which support `doc_values`.

Why don't we use `raw` https://www.elastic.co/guide/en/elasticsearch/guide/master/multi-fields.html for sorting?

datahub/search/dsl_utils.py
elcct: We don't. I will remove it.

datahub/search/dsl_utils.py
elcct: I had the same question, so I will add it. Thanks

elcct:
@marcofucci in 2.4 `fielddata` is enabled by default on those fields. I think the first step would be to get compatibility first and then do optimisations specific to the platform.
Regarding multi fields we just use multiple fields in the model as this was more convenient (in a `fieldname_variation` format. As a next step we could add a keyword field for the purpose of sorting. 
But if we want to have case insensitive sorting, then we need to keep `fielddata`, as `Keyword` is case sensitive.

elcct:
@reupen I don't know about instance. But having new env variable sounds good. I will add that.

marcofucci:
@elcct 👍  makes sense to refactor in steps. 
In 5.x there seem to be a new [normalizer](https://discuss.elastic.co/t/best-practice-of-case-insensitive-keyword-mapping-in-es-5-x/74015) thing which we could use.

PR #441, opened 2017-09-05
Add ability to cancel a quote

marcofucci:
This adds the ability to reopen an order and cancel a quote.

A new endpoint POST `v3/order/<order-id>/quote/cancel` has been created.

The reason behind POSTing to cancel instead of DELETing the resource is that we do want the cancelled quote to still be referenced by the order so that it can be used to check its status.

We do not delete any quotes for the time being as they are going to be useful in absence of audit trail.

A cancelled quote can be replaced by a new one.


datahub/omis/order/test/test_models.py
```diff
@@ -128,6 +133,58 @@ def test_without_committing(self):
         assert not Quote.objects.count()
 
 
+class TestReopen:
+    """Tests for when an order is reopened."""
+
+    def test_without_quote(self):
+        """
+        Test that if an order without quote is reopened, nothing happens as
+        the quote is already open.
```
reupen: > the quote is already open
Is that right or should it say 'order is already open'?

marcofucci: Nice catch!

marcofucci:
@reupen is it a 👍  for you too? Just double-checking to be sure.

marcofucci:
@reupen added some docs ;)

reupen:
Looks good :)

marcofucci:
need to wait for your bugfix before merging ;)

PR #444, opened 2017-09-05
Refactor quote creation

marcofucci:
This refactors the quote creation logic to make it more obvious and consistent:
- a quote gets created by the `QuoteManager`
- the classmethods to populate the quote have been moved to `utils.py`
- `order.generate_quote` has now the same signature as `order.reopen`

### DIFF
--- a/datahub/omis/quote/manager.py
+++ b/datahub/omis/quote/manager.py
@@ -1,0 +1,26 @@
+from django.db import models
+
+from .utils import generate_quote_content, generate_quote_reference
+
+
+class QuoteManager(models.Manager):
+    """Custom Quote Manager."""
+
+    def create_from_order(self, order, by, commit=True):
+        """
+        :param order: Order instance for this quote
+        :param by: who made the action
+        :param commit: True if the changes have to be committed
+
+        :returns: Quote object generated from the order
+        """
+        quote = self.model(
+            created_by=by,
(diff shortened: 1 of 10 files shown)


datahub/omis/quote/manager.py
```diff
@@ -0,0 +1,26 @@
...
+    def create_from_order(self, order, by, commit=True):
+        """
+        :param order: Order instance for this quote
+        :param by: who made the action
+        :param commit: True if the changes have to be committed
+
+        :returns: Quote object generated from the order
+        """
+        quote = self.model(
+            created_by=by,
```
reupen: Shouldn't it set `modified_by` too for consistency with other endpoints etc.?

marcofucci: A quote can be accepted by the client who is not an adviser and therefore it cannot be a valid value. I don't know if setting `modified_by` only if the actor is an adviser seems obvious enough.
That's the reason why I left it out.

What do you think?

reupen: Ah, that makes sense. Will there be any other cases where advisers update quotes? Null could represent an update that's not been made by an adviser, but it's only a useful distinction to make if there are actually updates being made by advisers.

marcofucci: Yeah I don't think it adds much value in this case as the quote doesn't get updated.

PR #445, opened 2017-09-05
[DH-559] Add get event and list events endpoints

reupen:


datahub/event/test/test_views.py
```diff
@@ -223,3 +296,15 @@ def test_create_blank_failure(self):
             'event_type': ['This field may not be null.'],
             'name': ['This field may not be blank.']
         }
+
+
+def _get_canonical_response_data(response):
+    return _canonicalise_response_data(response.json())
```
elcct: That name feels weird. Is it really canonical? Would something like `format_response_data` be better?

reupen: It's not _the_ canonical form, but it's what the tests defined the canonical form to be. I don't think format describes what it does, but could go with normalise.

PR #450, opened 2017-09-06
Add expires_on field to OMIS Quote

marcofucci:
This adds a new field `expires_on` to an OMIS quote.
The value is auto-generated from the delivery date when creating a quote and an error is raised if the delivery date is too close and the resulting expiry date would be in the past.


datahub/omis/quote/constants.py
```diff
@@ -0,0 +1,5 @@
+# used to calculate the quote expiry date, that is whichever is earliest of
+# [delivery date - QUOTE_EXPIRY_DAYS_BEFORE_DELIVERY days] OR
+# [date quote created + QUOTE_EXPIRY_DAYS_FROM_NOW days]
+QUOTE_EXPIRY_DAYS_BEFORE_DELIVERY = 21
```
elcct: Shouldn't we use environment variables for such constants? Or is it unlikely to change?

marcofucci: Yeah very unlikely to change. These haven't been signed off yet so I just wanted to make them easy to find.

PR #455, opened 2017-09-08
[DH-557] Create Event search endpoint.

elcct:
It creates Event search endpoints.

I moved serializers from Interactions and view from OMIS a level up and reused it across other search endpoints.

It doesn't have serializer for filters - I think this could be done in another PR.


datahub/search/company/serializers.py
```diff
@@ -0,0 +1,31 @@
+from ..serializers import SearchSerializer
+
+
+class SearchCompanySerializer(SearchSerializer):
+    """Serialiser used to validate event search POST bodies."""
```
reupen: Docstring need updating?

elcct: Ah I missed that. Updated, thanks!

datahub/search/event/tests/test_views.py
```diff
@@ -0,0 +1,114 @@
...
+from rest_framework.reverse import reverse
+
+from datahub.core.test_utils import APITestMixin
+from datahub.event.test.factories import EventFactory
+
+
+@pytest.fixture
+def setup_data():
+    """Sets up data for the tests."""
+    EventFactory()
```
marcofucci: `EventFactory.create_batch(2)` ;)

datahub/search/event/tests/test_views.py
```diff
@@ -0,0 +1,114 @@
...
+        url = reverse('api-v3:search:basic')
+        response = self.api_client.get(url, {
+            'term': term,
+            'entity': 'event'
+        })
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] > 0
+
+    def test_events(self, setup_es, setup_data):
```
marcofucci: Maybe change to `test_aggregations` ?

datahub/search/omis/serializers.py
```diff
@@ -0,0 +1,11 @@
+from ..serializers import SearchSerializer
+
+
+class SearchOrderSerializer(SearchSerializer):
+    """Serialiser used to validate OMIS search POST bodies."""
+
+    DEFAULT_ORDERING = 'created_on:desc'
+
+    SORT_BY_FIELDS = (
+        'created_on'
```
marcofucci: missing trailing comma

datahub/search/event/apps.py
```diff
@@ -0,0 +1,20 @@
...
+
+class EventSearchApp(SearchApp):
+    """SearchApp for events."""
+
+    name = 'event'
+    ESModel = Event
+    view = SearchEventAPIView
+    queryset = DBEvent.objects.prefetch_related(
+        'address_country',
+        'location_type',
```
reupen: Looks like event_type is missing?

elcct: Good catch!

datahub/search/event/tests/test_views.py
elcct: done :)

datahub/search/event/tests/test_views.py
elcct: Good point. Done.

datahub/search/omis/serializers.py
elcct: Nooooo.... Good catch :) Fixed

datahub/search/serializers.py
```diff
@@ -0,0 +1,53 @@
...
+
+class LimitOffsetSerializer(serializers.Serializer):
+    """Serialiser used to validate limit/offset values in POST bodies."""
+
+    offset = serializers.IntegerField(default=0, min_value=0, max_value=MAX_RESULTS - 1)
+    limit = serializers.IntegerField(default=api_settings.PAGE_SIZE, min_value=1)
+
+
+class SearchSerializer(LimitOffsetSerializer):
+    """Serialiser used to validate interaction search POST bodies."""
```
reupen: Docstring to update here (has 'interaction' in it)

datahub/search/views.py
```diff
@@ -73,3 +74,77 @@ def get(self, request, format=None):
...
+            })
+
+        return filters, ranges
+
+    def post(self, request, format=None):
+        """Performs search."""
+        data = request.data.copy()
+
+        # to support legacy paging parameters that can be in query_string
+        data.update(request.query_params)
```
reupen: I think it would make sense for the request body to take priority over the query string here? And to just extract the two fields that are needed from the query string?

datahub/search/views.py
```diff
@@ -73,3 +74,77 @@ def get(self, request, format=None):
...
+class SearchAPIView(APIView):
+    """Filtered investment project search view."""
+
+    FILTER_FIELDS = []
+    REMAP_FIELDS = {}
+
+    serializer_class = SearchSerializer
+    entity = None
+
+    aggregations = False
```
reupen: Might be nice to give this a more boolean-like name.

datahub/search/views.py
elcct: We don't use `query_params` except for pagination, so I thought that wouldn't matter. Since this is supporting legacy behaviour I am happy to extract just the parameters needed for pagination here.

datahub/search/views.py
```diff
@@ -73,3 +74,77 @@ def get(self, request, format=None):
...
+
+        filters, ranges = self.get_filtering_data(request)
+
+        results = elasticsearch.get_search_by_entity_query(
+            entity=self.entity,
+            term=validated_data['original_query'],
+            filters=filters,
+            ranges=ranges,
+            field_order=validated_data['sortby'],
+            aggs=self.FILTER_FIELDS,
```
reupen: Does aggs need to be passed if `self.aggregations` is false(y)?

elcct: This is good point - it originally meant to only aggregate fields specified in `self.aggregates`, it looks like I forgot to change it to just use `self.FILTER_FIELDS` halfway through

datahub/search/serializers.py
elcct: Nice. Updated.

datahub/search/views.py
elcct: Changed so that only paging parameters are being looked into and copied if don't exist in request body.

datahub/search/views.py
elcct: Changed to `include_aggregations`.

datahub/search/views.py
elcct: Changed to `include_aggregations` boolean.

datahub/search/views.py
```diff
@@ -111,7 +111,11 @@ def post(self, request, format=None):
         data = request.data.copy()
 
         # to support legacy paging parameters that can be in query_string
-        data.update(request.query_params)
+        for legacy_query_params in ('limit', 'offset',):
```
reupen: s should probably be dropped from legacy_query_params.

PR #456, opened 2017-09-08
Add OMIS order status

marcofucci:
This:
- adds a new `order.status` char field
- adds `OrderStatus` enum constants
- sets status to `draft` when an order is created
- sets status to `quote_awaiting_acceptance` when a quote is created
- sets status back to `draft` if an order is reopened
- adds validation logic that you can create a quote or reopen an order only if the order is in the right state
- includes the status value in the response body


datahub/search/omis/models.py
```diff
@@ -11,6 +11,7 @@ class Order(DocType, MapDBModelToDict):
 
     id = String(index='not_analyzed')
     reference = String(analyzer='lowercase_keyword_analyzer')
+    status = String(analyzer='lowercase_keyword_analyzer')
```
elcct: You could use `CaseInsensitiveKeywordString` from `dsl_utils`.

marcofucci: 👍  missed that

datahub/omis/order/constants.py
```diff
@@ -0,0 +1,44 @@
+from enum import Enum
+from functools import lru_cache
+
+from datahub.core.constants import Constant
+
+
+class OrderStatus(Enum):
+    """Order Status."""
+
+    draft = Constant('Draft', 'draft')
```
reupen: Just wondering if we need Constant here or could use the built-in Enum features ([\_\_members\_\_](https://docs.python.org/3.6/library/enum.html#iteration) etc.)?

datahub/omis/order/test/test_models.py
```diff
@@ -186,6 +271,24 @@ def test_with_already_cancelled_quote(self):
...
+            OrderStatus.paid,
+            OrderStatus.complete,
+            OrderStatus.cancelled,
+        )
+    )
+    def test_fails_if_order_not_in_allowed_status(self, disallowed_status):
+        """Test that if the order is in a disallowed status, it cannot be reopened."""
+        order = OrderFactory(status=disallowed_status.value.id)
+        with pytest.raises(Conflict):
+            order.generate_quote(by=None)
```
reupen: Should this be calling `order.reopen()`?

datahub/omis/order/test/test_validators.py
```diff
@@ -254,3 +256,95 @@ def test_with_cancelled_quote(self):
...
+        )
+
+        validator.set_instance(order)
+
+        try:
+            validator()
+        except Exception:
+            pytest.fail('Should not raise a validator error')
+
+    def test_fails_with_allowed_statuses_as_enums(self):
```
reupen: Could consider using parametrize with these three tests?

datahub/omis/order/validators.py
```diff
@@ -125,3 +125,31 @@ def __call__(self, data=None):
...
+
+    message = 'The action cannot be performed in the current status {0}.'
+
+    def __init__(self, allowed_statuses):
+        """Constructor."""
+        self.allowed_statuses = allowed_statuses
+
+    def set_instance(self, instance):
+        """Set the current instance."""
+        self.instance = instance
```
reupen: I think it would be good to set `self.instance` to None in \_\_init\_\_() so that the attribute always exists.

datahub/omis/order/constants.py
marcofucci: yeah I started that way and I'm not 100% convinced `Constant` is right.

However, `OrderStatus.quote_accepted` returning `Quote accepted` doesn't seem right either.

The value of `OrderStatus.quote_accepted` should really be `quote_accepted` so I'd expect `OrderStatus.quote_accepted` to return `quote_accepted` in that case.

Don't you think?

datahub/omis/order/test/test_models.py
marcofucci: 👍  forgot

datahub/omis/order/test/test_validators.py
marcofucci: very nice, thanks!

datahub/omis/order/validators.py
marcofucci: true, done

datahub/omis/order/constants.py
reupen: Not sure I'm following. OrderStatus.quote_accepted would be an OrderStatus instance, it'd have name and value attributes, no?

marcofucci: yeah sorry that's what I meant.
```
> OrderStatus.quote_accepted.name
> quote_accepted

> OrderStatus.quote_accepted.value
> Quote accepted
```

`OrderStatus.quote_accepted.value` should return the value, not a description so it should return `quote_accepted`.

reupen: You could do this if you want to add additional attributes:
https://docs.python.org/3.6/library/enum.html#planet

Works out similar to using Constant but without having to go `.value`. (`self.id = self.name` works so the IDs don't need to be repeated.)

Or Choices from model_utils would probably do the job too.

marcofucci: Yes I was about to say, I am going to use `Choices`. I wanted to keep things consistent byt we are probably going to get rid of Constants.

marcofucci:
@reupen it looks a lot better now with the `Choices`, have a look

PR #458, opened 2017-09-10
Only allow editing an order when in draft

marcofucci:
This disallow people from editing an order when it's not in draft including when changing subscribers or assignees.

### DIFF
--- a/datahub/omis/order/test/test_validators.py
+++ b/datahub/omis/order/test/test_validators.py
@@ -43,7 +43,7 @@ class TestContactWorksAtCompanyValidator:
                 'company': company
             })
         except Exception:
-            pytest.fail('Should not raise a validator error')
+            pytest.fail('Should not raise a validator error.')
 
     def test_contact_not_from_company(self):
         """
@@ -83,7 +83,7 @@ class TestContactWorksAtCompanyValidator:
                 'main_company': company
             })
         except Exception:
-            pytest.fail('Should not raise a validator error')
+            pytest.fail('Should not raise a validator error.')
 
 
 class TestReadonlyAfterCreationValidator:
@@ -104,7 +104,7 @@ class TestReadonlyAfterCreationValidator:
                 'field2': 'some value',
             })
         except Exception:
-            pytest.fail('Should not raise a validator error')
+            pytest.fail('Should not raise a validator error.')
 
     def test_cannot_change_after_creation(self):
         """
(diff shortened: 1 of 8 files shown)


datahub/omis/order/test/test_validators.py
```diff
@@ -298,3 +298,45 @@ def test_validation_fails(self):
...
+                OrderStatus.cancelled
+            ),
+            order_required=False
+        )
+        validator.set_instance(None)
+
+        try:
+            validator()
+        except Exception:
+            pytest.fail('Should not raise a validator error')
```
elcct: Should end with a dot?

datahub/omis/order/validators.py
```diff
@@ -144,17 +144,38 @@ class OrderInStatusValidator:
...
+        This hook is called by the serializer instance, prior to the validation call being made.
+
+        It sets self.instance searching for an order in the following order:
+            - serializer.context['order]
+            - serializer.instance
+        """
+        if 'order' in serializer.context:
+            self.set_instance(serializer.context['order'])
+        else:
+            self.set_instance(getattr(serializer, 'instance', None))
```
reupen: Do we need getattr() here as self.instance is set in BaseSerializer.\_\_init\_\_()?

marcofucci: It's more for covering the case where `serializer.instance` doesn't exist. 
Are you saving that it would be better:
```
elif hasattr(serializer, 'instance'):
   self.set_instance(serializer.instance)
```

reupen: I was more questioning why it wouldn't exist, since the attribute should exist on all serialisers I thought?

marcofucci: True, that line is from the DRF codebase so there might be cases were it wouldn't exit I guess.

reupen: The docs don't though :)
http://www.django-rest-framework.org/api-guide/validators/#using-set_context

Not a big deal, but if don't why why it wouldn't exist then assuming None might not necessarily be safe.

marcofucci: In case of updates `serializer.instance` doesn't exist I believe

reupen: It should exist but be None. `BaseSerializer.__init__()` always sets self.instance.

marcofucci: hm I guess you are right. It doesn't exist on a serializer Field but it should always exist in a Serializer. Going to remove the check.

PR #459, opened 2017-09-10
Add Purchase Order Number for OMIS orders

marcofucci:
This:
- adds a new `Order.po_number` field
- adds the ability to specify a value when creating or patching
- includes the field in the response body


datahub/omis/order/test/views/test_order_details.py
```diff
@@ -123,6 +125,7 @@ def test_success_minimal(self):
         assert response.json()['description'] == ''
         assert response.json()['contacts_not_to_approach'] == ''
         assert response.json()['delivery_date'] is None
+        assert response.json()['po_number'] == ''
```
elcct: Do we need to call `response.json()` for each assert?

datahub/omis/order/models.py
```diff
@@ -111,6 +111,8 @@ class Order(BaseModel):
         on_delete=models.SET_NULL
     )
 
+    po_number = models.CharField(max_length=100, blank=True)
```
elcct: `po_number` is really going to be max 100 characters long?

marcofucci: The CDMS field is 100 char long so I think defining as 100 here as well is a no brainer.

datahub/omis/order/test/views/test_order_details.py
marcofucci: `response.json()` is cached so there's no harm in doing that

datahub/omis/order/models.py
elcct: Didn't know. Great.

PR #463, opened 2017-09-12
Add OMIS HourlyRate

marcofucci:
This adds the table `HourlyRate` and the fk `Order.hourly_rate`.

The fk has a default value which has to be in the db so the migration creates it before defining it.

Also, fixtures for the HourlyRate values have been added and included in the `load_omis_metadata` command.

The `vat_value` has been defined as a `DecimalField` as any other type seemed weird whilst the `rate_value` is an int holding money values in pence.


datahub/omis/order/models.py
```diff
@@ -23,6 +26,28 @@ class ServiceType(BaseOrderedConstantModel, DisableableModel):
     """
 
 
+class HourlyRate(BaseConstantModel):
+    """
+    Values for the hourly rates used to calculate order pricing and for the
+    current VAT to apply.
+    """
+
```
reupen: Just wondering how useful the name field really is – looks like it could probably be generated from the other fields?

datahub/omis/order/models.py
```diff
@@ -23,6 +26,28 @@ class ServiceType(BaseOrderedConstantModel, DisableableModel):
...
+    created_on = models.DateTimeField(auto_now_add=True)
+    modified_on = models.DateTimeField(auto_now=True)
+
+    rate_value = models.PositiveIntegerField(
+        help_text='Rate in pence. E.g. 1 pound should be stored as 100 (100 pence).'
+    )
+    vat_value = models.DecimalField(
```
reupen: Just wondering if there was a specific reason for using integer for one of these fields and decimal for the other?

datahub/omis/order/models.py
marcofucci: That field is from CDMS and I agree it could be dynamic but I think it's useful to keep it this way as they might want a more descriptive name if they are gonna allow people to choose between a few of them.

datahub/omis/order/models.py
marcofucci: `rate_value` is a money value whilst `vat_value` is a _percentage_ value (between 0 and 100) so they are different types really. 

Using ints for percentages doesn't look right.

I could potentially use decimals for `rate_value` as well. I guess ints for money makes moving data from/to json API slightly easier though.
What do you think?

reupen: If you use Decimal it should get serialised as a string. So it might depend on what the front end is doing with it. (Of course in JS normal numbers are floats.)

marcofucci: Yeah I'd like to keep money as integers.

PR #466, opened 2017-09-13
Add VAT fields to OMIS orders

marcofucci:
This:
- adds the fields:
	- `vat_status` = (eu, outside_eu or uk)
	- `vat_number` (only if vat_status is eu)
	- `vat_verified` (only if vat_status is eu)
- validates that
	- if vat_status = eu, vat_verified has to be specified (can be True or False, by default it's null)
	- if vat_status = eu and vat_verified = True, vat_number has to be specified as well

The validation is triggered only when generating a quote to allow people to fill in an order incrementally based on the information they have at the time and validate just at the end when moving to a new state.


datahub/omis/order/constants.py
```diff
@@ -15,3 +15,10 @@
 
 
 DEFAULT_HOURLY_RATE = '7e1ca5c3-dc5a-e511-9d3c-e4115bead28a'
+
+
+VatStatus = Choices(
+    ('eu', 'Europe'),
```
reupen: Should this be European Union? And doesn't it overlap with UK?

datahub/omis/order/serializers.py
```diff
@@ -133,6 +137,23 @@ def validate_primary_market(self, country):
...
+        vat_status = data_combiner.get_value('vat_status')
+        if vat_status and vat_status != VatStatus.eu:
+            data['vat_number'] = ''
+            data['vat_verified'] = None
+
+        return data
+
+    def validate(self, data):
+        """Add extra logic to the default DRF one."""
+        data = super().validate(data)
```
reupen: Not sure the super call is that helpful here, the DRF examples don't do that:
http://www.django-rest-framework.org/api-guide/serializers/#object-level-validation

datahub/omis/order/validators.py
```diff
@@ -71,6 +74,50 @@ def __call__(self, data):
                     })
 
 
+class VATValidator:
```
reupen: Why VatStatus but VATValidator? :)

datahub/omis/order/constants.py
marcofucci: I'll change it to _EU_ if I'm using _UK_ it makes more sense.

Re. the logic:
- if the company is in the UK => we always apply VAT
- if the company is in EU but not UK => if we can verify the VAT number, we don't apply VAT otherwise we do apply it
- if the company is outside EU => we never apply VAT

I guess I could replace _'Europe'_ with (_'EU excluding UK'_)

datahub/omis/order/serializers.py
marcofucci: 👍  I had the same feeling but added just to be safe... I'm going to remove it.

datahub/omis/order/validators.py
marcofucci: Ha! I'll go with VATStatus then.

PR #467, opened 2017-09-13
Add OAuth scopes

reupen:
This does a couple of things:

* Adds the ability to restrict scopes for OAuth applications. The background to this is covered in https://github.com/evonove/django-oauth-toolkit/pull/395
* Defines a scope for the `v3/investment/from` endpoint.

The purpose of this is to be able to add client credentials OAuth apps with restricted scope for pure machine-to-machine apps (where users aren't involved).

The Application model in django-oauth-toolkit is swappable, but it would be a hassle to swap it out now so I went with an additional model. There is a question of where to put that model – I went with the core app, though I guess it could go in a new app. There will probably be app-specific scopes coming natively to django-oauth-toolkit at some point so that part of the functionality could probably be removed if/when that happens.


datahub/core/models.py
```diff
@@ -95,3 +97,20 @@ class BaseOrderedConstantModel(BaseConstantModel):
...
         abstract = True
         ordering = ('order', )
+
+
+class OAuthApplicationScope(models.Model):
+    """Allowed scopes for an OAuth application."""
+
+    application = models.OneToOneField(oauth2_settings.APPLICATION_MODEL, on_delete=models.CASCADE)
+    scopes = ArrayField(models.CharField(
+        max_length=settings.CHAR_FIELD_MAX_LENGTH),
```
elcct: That break looks odd.

datahub/investment/views.py
```diff
@@ -111,6 +114,8 @@ class IProjectModifiedSinceViewSet(IProjectViewSet):
     """View set for the modified-since endpoint (intended for use by Data Hub MI)."""
 
     pagination_class = _SinglePagePaginator
+    permission_classes = [AnyOfChainer(IsAuthenticated, TokenHasScope)]
```
marcofucci: Why do you need `IsAuthenticated` here?

reupen: Ah, it's complicated.

It's currently the default permission class:

https://github.com/uktrade/data-hub-leeloo/blob/19fc66b44b1ca021d92b91eb3039bf5fc399aadc/config/settings/common.py#L168

I wanted the browseable API to still work. django-oauth-toolkit provides a class `IsAuthenticatedOrTokenHasScope` which does that. However, for front-end requests it checks the scope since they are using OAuth. Although this endpoint isn't used by the front end, I still didn't want to interfere with front-end requests to keep things simple (in case we later have endpoints used both by the front end and other machine-to-machine services).

Using `IsAuthenticatedOrTokenHasScope` and making `ApplicationScopesBackend` default to all scopes if there is no `OAuthApplicationScope` entry here would've also done the same thing, but I thought keeping the scopes away from the front end was simpler.

elcct:
@reupen Can you test the case when token has expired and needs to be renewed?

datahub/investment/views.py
elcct: Wouldn't be better to give users required scopes and use one permission class?

marcofucci: I agree with @elcct, having different behaviours is confusing.

If we are going to start using oauth2 for machine to machine communications and token based permissions, we should extend the token permissions to all the endpoints to make it consistent.

config/settings/common.py
```diff
@@ -135,6 +135,15 @@
     'django.contrib.auth.backends.ModelBackend'
 ]
 
+# django-oauth-toolkit settings
+
+OAUTH2_PROVIDER = {
+    'SCOPES': {
+        'investment:read': 'Ability to read investment projects',
```
marcofucci: I'm not sure 'investment:read' is a good name for the scope, it doesn't look like a generic read in your case. You can't read from the other investment endpoints with this scope so it might be misleading.

marcofucci:
If would be good to:
- update the README with how to set up this new oauth2 application
- maybe test that the token with client credentials does not have access to an endpoint with default permissions
- maybe use a separate django app

reupen:
Will replace with a new PR.

PR #468, opened 2017-09-13
[DH-603] Search export endpoints.

elcct:
Adds new endpoints that stream all search results in CSV format.

For example `/search/company/export`.

Endpoint accepts the same parameters as filtered search except that `offset` and `limit` parameters are being ignored.


datahub/search/company/test/test_views.py
```diff
@@ -179,3 +181,76 @@ def test_companies_no_term(self, setup_es, setup_data):
         response = self.api_client.get(url, {})
 
         assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+
+class TestSearchExport(APITestMixin):
+    """Tests search export views."""
+
+    def test_search_company(self, setup_es, setup_data):
```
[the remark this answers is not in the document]
elcct: Good point, changed to `test_company_export`.

datahub/search/company/test/test_views.py
```diff
@@ -179,3 +181,76 @@ def test_companies_no_term(self, setup_es, setup_data):
...
+        })
+
+        assert response.status_code == status.HTTP_200_OK
+
+        filename = slugify(term)
+
+        # checks if filename includes our search term
+        assert filename in response['Content-Disposition']
+
+        csv_file = csv.reader(line.decode('utf-8') for line in response.streaming_content)
```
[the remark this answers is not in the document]
elcct: I didn't think it will make much difference for this test, but it actually looks cleaner, so changed.

datahub/search/utils.py
```diff
@@ -0,0 +1,6 @@
+class Echo(object):
```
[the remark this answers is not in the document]
elcct: True. Removed.

datahub/search/views.py
```diff
@@ -155,3 +159,57 @@ def post(self, request, format=None):
...
 
         return Response(data=response)
+
+
+class SearchExportAPIView(SearchAPIView):
+    """Returns CSV file with all search results."""
+
+    def format_cell(self, cell):
+        """Gets cell or name from cell and flattens the cell if necessary."""
+        if isinstance(cell, dict):
```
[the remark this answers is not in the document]
elcct: Done.

datahub/search/views.py
```diff
@@ -155,3 +159,57 @@ def post(self, request, format=None):
...
+            if 'name' in cell:
+                return cell['name']
+
+        if isinstance(cell, list):
+            return ','.join([self.format_cell(item) for item in cell])
```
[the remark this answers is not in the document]
elcct: I keep forgetting that! Thanks :)

datahub/search/views.py
```diff
@@ -155,3 +159,57 @@ def post(self, request, format=None):
...
+
+        return cell
+
+    def get_row(self, query):
+        """Gets formatted row of results."""
+        row = 0
+        # we want to keep the same order of rows that is in the search results
+        query.params(preserve_order=True)
+        for hit in query.scan():
```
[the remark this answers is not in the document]
elcct: Indeed. Fixed.

datahub/search/urls.py
```diff
@@ -13,6 +13,15 @@
...
+            rf'^search/{search_app.name}$',
+            search_app.view.as_view(),
+            name=search_app.name
+        ),
+        url(
+            rf'^search/{search_app.name}/export$',
+            search_app.export_view.as_view(),
+            name=f'{search_app.name}-export'
+        )
+    ])
```
marcofucci: This should be refactored a bit, it should define the search route if `search_app.view` exists and it should define the search export route if `search_app.export_view` exists. 

At the moment skips both and doesn't check for export_view.

datahub/search/omis/apps.py
```diff
@@ -12,6 +12,7 @@ class OrderSearchApp(SearchApp):
     name = 'order'
     ESModel = Order
     view = SearchOrderAPIView
+    export_view = SearchOrderExportAPIView
```
marcofucci: We don't have any user needs that requires an export to CSV yet (maybe in the future) so the export functionality should not be implemented for OMIS orders.

datahub/search/views.py
```diff
@@ -155,3 +159,55 @@ def post(self, request, format=None):
...
+            entity=self.entity,
+            term=validated_data['original_query'],
+            filters=filters,
+            ranges=ranges,
+            field_order=validated_data['sortby'],
+        )
+
+        filename = slugify(f"{self.entity.__name__}-{validated_data['original_query']}")
+
+        writer = csv.writer(Echo())
```
reupen: I'd consider using DictWriter since there's less scope for things to go wrong.

datahub/search/urls.py
elcct: Good catch, I missed that :( Fixed.

datahub/search/omis/apps.py
elcct: I have removed it. Thanks.

datahub/search/views.py
```diff
@@ -155,3 +159,55 @@ def post(self, request, format=None):
...
+
+        results = elasticsearch.get_search_by_entity_query(
+            entity=self.entity,
+            term=validated_data['original_query'],
+            filters=filters,
+            ranges=ranges,
+            field_order=validated_data['sortby'],
+        )
+
+        filename = slugify(f"{self.entity.__name__}-{validated_data['original_query']}")
```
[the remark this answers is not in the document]
elcct: The ticket only mentions adding `original_query` to the file name and example also has entity name as a prefix. "all" is misleading when using filters - I would rather leave it blank and this could be amended later.

datahub/search/company/test/test_views.py
```diff
@@ -179,3 +181,78 @@ def test_companies_no_term(self, setup_es, setup_data):
...
+                'website',
+                'trading_address_1',
+                'trading_address_2',
+                'trading_address_town',
+                'trading_address_county',
+                'trading_address_postcode',
+                'headquarter_type',
+                'classification',
+                'parent',
+                'one_list_account_owner'} == set(rows[0].keys())
```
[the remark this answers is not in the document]
elcct: Order at the moment is not defined, that's why I am not checking it.

datahub/search/views.py
elcct: True, I replaced it with DictWriter.

datahub/search/company/test/test_views.py
```diff
@@ -179,3 +181,78 @@ def test_companies_no_term(self, setup_es, setup_data):
...
+
+        # checks if we have a company we look for in the CSV file
+        data_row = ','.join(rows[0].values())
+        assert 'abc defg' in data_row
+        assert 'United States' in data_row
```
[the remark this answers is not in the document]
elcct: Good point. Fixed.

datahub/search/views.py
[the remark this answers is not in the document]
reupen: Actually it says 'search data criteria' which isn't really defined (I mean that could include filters). Yep good point I forgot about the filters.

Still I think it's good to have 'data-hub' and probably the date in there so the user can identify where the file is from later on and how old it is. We can take this to Jira...

datahub/search/company/test/test_views.py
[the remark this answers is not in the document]
reupen: If we don't have an order to use then we could just sort them alphabetically. At least they'd not be guaranteed to change order between exports.

datahub/search/views.py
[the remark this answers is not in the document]
elcct: I am happy to add data-hub. In terms of date, there is too many questions (just day? hours, seconds? which time zone? am/pm or 24h etc) and you are likely to have date in your file meta data.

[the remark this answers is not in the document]
reupen: Date to me would be e.g. 2017-01-01, but yeah you're right we don't know the user's date since we don't know the time zone.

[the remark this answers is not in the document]
elcct: I updated Jira for clarification.

datahub/search/company/test/test_views.py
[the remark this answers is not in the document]
elcct: That makes sense. Done.

elcct:
@reupen regarding mixed up columns - use of DictWriter fixed that.

reupen:
I'm not getting the contents of companies_house_data through correctly – might just be easier to omit it?

datahub/search/views.py
```diff
@@ -155,3 +161,75 @@ def post(self, request, format=None):
...
+
+    def _clean_fieldnames(self, fieldnames):
+        """Remove special fields from the export."""
+        return [field for field in fieldnames
+                if not any(field.endswith(suffix) for suffix in self.IGNORED_SUFFIXES)]
+
+    def _format_cell(self, cell):
+        """Gets cell or name from cell and flattens the cell if necessary."""
+        if isinstance(cell, dict):
+            if 'name' in cell:
```
reupen: Looks a bit odd here – the ifs could be combined, or there might be a missing return.

datahub/search/company/test/test_elasticsearch.py
```diff
@@ -281,11 +281,13 @@ def test_get_search_by_entity_query():
             'lte': date
         }
     }
-    query = elasticsearch.get_search_by_entity_query(
-        term='test',
-        filters=filters,
-        ranges=ranges,
-        entity=ESCompany,
+    query = elasticsearch.limit_search_query(
+        elasticsearch.get_search_by_entity_query(
```
reupen: Might be nice to unnest that function call.

datahub/search/interaction/apps.py
```diff
@@ -1,7 +1,8 @@
 from datahub.interaction.models import Interaction as DBInteraction
 from datahub.search.apps import SearchApp
 from datahub.search.interaction.models import Interaction
-from datahub.search.interaction.views import SearchInteractionAPIView
+from datahub.search.interaction.views import SearchInteractionAPIView, \
```
reupen: `(` and `)` for continuation here please :)

datahub/search/views.py
elcct: Yes, I am working on that now.

datahub/search/interaction/apps.py
elcct: Keep forgetting that :) Thanks

datahub/search/company/test/test_elasticsearch.py
elcct: Done.

datahub/search/views.py
elcct: Added all cases for that dict now.

datahub/search/company/test/test_views.py
```diff
@@ -179,3 +187,160 @@ def test_companies_no_term(self, setup_es, setup_data):
...
+
+    @staticmethod
+    def _get_random_constant_id(constant):
+        """Gets random constant id."""
+        return random.choice(list(constant)).value.id
+
+    @staticmethod
+    def _get_random_list_of_constant_ids(constant, max=10):
+        """Gets list of random constant ids."""
+        return {TestSearchExport._get_random_constant_id(constant) for _ in range(max)}
```
marcofucci: I'd prefer `self....` instead of `TestSearchExport....`

datahub/search/company/test/test_views.py
```diff
@@ -179,3 +187,160 @@ def test_companies_no_term(self, setup_es, setup_data):
...
+            data['uk_region_id'] = TestSearchExport._get_random_constant_id(constants.UKRegion)
+        if archived:
+            data.update({
+                'archived_by': AdviserFactory(),
+                'archived_on': factory.Faker('date'),
+                'archived_reason': factory.Faker('text'),
+            })
+
+        company = CompanyFactory(
+            **data
```
marcofucci: Instead of manually populating the CompanyFactory why don't you update the Factory definition in the first place which should in theory set random values?

datahub/search/company/test/test_views.py
```diff
@@ -179,3 +187,160 @@ def test_companies_no_term(self, setup_es, setup_data):
...
+                'trading_address_county',
+                'trading_address_postcode',
+                'trading_address_town',
+                'turnover_range',
+                'uk_based',
+                'uk_region',
+                'website'] == csv_file.fieldnames
+
+        for row in rows:
+            company = Company.objects.get(pk=uuid.UUID(row['id']))
```
marcofucci: You don't have do use `uuid.UUID(...)`, `pk=row['id']` is just enough

datahub/search/company/test/test_views.py
elcct: I had a feeling that it is too much hassle to achieve what I wanted through Factory definition and that would affect entire project, probably slowing tests from creating all those extra models and adding unknown other side effects.

datahub/search/company/test/test_views.py
elcct: Changed.

datahub/search/company/test/test_views.py
elcct: That's right. Changed.

PR #469, opened 2017-09-15
Add OMIS pricing

marcofucci:
This:
- adds pricing fields to the order model
- updates those values when an order or an assignee is saved
- updates the order endpoints to return those values
- updates the search to include those fields. Only the total_cost is indexed as it will be used for searching in the near future
- updates the content of the quote to include the pricing values


datahub/omis/order/pricing.py
```diff
@@ -0,0 +1,138 @@
...
+    if order.vat_status == VATStatus.uk:
+        return True
+    if order.vat_status == VATStatus.outside_eu:
+        return False
+
+    assert order.vat_status == VATStatus.eu
+    return not order.vat_verified
+
+
+def _calculate_pricing(estimated_time, hourly_rate, vat_value, discount_value):
```
reupen: It's hard to follow what the units are for these variables, could append to the names or add to the docstring?

datahub/omis/order/pricing.py
```diff
@@ -0,0 +1,138 @@
...
+    Calculate the pricing using the given base values.
+
+    :returns: pricing values
+    """
+    if estimated_time == 0:
+        return ZERO_PRICING
+
+    net_cost = round((estimated_time / 60) * hourly_rate)
+    subtotal_cost = max(net_cost - discount_value, 0)
+    vat_cost = round(float(vat_value) * subtotal_cost * 0.01)
```
reupen: Seems like the rounding is doing two things:
1) Killing floating-point inaccuracies
2) Rounding fractional pennies

Is that right? For 2) is the rounding right in terms of VAT?

datahub/omis/order/test/test_pricing.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        Test that if the order doesn't have the right VAT fields populated,
+        it raises a ValidationError.
+        """
+        order = Order(vat_status=None)
+
+        with pytest.raises(ValidationError):
+            should_vat_be_applied(order)
+
+    def test_outside_eu_shouldnt(self):
```
reupen: Could consider paramet(e)rise these.

datahub/omis/order/models.py
```diff
@@ -149,10 +149,24 @@ class Order(BaseModel):
     )
     discount_value = models.PositiveIntegerField(default=0)
     discount_label = models.CharField(max_length=100, blank=True)
+
     vat_status = models.CharField(max_length=100, choices=VATStatus, blank=True)
     vat_number = models.CharField(max_length=100, blank=True)
     vat_verified = models.NullBooleanField()
 
+    net_cost = models.PositiveIntegerField(
+        default=0, help_text='Tot hours * hourly rate in pence.'
```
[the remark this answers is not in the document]
marcofucci: Ha, true

datahub/omis/order/pricing.py
```diff
@@ -0,0 +1,138 @@
...
+def can_pricing_be_calculated(order):
+    """
+    :returns: True if the pricing for `order` can be calculated
+        (e.g. all the fields required are filled in)
+    """
+    try:
+        _validate_vat(order)
+    except ValidationError:
+        return False
+    else:
```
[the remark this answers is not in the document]
marcofucci: 👍

datahub/omis/order/pricing.py
marcofucci: 👍  for docstring

datahub/omis/order/pricing.py
marcofucci: That's correct.

For the VAT, `vat_value` is a `Decimal` so I need to convert it to float and then round up or down the final value.
I'm make sure our PO checks the values, I'm not too worried as the error could be max 1p and they can see the price before proceeding but better to double-check.

datahub/omis/order/test/test_pricing.py
marcofucci: yep good point

PR #470, opened 2017-09-15
Add UK region to events

reupen:


marcofucci:
It looks good but some tests are failing

PR #474, opened 2017-09-18
Add OMIS Order.public_token field

marcofucci:
This adds a new field ~`public_id`~ `public_token` to the order model used to access the order via the public facing OMIS app.

The field is a random string composed of 50 uppercase, lowercase letters, digits and symbols.
A notification with a link to the public facing OMIS app is sent when a quote is generated.

As the public_token field is mandatory, this creates 3 migrations:

1. adds the field as optional
2. populates all the existing records
3. makes the field required


datahub/omis/order/models.py
```diff
@@ -206,18 +211,34 @@ def gen():
...
         return generate_reference(model=cls, gen=gen)
 
+    @classmethod
+    def generate_public_id(cls):
+        """
+        :returns: a random unused public id of form
+            <50 uppercase/lowercase letters and numbers>
+        :raises RuntimeError: if no public_id can be generated
+        """
+        gen = partial(get_random_string, length=50)
```
reupen: I don't see get_random_string() in the Django docs – do we want to use it over secrets.token_urlsafe()?

https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe

marcofucci: `get_random_string` was introduced many many years ago and it's not going away.

I'm using it for all the other references.
It lets you control the format of the generated string which is quite handy for all other use cases. 

I could you `token_urlsafe` just for this instance though as I don't care about the format.

marcofucci: I also don't like the name of the field `public_id`. 
What do you think about `public_token`?

reupen: > I'm using it for all the other references.
Yeah, but this is a token for a URL isn't it? I.e. different from order numbers etc.

>  What do you think about public_token or even client_token?
public_token sounds good to me :+1:

marcofucci: 👍  to both, thanks!

datahub/omis/order/models.py
```diff
@@ -206,18 +212,34 @@ def gen():
 
         return generate_reference(model=cls, gen=gen)
 
+    @classmethod
+    def generate_public_token(cls):
+        """
+        :returns: a random unused public token of form
+            <50 uppercase/lowercase letters, digits and symbols>
```
reupen: Docstring to update here

marcofucci: What do you mean?

reupen: Yep sorry, it's correct, just misread it :+1:

PR #475, opened 2017-09-19
[DH-655] Event filters 

elcct:
This adds following filters to event: name, organiser name, address country, start date.

### DIFF
--- a/datahub/search/dsl_utils.py
+++ b/datahub/search/dsl_utils.py
@@ -22,6 +22,18 @@ def contact_or_adviser_mapping(field, include_dit_team=False):
     return Nested(properties=props)
 
 
+def contact_or_adviser_partial_mapping(field):
+    """Mapping for Adviser/Contact fields that allows partial matching."""
+    props = {
+        'id': KeywordString(),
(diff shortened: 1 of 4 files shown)


datahub/search/dsl_utils.py
```diff
@@ -22,6 +22,18 @@ def contact_or_adviser_mapping(field, include_dit_team=False):
     return Nested(properties=props)
 
 
+def contact_or_adviser_partial_mapping(field):
+    """Mapping for Adviser/Contact fields that allows partial matching."""
+    props = {
+        'id': KeywordString(),
```
marcofucci: should this be `index='no'` instead of `not_analyzed`?
We are probably never gonna search for a related id, no?

marcofucci: Ignore me, we might want to filter based on it

elcct: We do search other items by `id`, I could see the case when we would want to find event by exact organiser - for example if we get to have auto complete dropdown.

elcct: I didn't see your second comment when was replying to your first one :)

PR #476, opened 2017-09-19
Add billing address to OMIS orders

marcofucci:
This:
- adds billing name, email, phone and address to OMIS orders
- validates the billing address only if any of the billing address fields is set
- adds those fields to the order endpoints during creation and updates
- adds those fields to the search mapping


datahub/omis/order/serializers.py
```diff
@@ -43,6 +44,8 @@ class OrderSerializer(serializers.ModelSerializer):
 
     delivery_date = serializers.DateField(required=False, allow_null=True)
 
+    billing_address_country = NestedRelatedField(Country, required=False)
```
reupen: Should we have allow_null=True?

datahub/omis/order/models.py
```diff
@@ -173,6 +175,22 @@ class Order(BaseModel):
         default=0, help_text='Subtotal + VAT cost in pence.'
     )
 
+    billing_contact_name = models.CharField(max_length=MAX_LENGTH, blank=True)
+    billing_email = models.EmailField(max_length=MAX_LENGTH, blank=True)
+    billing_phone = models.CharField(max_length=150, blank=True)
```
reupen: Not a big deal, but `billing_telephone_number` or `billing_telephone` would probably be more consistent with other models.

datahub/omis/order/validators.py
```diff
@@ -293,3 +293,80 @@ def __call__(self, data=None):
             raise Conflict(
                 self.message.format(self.instance.get_status_display())
             )
+
+
+class AddressValidation:
```
reupen: AddressValidator?

datahub/omis/order/models.py
marcofucci: Hm I should have changed the `contact_phone` field as well though...

PR #478, opened 2017-09-19
[DH-667] Adds sortby start/end_date.

elcct:
As in the title.

### DIFF
--- a/datahub/search/event/tests/test_views.py
+++ b/datahub/search/event/tests/test_views.py
@@ -131,7 +131,7 @@ class TestSearch(APITestMixin):
 
     def test_search_event_date(self, setup_es, setup_data):
         """Tests start_date filter."""
-        start_date = datetime.datetime(2017, 7, 2)
+        start_date = datetime.date(2017, 7, 2)
         event = EventFactory(
             start_date=start_date,
         )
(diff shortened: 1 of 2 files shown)


datahub/search/event/tests/test_views.py
```diff
@@ -150,6 +150,77 @@ def test_search_event_date(self, setup_es, setup_data):
         assert len(response.data['results']) == 1
         assert response.data['results'][0]['id'] == str(event.id)
 
+    def test_search_event_sortby_start_date(self, setup_es, setup_data):
+        """Tests sort by start_date desc."""
+        start_date_a = datetime.datetime(2011, 9, 29)
```
reupen: Why do we need datetime for a date field?

datahub/search/event/tests/test_views.py
```diff
@@ -150,6 +150,77 @@ def test_search_event_date(self, setup_es, setup_data):
...
+            'original_query': '',
+            'start_date_after': start_date_a,
+            'start_date_before': start_date_b,
+            'sortby': 'start_date:desc',
+        })
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == 2
+        assert len(response.data['results']) == 2
+        assert response.data['results'][0]['start_date'] == start_date_b.strftime('%Y-%m-%d')
```
reupen: If we use a date for start_date_b, how about `start_date_b.isoformat()` instead of `start_date_b.strftime('%Y-%m-%d')`?

datahub/search/event/tests/test_views.py
elcct: Absolutely nothing :) Fixed

datahub/search/event/tests/test_views.py
elcct: That works too. Thanks!

PR #479, opened 2017-09-20
Add OAuth scopes (v2)

reupen:
Replaces https://github.com/uktrade/data-hub-leeloo/pull/467

* Adds the ability to specify and restrict scopes for OAuth applications
* Defines scopes for all views
* Fixes the login process for tests so that it no longer uses APIClient.force_authenticate()

Once merged and deployed we'll need to specify the scope for the front end as otherwise it'll get no scopes.


README.md
```diff
@@ -250,3 +250,19 @@ pip-compile --upgrade --output-file requirements.txt requirements.in
 This must be run whenever `requirements.in` is edited.
 
 Dependencies should still be installed using `requirements.txt`.
+
+# OAuth2 applications
+
+To set up access to the back end for the internal front end, in Django admin:
+
+* Add a new application (in section Django OAuth Toolkit) with client type 'Confidential' and authorisation grant type 'Resource owner password-based'
```
elcct: `(in Django OAuth Toolkit section)` would read better

datahub/oauth/scopes.py
```diff
@@ -0,0 +1,61 @@
...
+from datahub.core.utils import StrEnum
+from datahub.oauth.models import OAuthApplicationScope
+
+
+class Scope(StrEnum):
+    """Defined OAuth scopes."""
+
+    mi = 'mi'
+    internal_front_end = 'internal-front-end'
+    public_front_end = 'public-front-end'
```
marcofucci: I would probably call this `public_omis_front_end` as even if we had a public something else, we wouldn't want them to share the same scope and therefore permissions.

datahub/oauth/test/test_scopes.py
```diff
@@ -0,0 +1,129 @@
...
+from datahub.core.test_utils import APITestMixin
+from datahub.oauth.scopes import SCOPES_DESCS
+from datahub.oauth.test.factories import AccessTokenFactory, OAuthApplicationScopeFactory
+
+
+pytestmark = pytest.mark.django_db
+
+
+@mock.patch.dict(SCOPES_DESCS, {
+    'test_scope_1': 'Scope for testing 1.',
```
marcofucci: The fact that we are relying on this `'test_scope_1'` being exactly like the one in `datahub.core.test.support.views.MyDisableableModelViewset.required_scopes[0]` is a bit fragile.
E.g. if you change that required_scopes it's not obvious that a different app depends on it implicitly.

There might be a better way but I can only think of maybe using datahub.core.test.support.views.MyDisableableModelViewset.required_scopes[0] here instead of hardcoding the string.

datahub/core/test_utils.py
```diff
@@ -36,31 +39,33 @@ def user(self):
...
-        """
-        Login using the OAuth2 authentication.
+        """An API client with internal-front-end scope."""
+        return self.create_api_client()
 
-        1) Create an application instance, if necessary
-        2) Generate the token
-        3) Add the auth credentials to the header
-        """
+    def create_api_client(self, scope=Scope.internal_front_end, *additional_scopes):
```
marcofucci: This is always going to use `self.application` which always creates a `grant_type == Application.GRANT_PASSWORD`.

We should really test with Client Credentials in some cases (e.g. MI) and create a token without user.

config/settings/common.py
```diff
@@ -155,8 +163,12 @@
     'DEFAULT_VERSIONING_CLASS': 'rest_framework.versioning.NamespaceVersioning',
     'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',
     'PAGE_SIZE': 100,
-    'DEFAULT_AUTHENTICATION_CLASSES': ['oauth2_provider.contrib.rest_framework.OAuth2Authentication'],
-    'DEFAULT_PERMISSION_CLASSES': ['rest_framework.permissions.IsAuthenticated'],
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+        'oauth2_provider.contrib.rest_framework.OAuth2Authentication'
+    ],
+    'DEFAULT_PERMISSION_CLASSES': [
+        'oauth2_provider.contrib.rest_framework.IsAuthenticatedOrTokenHasScope'
```
marcofucci: I don't think this matches what we want.

We want:
1. you always have to have a valid token with valid scope for the resource
2. if you are authenticate, we want it to use the `IsAuthenticated` logic as well.

We don't want an OR otherwise you just need to be logged in to be able to access all the endpoint.

marcofucci:
It looks good! I added a couple of comments.

config/settings/common.py
reupen: It's not really an OR. If you're logged in using an OAuth token it validates the scopes. If you're logged in a session (e.g. using Django admin) it won't validate the scopes (e.g. when using the browseable API).

It always validates the scopes when using an OAuth token.

In production we SessionAuthentication is turned off. It's only on in other environments.

marcofucci: Oh yeah, you are right! 👍  the name `IsAuthenticated` is a bit misleading but it's OK, ignore me.

datahub/core/test_utils.py
```diff
@@ -61,24 +61,26 @@ def api_client(self):
         """An API client with internal-front-end scope."""
         return self.create_api_client()
 
-    def create_api_client(self, scope=Scope.internal_front_end, *additional_scopes):
+    def create_api_client(self, scope=Scope.internal_front_end, *additional_scopes,
+                          grant_type=Application.GRANT_PASSWORD):
         """Creates an API client associated with an OAuth token with the specified scope."""
         token = self.get_token(scope, *additional_scopes)
```
marcofucci: You forgot to pass in the `grant_type` ;-)

marcofucci:
Looks very good! 👍

PR #481, opened 2017-09-21
Add OMIS order get endpoint for public facing app

marcofucci:
This adds `GET /v3/omis/public/order/<public-token>` which returns the details of a order.

The response data is slightly different from the ordinary order endpoints as the public facing app needs less details.
Also, the nested company is expanded as the app needs extra data and adding an extra company endpoint is overkilled.

An order can only be accessed by its public token and only if its status is not in draft/cancelled.

### DIFF
--- a/datahub/omis/order/serializers.py
+++ b/datahub/omis/order/serializers.py
@@ -185,6 +185,65 @@ class OrderSerializer(serializers.ModelSerializer):
         return data
 
 
+class CompanyWithAddressSerializer(serializers.ModelSerializer):
+    """
+    Read-only DRF Serializer for Company with id, name an registered address.
+    """
+
+    registered_address_country = NestedRelatedField(Country)
+
+    class Meta:  # noqa: D101
+        model = Company
+        fields = (
+            'id',
+            'name',
+            'registered_address_1',
+            'registered_address_2',
+            'registered_address_county',
+            'registered_address_postcode',
+            'registered_address_town',
+            'registered_address_country',
+        )
+        read_only_fields = fields
+
+
+class PublicOrderSerializer(serializers.ModelSerializer):
+    """DRF serializer for public facing API."""
+
+    company = CompanyWithAddressSerializer()
+    contact = NestedRelatedField(Contact)
+    billing_address_country = NestedRelatedField(Country)
+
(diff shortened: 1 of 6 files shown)


datahub/omis/order/urls.py
```diff
@@ -15,7 +15,7 @@
 })
 
 
-urlpatterns = [
+internal_frontent_urls = [
```
reupen: Typo here :)

marcofucci: Ha! :grimacing:

datahub/core/serializers.py
```diff
@@ -102,6 +102,8 @@ def _value_for_json(val):
     """Returns a JSON-serialisable version of a value."""
     if isinstance(val, UUID):
         return str(val)
+    if all(hasattr(val, attr) for attr in ['__class__', 'id', 'name']):
+        return {'id': val.id, 'name': val.name}
```
reupen: Interesting that the UUID here serialises correctly without an str conversion?

datahub/omis/order/serializers.py
```diff
@@ -185,6 +185,54 @@ def validate(self, data):
         return data
 
 
+class PublicOrderSerializer(serializers.ModelSerializer):
+    """DRF serializer for public facing API."""
+
+    company = NestedRelatedField(
```
reupen: Interesting. I'm OK with this but a nested serialiser might've been an option since it's a bit complex and it's read-only.

datahub/core/serializers.py
elcct: Learned that recently

```
>>> id = uuid.uuid4()
>>> f'{id}'
'd965aac0-36c9-4963-a0fe-7b9192e7d106'
```

reupen: Yeah, but not with json.dumps().

elcct: True :) Thanks

datahub/omis/order/serializers.py
marcofucci: True

marcofucci:
@reupen @elcct I've undone the changes to `NestedRelatedField` and used a proper Serializer for the company field instead.
Please have another quick look.

PR #483, opened 2017-09-21
Add ability to redirect to specific email address

marcofucci:
In environments != live, it's handy to send all the notifications to a specific email address so that:
1. we don't accidentally send emails
2. we can test that the logic works


config/settings/common.py
```diff
@@ -198,6 +198,10 @@
 DATAHUB_FRONTEND_BASE_URL = env('DATAHUB_FRONTEND_BASE_URL', default='http://localhost:3000')
 
 # OMIS
+
+# if set, all the notifications will be sent to this address instead of the
+# intended recipient, useful for environments != live
+OMIS_NOTIFICATION_ROUTE_TO = env('OMIS_NOTIFICATION_ROUTE_TO', default='')
```
elcct: I would prefer `OMIS_NOTIFICATION_OVERRIDE_EMAIL` as `ROUTE_TO` is less descriptive.

marcofucci: How about `OMIS_NOTIFICATION_OVERRIDE_RECIPIENT` ? With _override email_ it's not clear if it's the sender or recipient's email being overridden.

elcct: Even better. How about `OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL` ?

marcofucci: 👌

marcofucci:
@elcct 👍  or 👎

PR #493, opened 2017-09-25
[DH-710] Adds default sort by created_on desc to contact and investme…

elcct:
…nt_project lists.

As in the title, list of contacts and investment projects is ordered by created_on desc.

### DIFF
--- a/datahub/company/test/test_contact_views.py
+++ b/datahub/company/test/test_contact_views.py
@@ -1,3 +1,5 @@
+from datetime import date
+
 import pytest
 import reversion
 from django.utils.timezone import now
@@ -526,13 +528,20 @@ class TestContactList(APITestMixin):
 
     def test_contacts_are_sorted_by_created_on_desc(self):
         """Test contacts are sorted by created on desc."""
-        contacts = ContactFactory.create_batch(5)
+        datetimes = [date(year, 1, 1) for year in range(2015, 2030)]
+        contacts = []
+
+        for creation_datetime in datetimes:
+            with freeze_time(creation_datetime):
+                contacts.append(
+                    ContactFactory()
+                )
 
         url = reverse('api-v3:contact:list')
         response = self.api_client.get(url)
 
         assert response.status_code == status.HTTP_200_OK
-        assert response.data['count'] == 5
+        assert response.data['count'] == len(contacts)
         response_data = response.json()['results']
 
         contacts = sorted(contacts, key=lambda key: key.created_on, reverse=True)
--- a/datahub/company/test/test_contact_views.py
+++ b/datahub/company/test/test_contact_views.py
@@ -1,3 +1,5 @@
+from datetime import date
+
 import pytest
 import reversion
 from django.utils.timezone import now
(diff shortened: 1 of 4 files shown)


datahub/investment/views.py
```diff
@@ -4,8 +4,11 @@
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 from django_filters import IsoDateTimeFilter
-from django_filters.rest_framework import DjangoFilterBackend, FilterSet
+from django_filters.rest_framework import (
```
reupen: Accidental change?

marcofucci:
Why we are not using search for these?

datahub/investment/views.py
elcct: Yes, thanks for spotting :)

elcct:
@marcofucci that's something to consider in the future as indeed search is capable of delivering those results. The ticket is only meant to fix acceptance tests.

datahub/company/test/test_contact_views.py
```diff
@@ -524,6 +526,27 @@ def test_all(self):
         assert response.status_code == status.HTTP_200_OK
         assert response.data['count'] == 5
 
+    def test_contacts_are_sorted_by_created_on_desc(self):
+        """Test contacts are sorted by created on desc."""
+        contacts = (
+            ContactFactory(created_on=date(2010, 1, 10)),
```
reupen: Does this work as expected? I think Django might be overriding created_on. Would need to call refresh_from_db() on the model instances to check though.

marcofucci: yeah it's better to check the `created_on` dates in your tests @elcct instead of the `ids`. 
Also mixing up the objects to create (e.g. 2011, 2008, 2012 ...) could help preventing other issues.

elcct: I should have double check that - indeed `created_on` here has no effect. I have removed it.

elcct: @marcofucci re ids, before comparing I am sorting list of objects by `created_on` in descending order, so comparing by id just ensures the result is also sorted. I could perhaps update the list with more differing dates, but created objects already have non equal dates.

marcofucci: I just meant that checking dates would have spotted the issue and using mixed creation dates would avoid potential other problems like some other sorting that shadow the `created_on` (e.g. sequential `id` if we were not using uuid).

However 👍  as it's a small detail and at the moment the current code does not have any of these problems.

datahub/company/test/test_contact_views.py
```diff
@@ -524,6 +524,21 @@ def test_all(self):
         assert response.status_code == status.HTTP_200_OK
         assert response.data['count'] == 5
 
+    def test_contacts_are_sorted_by_created_on_desc(self):
+        """Test contacts are sorted by created on desc."""
+        contacts = ContactFactory.create_batch(5)
```
reupen: Would prefer to use freezegun here, since it'll be a bit more predictable. There's a simple test here:

https://github.com/uktrade/data-hub-leeloo/blob/cb29df2c6627a4e16d2c6e7f7213faf6fee26bdd/datahub/dashboard/test/test_views.py#L17-L24

elcct: That's great. Thanks for hint. Updated.

PR #494, opened 2017-09-25
Create manifest.yml for CF

nusnewob:


manifest.yml
```diff
@@ -0,0 +1,4 @@
+---
+applications:
+- buildpack: python_buildpack
+  command:  python manage.py distributed_migrate --noinput && gunicorn config.wsgi --config config/gunicorn.py
```
reupen: Isn't the Procfile used for the commands?

nusnewob:
Procfile is not full supported on current version CF API, this only should
be temp workaround.


Bowen Sun
BowenSun@gmail.com

PR #497, opened 2017-09-27
OMIS - populate billing fields

marcofucci:
This populates all the billing details / address fields from company and contact when generating a quote.
If billing fields have been set by the user, the will not be overridden.

As the billing fields will always be present and populated, we don't need to return the company registered address any more so we can remove it from the public get order endpoint.

### DIFF
--- a/datahub/omis/order/utils.py
+++ b/datahub/omis/order/utils.py
@@ -1,0 +1,51 @@
+def populate_billing_data(order):
+    """
+    Populate the order.billing_* fields from the company/contact if not set already.
+
+    :param order: Order to change if needed
+    :returns: order with billing_* fields filled in
+    """
+    contact = order.contact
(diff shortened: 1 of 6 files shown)


datahub/omis/order/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    }
+    order_billing_details = {
+        field_name: getattr(order, field_name)
+        for field_name in default_billing_details
+        if getattr(order, field_name)
+    }
+
+    # get default and current order values of billing address
+    default_billing_address = {
+        'billing_address_1': company.registered_address_1 or '',
```
reupen: At least as it stands, these fields aren't really used for UK companies with CH data as the CH data is used instead. So that might need to be considered here...

marcofucci: I'm now using the companies house registered address if `company.company_number` exists.

PR #498, opened 2017-09-27
Create an invoice object when accepting a quote

marcofucci:
This
- creates a new omis-invoice django app
- creates an invoice object when accepting a quote
- links the invoice to the order

The invoice number is a sequential number stored as string of form:
`<year><month><day><4 digit sequential number> e.g. 201702280001`


datahub/omis/invoice/utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+from datahub.omis.core.utils import generate_reference
+
+
+def generate_invoice_number():
+    """
+    :returns: an unused invoice number
+    :raises RuntimeError: if no reference can be generated.
+    """
+    from .models import Invoice
```
reupen: Any reason for this being a local import?

datahub/omis/invoice/models.py
```diff
@@ -0,0 +1,39 @@
...
+from .manager import InvoiceManager
+
+MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class Invoice(BaseModel):
+    """Details of an invoice."""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    invoice_number = models.CharField(max_length=100)
```
reupen: Should this have a unique constraint?

datahub/omis/invoice/utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+    current_date = now()
+    prefix = datetime.strftime(current_date, '%Y%m%d')
+
+    def gen():
+        start_count = Invoice.objects.filter(created_on__date=now().date()).count()
+
+        while True:
+            start_count += 1
+            yield f'{start_count}'.zfill(4)
```
reupen: How about `f'{start_count:04}'?

datahub/omis/invoice/manager.py
```diff
@@ -0,0 +1,24 @@
...
+
+class InvoiceManager(models.Manager):
+    """Custom Invoice Manager."""
+
+    def create_populated(self):
+        """
+        :returns: Invoice object generated from the order
+        """
+        return self.create(
+            invoice_number=generate_invoice_number(),
```
reupen: Since there's no randomness in generate_invoice_number() it seems a bit vulnerable to race conditions.

Probably want to try and avoid it – if you can't change the algorithm there are always database-level locks...

datahub/omis/invoice/utils.py
marcofucci: Circular imports, I don't really like it but I can't find any other way.
Can you think of anything better?

datahub/omis/invoice/models.py
marcofucci: 👍  missed that

datahub/omis/invoice/utils.py
marcofucci: very nice 👍

datahub/omis/invoice/manager.py
marcofucci: Very true. 
It's very unlikely, that's why I didn't implement it but maybe I will have a think about using db locks or `<year><month><day><hour><minute><sec><millis>` (which doesn't exclude race conditions either though :( ).

marcofucci: @reupen 
So the current db says that on a busy day (2013), there were 40 invoices created.
That seems pretty low to cause race conditions (2 invoices at exactly the same time you select and check for unused number) but it could still be good to plan for it.

Two options:

1. I can use `start_count = len(Invoice.objects.filter(created_on__date=now().date()))` instead of the current logic to lock the rows. This completely eliminates the problem of race conditions *except* for the fist invoice of the day as there aren't any rows to lock.

2. I can add a line `list(Invoice.objects.select_for_update().all())` which doesn't do anything apart from locking all the rows. This completely eliminates the problem of race conditions in all cases but it's a bit ugly as the line doesn't do much to the code.

Which one do you think it's better?

elcct: Or can you retry if it fails to create?

reupen: It's less of a problem with the unique constraint anyway. I was more concerned with it silently going wrong.

marcofucci: @elcct it's a bit more complicated as that code is wrapped around `transaction.atomic`.

I've implemented option 1 and documented that race conditions could still occur.

elcct: Could be dealt with level above.

PR #503, opened 2017-09-29
Fix non-random invoice number in tests

marcofucci:
`Faker('text')` does not generate random values and this was causing some tests to fail because of unique constraints.
Switched to using ~`get_random_string`~ `Faker('pystr')` instead.


marcofucci:
That's a very strange name `pystr`. OK I've switched to that instead.

PR #504, opened 2017-09-29
Add event field to interactions

reupen:


datahub/interaction/test/factories.py
```diff
@@ -41,6 +42,27 @@ class Meta:  # noqa: D101
...
 
+    kind = Interaction.KINDS.service_delivery
+    created_by = factory.SubFactory(AdviserFactory)
+    modified_by = factory.SubFactory(AdviserFactory)
+    company = factory.SubFactory(CompanyFactory)
+    contact = factory.SubFactory(ContactFactory)
+    event = factory.SubFactory(EventFactory)
+    subject = 'foo'
+    date = now()
+    notes = 'Bar'
```
marcofucci: Not a big deal but it's sometimes useful to have fuzzy attributes instead of a fixed value.
It's for example handy for creating a few records in the dev environment.

reupen: :+1: 

It was based on the old service delivery factory below – will improve it in another PR.

PR #509, opened 2017-10-02
[DH-798] Expose is_disabled property on all metadata.

elcct:
This adds `disabled_on` field to all metadata. It allows to use `is_disabled` query parameter to return either disabled or active metadata - and in case of absence of `is_disabled` it will return all.


marcofucci:
I'm not sure about the logic behind `is_disabled`.

My main concerns:
- `is_disabled` should be time based, _is disabled_ since when? From the name _is_disabled_ is not clear
- the idea is to use a date as base value for the is_disabled value. E.g. if a record has a fk to a metadata object, all the possible valid values are the ones that were not disabled _at the time of creation of that record_ and not since _now_. That's why I added the `was_disabled_on`.
- The metadata endpoints should always return all the values and the frontend should filter the values based on the point 2 (creation date of a particular record). Plus the backend should really validate that the linked metadata was not disabled when creating/saving an object. I know this is hard to generalise but it would be good to try at least.

I'm not sure it all makes sense :/

elcct:
I think this is how it works. I am sorry for not giving more details. `is_disabled` is a computed field based on `disabled_on` field. Our requirement is to have a simple filter to get "active" elements. Those that require `disabled_on` being exposed, can provide different serializer (like in OMIS). Metadata always return all the values by default, but for front end convenience it is possible to just get active ones using `is_disabled` parameter.
Validation should come in next piece of work for particular entity that uses metadata.

marcofucci:
Sorry, I didn't explain myself well.

Let's say that you have a model MyModel with a fk to a metadata MyModelType model.

A record of MyModel (record A) is created with a MyModelType (typeA) on 01/01/2017.

The typeA gets disabled on 01/02/2017.

Today is 01/10/2017, typeA is now disabled.
However, when editing the record A, the typeA shouldn't be disabled as it was not when the record A got created.
Editing the record A should still show the data and save without problems typeA.
The state is frozen at the creation time of the record.

For this reason, `is_disabled` does not make sense as a model property or as a serializer field as it's a state with value depending on the time you request.
"Disabled, when? Now? When the record got created?"

The requirement to have a simple filter to get "active" elements depends on the creation time of the record.

I think we shouldn't add the `is_disabled` property or the serializer field but we should use the disabled_on value instead and the frontend will deal with showing/hiding things depending on the creation date of the record.

In the same way, we shouldn't add any is_disabled filter in the GET metadata endpoint as it's the responsibility of the frontend to hide some disabled values and show some others disabled depending on the frozen time of the record.

elcct:
Thank you for more thorough explanation. I see what you mean. We have not considered this during planning, it was only important if MyModelType is disabled at all and if you decide to save record with related MyModelType that is disabled, it should save provided there is no change that field. I can see that eventually this would led to what you describe, so indeed `is_disabled` makes not much sense. 
I am going to expose `disabled_on` field instead and remove the filter.

PR #513, opened 2017-10-04
Add django admin configs for OMIS orders

marcofucci:


datahub/omis/order/admin.py
```diff
@@ -0,0 +1,21 @@
+from django.contrib import admin
+
+from .models import Order, OrderAssignee, OrderSubscriber
+
+
+@admin.register(Order)
+class OrderAdmin(admin.ModelAdmin):
+    """Admin for orders."""
+
+    search_fields = ('reference',)
```
reupen: You might need raw_id_fields for things like companies and contacts for performance reasons.

marcofucci: very good point, thanks!

PR #514, opened 2017-10-04
[DH-822] Adds kind, company and contact filters to interaction.

elcct:


datahub/search/interaction/tests/test_views.py
```diff
@@ -177,6 +178,51 @@ def test_search_original_query(self, interactions, term):
             'modified_on': interaction.modified_on.isoformat(),
         }]
 
+    def test_filter_by_kind(self, setup_es):
+        """Tests filtering interaction by kind."""
+        interaction = InteractionFactory(
+            kind=Interaction.KINDS.service_delivery,
```
reupen: I think we need some interactions that shouldn't match in here as well (same for the other tests).

ztolley:
Does it add investment project? This filter is also required so we can list interactions for an investment project.

elcct:
@ztolley it doesn't add investment project as filter, because it is already implemented.

datahub/search/interaction/tests/test_views.py
elcct: I now added more Interactions that don't match.

datahub/search/interaction/views.py
```diff
@@ -12,7 +12,10 @@ class SearchInteractionParams:
     serializer_class = SearchInteractionSerializer
 
     FILTER_FIELDS = (
+        'kind',
+        'company',
         'company_name',
+        'contact',
         'contact_name',
         'dit_adviser_name',
         'dit_team',
```
ztolley: i didn't see the investment_project here to assumed it wasn't a supported filter

PR #515, opened 2017-10-05
Update interaction validation

reupen:
Adds some checks to reject fields that are invalid for the interaction kind. Also corrects contact to be required.

This adds a RulesBasedValidator class-level validator to do cross-field validation in a more declarative fashion. It's similar to how cross-field validation was done for investment, but more generic.

I'm planning to split up validate_utils into two modules in a separate PR.

### DIFF
--- a/datahub/core/test/test_validate_utils.py
+++ b/datahub/core/test/test_validate_utils.py
@@ -61,16 +61,20 @@ def test_validation_condition(data, field, op, args, res):
     assert condition(combiner) == res
 
 
-@pytest.mark.parametrize('data,field,op,condition,res', (
-    ({'colour': 'red', 'valid': True}, 'valid', bool, lambda x: True, True),
-    ({'colour': 'red', 'valid': False}, 'valid', bool, lambda x: True, False),
-    ({'colour': 'red', 'valid': True}, 'valid', bool, lambda x: False, True),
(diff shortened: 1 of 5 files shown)


datahub/core/test/test_validate_utils.py
```diff
@@ -48,6 +50,74 @@ def test_is_blank(value, blank):
...
+@pytest.mark.parametrize('data,field,op,condition,res', (
+    ({'colour': 'red', 'valid': True}, 'valid', bool, lambda x: True, True),
+    ({'colour': 'red', 'valid': False}, 'valid', bool, lambda x: True, False),
+    ({'colour': 'red', 'valid': True}, 'valid', bool, lambda x: False, True),
+    ({'colour': 'red', 'valid': False}, 'valid', bool, lambda x: False, True),
+))
+def test_validation_rule(data, field, op, condition, res):
+    """Tests ValidationRule for various cases."""
+    combiner = Mock(spec_set=DataCombiner, get_value=lambda field_: data[field_])
+    rule = ValidationRule('error_key', field, op, condition=condition)
```
elcct: Should that also test `operator_extra_args` ?

reupen: Yep, good point.

PR #516, opened 2017-10-05
[DH-805] Create option in Django Admin to allow to disable events.

elcct:
This creates `EventAdmin` for `Event` with ability to disable (or enable) given entity either one by one or in bulk (with confirmation step).


reupen:
Are we able to add some tests?

elcct:
Good idea, yes.

elcct:
@reupen added tests now.

reupen:
Looks like this disables event types instead of events?

templates/admin/action_confirmation.html
```diff
@@ -0,0 +1,46 @@
+{% extends "admin/base_site.html" %}
```
reupen: Do we need an acknowledgement about what this is based on?

elcct:
I thought this is about EventType, but after re-reading the requirement I think this is actually about the Event itself. I am going to change it to work with the Event instead. Thanks!

templates/admin/action_confirmation.html
elcct: Not sure what kind of acknowledgement you have in mind. I have added a comment that it is based on Django delete confirmation page.

elcct:
@reupen updated admin to work with Event model instead of EventType. Extended Event with `disabled_on` field.

datahub/event/admin.py
```diff
@@ -1,28 +1,129 @@
...
+        'organiser',
+        'lead_team',
+        'teams',
+        'related_programmes',
+        'service',
+        'disabled_on',
+    )
+    list_display = ('name', 'disabled_on',)
+    list_editable = ('disabled_on',)
+    list_filter = (DisabledOnFilter,)
```
reupen: I don't think there's enough information in the grid to decide whether to disable the events or not... I think we should display start date and end date in the grid.

Also need a filter for start date that has options like occurred more than three months ago, six months ago, a year ago etc. Looks simple using DateFieldListFilter but could be a separate PR.

Do we need disabled_on editable in the list now it's on event instead of event type?

datahub/event/admin.py
```diff
@@ -1,28 +1,129 @@
...
+
+    @confirm_action(
+        title='Disable selected',
+        action_message='disable selected'
+    )
+    def disable_selected(self, request, queryset):
+        """Disables selected objects."""
+        return queryset.update(disabled_on=datetime.utcnow())
+
+    disable_selected.short_description = 'Disable selected'
```
reupen: 'Disable selected events' and 'Enable select events' is more consistent with the delete option.

datahub/event/admin.py
```diff
@@ -1,28 +1,129 @@
...
         'organiser',
         'modified_by',
         'created_by',
     )
+
+    actions = ('disable_selected', 'enable_selected',)
+
+    @confirm_action(
+        title='Disable selected',
+        action_message='disable selected'
```
reupen: Should be 'disable the selected events' to complete the sentence in the confirmation?

templates/admin/action_confirmation.html
```diff
@@ -0,0 +1,47 @@
...
+        <a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
+        &rsaquo; <a href="{% url 'admin:app_list' app_label=opts.app_label %}">{{ opts.app_config.verbose_name }}</a>
+        &rsaquo; <a href="{% url opts|admin_urlname:'changelist' %}">{{ opts.verbose_name_plural|capfirst }}</a>
+        &rsaquo; {{ title }}
+    </div>
+{% endblock %}
+
+{% block content %}
+    <p>Are you sure you want to {{ action_message }}?</p>
+    <h2>{% trans "Objects" %}</h2>
```
reupen: Could get the model verbose name from the context and use that instead of 'Objects'? Something like `opts.verbose_name_plural|capfirst`?

datahub/event/admin.py
```diff
@@ -1,28 +1,129 @@
...
-    search_fields = ('name', 'pk')
+
+def confirm_action(title, action_message):
+    """Decorator that adds confirmation step before modifying selected objects."""
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(self, request, queryset):
+            if request.POST.get('confirm') == 'yes':
+                updated = func(self, request, queryset)
+                self.message_user(request, f'{updated} objects updated.')
```
reupen: Could use the model verbose name instead of 'objects'?

datahub/event/admin.py
elcct: I added more fields and date filter. Thanks :)

Yes, we need ability to edit individually and in bulk.

datahub/event/admin.py
elcct: Good point. Added.

datahub/event/admin.py
elcct: Looks better indeed.

templates/admin/action_confirmation.html
elcct: Django uses "Objects" in confirmation page, but changed to plural model name.

datahub/event/admin.py
elcct: Updated.

templates/admin/action_confirmation.html
reupen: Yeah but for the delete confirmation there are cascaded deletes so different types of object are listed. Don't think the heading helps much here actually but that's not important.

datahub/event/admin.py
```diff
@@ -82,9 +83,19 @@ class EventAdmin(BaseModelVersionAdmin):
...
+        'end_date',
+        'lead_team',
+        'service',
+        'disabled_on',
+    )
     list_editable = ('disabled_on',)
-    list_filter = (DisabledOnFilter,)
+    list_filter = (
+        DisabledOnFilter,
+        ('start_date', DateFieldListFilter),
```
reupen: I don't think DateFieldListFilter will be much use unless we can say older than three months, six months etc. But not needed for this PR.

PR #520, opened 2017-10-06
Make subject sortable.

elcct:
Field that we want to sort by must not be analysed. I created separated field `subject_english` with English analyzer for search, enabling `subject` field to be sortable.


datahub/search/interaction/models.py
```diff
@@ -45,6 +46,7 @@ class Interaction(DocType, MapDBModelToDict):
 
     SEARCH_FIELDS = [
         'subject',
+        'subject_english',
```
reupen: Do both subjects need to be here?

elcct: Don't need to, but I thought this could give more score to exact match.

PR #522, opened 2017-10-06
Move core validators to new module

reupen:
Splits validate_utils into two modules.


datahub/core/validators.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class ValidationRule:
+    """Validation rule."""
+
+    def __init__(self,
+                 error_key: str,
+                 field: str,
+                 operator_: Callable,
+                 operator_extra_args: Sequence = (),
```
marcofucci: I know it was introduced before but you are not using `operator_extra_args` anywhere so should we get rid of it? Also, if needed `partial` could be used instead.

marcofucci: I know see that you have it there to make is similar to the `Condition` object.
I read nice when using `Condition('kind', eq, (Interaction.KINDS.interaction,))` but not much when using `ValidationRule`.

I find the choice of names / logic a bit confusing but I'll leave it to you.

reupen: Yeah it could go for now. I think it'll need more thought anyway as some of those rules would involve other fields rather than just constants. I'll just remove it for the minute.

Partial wouldn't do the same thing though, because you'd have to bind the first argument (so it'd have to be a lambda).

I'll think about the names for a separate PR (let me know if you have any suggestions).

marcofucci: Haven't thought about it for long but I see a _rule_ class, a _rule with condition_ class (condition being a rule instance) and a "field validation" with a rule (rule or rule with a condition).
You'll definitely find a better design though and it's not a massive priority tbh.

PR #532, opened 2017-10-10
Add lead_team, teams and disabled_on filter to event search.

elcct:
As in the title.

### DIFF
--- a/datahub/search/event/tests/test_views.py
+++ b/datahub/search/event/tests/test_views.py
@@ -8,6 +8,7 @@ from datahub.company.test.factories import AdviserFactory
 from datahub.core import constants
 from datahub.core.test_utils import APITestMixin
 from datahub.event.test.factories import EventFactory
+from datahub.metadata.test.factories import TeamFactory
 
 
 @pytest.fixture
@@ -47,7 +48,6 @@ class TestSearch(APITestMixin):
 
         url = reverse('api-v3:search:event')
         response = self.api_client.post(url, {
-            'original_query': '',
             'offset': 1,
             'limit': 1,
         })
@@ -77,7 +77,6 @@ class TestSearch(APITestMixin):
         url = reverse('api-v3:search:event')
 
         response = self.api_client.post(url, {
-            'original_query': '',
             'name': event_name[:5],
         })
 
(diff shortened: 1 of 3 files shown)


datahub/search/event/views.py
```diff
@@ -12,22 +12,28 @@ class SearchEventParams:
     serializer_class = SearchEventSerializer
 
     FILTER_FIELDS = (
+        'address_country',
+        'disabled_on_after',
```
reupen: Do these cover 'disabled_on is not null'?

datahub/search/event/tests/test_views.py
```diff
@@ -149,6 +145,83 @@ def test_search_event_address_country(self, setup_es, setup_data):
...
+
+    def test_search_event_teams(self, setup_es, setup_data):
+        """Tests teams filter."""
+        url = reverse('api-v3:search:event')
+
+        team_a = TeamFactory()
+        team_b = TeamFactory()
+        for _ in range(5):
+            EventFactory(
+                teams=[team_a]
```
reupen: Would be nice to have some events with multiple teams here as well.

datahub/search/event/views.py
elcct: This is a very good point. I have added `disabled_on_exists` filter to cover this case. This one accepts boolean as a parameter and is useful to get just disabled events or only active.

datahub/search/event/tests/test_views.py
```diff
@@ -149,7 +145,133 @@ def test_search_event_address_country(self, setup_es, setup_data):
...
+    def test_search_event_lead_team(self, setup_es):
+        """Tests lead_team filter."""
+        url = reverse('api-v3:search:event')
+
+        team = TeamFactory()
+        for _ in range(5):
+            EventFactory()
+            EventFactory(
+                lead_team_id=team.id
+            )
```
marcofucci: isn't this the same as this?
```
EventFactory.create_batch(5)
EventFactory.create_batch(5, lead_team_id=team.id)
```

datahub/search/event/tests/test_views.py
elcct: I have added more teams.

datahub/search/event/tests/test_views.py
elcct: Indeed! Thanks :)

PR #540, opened 2017-10-12
Re-add investment project anonymous description

reupen:


elcct:
Should this also include entry in the search model?

reupen:
@elcct Yep, added now.

PR #549, opened 2017-10-17
Add is_event virtual field to interaction

reupen:
This is to resolve a problem around validation for the 'Is this an event?' question in the front end, as this isn't a field in the model (as it can be inferred by checking whether the event field is set). We agreed to solve this by adding it as a virtual field and performing some validation around it at the API level, so this is what this PR does.


elcct:
Do we need this in search model?

reupen:
Hmm, I'll add it for consistency.

PR #550, opened 2017-10-17
[DH-921] Add dit_adviser Interaction filter & default sort order by date desc.

elcct:
This adds missing `dit_adviser` Interaction filter and sets default sorting to `date:desc`.


datahub/search/interaction/tests/test_views.py
```diff
@@ -184,15 +213,12 @@ def test_search_original_query(self, interactions, term):
...
-
-        for _ in range(10):
-            InteractionFactory(
-                kind=Interaction.KINDS.interaction,
-            )
-            service_deliveries.append(InteractionFactory(
-                kind=Interaction.KINDS.service_delivery,
-            ))
+        InteractionFactory.create_batch(10, kind=Interaction.KINDS.interaction),
+        service_deliveries = InteractionFactory.create_batch(
```
reupen: Use ServiceDeliveryFactory?

datahub/search/interaction/tests/test_views.py
```diff
@@ -77,6 +79,33 @@ def test_offset(self):
...
+            datetime(2017, 2, 4, 13, 15, 0),
+            datetime(2017, 1, 4, 11, 23, 10),
+            datetime(2017, 9, 29, 3, 25, 15),
+            datetime(2017, 7, 5, 11, 44, 33),
+            datetime(2017, 2, 1, 18, 15, 1),
+        )
+        date_iter = iter(dates)
+        InteractionFactory.create_batch(
+            len(dates),
+            date=factory.LazyFunction(lambda: next(date_iter))
```
reupen: What about factory.Iterator?

datahub/search/interaction/tests/test_views.py
elcct: I should have looked for its existence. Thanks :)

datahub/search/interaction/tests/test_views.py
elcct: That's exactly what I was looking for :) Thanks! Fixed.

datahub/search/interaction/tests/test_views.py
```diff
@@ -287,15 +315,17 @@ def test_filter_by_contact_id(self, setup_es):
...
-            ))
-            InteractionFactory(
-                contact=contacts[i]
-            )
+        name_str = partial(get_random_string, 16)
+
+        contacts = ContactFactory.create_batch(
+            10,
+            first_name=factory.LazyFunction(name_str),
+            last_name=factory.LazyFunction(name_str)
```
reupen: I know it was like this before, but looks like we could use factory.Faker() here (with 'first_name'/'last_name') (in fact ContactFactory should be changed).

datahub/search/interaction/tests/test_views.py
```diff
@@ -237,10 +260,14 @@ def test_filter_by_company_id(self, setup_es):
 
     def test_filter_by_company_name(self, setup_es):
         """Tests filtering interaction by company name."""
-        companies = []
-        for i in range(10):
-            companies.append(CompanyFactory(name=get_random_string(16)))
-            InteractionFactory(company=companies[i])
+        companies = CompanyFactory.create_batch(
+            10,
+            name=factory.LazyFunction(partial(get_random_string, 16))
```
reupen: Similar to below (looks like CompanyFactory could use `factory.Faker('company')`, but could be fixed in a separate PR).

datahub/search/interaction/tests/test_views.py
elcct: Great idea. I changed ContactFactory.

datahub/search/interaction/tests/test_views.py
elcct: Also changed CompanyFactory, thanks :)

PR #552, opened 2017-10-18
Remove test data where it is not necessary.

elcct:
As in the title. Extra data can for example create company name similar to the one created in `test_search_contact_by_partial_company_name` and thus making it fail. Either I could change the test to look up if any result is a match or leave the test and remove extra test data. As a precaution I removed this from other tests as well, where extra test data is not necessary to be there.


reupen:
Do some of these need some deliberately not-matching data as well?

elcct:
I think these tests need some improvements in general and also it is worth adding some that test non matching queries.

PR #554, opened 2017-10-18
[DH-956] Rename address fields in Investment Project.

elcct:
This renames address fields in Investment Project to make it consistent with address fields in other entities.


datahub/investment/migrations/0020_auto_20171018_1506.py
```diff
@@ -0,0 +1,35 @@
...
+            new_name='address_1',
+        ),
+        migrations.RenameField(
+            model_name='investmentproject',
+            old_name='address_line_2',
+            new_name='address_2',
+        ),
+        migrations.RenameField(
+            model_name='investmentproject',
+            old_name='address_line_3',
```
reupen: Looks like the migration has swapped address_3/postcode in the new column names.

fixtures/test_data.yaml
```diff
@@ -561,9 +561,9 @@
     project_manager: d7493b4e-5d7b-4834-98d9-28b78a74052a
     project_assurance_adviser: 5a644146-5298-4741-91f3-d5d7558adf47
     uk_company: a73efeba-8499-11e6-ae22-56b6b6499611
-    address_line_1: 10 Eastings Road
-    address_line_2: London
-    address_line_postcode: W1 2AA
+    address_1: 10 Eastings Road
+    address_2: London
```
reupen: Should probably be address_town here.

datahub/investment/migrations/0020_auto_20171018_1506.py
elcct: Wow, good catch! I'll update in a minute. Thanks.

datahub/investment/validate.py
```diff
@@ -35,9 +35,9 @@
     'non_fdi_r_and_d_budget': Stage.verify_win.value,
     'new_tech_to_uk': Stage.verify_win.value,
     'export_revenue': Stage.verify_win.value,
-    'address_line_1': Stage.verify_win.value,
-    'address_line_2': Stage.verify_win.value,
-    'address_line_postcode': Stage.verify_win.value,
+    'address_1': Stage.verify_win.value,
+    'address_2': Stage.verify_win.value,
```
reupen: Should probably be address_town here instead of address_2 now.

ztolley: I think we need address_1, address_2, address_town and address_postcode

reupen: We don't require address_2 anywhere else. There won't always be a second line.

fixtures/test_data.yaml
elcct: True, it looks like it should be `address_town` instead.

datahub/investment/validate.py
elcct: I changed it to `address_town` as well. `address_2` is now not required.

ztolley:
You need address_1 AND address_2 for some addresses IMHO, plus all the other address entries have 1 and 2. I know for my home address you need two

elcct:
@ztolley `address_2` do exist, but it is not required.

PR #567, opened 2017-10-23
Add archived_documents_url_path to OMIS orders

marcofucci:
This adds a new field called ~`archived_documents_url`~ `archived_documents_url_path` with the url to an external service holding documents for this order.


datahub/omis/order/test/views/test_order_details.py
```diff
@@ -817,7 +823,9 @@ class TestViewOrderDetails(APITestMixin):
 
     def test_get(self):
         """Test getting an existing order."""
-        order = OrderFactory()
+        order = OrderFactory(
+            archived_documents_url='/documents/123'
```
reupen: Is that really a URL? :P

datahub/omis/order/models.py
```diff
@@ -219,6 +219,10 @@ class Order(BaseModel):
         blank=True, editable=False,
         help_text='Legacy field. Can DIT speak to the contacts?'
     )
+    archived_documents_url = models.CharField(
```
reupen: Do we not want URLField for this?

datahub/omis/order/test/views/test_order_details.py
marcofucci: yes, the domain is probably going to be an env var as so that we can change it without needing to migrate all the records.

datahub/omis/order/models.py
marcofucci: no domain in the db means the url is not valid like this.

datahub/omis/order/test/views/test_order_details.py
reupen: It's still not a URL though. It's a URL path.

marcofucci: Got it, so you mean I should probably rename it to `archived_documents_url_path`?

datahub/omis/order/models.py
reupen: See other comment – URL sounds like it contains a URL. Something like `archived_documents_url_path` would be clearer.

marcofucci: 👍

datahub/omis/order/test/views/test_order_details.py
elcct: What about uri? url_path suggests that it will not contain the host address.

datahub/omis/order/models.py
elcct: It looks more like a `URI reference` - https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

datahub/omis/order/test/views/test_order_details.py
marcofucci: It will not contain the host address so `...url_path` looks better I suppose, no?

elcct: Ah, sorry I thought it may contain host address.

marcofucci: 👍

PR #568, opened 2017-10-23
add Mercury Ltd ch fixtures for UAT tests

feedmypixel:
We also need the ch data for the Mercury Ltd fixture in UAT tests. This work adds that information in.


feedmypixel:
> Would prefer to load test_ch_data.yaml before test_data.yaml (but it doesn't make much difference at the moment).

done

PR #569, opened 2017-10-23
Add trading_name to global search.

elcct:
As in the title, it adds `trading_name` (alias for `alias` :) ) to the global search. It also has a small cleanup of tests and small improvement to queries.

### DIFF
--- a/datahub/search/company/test/test_elasticsearch.py
+++ b/datahub/search/company/test/test_elasticsearch.py
@@ -14,26 +14,21 @@ def test_get_basic_search_query():
                     {
                         'match_phrase': {
                             'name_keyword': {
-                                'query': 'test', 'boost': 2
+                                'query': 'test',
+                                'boost': 2
                             }
                         }
                     }, {
                         'match_phrase': {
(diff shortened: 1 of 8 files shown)


datahub/search/company/test/test_elasticsearch.py
```diff
@@ -14,35 +14,29 @@ def test_get_basic_search_query():
...
                         }
                     }, {
                         'match': {
                             'address_1': 'test'
                         }
                     }, {
                         'match': {
-                            'address_2': 'test'
-                        }
+                            'address_2': 'test'}
```
marcofucci: You might have forgotten to indent the closing bracket

elcct: Thanks, updated.

elcct:
@reupen what do you think about changing model's field from `alias` to `trading_name` ?

reupen:
> @reupen what do you think about changing model's field from alias to trading_name ?

Yeah, that seems fair, probably better as a separate PR though.

PR #570, opened 2017-10-23
add the company_category property to ch fixture data

feedmypixel:
We need this added for the UAT tests to be able to create companies from the fixtures


fixtures/test_ch_data.yaml
```diff
@@ -14,3 +14,4 @@
     registered_address_town: Y Ffor
     registered_address_postcode: LL53 5RN
     incorporation_date: 2000-12-14
+    company_category: Company
```
reupen: The values we get from Companies House are things like `Private Limited Company`, `Public Limited Company` so this needs updating.

PR #577, opened 2017-10-25
Improve search filter validation.

elcct:
This adds validation to search filters using DRF serializers. I needed to add three serializers:

`RelaxedDateTimeField` - which handles date sent in different formats.
`StringUUIDField` - we can't use UUID in the ES queries, so this only ensures that given string can be parsed as UUID and returns a string.
`SingleOrListField` - this extends `ListField` serializer. If data input is a string, it creates a list of one element. This is to handle case when filter can be either single value or an array.

It changes behaviour in case of date validation, as it now returns which field has a problem, not an error in general.

### DIFF
--- a/datahub/search/company/test/test_views.py
+++ b/datahub/search/company/test/test_views.py
@@ -60,6 +60,26 @@ class TestSearch(APITestMixin):
         assert len(response.data['results']) == 1
         assert response.data['results'][0]['trading_address_country']['id'] == united_states_id
 
+    def test_multiple_trading_address_country_filter(self, setup_data):
+        """Tests multiple trading address countries filter."""
+        term = 'abc defg'
+
+        url = reverse('api-v3:search:company')
+        united_states_id = constants.Country.united_states.value.id
+        united_kingdom_id = constants.Country.united_kingdom.value.id
+
+        response = self.api_client.post(url, {
+            'original_query': term,
+            'trading_address_country': [united_states_id, united_kingdom_id],
+        }, format='json')
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == 2
+        assert len(response.data['results']) == 2
+        country_ids = {result['trading_address_country']['id']
+                       for result in response.data['results']}
+        assert country_ids == {united_kingdom_id, united_states_id}
+
     def test_trading_name_filter(self, setup_es):
         """Tests detailed company search."""
         trading_name = 'Hello World'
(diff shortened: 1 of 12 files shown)


elcct:
My reason is that at the moment date fields in the FE are free text and people can type in date in various different formats, `DateTimeField` was not giving enough flexibility.

datahub/search/company/test/test_views.py
```diff
@@ -80,6 +100,16 @@ def test_trading_name_filter(self, setup_es):
...
 
+    def test_null_filter(self, setup_es):
+        """Tests filter with null value."""
+        url = reverse('api-v3:search:company')
+
+        response = self.api_client.post(url, {
+            'uk_region': None,
+        }, format='json')
+
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
```
reupen: Would be nice to check the response body as well.

datahub/search/serializers.py
```diff
@@ -1,8 +1,54 @@
...
+
+from dateutil.parser import parse as dateutil_parse
 from rest_framework import serializers
 from rest_framework.settings import api_settings
+
 from datahub.search.elasticsearch import MAX_RESULTS
 
 
+class DateTimeyField(serializers.Field):
+    """String Date time field."""
```
reupen: Would be nice to say something here about what the purpose of this is (compared to the DRF DateTimeField).

Also might be nice to chose a name for it that's a bit more different from `DateTimeField` (`RelaxedDateTimeField` ?)

datahub/search/serializers.py
```diff
@@ -1,8 +1,54 @@
...
+    """Field can be single instance or list."""
+
+    def to_internal_value(self, data):
+        """If data is str, creates a list."""
+        if isinstance(data, str):
+            data = [data]
+        return super().to_internal_value(data)
+
+
+class StringUUIDField(serializers.Field):
```
reupen: Are we able to reuse UUIDField? Either using inheritance or composition? (Just need to convert the return value of its to_internal_value() to a string.)

datahub/search/company/test/test_views.py
elcct: Good idea, added.

datahub/search/serializers.py
elcct: I have updated the comment and changed the name, as the one you proposed is better. Thanks.

datahub/search/serializers.py
elcct: Indeed, UUIDField can be reused. Updated.

PR #586, opened 2017-10-30
Feature/simple permissions

PeterNociar:
Extended Django permissions to include read.
Added connection between user and permission groups through Team-> TeamRole->Group + fixtures


datahub/company/test/test_advisor_permissions.py
```diff
@@ -0,0 +1,42 @@
...
+        url_adviser_list = reverse('api-v1:advisor-list')
+        CompanyFactory()
+        CompanyFactory()
+        url_companies = reverse('api-v3:company:collection')
+        ContactFactory()
+        url_contacts = reverse('api-v3:contact:list')
+        EventFactory()
+        url_events = reverse('api-v3:event:collection')
+
+        response = self.api_client.get(url_adviser_list)
```
elcct: I think it would be better to have separate test for each list, maybe could use parametrized test.

PeterNociar: Ok.

PeterNociar: Fixed

datahub/core/auth.py
```diff
@@ -147,3 +149,31 @@ def _submit_form(session, source, url=None, params=None):
         assert form_action != html_parser('form').attr('action')
 
         return resp
+
+
+class TeamModelPermissionsBackend(CDMSUserBackend):
+    """Extension of CDMSUserBackend to include a team based permissions for user"""
+
+    def _get_team_permissions(self, user_obj):
```
marcofucci: I would add a comment saying that this method is constructed dynamically and called by the django core method `_get_permissions` so that it doesn't get accidentally deleted.

datahub/investment/models.py
```diff
@@ -327,6 +333,14 @@ class IProjectDocument(BaseModel, ArchivableModel):
     filename = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH)
     document = models.OneToOneField(Document, on_delete=models.PROTECT)
 
+    class Meta:
+        verbose_name = 'investment project document'
+        verbose_name_plural = 'investment project documents'
+        unique_together = (
+            ('project', 'doc_type', 'filename'),
+        )
+        permissions = (('read_iprojectdocument', 'Can read project document'),)
```
reupen: Permission description should match model verbose name.

datahub/omis/order/migrations/0002_adding_read_permissions.py
```diff
@@ -0,0 +1,19 @@
...
+# Generated by Django 1.11.6 on 2017-10-27 09:33
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('order', '0001_squashed_0030_cancellation'),
```
marcofucci: A new migration has been added so you'll have to solve migration conflicts. 
I will soon delete all the old migrations that got squashed as well.

datahub/core/permissions.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.permissions import DjangoModelPermissions
+
+
+class CrudPermission(DjangoModelPermissions):
```
marcofucci: Should we call it `DjangoCrudPermissions` to make it more obvious?

config/settings/common.py
```diff
@@ -165,7 +164,8 @@
         'oauth2_provider.contrib.rest_framework.OAuth2Authentication'
     ],
     'DEFAULT_PERMISSION_CLASSES': [
-        'oauth2_provider.contrib.rest_framework.IsAuthenticatedOrTokenHasScope'
+        'oauth2_provider.contrib.rest_framework.IsAuthenticatedOrTokenHasScope',
```
marcofucci: Isn't adding `permission_classes = (IsAuthenticatedOrTokenHasScope,)` to each individual view unnecessary if we have this? If so, should we only keep this and delete the others?

datahub/company/test/test_advisor_permissions.py
```diff
@@ -0,0 +1,32 @@
...
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.company.test.factories import AdviserFactory, CompanyFactory, ContactFactory
+from datahub.core.test_utils import APITestMixin, get_test_user
+from datahub.event.test.factories import EventFactory
+from datahub.metadata.models import Team
+
+
+class TestAdviserPermissions(APITestMixin):
```
marcofucci: I'm not sure this is the right place for these tests. The views are not all from this django app and it depends on the content of the teams fixture.

I suggest we should maybe try create some generic tests in `core/auth/test/...` which basically tests the new `TeamModelPermissionsBackend` using some testing urls/views and testing perms. What do you think?

datahub/metadata/models.py
```diff
@@ -55,7 +57,14 @@ class Role(BaseConstantModel):
 class TeamRole(BaseConstantModel):
     """Team role."""
 
-    pass
+    team_role_groups = models.ManyToManyField(
+        Group,
+        verbose_name=_('Team role permission groups'),
```
reupen: The verbose_name should start with a lower case letter.

datahub/metadata/models.py
```diff
@@ -55,7 +57,14 @@ class Role(BaseConstantModel):
...
+        blank=True,
+        help_text=_('Permission groups associated with this team.'),
+        related_name='teamrole_set',
```
reupen: The convention in the rest of the repo is to use things like `team_roles` for `related_name`, I think we should be consistent about it...

datahub/search/company/views.py
```diff
@@ -41,6 +43,12 @@ class SearchCompanyParams:
 class SearchCompanyAPIView(SearchCompanyParams, SearchAPIView):
     """Filtered company search view."""
 
+    permission_required = 'company.read_company'
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
```
reupen: Isn't `permission_classes = (IsAuthenticatedOrTokenHasScope,)` already set in the base class SearchAPIView? (Same comment for the other search views as well.)

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,5046 @@
...
-- model: metadata.teamrole
-  pk: 64329c18-6095-e211-a939-e4115bead28a
-  fields: {name: "Regional Investment Team"}
+- model: auth.group
+  fields:
+    name: LEP
+    permissions:
+    - [add_advisor, company, advisor]
+    - [change_advisor, company, advisor]
+    - [delete_advisor, company, advisor]
```
reupen: Don't think we want any delete permissions in these groups?

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,5046 @@
...
+    - [read_advisor, company, advisor]
+    - [add_companieshousecompany, company, companieshousecompany]
+    - [change_companieshousecompany, company, companieshousecompany]
```
reupen: companieshousecompany should be read-only (it's just populated using the sync_ch manage.py command, not edited by users).

config/settings/common.py
PeterNociar: That is actually to remove the CrudPermission permission class. It is only used on the views that don't have queryset eg. search.

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,5046 @@
...
+    - [read_companieshousecompany, company, companieshousecompany]
+    - [add_company, company, company]
+    - [change_company, company, company]
+    - [delete_company, company, company]
+    - [read_company, company, company]
+    - [add_contact, company, contact]
+    - [change_contact, company, contact]
+    - [delete_contact, company, contact]
+    - [read_contact, company, contact]
+    - [add_session, sessions, session]
```
reupen: Do we need session here? We don't use Django sessions for users.

datahub/core/auth.py
PeterNociar: Ok

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,5046 @@
...
+    - [change_businesslead, leads, businesslead]
+    - [delete_businesslead, leads, businesslead]
+    - [read_businesslead, leads, businesslead]
+    - [add_businesstype, metadata, businesstype]
+    - [change_businesstype, metadata, businesstype]
+    - [delete_businesstype, metadata, businesstype]
+    - [add_companyclassification, metadata, companyclassification]
+    - [change_companyclassification, metadata, companyclassification]
+    - [delete_companyclassification, metadata, companyclassification]
+    - [add_country, metadata, country]
```
reupen: We shouldn't have add/change/delete for metadata models for normal users (also recent metadata models are in the relevant app instead of metadata).

Do we need a DIT_admin group as well? ( @jim68000 ?)

datahub/company/test/test_advisor_permissions.py
PeterNociar: I don't have any strong preference about this. I thought it was closest to the adviser model as its testing the permission on a user, but I can move it.  Where would you recommend it to go? The 
`code/auth` is a file atm. So I can create a 
`code/
    test.py
    backends.py`
Would that be ok ?

datahub/company/test/test_advisor_permissions.py
```diff
@@ -0,0 +1,32 @@
...
+            (ContactFactory, 'api-v3:contact:list', status.HTTP_200_OK),
+            (EventFactory, 'api-v3:event:collection', status.HTTP_403_FORBIDDEN)
+        ))
+    def test_adviser_list_view(self, factory, reverse_url, expected_status):
+        """User different permissions for LEP team role user."""
+        team = Team.objects.filter(role__team_role_groups__name='LEP').first()
+        self._user = get_test_user(team=team)
+
+        url = reverse(reverse_url)
+        factory()
```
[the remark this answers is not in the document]
PeterNociar: I just wanted to be thorough and test views with data, but I can remove it ;)

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,5046 @@
...
+    - [add_title, metadata, title]
+    - [change_title, metadata, title]
+    - [delete_title, metadata, title]
+    - [add_turnoverrange, metadata, turnoverrange]
+    - [change_turnoverrange, metadata, turnoverrange]
+    - [delete_turnoverrange, metadata, turnoverrange]
+    - [add_ukregion, metadata, ukregion]
+    - [change_ukregion, metadata, ukregion]
+    - [delete_ukregion, metadata, ukregion]
+    - [add_oauthapplicationscope, oauth, oauthapplicationscope]
```
reupen: This OAuth stuff needs checking as well – doesn't look like it should be here...

datahub/core/permissions.py
PeterNociar: Sure.

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,5046 @@
...
+    - [delete_ordersubscriber, order, ordersubscriber]
+    - [add_servicetype, order, servicetype]
+    - [change_servicetype, order, servicetype]
+    - [delete_servicetype, order, servicetype]
+    - [add_session, sessions, session]
+    - [change_session, sessions, session]
+    - [delete_session, sessions, session]
+- model: auth.group
+  fields:
+    name: DIT_staff
```
reupen: How about just `DIT staff` (no underscore)?

datahub/omis/order/migrations/0002_adding_read_permissions.py
PeterNociar: I have recreated all the migrations on rebase (couple of times :) ) This was working ok for me

datahub/core/auth.py
```diff
@@ -147,3 +149,31 @@ def _submit_form(session, source, url=None, params=None):
...
+        teams they belong to based on groups associated to team roles.
+        """
+        return self._get_permissions(user_obj, obj, 'team')
+
+    def get_all_permissions(self, user_obj, obj=None):
+        """
+        Because of using cache in the parent class, its hard to extend using super()
+        so the code is slightly duplicated
+        """
+        if not user_obj.is_active or user_obj.is_anonymous or obj is not None:
```
reupen: Can't find this file in `codecov.io`, is this covered by tests?

config/settings/common.py
marcofucci: oh yeah missed that 👍

datahub/company/models/adviser.py
```diff
@@ -115,3 +115,4 @@ class Meta:
             models.Index(fields=['first_name', 'last_name']),
         ]
         verbose_name = 'adviser'
+        permissions = (('read_advisor', 'Can read advisor'),)
```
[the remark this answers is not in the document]
PeterNociar: Sure.

datahub/core/auth.py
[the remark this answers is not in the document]
PeterNociar: Done

datahub/core/auth.py
PeterNociar: Yes this is the method used to get a user permission for all of the API

datahub/core/test_utils.py
```diff
@@ -23,6 +24,13 @@ def get_test_user():
             email='Testo@Useri.com',
             date_joined=now(),
         )
+        if team is None:
+            test_user.dit_team = Team.objects.filter(
```
reupen: Can this use `Team.objects.get()`?

datahub/search/company/views.py
PeterNociar: This is to remove the Model based  permissions check on the search apis as they dont have a query_set which is required

datahub/core/auth.py
```diff
@@ -147,3 +149,31 @@ def _submit_form(session, source, url=None, params=None):
...
+            return set()
+        if not hasattr(user_obj, '_perm_cache'):
+            user_obj._perm_cache = self.get_user_permissions(user_obj)
```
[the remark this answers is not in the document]
PeterNociar: Yes and that is what we need.  _perm_cache holds just the permissions of the user. And the original Django method was doing the same but only for user and groups. I just added team permissions

datahub/core/test_utils.py
PeterNociar: .get() needs a unique id. I didn't want to hardcode pk there

datahub/search/company/views.py
PeterNociar: true. Removing

datahub/core/test_utils.py
reupen: QuerySet.get() takes keyword arguments, it doesn't need the primary key.

https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get

datahub/core/auth.py
PeterNociar: Done

datahub/core/test_utils.py
PeterNociar: Yes, but they need to produce a unique result of only one object.

datahub/core/auth.py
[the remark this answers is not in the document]
reupen: I'm referring to `_user_perm_cache`. `self.get_user_permissions()` returns `self_user_perm_cache` and then it's mutated. `user_obj._perm_cache` and `user_obj._user_perm_cache` end up being the same object.

Would suggest creating a new object instead of mutating what `self.get_user_permissions(user_obj)` returns.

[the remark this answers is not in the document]
PeterNociar: Oh, Now I see what you mean. Yes, you are right. Its what django is doing, so I hope it won't break any special django behaviour. But yes, I'll fix that

datahub/core/auth/test/test_advisor_permissions.py
```diff
@@ -0,0 +1,28 @@
...
+            ('api-v3:event:collection', status.HTTP_403_FORBIDDEN)
+        ))
+    def test_adviser_list_view(self, reverse_url, expected_status):
+        """User different permissions for LEP team role user."""
+        team = Team.objects.filter(role__team_role_groups__name='LEP').first()
+        self._user = get_test_user(team=team)
+        url = reverse(reverse_url)
+
+        response = self.api_client.get(url)
+        assert response.status_code == expected_status
```
marcofucci: Sorry for the delay but: 

1. I actually meant that the test should be in the `datahub/core/test/` folder instead 😬 

2. this test file should not use any datahub specific views e.g. `'dashboard:intelligent-homepage'`, `'api-v1:advisor-list'`, etc. This django app is encapsulated and shouldn't know anything about the other ones. Have a look at the `datahub/core/test/support`, it has testing urls and views etc. You can use or extend those but generic code/logic.

3. The test should probably use generic team roles as well. The `LEP` fixture is a specific implementation and can change in the future, it's better to test the abstract logic with generated objects from factories instead.

datahub/core/auth.py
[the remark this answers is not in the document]
PeterNociar: Done

datahub/core/test_utils.py
reupen: Oh, yes, I see now (misread what it was doing) :+1:

datahub/omis/order/migrations/0002_adding_read_permissions.py
marcofucci: There's already a 0002 migration, I've removed the old squashed migrations now so it should be clearer. I suggest:
- `./manage.py migrate order 0001`
- delete `0002_adding_read_permissions.py`
- rebase
- re-make migrations `./manage.py makemigrations order`

The resulting mig file should be 0003.

datahub/search/interaction/views.py
```diff
@@ -42,6 +42,10 @@ class SearchInteractionParams:
 class SearchInteractionAPIView(SearchInteractionParams, SearchAPIView):
     """Filtered interaction search view."""
 
+    permission_required = 'interaction.read_interaction'
+
 
 class SearchInteractionExportAPIView(SearchInteractionParams, SearchExportAPIView):
     """Filtered interaction search export view."""
+
+    permission_required = 'company.read_company'
```
marcofucci: Shouldn't this be `'interaction.read_interaction'` ?

marcofucci: Also, don't we need `permission_required` for `search/investment/views.py` and `search/omis/views.py` as well?

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [change_advisor, company, advisor]
+    - [read_advisor, company, advisor]
+    - [read_companieshousecompany, company, companieshousecompany]
+    - [add_company, company, company]
+    - [change_company, company, company]
+    - [delete_company, company, company]
+    - [read_company, company, company]
+    - [add_contact, company, contact]
+    - [change_contact, company, contact]
+    - [delete_contact, company, contact]
```
marcofucci: @reupen @elcct do we support deleting a contact?

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [read_contact, company, contact]
+    - [add_invoice, omis-invoice, invoice]
+    - [change_invoice, omis-invoice, invoice]
+    - [delete_invoice, omis-invoice, invoice]
```
marcofucci: get rid of the add_invoice/change_invoice/delete_invoice please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [read_invoice, omis-invoice, invoice]
+    - [add_market, omis-market, market]
+    - [change_market, omis-market, market]
+    - [delete_market, omis-market, market]
```
marcofucci: get rid of add_market/change_market/delete_market please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [add_payment, omis-payment, payment]
+    - [change_payment, omis-payment, payment]
+    - [delete_payment, omis-payment, payment]
```
marcofucci: get rid of change_payment/delete_payment please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [read_payment, omis-payment, payment]
+    - [add_quote, omis-quote, quote]
+    - [change_quote, omis-quote, quote]
+    - [delete_quote, omis-quote, quote]
```
marcofucci: get rid of delete_quote please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [read_quote, omis-quote, quote]
+    - [add_hourlyrate, order, hourlyrate]
+    - [change_hourlyrate, order, hourlyrate]
+    - [delete_hourlyrate, order, hourlyrate]
```
marcofucci: get rid of add_hourlyrate/change_hourlyrate/delete_hourlyrate please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [add_order, order, order]
+    - [change_order, order, order]
+    - [delete_order, order, order]
```
marcofucci: get rid of delete_order please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [read_order, order, order]
+    - [add_orderassignee, order, orderassignee]
+    - [change_orderassignee, order, orderassignee]
+    - [delete_orderassignee, order, orderassignee]
+    - [add_ordersubscriber, order, ordersubscriber]
+    - [change_ordersubscriber, order, ordersubscriber]
+    - [delete_ordersubscriber, order, ordersubscriber]
+    - [add_servicetype, order, servicetype]
+    - [change_servicetype, order, servicetype]
+    - [delete_servicetype, order, servicetype]
```
marcofucci: get rid of add_servicetype/change_servicetype/delete_servicetype please

fixtures/metadata/teams.yaml
```diff
@@ -1,3453 +1,4937 @@
...
+    - [add_revision, reversion, revision]
+    - [change_revision, reversion, revision]
+    - [delete_revision, reversion, revision]
+    - [add_version, reversion, version]
+    - [change_version, reversion, version]
+    - [delete_version, reversion, version]
+
```
marcofucci: please get rid of all the ones I listed in DA here as well

PeterNociar: ok

PeterNociar: done

datahub/core/auth.py
```diff
@@ -147,3 +149,35 @@ def _submit_form(session, source, url=None, params=None):
...
 
         return resp
+
+
+class TeamModelPermissionsBackend(CDMSUserBackend):
+    """Extension of CDMSUserBackend to include a team based permissions for user"""
+
+    def _get_team_permissions(self, user_obj):
+        """
+        This method is called by the ModelBackend _get_permissions() dynamicaly
```
elcct: A typo, should be `dynamically`.

PeterNociar:
I think all comments are addressed.

marcofucci:
I think that many `delete_` (and some `change_`) permissions in `teams.yaml` can be deleted but @reupen and @elcct know which ones, I've specified the OMIS ones that we don't need already.

datahub/core/permissions.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Check the permission_required on view against User Permissions
+    """
+
+    def has_permission(self, request, view):
+        """
+        Return `True` if permission is granted or there is no
+        required_permission set for view, `False` otherwise.
+        """
+        if hasattr(view, 'permission_required'):
```
reupen: Is this check needed – if you specify the class and don't set `permission_required` then shouldn't that be an error?

Also `permission_required` sounds like a boolean to me – you could swap the two words to make it `required_permission`.

datahub/core/permissions.py
```diff
@@ -0,0 +1,27 @@
...
+            try:
+                return request.user.has_perm(view.permission_required)
+            except AttributeError:
```
reupen: What is this protecting against? It's not obvious. If it's whether request.user is None, then I'd go for
`return request.user and request.user.has_perm(view.permission_required)` (similar to the permission classes in DRF).

PeterNociar: No its when the view.permission_required is not set

datahub/omis/invoice/test/views/test_public_invoice_details.py
```diff
@@ -131,6 +131,10 @@ def test_verbs_not_allowed(self, verb):
         """Test that makes sure the other verbs are not allowed."""
         order = OrderWithAcceptedQuoteFactory()
 
+        self._user = get_test_user()
```
reupen: Actually grant type GRANT_CLIENT_CREDENTIALS is not supposed to have a user. Looks like the logic in APITestMixin.get_token() isn't quite right there.

datahub/core/permissions.py
PeterNociar: the `permission_required` is already used in class-based views with a PermissionRequiredMixins Which wasn't working in our case as it was overriding the dispatch method and some of our mixins or parent classes were not calling super() ... 
So I wanted to keep it the same

datahub/search/company/views.py
```diff
@@ -41,6 +42,12 @@ class SearchCompanyParams:
 class SearchCompanyAPIView(SearchCompanyParams, SearchAPIView):
     """Filtered company search view."""
 
+    permission_classes = SearchAPIView.permission_classes + (UserHasPermissions, )
```
reupen: From a quick glance it looks like it would be cleaner to just specify both permission classes in SearchAPIView – aren't they common to all of the search views? You could add `permission_required = None` there to indicate it needs to be overridden.

reupen:
What prep work do we need to do in our environments before merging this (e.g. making sure all users are in teams)?

datahub/core/permissions.py
PeterNociar: For the first part. Ok  I can make it fail

datahub/metadata/models.py
```diff
@@ -55,7 +57,14 @@ class Role(BaseConstantModel):
 class TeamRole(BaseConstantModel):
     """Team role."""
 
-    pass
+    team_role_groups = models.ManyToManyField(
```
marcofucci: why don't we call this just `groups`?

datahub/omis/invoice/test/views/test_public_invoice_details.py
marcofucci: yes, very true!

datahub/omis/order/test/views/test_public_order_details.py
```diff
@@ -141,6 +141,10 @@ def test_verbs_not_allowed(self, verb):
             status=OrderStatus.quote_awaiting_acceptance
         )
 
+        self._user = get_test_user()
+        self._user.is_superuser = True
+        self._user.save()
```
marcofucci: same here, no user

datahub/omis/payment/test/views/test_public_payments.py
```diff
@@ -93,6 +93,10 @@ def test_verbs_not_allowed(self, verb):
         """Test that makes sure the other verbs are not allowed."""
         order = OrderPaidFactory()
 
+        self._user = get_test_user()
+        self._user.is_superuser = True
+        self._user.save()
```
marcofucci: same here, no user

datahub/omis/quote/test/views/test_public_quote_details.py
```diff
@@ -139,6 +139,10 @@ def test_verbs_not_allowed(self, verb):
             status=OrderStatus.quote_awaiting_acceptance
         )
 
+        self._user = get_test_user()
+        self._user.is_superuser = True
+        self._user.save()
```
marcofucci: same here, no user

datahub/core/permissions.py
reupen: :+1: on permission_required

PeterNociar:
Even if the users are not part of the team, they will just lose permission to most of the APIs (which is expected) but after they are added to the team, they will regain the permissions, so all should be ok.

datahub/core/permissions.py
```diff
@@ -0,0 +1,24 @@
+from django.core.exceptions import ImproperlyConfigured
+from rest_framework.permissions import BasePermission, DjangoModelPermissions
+
+
+class DjangoCrudPermission(DjangoModelPermissions):
+    """Extension of Permission class to include read permissions"""
+
+    perms_map = DjangoModelPermissions.perms_map
```
reupen: Would prefer to make a copy of DjangoModelPermissions.perms_map here.

reupen:
@marcofucci @elcct Any other comments?

datahub/core/permissions.py
```diff
@@ -0,0 +1,24 @@
+from django.core.exceptions import ImproperlyConfigured
+from rest_framework.permissions import BasePermission, DjangoModelPermissions
+
+
+class DjangoCrudPermission(DjangoModelPermissions):
+    """Extension of Permission class to include read permissions"""
+
+    perms_map = DjangoModelPermissions.perms_map.copy()
+    perms_map['GET'].append('%(app_label)s.read_%(model_name)s')
```
marcofucci: Not a big deal but I find:
```
perms_map = {
    **DjangoModelPermissions.perms_map,
    'GET': '%(app_label)s.read_%(model_name)s'
}
```
more readable

datahub/core/permissions.py
```diff
@@ -0,0 +1,24 @@
...
+
+class UserHasPermissions(BasePermission):
+    """
+    Check the permission_required on view against User Permissions
+    """
+
+    def has_permission(self, request, view):
+        """
+        Return `True` if permission is granted or there is no
+        required_permission set for view, `False` otherwise.
```
marcofucci: > or there is no required_permission set for view

Do you mean if permission_required is an empty list? Because if the view doesn't have permission_required an exception is raised instead (also required_permission should be permission_required?)

datahub/interaction/test/test_views.py
```diff
@@ -6,17 +6,27 @@
...
     EventServiceDeliveryFactory, InteractionFactory, ServiceDeliveryFactory
 )
 from datahub.investment.test.factories import InvestmentProjectFactory
+from datahub.metadata.test.factories import TeamFactory
 
 
 class TestInteractionV3(APITestMixin):
     """Tests for v3 interaction views."""
 
+    def test_intereaction_no_permissions(self):
```
marcofucci: typo _intereaction_ :)

datahub/search/omis/test/test_views.py
```diff
@@ -75,6 +76,14 @@ def setup_data(setup_es):
 class TestSearchOrder(APITestMixin):
     """Test specific search for orders."""
 
+    def test_order_search_no_permissions(self):
```
marcofucci: the _order_search_ bit already part of the class name so it's redundant. Can you rename it to _test_no_permissions_ instead?

datahub/core/permissions.py
PeterNociar: Oh yes, the logic changed since last time as per Reupen's comment. Comment fixed.

datahub/core/permissions.py
PeterNociar: Well, this version is force overriding the GET list of a parent.

marcofucci: Well, that's exactly what those lines do but as I said, not a biggie 👍

PR #587, opened 2017-10-30
Fix/event dates mandatory

elcct:
This makes `start_date` and `end_date` Event fields mandatory ~when creating an Event.~ at all times.

### DIFF
--- a/datahub/event/serializers.py
+++ b/datahub/event/serializers.py
@@ -13,8 +13,6 @@ class EventSerializer(serializers.ModelSerializer):
 
     default_error_messages = {
         'lead_team_not_in_teams': ugettext_lazy('Lead team must be in teams array.'),
-        'end_date_without_start_date': ugettext_lazy('Cannot have an end date without a start '
-                                                     'date.'),
         'end_date_before_start_date': ugettext_lazy('End date cannot be before start date.'),
         'uk_region_non_uk_country': ugettext_lazy('Cannot specify a UK region for a non-UK '
                                                   'country.')
--- a/datahub/event/serializers.py
+++ b/datahub/event/serializers.py
@@ -19,7 +19,7 @@ class EventSerializer(serializers.ModelSerializer):
         'uk_region_non_uk_country': ugettext_lazy('Cannot specify a UK region for a non-UK '
                                                   'country.')
     }
-    end_date = serializers.DateField(required=False, allow_null=True)
+    end_date = serializers.DateField()
     event_type = NestedRelatedField('event.EventType')
     location_type = NestedRelatedField('event.LocationType', required=False, allow_null=True)
     organiser = NestedAdviserField(required=False, allow_null=True)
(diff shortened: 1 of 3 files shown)


datahub/event/serializers.py
```diff
@@ -31,6 +31,16 @@ class EventSerializer(serializers.ModelSerializer):
...
+    start_date = serializers.DateField(required=False, allow_null=True)
+
+    def __init__(self, *args, **kwargs):
+        """Makes start_date and end_date fields required when creating an Event."""
+        super().__init__(*args, **kwargs)
+
+        if not self.instance:
+            for field in ('start_date', 'end_date', ):
+                self.fields[field].required = True
+                self.fields[field].allow_null = True
```
marcofucci: I think this allows you to create an event with start/end dates (as they are mandatory) but then you can patch them with None values and reset them, no?

elcct: That's correct, but the story I work on only says about creation time. I have sent a question to clarify the update behaviour. This could be modified later.

elcct: @marcofucci I got it clarified, that it should be required at all times, so changed it now.

datahub/event/serializers.py
```diff
@@ -31,6 +31,7 @@ class EventSerializer(serializers.ModelSerializer):
         'event.Programme', many=True, required=False, allow_empty=True
     )
     service = NestedRelatedField('metadata.Service')
+    start_date = serializers.DateField()
```
marcofucci: not sure if start_date/end_date were optional in the past, in this case you might have to use 
@reupen's `RequiredUnlessAlreadyBlankValidator` ?

datahub/event/serializers.py
```diff
@@ -63,14 +64,12 @@ def _validate_lead_team(self, combiner):
...
         start_date = combiner.get_value('start_date')
         end_date = combiner.get_value('end_date')
 
-        if end_date:
-            if not start_date:
-                errors['end_date'] = self.error_messages['end_date_without_start_date']
-            elif end_date < start_date:
-                errors['end_date'] = self.error_messages['end_date_before_start_date']
+        if start_date and end_date and end_date < start_date:
+            errors['end_date'] = self.error_messages['end_date_before_start_date']
```
marcofucci: I think you can delete the error_message `end_date_without_start_date`

datahub/event/serializers.py
elcct: I asked about this case and user should be required to fill the dates when editing an Event that doesn't have dates.

datahub/event/serializers.py
elcct: Good catch :) Removed.

datahub/event/test/test_views.py
```diff
@@ -353,9 +364,11 @@ def test_create_blank_failure(self):
             'address_1': '',
             'address_country': None,
             'address_town': '',
+            'end_date': '',
```
reupen: I think the intention was to use the value considered blank for the field type (to make sure it's rejected). So None for dates rather than ''.

elcct: I thought input from front end could be ''. Fixed.

PR #588, opened 2017-10-30
Add ability to mark an order as complete

marcofucci:
This adds the ability to mark an order as complete via a new endpoint `/v3/omis/order/<pk>/complete`

### DIFF
--- a/datahub/omis/order/test/test_models.py
+++ b/datahub/omis/order/test/test_models.py
@@ -17,8 +17,10 @@ from datahub.omis.payment.models import Payment
 from datahub.omis.quote.models import Quote
 
 from .factories import (
+    OrderAssigneeCompleteFactory,
     OrderAssigneeFactory,
     OrderFactory,
+    OrderPaidFactory,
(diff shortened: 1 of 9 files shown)


datahub/omis/order/test/test_models.py
```diff
@@ -472,6 +474,82 @@ def test_validation_error_if_amounts_less_then_total_cost(self):
...
+        """
+        order = OrderPaidFactory(status=allowed_status, assignees=[])
+        OrderAssigneeCompleteFactory(order=order)
+        adviser = AdviserFactory()
+
+        try:
+            with freeze_time('2018-07-12 13:00'):
+                order.complete(by=adviser)
+        except Exception:
+            pytest.fail('Should not raise a validator error.')
```
elcct: Should this not assume this is ValidationError exception?

marcofucci: It should actually be `Exception` as it can raise different exceptions so changed the text instead to make it clearer.

PR #591, opened 2017-10-31
Delete old and squashed OMIS order migrations

marcofucci:


reupen:
What about removing `replaces =` from `0001_squashed_0030_cancellation` (and updating the first payments migration)?

https://docs.djangoproject.com/en/1.11/topics/migrations/#squashing-migrations

marcofucci:
@reupen thanks for spotting that, I completely missed it :/

PR #594, opened 2017-11-01
[DH-881] Turn on USE_TZ

reupen:
The main effective change is that datetimes returned by the API have a Z at the end to indicate they're in UTC, and Django and DRF uses datetimes time-zone-aware internally.

The main snag was that the freezegun context manager returns a naive datetime (breaking comparisons based on it) – simplest workaround was to get the frozen time using django.utils.timezone.now().


reupen:
Have put back to work in progress – just spotted some more warnings showing up in tests to look at

reupen:
@marcofucci @elcct Have updated this now.

marcofucci:
Do we need the `Z` in the `freeze_time` calls?
Shouldn't it always use `tz_offset=0` if you don't pass anything?

datahub/core/test/test_models.py
```diff
@@ -11,23 +11,23 @@ class TestDisableableModel:
...
 
         obj = MyDisableableModel(disabled_on=None)
         assert not obj.was_disabled_on(date_on)
 
     def test_was_disabled(self):
         """If disabled_on < date_on, return True."""
-        disabled_on = dateutil_parse('2017-01-01 12:00:00')
-        date_on = dateutil_parse('2017-01-01 13:00:00')
+        disabled_on = dateutil_parse('2017-01-01 12:00:00Z')
+        date_on = dateutil_parse('2017-01-01T13:00:00Z')
```
marcofucci: Can we always include `T` or always exclude it? It helps trying to be consistent.

datahub/omis/order/test/views/test_order_details.py
```diff
@@ -851,7 +851,7 @@ def test_fails_with_incomplete_billing_address(self):
 class TestMarkOrderAsComplete(APITestMixin):
     """Test cases for marking an order as complete."""
 
-    @freeze_time('2017-04-18 13:00')
+    @freeze_time('2017-04-18 13:00Z')
```
marcofucci: `Z` not needed?

reupen:
@marcofucci I've made those more consistent, have a look now.

reupen:
@marcofucci Could you also have a look at the logic around Order.paid_on (if you haven't already)?

PR #595, opened 2017-11-01
Additional Investment Project fields.

elcct:
This adds three new metadata fields to the Investment Project.

`investor_type`, `level_of_engagement` and `specific_programme`

As I added these metadata fields in the `investment` app, probably it is worth moving the investment's metadata fields that exist in `metadata` app to `investment` app. - perhaps in another PR?


datahub/core/management/commands/loadmetadata.py
```diff
@@ -29,6 +30,12 @@
     EVENTS_FIXTURE_DIR / 'event_types.yaml',
 )
 
+INVESTMENTS_FIXTURES = (
```
reupen: Could register these using datahub.metadata.fixtures.Fixture (like in the interaction app).

datahub/investment/fixtures/specific_programmes.yaml
```diff
@@ -0,0 +1,69 @@
+- model: investment.specificprogramme
```
reupen: We might need to include inactive ones from CDMS (I see four) and set disabled_on for them... @jim68000 can confirm.

reupen:
The serialisers and views will need updating to add the fields as well.

> As I added these metadata fields in the investment app, probably it is worth moving the investment's metadata fields that exist in metadata app to investment app. - perhaps in another PR?

Would be nice I think that'd probably be better left till after November as it's fairly involved :cry:

datahub/core/management/commands/loadmetadata.py
elcct: Thanks for reminding. Fixed.

elcct:
@reupen I have indeed forgot to include them in serialisers and views :blush:, now added.

datahub/investment/fixtures/specific_programmes.yaml
ztolley: Indeed. The spec wants all of them and the FE will filter out the inactive ones, unless the current record has a value that is inactive.

datahub/investment/test/factories.py
```diff
@@ -26,7 +29,9 @@ class InvestmentProjectFactory(factory.django.DjangoModelFactory):
     estimated_land_date = date(2020, 1, 1)
     investment_type_id = InvestmentType.commitment_to_invest.value.id
     referral_source_activity_id = ReferralSourceActivity.cold_call.value.id
-
+    investor_type_id = InvestorType.new_investor.value.id
+    level_of_involvement_id = Involvement.no_involvement.value.id
+    specific_programme_id = SpecificProgramme.space.value.id
```
marcofucci: @elcct @reupen I was thinking to get rid of all these constants or at least not keep adding them and use something like `factory.LazyFunction(SpecificProgramme.objects.first)` instead. We could even use a random one `factory.LazyFunction(SpecificProgramme.objects.order_by('?').first)` and/or create a util thing to make is nicer e.g. `factory_utils.ForeignKey(SpecificProgramme)`.

What do you think?

marcofucci: yep for those, we could have the util class get an exact querystring so that we can get the country by name for example?

elcct: I think having the util class etc. is a good way forward. I also think that could be good for separate PR as a general metadata cleanup.

PR #596, opened 2017-11-01
Fix Investment Project status filter.

elcct:
- adds missing `status` field to the ES model
- changes serializer to accept multiple status filters.
- enhances test to check for multpile status filter

### DIFF
--- a/datahub/search/investment/test/test_views.py
+++ b/datahub/search/investment/test/test_views.py
@@ -20,7 +20,7 @@ def setup_data(setup_es):
         InvestmentProjectFactory(
             name='abc defg',
             description='investmentproject1',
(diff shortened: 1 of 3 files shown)


datahub/search/investment/test/test_views.py
```diff
@@ -44,6 +44,21 @@ def setup_data(setup_es):
             uk_region_locations=[
                 constants.UKRegion.north_west.value.id,
             ]
+        ),
+        InvestmentProjectFactory(
+            name='won project',
+            description='investmentproject3',
+            estimated_land_date=datetime.datetime(2057, 6, 13, 9, 44, 31, 62870),
```
reupen: `estimated_land_date` should be a date (rather than a datetime).

elcct: True. Fixed.

PR #597, opened 2017-11-02
Omis cancel order

marcofucci:
Add ability to cancel an order with `/v3/order/<order-pk>/cancel` and passing a cancellation reason.


datahub/omis/order/serializers.py
```diff
@@ -206,6 +217,24 @@ def complete(self):
         return self.instance
 
 
+class CancelOrderSerializer(OrderSerializer):
+    """DRF serializer for cancelling an order."""
+
+    cancellation_reason = NestedRelatedField(CancellationReason, required=True)
```
reupen: Isn't `required=True` the default?

datahub/omis/order/test/test_models.py
```diff
@@ -551,6 +552,74 @@ def test_validation_error_if_not_all_actual_time_set(self):
...
+        Test that the order can be cancelled if it's in one of the allowed statuses.
+        """
+        reason = CancellationReason.objects.order_by('?').first()
+        order = OrderFactory(status=allowed_status)
+        adviser = AdviserFactory()
+
+        try:
+            with freeze_time('2018-07-12 13:00'):
+                order.cancel(by=adviser, reason=reason)
+        except Exception:
```
reupen: Is this really needed? I mean the test will fail if it raises an exception and there are other assertions below.

datahub/omis/order/serializers.py
marcofucci: 👍

datahub/omis/order/test/test_models.py
marcofucci: hm maybe you are right

PR #598, opened 2017-11-02
[DH-881] Add support for single sign-on (SSO) using RFC 7662

reupen:
This includes prepending data-hub: to the existing OAuth2 scopes so that they make sense within the central authentication broker.

Existing OAuth flows should continue to work alongside the token introspection flow.


marcofucci:
I'm not sure to understand, where does the `http://sso` come from?

datahub/oauth/migrations/0003_add_data_hub_to_scopes.py
```diff
@@ -0,0 +1,43 @@
...
+        app_scope.scopes = [SCOPES_MAPPING.get(scope, scope) for scope in app_scope.scopes]
+        app_scope.save()
+
+    # Note: this does not handle tokens with multiple scopes, however no such tokens were in use
+    # pre-SSO
+    for old_scope, new_scope in SCOPES_MAPPING.items():
+        AccessToken.objects.filter(scope=old_scope).update(scope=new_scope)
+
+
+def _map_scope(scope):
```
elcct: Is this being used?

reupen:
> I'm not sure to understand, where does the http://sso come from?

Sorry, I just put dummy data in docker-compose.yaml and the other files following the pattern of CDMS_AUTH_URL and other values. What do you suggest?

datahub/oauth/migrations/0003_add_data_hub_to_scopes.py
reupen: Ah, nope, I'll get rid of it :+1:

marcofucci:
@reupen I see, that's fine 👍

reupen:
Have updated the front-end instructions in the README as well (it won't be true until the front end changes are merged).

.circleci/config.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       AWS_ACCESS_KEY_ID: foo
       AWS_SECRET_ACCESS_KEY: bar
       DOCUMENTS_BUCKET: baz
+      SSO_ENABLED: 'True'
```
feedmypixel: @reupen is it possible to have this as an env? As I will need to overwrite it in `data-hub-frontend` CircleCi jobs.

`SSO_ENABLED: ${SSO_ENABLED}` with a default in [config/settings/sample.env](https://github.com/uktrade/data-hub-leeloo/blob/develop/config/settings/sample.env)

PR #599, opened 2017-11-02
Added missing fields in serializer.

elcct:
Adds missing fields in serialiser. I forgot to add them to the Meta fields :blush:


reupen:
Can we update one of the tests to make sure the fields are there?

elcct:
Good idea, so mistake like that won't happen again :) Added.

PR #600, opened 2017-11-03
Restore further_info and existing_agents fields

marcofucci:
The OMIS order fields further_info and existing_agents are not legacy fields any more as we realised that we need them.

This restores them and allow them to be changes.

### DIFF
--- a/datahub/omis/order/migrations/0001_squashed_0030_cancellation.py
+++ b/datahub/omis/order/migrations/0001_squashed_0030_cancellation.py
@@ -158,7 +158,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='order',
             name='contacts_not_to_approach',
-            field=models.TextField(blank=True, help_text='Are there contacts that DIT should not approach?'),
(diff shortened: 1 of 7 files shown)


datahub/omis/order/migrations/0001_squashed_0030_cancellation.py
```diff
@@ -168,12 +168,12 @@ class Migration(migrations.Migration):
...
             model_name='order',
             name='existing_agents',
-            field=models.TextField(blank=True, editable=False, help_text='Legacy field. Details of any existing agents.'),
+            field=models.TextField(blank=True, help_text='Contacts the company already has in the market.'),
         ),
         migrations.AddField(
             model_name='order',
             name='further_info',
-            field=models.TextField(blank=True, editable=False, help_text='Legacy field. Further information.'),
+            field=models.TextField(blank=True, help_text='Additional notes and useful information.'),
```
elcct: Is "and useful information" bit redundant? Also "useful" is subjective.

marcofucci: That is the label in the frontend interface and I think it's been through our usability testing. 
I'll keep an eye and change it if that stops being the case.

PR #603, opened 2017-11-03
Create an Access token to bypass SSO in datahub-frontend UAT

feedmypixel:
This work enables an Access token (with a value provided via `$OAUTH2_DEV_TOKEN`) to be created within the backend for CircleCi User Acceptance tests. The same access token value is then used in `datahub-frontend` to byPass SSO on CircleCi.

### Of Note
To cover not running SSO in UAT for `data-hub-frontend`:
- @r4vi and @lgarvey have spoken about an SSO stub. When this has been built we can look at integration testing for `data-hub-frontend` -> `SSO`
- I have asked the SSO team if they can add a healthcheck endpoint so we can add this to the `data-hub-frontend` health check and in the future potentially a separately hosted `data-hub-frontend` status page.

This work is related to [#877](https://github.com/uktrade/data-hub-frontend/pull/877)

This branch is green when run with `data-hub-frontend` on CircleCi in job: https://circleci.com/gh/uktrade/data-hub-frontend/9987

### DIFF
--- a/setup-uat.sh
+++ b/setup-uat.sh
@@ -7,23 +7,22 @@ python /app/manage.py migrate
 python /app/manage.py loadmetadata
 python /app/manage.py load_omis_metadata
 
-echo "from oauth2_provider.models import Application;
-from datahub.oauth.models import OAuthApplicationScope;
-app = Application.objects.create(
-    name='circleci',
-    client_id='${API_CLIENT_ID}',
-    client_secret='${API_CLIENT_SECRET}',
-    client_type=Application.CLIENT_CONFIDENTIAL,
-    authorization_grant_type=Application.GRANT_PASSWORD
-);
-OAuthApplicationScope.objects.create(application=app, scopes=['data-hub:internal-front-end'])" | /app/manage.py shell
+echo "import datetime
+from oauth2_provider.models import AccessToken
+from django.utils.timezone import now
+from datahub.company.models import Advisor
 
-echo "from datahub.company.models import Advisor;
-Advisor.objects.create_user(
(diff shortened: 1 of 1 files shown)


r4vi:
@feedmypixel you can try make fakesso if in circle if you like https://github.com/r4vi/fakesso

setup-uat.sh
```diff
@@ -7,23 +7,33 @@ python /app/manage.py migrate
...
 python /app/manage.py load_omis_metadata
 
-echo "from oauth2_provider.models import Application;
-from datahub.oauth.models import OAuthApplicationScope;
+echo "import datetime
+from oauth2_provider.models import AccessToken, Application
+from datahub.oauth.models import OAuthApplicationScope
+from datahub.company.models import Advisor
+
 app = Application.objects.create(
```
reupen: You should be able to get rid of the Application if you're creating an access token directly.

feedmypixel: > You should be able to get rid of the Application if you're creating an access token directly.

Do we not need the `app` as an argument for `AccessToken.objects.create()` see line 33

feedmypixel:
> @feedmypixel you can try make fakesso if in circle if you like https://github.com/r4vi/fakesso

@r4vi great thanks, I'm not gonna integrate this s this piece of work but will come back to it later in the week and give you a shout if there are any questions.

r4vi:
@feedmypixel I've tried it with export-wins and it seems to work 
peek 2017-11-06 10-46

feedmypixel:
> @feedmypixel I've tried it with export-wins and it seems to work

awesome :)

PR #607, opened 2017-11-07
User without team now returns proper 403

PeterNociar:
... if he has no permission, rather than blowing up

### DIFF
--- a/datahub/core/auth.py
+++ b/datahub/core/auth.py
@@ -159,9 +159,9 @@ class TeamModelPermissionsBackend(CDMSUserBackend):
         This method is called by the ModelBackend _get_permissions() dynamically
         as part of aggregating user, group and team permissions
         """
-        try:
+        if user_obj.dit_team and user_obj.dit_team.role:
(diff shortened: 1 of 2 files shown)


datahub/core/auth.py
```diff
@@ -159,7 +159,11 @@ def _get_team_permissions(self, user_obj):
         This method is called by the ModelBackend _get_permissions() dynamically
         as part of aggregating user, group and team permissions
         """
-        groups = user_obj.dit_team.role.groups.all()
+        try:
+            groups = user_obj.dit_team.role.groups.all()
+        except AttributeError:
```
elcct: Wouldn't be better to check if attribute exists rather than catching an Exception?

reupen: Yeah I agree, I would prefer `if user_obj.dit_team and user_obj.dit_team.role:`

It's explicit and more specific.

PeterNociar: I thought that this is more "pythonic" but ok :)

PR #609, opened 2017-11-07
Include user's permissions in whoami endpoint response.

PeterNociar:
Adding a set of permissions into a who_am_i API response to allow FE to filter menus for different users.

### DIFF
--- a/datahub/core/permissions.py
+++ b/datahub/core/permissions.py
@@ -1,6 +1,6 @@
+from collections import defaultdict
 from django.core.exceptions import ImproperlyConfigured
 from rest_framework.permissions import BasePermission, DjangoModelPermissions
-from collections import defaultdict
 
 
 class DjangoCrudPermission(DjangoModelPermissions):
@@ -26,6 +26,9 @@ class UserHasPermissions(BasePermission):
 
 
 def serialize_permissions(permissions):
+    """
+    Serialize permissions into simplified structure
+    """
     formatted_permissions = defaultdict(list)
     for perm in permissions:
         app, action_model = perm.split('.', 1)
@@ -34,4 +37,3 @@ def serialize_permissions(permissions):
         formatted_permissions[model].append(action)
 
     return formatted_permissions
-
--- a/datahub/core/permissions.py
+++ b/datahub/core/permissions.py
@@ -1,5 +1,6 @@
 from django.core.exceptions import ImproperlyConfigured
 from rest_framework.permissions import BasePermission, DjangoModelPermissions
+from collections import defaultdict
 
 
 class DjangoCrudPermission(DjangoModelPermissions):
@@ -22,3 +23,14 @@ class UserHasPermissions(BasePermission):
         if not hasattr(view, 'permission_required'):
             raise ImproperlyConfigured()
         return request.user and request.user.has_perm(view.permission_required)
+
+
+def serialize_permissions(permissions):
+    formatted_permissions = defaultdict(lambda: defaultdict(list))
+    for perm in permissions:
+        app, action_model = perm.split('.')
+        action, model = action_model.split('_')
+        formatted_permissions[app][model].append(action)
+
+    return formatted_permissions
+
--- a/datahub/core/permissions.py
+++ b/datahub/core/permissions.py
@@ -1,4 +1,3 @@
-from collections import defaultdict
 from django.core.exceptions import ImproperlyConfigured
 from rest_framework.permissions import BasePermission, DjangoModelPermissions
 
@@ -25,15 +24,7 @@ class UserHasPermissions(BasePermission):
         return request.user and request.user.has_perm(view.permission_required)
 
 
-def serialize_permissions(permissions):
     """
-    Serialize permissions into simplified structure
     """
-    formatted_permissions = defaultdict(list)
-    for perm in permissions:
-        app, action_model = perm.split('.', 1)
-        action, model = action_model.split('_', 1)
 
-        formatted_permissions[model].append(action)
 
-    return formatted_permissions
--- a/datahub/core/permissions.py
+++ b/datahub/core/permissions.py
@@ -26,11 +26,12 @@ class UserHasPermissions(BasePermission):
 
 
 def serialize_permissions(permissions):
-    formatted_permissions = defaultdict(lambda: defaultdict(list))
+    formatted_permissions = defaultdict(list)
     for perm in permissions:
-        app, action_model = perm.split('.')
-        action, model = action_model.split('_')
-        formatted_permissions[app][model].append(action)
+        app, action_model = perm.split('.', 1)
+        action, model = action_model.split('_', 1)
+
+        formatted_permissions[model].append(action)
 
     return formatted_permissions
 
--- a/datahub/core/permissions.py
+++ b/datahub/core/permissions.py
@@ -22,9 +22,3 @@ class UserHasPermissions(BasePermission):
         if not hasattr(view, 'permission_required'):
             raise ImproperlyConfigured()
         return request.user and request.user.has_perm(view.permission_required)
-
-
-    """
-    """
-
-
--- a/datahub/core/test/factories.py
+++ b/datahub/core/test/factories.py
@@ -60,3 +60,14 @@ class GroupFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = 'auth.Group'
+
+
+class PermissionFactory(factory.django.DjangoModelFactory):
+    """Permission Factory"""
+
+    name = factory.Faker('name')
+    codename = factory.Faker('codename')
+    content_type = factory.SubFactory()
+
+    class Meta:
+        model = 'auth.Permission'
--- a/datahub/core/test/factories.py
+++ b/datahub/core/test/factories.py
@@ -66,8 +66,8 @@ class GroupFactory(factory.django.DjangoModelFactory):
 class PermissionFactory(factory.django.DjangoModelFactory):
     """Permission Factory"""
 
-    name = factory.Faker('name')
-    codename = factory.Faker('name')
+    name = factory.Faker('word')
+    codename = factory.Faker('word')
     content_type = factory.Iterator(ContentType.objects.all())
 
     class Meta:
--- a/datahub/core/test/factories.py
+++ b/datahub/core/test/factories.py
@@ -1,6 +1,7 @@
 from functools import wraps
 
 import factory
+from django.contrib.contenttypes.models import ContentType
 
 
 def to_many_field(wrapped_func):
@@ -66,8 +67,8 @@ class PermissionFactory(factory.django.DjangoModelFactory):
     """Permission Factory"""
 
     name = factory.Faker('name')
-    codename = factory.Faker('codename')
-    content_type = factory.SubFactory()
+    codename = factory.Faker('name')
+    content_type = factory.Iterator(ContentType.objects.all())
 
     class Meta:
         model = 'auth.Permission'
--- a/datahub/user/serializers.py
+++ b/datahub/user/serializers.py
@@ -7,6 +7,7 @@ class WhoAmISerializer(serializers.ModelSerializer):
     """Adviser serializer for that includes a permissions"""
 
     name = serializers.CharField()
+    permissions = serializers.SerializerMethodField()
 
     class Meta:
         model = Advisor
@@ -20,13 +21,14 @@ class WhoAmISerializer(serializers.ModelSerializer):
             'contact_email',
             'telephone_number',
             'dit_team',
+            'permissions',
         )
         depth = 2
 
-    def serialize_permissions(self, permissions):
+    def get_permissions(self, obj):
         """Serialize permissions into simplified structure."""
         formatted_permissions = {}
-        for perm in permissions:
+        for perm in obj.get_all_permissions():
             app, action_model = perm.split('.', 1)
             action, model = action_model.split('_', 1)
 
@@ -36,11 +38,3 @@ class WhoAmISerializer(serializers.ModelSerializer):
                 formatted_permissions[model] = [action]
 
         return formatted_permissions
-
-    @property
-    def data(self):
-        """Data property."""
-        ret = super().data
-        ret['permissions'] = self.serialize_permissions(self.instance.get_all_permissions())
-
-        return ret
--- a/datahub/user/serializers.py
+++ b/datahub/user/serializers.py
@@ -1,5 +1,3 @@
-from collections import defaultdict
-
 from rest_framework import serializers
 
 from datahub.company.models.adviser import Advisor
@@ -28,11 +26,12 @@ class WhoAmISerializer(serializers.ModelSerializer):
 
     def get_permissions(self, obj):
         """Serialize permissions into simplified structure."""
-        formatted_permissions = defaultdict(dict)
+        formatted_permissions = {}
         for perm in obj.get_all_permissions():
             _, action_model = perm.split('.', 1)
             action, model = action_model.split('_', 1)
 
-            formatted_permissions[model][action] = True
+            model_dict = formatted_permissions.setdefault(model, {})
+            model_dict[action] = True
 
         return formatted_permissions
--- a/datahub/user/serializers.py
+++ b/datahub/user/serializers.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from rest_framework import serializers
 
 from datahub.company.models.adviser import Advisor
@@ -6,7 +8,6 @@ from datahub.company.models.adviser import Advisor
 class WhoAmISerializer(serializers.ModelSerializer):
     """Adviser serializer for that includes a permissions"""
 
-    name = serializers.CharField()
     permissions = serializers.SerializerMethodField()
 
     class Meta:
@@ -27,14 +28,11 @@ class WhoAmISerializer(serializers.ModelSerializer):
 
     def get_permissions(self, obj):
         """Serialize permissions into simplified structure."""
-        formatted_permissions = {}
+        formatted_permissions = defaultdict(dict)
         for perm in obj.get_all_permissions():
-            app, action_model = perm.split('.', 1)
+            _, action_model = perm.split('.', 1)
             action, model = action_model.split('_', 1)
 
-            if model in formatted_permissions:
-                formatted_permissions[model][action] = True
-            else:
-                formatted_permissions[model] = {action: True}
+            formatted_permissions[model][action] = True
 
         return formatted_permissions
--- a/datahub/user/serializers.py
+++ b/datahub/user/serializers.py
@@ -1,0 +1,46 @@
+from rest_framework import serializers
+
+from datahub.company.models.adviser import Advisor
+
+
+class WhoAmISerializer(serializers.ModelSerializer):
+    """Adviser serializer for that includes a permissions"""
+
+    name = serializers.CharField()
+
+    class Meta:
+        model = Advisor
+        fields = (
+            'id',
+            'name',
+            'last_login',
+            'first_name',
+            'last_name',
+            'email',
+            'contact_email',
+            'telephone_number',
+            'dit_team',
+        )
+        depth = 2
+
+    def serialize_permissions(self, permissions):
+        """Serialize permissions into simplified structure."""
+        formatted_permissions = {}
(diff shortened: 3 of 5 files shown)


datahub/core/permissions.py
```diff
@@ -22,3 +23,17 @@ def has_permission(self, request, view):
         if not hasattr(view, 'permission_required'):
             raise ImproperlyConfigured()
         return request.user and request.user.has_perm(view.permission_required)
+
+
+def serialize_permissions(permissions):
+    """
+    Serialize permissions into simplified structure
```
elcct: That comment could fit in single line? Also missing dot at the end (as couple of others).

datahub/user/views.py
```diff
@@ -1,13 +1,27 @@
...
-    return Response(data=serializer.data)
+    data = serializer.data
+    data['team'] = request.user.dit_team.name
+    data['team_role'] = request.user.dit_team.role.name
+
+    permissions = set()
+    for backend in auth.get_backends():
+        try:
+            permissions.update(backend.get_all_permissions(user_obj=request.user))
+        except AttributeError:
```
elcct: Same as before, I think it would be better to check for attribute rather than catching an exception.

datahub/user/views.py
```diff
@@ -1,13 +1,27 @@
...
 
 @api_view()
 @permission_classes([IsAuthenticated])
 def who_am_i(request):
     """Return the current user. This view is behind a login."""
     serializer = AdviserSerializer(request.user)
-    return Response(data=serializer.data)
+    data = serializer.data
+    data['team'] = request.user.dit_team.name
+    data['team_role'] = request.user.dit_team.role.name
```
marcofucci: can't you define and use a serializer which extends from `AdviserSerializer`?

datahub/user/test/test_views.py
```diff
@@ -1,19 +1,43 @@
...
 
         assert response.status_code == status.HTTP_200_OK
         assert response.data['name'] == user_test.name
         assert response.data['first_name'] == user_test.first_name
         assert response.data['id'] == str(user_test.pk)
+        assert response.data['team'] == team.name
+        assert response.data['team_role'] == role.name
+        assert response.data.get('permissions') is not None
+        for model in (model_name_1, model_name_2):
+            assert action in response.data['permissions'][model]
```
marcofucci: Can you change it to assert the whole response instead of the single elements? This way we can use it as documentation as well, that is, we can see immediately that the response body looks like.

PeterNociar: Ok

datahub/core/permissions.py
```diff
@@ -22,3 +23,17 @@ def has_permission(self, request, view):
...
+    """
+    formatted_permissions = defaultdict(list)
```
reupen: I think I'd have gone for using dict,setdefault() over using defaultdict since the dict is being returned from this function.

datahub/core/permissions.py
```diff
@@ -22,3 +23,17 @@ def has_permission(self, request, view):
...
+    for perm in permissions:
+        app, action_model = perm.split('.', 1)
+        action, model = action_model.split('_', 1)
```
reupen: Just thinking this split might constrain how we can use permissions in future. @marcofucci  @elcct what do you think?

datahub/core/test/factories.py
```diff
@@ -60,3 +61,14 @@ class GroupFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = 'auth.Group'
+
+
+class PermissionFactory(factory.django.DjangoModelFactory):
+    """Permission Factory"""
+
+    name = factory.Faker('name')
+    codename = factory.Faker('name')
```
reupen: Not sure 'name' is a good choice here?

datahub/user/views.py
reupen: Which attribute not existing is it protecting against :thinking:

datahub/core/permissions.py
marcofucci: Not sure what the response is, that's why it would be good to have the full content in the tests https://github.com/uktrade/data-hub-leeloo/pull/609#discussion_r149362439

PeterNociar: This is bbecause of how django permissions are formated.

perms_map = {
        'GET': [],
        'OPTIONS': [],
        'HEAD': [],
        'POST': ['%(app_label)s.add_%(model_name)s'],
        'PUT': ['%(app_label)s.change_%(model_name)s'],
        'PATCH': ['%(app_label)s.change_%(model_name)s'],
        'DELETE': ['%(app_label)s.delete_%(model_name)s'],
    }

datahub/core/permissions.py
PeterNociar: Sure

PeterNociar: Hmmm, how would you use the setdefault() ?

datahub/user/views.py
PeterNociar: get_all_permissions

datahub/core/test/factories.py
PeterNociar: Ok found the magic incantation to make Faker work 
'word'

datahub/company/serializers.py
```diff
@@ -37,6 +38,46 @@ class Meta:
...
+            if model in formatted_permissions:
+                formatted_permissions[model].append(action)
+            else:
+                formatted_permissions[model] = [action]
+
+        return formatted_permissions
+
+    def get_permissions(self, user):
+        """Get permissions method."""
+        permissions = set()
```
reupen: Can't this use user.get_all_permissions()?

datahub/company/serializers.py
```diff
@@ -37,6 +38,46 @@ class Meta:
...
+    def data(self):
+        """Data property."""
+        ret = super().data
+        perms = self.get_permissions(self.instance)
+        ret['permissions'] = self.serialize_permissions(perms)
+        ret['team'] = self.instance.dit_team.name if self.instance.dit_team and \
+            self.instance.dit_team.name else None
+        ret['team_role'] = self.instance.dit_team.role.name if self.instance.dit_team and \
+            self.instance.dit_team.role and \
+            self.instance.dit_team.role.name else None
```
marcofucci: Maybe we can use
```
dit_team = NestedRelatedField(Team, read_only=True)
```
instead of do that in the overridden data?

datahub/company/serializers.py
```diff
@@ -37,6 +38,46 @@ class Meta:
...
+                permissions.update(backend.get_all_permissions(user_obj=user))
+        return permissions
+
+    @property
+    def data(self):
+        """Data property."""
+        ret = super().data
+        perms = self.get_permissions(self.instance)
+        ret['permissions'] = self.serialize_permissions(perms)
+        ret['team'] = self.instance.dit_team.name if self.instance.dit_team and \
```
reupen: Doesn't AdviserSerializer already have dit_team?

datahub/company/serializers.py
```diff
@@ -37,6 +38,46 @@ class Meta:
...
+            self.instance.dit_team.name else None
+        ret['team_role'] = self.instance.dit_team.role.name if self.instance.dit_team and \
```
reupen: You should be able to use a serializer field for these (with source and default specified as it's a nested attribute).

Would stick with consistent field names i.e. dit_team_role. Also would the serialized format consistent with other endpoints/properties i.e. {"id": ..., "name": ...} via NestedRelatedField.

datahub/company/serializers.py
marcofucci: oh yes  @reupen is right, adviserserializer should have dit_team

datahub/company/serializers.py
marcofucci: +1

datahub/user/test/test_views.py
```diff
@@ -1,19 +1,41 @@
...
+
+        user_test = get_test_user(team=team)
+        user_test.user_permissions.add(user_permission)
+
         url = reverse('who_am_i')
         response = self.api_client.get(url)
-        user_test = get_test_user()
 
         assert response.status_code == status.HTTP_200_OK
         assert response.data['name'] == user_test.name
```
reupen: I think what @marcofucci was suggesting was asserting on the entire response – so `assert response.json() == {...}`

marcofucci: yeah

datahub/company/serializers.py
```diff
@@ -37,6 +38,46 @@ class Meta:
         depth = 1
 
 
+class UserSerializer(AdviserSerializer):
```
reupen: Not sure on the class name since 'adviser' is just what we call users. I'd go for something like WhoAmISerializer and stick it in the same app as the view.

datahub/company/serializers.py
PeterNociar: In the email chain about the permissions 'team' and 'team_role' was in the latest example, but sure I can keep it dit_team

datahub/core/permissions.py
marcofucci: If the response body is something like:
```
'permissions': {
    'company': ['read_company', 'change_company'],
    'contact': ['read_contact', 'read_contact'],
    ...
}
```
it could work as it supports whatever custom permissions we add to a model.
If you guys like that, we shouldn't get the model by splitting the perm codename but get it from the ContentType attached to the Permission object instead.

@reupen what other option do you have in mind for the response?

reupen: > we shouldn't get the model by splitting the perm codename but get it from the ContentType attached to the Permission object instead.

Yeah that would be a lot better.

I prefer your suggestion to what is there currently.

Still I would point out that it's not the easiest thing for the FE to use. I'd expect something like:
```json
{
    "read_company": true,
    "read_contact": true
}
```
to be easier because it can just be indexed into. 

But I'm OK with your suggestion.

PeterNociar:
Any more comments?

datahub/user/serializers.py
```diff
@@ -0,0 +1,46 @@
...
+            else:
+                formatted_permissions[model] = [action]
+
+        return formatted_permissions
+
+    @property
+    def data(self):
+        """Data property."""
+        ret = super().data
+        ret['permissions'] = self.serialize_permissions(self.instance.get_all_permissions())
```
marcofucci: Maybe you can use http://www.django-rest-framework.org/api-guide/fields/#serializermethodfield instead of overriding `data`?

datahub/user/test/test_views.py
```diff
@@ -1,19 +1,69 @@
...
+                    'id': str(team.uk_region_id),
+                    'disabled_on': None,
+                    'name': 'East Midlands',
+                },
+                'country': {
+                    'id': str(team.country_id),
+                    'disabled_on': None,
+                    'name': 'France',
+                }
+            },
```
marcofucci: @reupen @elcct I know that it's like this in `AdviserSerializer` but shouldn't `dit_team` be more consistent with the rest?
Something like:
```
'dit_team': {
    'id': str(team.id),
    'name': 'Test Team',
    'role': {
        'id': str(role.id),
        'name': 'Test Role',
    },
    'uk_region': {
        'id': str(team.uk_region_id),
        'name': 'East Midlands',
    },
    'country': {
        'id': str(team.country_id),
        'name': 'France',
    }
}
```
Not even sure we need `uk_region` and `country` in dit_team but maybe.

What do you think?

datahub/user/test/test_views.py
```diff
@@ -1,19 +1,69 @@
...
+            'permissions': {
+                model_name_1: [action],
+                model_name_2: [action],
+            }}
```
marcofucci: @feedmypixel @yeahfro @teneightfive 

We are suggesting.

*Option 1:*
```
{
    ...
    'permissions': {
        'company': [
            'read_company',
            'change_company'
        ],
        'order': [
            'read_order'
        ],
    }
    ...
}
```

or *Option 2:*
```
{
    ...
    'permissions': [
        'read_company',
        'change_company',
        'read_order'
    ]
    ...
}
```

What do you guys think/prefer?

PeterNociar: At the moment it's 
{
    ...
    'permissions': {
        'company': ['read', 'change'],
        'order': ['read'],

marcofucci: I would keep the whole codename so `read_company`, `change_company`. It's a bit ugly but those names can be chosen and don't always have the '_'.

PeterNociar: Ok, in that case, we need to go for option 2 from your list, as we won't have the model name without parsing the _

datahub/user/serializers.py
PeterNociar: Great! I was looking for a way to do that. My google-fu was weak

datahub/user/test/test_views.py
marcofucci: If you have the perm codename, you can get the model name from the contenttype linked to it, no?

PeterNociar: Yeah, I can do that

PeterNociar: So Option 1?

datahub/user/test/test_views.py
reupen: Yeah the problem is that AdviserSerializer relates to v1 endpoints (`/adviser/`, `/dashboard/` and this one). So we need an AdviserSerializerV3 really. Though this endpoint hasn't been moved to v3 so it might be something to postpone (though I'd rather have moved this to v3 now).

datahub/user/test/test_views.py
PeterNociar: @feedmypixel @yeahfro @teneightfive @ztolley Any preference in the response format?

feedmypixel: @PeterNociar can you give some context as to what this work is about? Can you add a description please?

PeterNociar: Sure. In short, its providing a list of users permissions in who_am_i so that the FE can do the menus, filtering.

PeterNociar: Question is, what format do you want it to be. Marco posted 2 Options above. Do you want it to be just flat representation (Option 2 ) Or grouped by the model (Option 1)

PeterNociar: Or some 3rd option. If you prefer some other format

feedmypixel: I'm a little unsure what this means? and what effect this would have in the frontend? FYI I have had no discussions whatsoever with anyone about this so I am totally green on the functionality required in the frontend here.

What does for instance 
```
permissions.company: [
  read_company,
  change_company
]```

allow a user to do? does it mean that a user can `read` and `change` all companies across the application?


if so I feel a format in the frontend like the following would be preferable:

```
{
    ...
    permissions: {
        company: {
            read: true,
            change: true,
        },
        order: {
            read: true,
        },
    }
    ...
}
```

However if that is a pain we can manipulate the call in the frontend to give us this. So @marcofucci your `option 2` with the verbose names would be the best for us.

@teneightfive am I missing anything here?

Comment (yeahfro) [reply datahub/user/test/test_views.py]: Agree with @feedmypixel my preference would be:

```
{
    ...
    permissions: {
        company: {
            read: true,
            change: true,
        },
        order: {
            read: true,
        },
    }
    ...
}```

PeterNociar: @feedmypixel  You got it right. And sure, I can make it in the format you presented

datahub/user/test/test_views.py
```diff
@@ -1,19 +1,69 @@
...
+            'telephone_number': '',
+            'dit_team': {
+                'id': str(team.id),
+                'disabled_on': None,
+                'name': 'Test Team',
+                'role': {
+                    'id': str(role.id),
+                    'disabled_on': None,
+                    'name': 'Test Role',
+                    'groups': [group.id],
```
marcofucci: can we get rid of `groups`? We don't need it and the frontend my start relying on it if we include it.

datahub/user/serializers.py
```diff
@@ -0,0 +1,37 @@
...
+            'permissions',
+        )
+        depth = 2
+
+    def get_permissions(self, obj):
+        """Serialize permissions into simplified structure."""
+        formatted_permissions = {}
+        for perm in obj.get_all_permissions():
+            _, action_model = perm.split('.', 1)
+            action, model = action_model.split('_', 1)
```
marcofucci: I don't think this is a reliable way of parsing permissions.
E.g. with `company.add_company`, `company.` is guaranteed to be the django app name but `add_company` can really be anything in case of custom permissions.
I would therefore change that and get the content_type for that permission instead and return something like:
```
'permissions': {
    'company': {
        'add_company': True,
        'change_company': False
    },
}}
```
(note the full `add_company` perm name)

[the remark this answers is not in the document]
reupen: It might just be simpler to drop the model grouping – what is it adding/helping with?

datahub/user/test/test_views.py
jim68000: What would that add time-wise to this? I'd like to get rid of all the V1 endpoints, and this feels like an obvious one to solve

elcct: @jim68000 serializer for this endpoint has changed (to WhoAmISerializer), but maybe we could do v1 cleanup in another PR?

datahub/user/serializers.py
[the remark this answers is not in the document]
elcct: I agree that grouping doesn't feel like it is bringing value to the table, so I removed it.

Permissions now are just list of strings in `app_label.permission` format.

datahub/user/test/test_views.py
elcct: I have removed groups.

datahub/user/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+            'contact_email',
+            'telephone_number',
+            'dit_team',
+            'permissions',
+        )
+        depth = 2
+
+    def get_permissions(self, obj):
+        """Serialize permissions."""
+        return obj.get_all_permissions()
```
reupen: Don't we want to keep the 
```
{
  "permission_1": true,
  "permission_2'" false
}
```
format?

elcct: But wouldn't this be redundant? I mean if permission doesn't exist in the list, then user doesn't have it.

jim68000: That is the intention of the design: it means we don't have to synchronise all the permissions between front end and back end every time

reupen: > But wouldn't this be redundant? I mean if permission doesn't exist in the list, then user doesn't have it.

The format was easier for the front end to use (because it will get deserialised as an object rather than as an array). So, you might want to check with the FEDs again.

elcct: I think we decided to stay with an array of user's permission.

reupen:
@elcct Can the 'work in progress' label be removed from this now?

datahub/user/serializers.py
marcofucci: I thought we decided to keep the models as key and permissions as array of permissions, no?
Something like:
```
{
    'company: ['add_company', 'change_company'],
    'company: ['add_company'],
}
```

elcct: @marcofucci as @reupen pointed out

> It might just be simpler to drop the model grouping – what is it adding/helping with?

I agree that grouping doesn't really bring any value and endpoint is returning just an array of permissions. When I asked for clarification on that today, grouping was not present and chosen option 1) is an array:

> 1) Array of permission user has available e.g. ['permission_1', 'permission_3']

marcofucci: ah OK, I missed that.

datahub/user/test/test_views.py
```diff
@@ -1,19 +1,81 @@
...
         response = self.api_client.get(url)
-        user_test = get_test_user()
 
         assert response.status_code == status.HTTP_200_OK
-        assert response.data['name'] == user_test.name
-        assert response.data['first_name'] == user_test.first_name
-        assert response.data['id'] == str(user_test.pk)
+
+        response_data = response.json()
+        if 'permissions' in response_data:
```
marcofucci: you can probably get rid of this if, right?

elcct: I was thinking what if somehow response will have `permissions` missing, then test would fail ugly way.

PR #612, opened 2017-11-07
Add terms and conditions to OMIS quotes

marcofucci:
This adds a new model TermsAndConditions with the markdown content of the Ts and Cs to be used in a quote.
They are then linked from a quote and returned via the quote endpoints.
This is because two quotes have the same Ts and Cs.
However, as Ts and Cs do get updated from time to time, the logic uses the latest record when creating a quote.

This commit also includes the initial fixture for loading the first Ts and Cs content.


datahub/omis/quote/serializers.py
```diff
@@ -47,13 +55,23 @@ class Meta:
...
             'content',
+            'terms_and_conditions',
         ]
         read_only_fields = fields
 
 
 class PublicQuoteSerializer(serializers.ModelSerializer):
     """Public Quote DRF serializer."""
 
+    terms_and_conditions = serializers.SerializerMethodField()
```
reupen: I think `terms_and_conditions = serializers.CharField(source='terms_and_conditions.content', default=None)` should work as well.

marcofucci: Yeah that was my first version but if `terms_and_conditions` is NULL, it doesn't get added to the response body whilst I want ''
Do you know how I could do that?

reupen: Are you sure? I tried
```
terms_and_conditions = serializers.CharField(source='terms_and_conditions.content',
                                                 default='')
```
and test_get_without_ts_and_cs passed. There was a related in change in DRF 3.7.2 though.

marcofucci: oh yes it works, I might have missed the `default=''` I suppose 🙄 
Thanks!

PR #616, opened 2017-11-08
[DH-1082] Add export experience category to companies

reupen:


marcofucci:
`export_wins_category` doesn't seem like a good name. 
Is it somewhat related to the readiness of the company to export?
Should we use something more descriptive?

reupen:
I don't like the name either but it seems to be the terminology used and correlates with the guidance – unless @MillsWest has a suggestion.

Still looks like this requirement hasn't been properly thought through so this PR may change or go away.

reupen:
Have looked further and 'export experience category' is also used so that looks like a better choice here.

PR #617, opened 2017-11-08
OMIS notification adviser added

marcofucci:
This sends a notification to OMIS advisers when they are added to an order.

### DIFF
--- a/datahub/omis/notification/signal_receivers.py
+++ b/datahub/omis/notification/signal_receivers.py
@@ -2,13 +2,16 @@ from django.db.models.signals import post_save
 from django.dispatch import receiver
 
 from datahub.omis.notification.client import notify
-from datahub.omis.order.models import Order
+from datahub.omis.order.models import Order, OrderAssignee, OrderSubscriber
 from datahub.omis.order.signals import quote_generated
 
 
 @receiver(post_save, sender=Order, dispatch_uid='notify_post_save_order')
-def notify_post_save_order(sender, instance, created, **kwargs):
+def notify_post_save_order(sender, instance, created, raw=False, **kwargs):
     """Notify people that a new order has been created."""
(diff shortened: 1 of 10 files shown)


datahub/omis/notification/test/test_templates.py
```diff
@@ -68,3 +70,21 @@ def test_quote_awaiting_acceptance_for_contact(self, settings):
         order = OrderWithOpenQuoteFactory()
 
         notify.quote_generated(order)
+
+    def test_you_have_been_added_for_adviser(self, settings):
+        """
+        Test the notification for when an adviser is added to an order.
+        If the template variables have been changed in GOV.UK notifications this
+        is going to raise an exception.
```
elcct: Do we know what type of exception?

datahub/omis/notification/signal_receivers.py
```diff
@@ -17,3 +17,16 @@ def notify_post_save_order(sender, instance, created, **kwargs):
 def notify_post_quote_generated(sender, order, **kwargs):
     """Notify contact that a quote has been generated."""
     notify.quote_generated(order)
+
+
+@receiver(post_save, sender=OrderAssignee, dispatch_uid='notify_post_save_assignee')
+@receiver(post_save, sender=OrderSubscriber, dispatch_uid='notify_post_save_subscriber')
+def notify_post_save_order_adviser(sender, instance, created, **kwargs):
+    """Notify adviser that they have been added to the order"""
+    if created:
```
reupen: Probably want to skip when raw=True?

marcofucci: 🤔  never thought about raw=True. Is it only True when loading fixtures?

reupen: Yes (it comes from `Model.save_base()`).

PR #628, opened 2017-11-10
Remove CDMS authentication

reupen:
Also gets rid of the second test run as there are no live server tests left.


elcct:
Finally this is gone :)

PR #630, opened 2017-11-10
Update team disabled-on dates and names

reupen:
The first commit reformats the file to make diffs easier to read.

The second commit has the actual changes. So, look at the second commit to see what the intended value changes are.


marcofucci:
I found the previous version of `permissions` easier to read as each item is not really a list logically

reupen:
> I found the previous version of permissions easier to read as each item is not really a list logically

Yep, I know. It was just how `ruamel.yaml` formatted it. We could change the permissions section back I guess.

PR #633, opened 2017-11-13
Add investment project permissions for LEPs

PeterNociar:
Restricting collections and detail page API's for investment projects


datahub/investment/permissions.py
```diff
@@ -0,0 +1,31 @@
...
+    InvestmentProject through the user's team.
+    """
+
+    base_permission_template = 'investment.{method}_associated_investmentproject'
+
+    def is_associated(self, request, view, obj):
+        """Check for connection."""
+        if not isinstance(obj, InvestmentProject):
+            raise ImproperlyConfigured(f'{self} needs to be used on InvestmentProject view')
+        return obj.team_members.filter(adviser__dit_team=request.user.dit_team).count() > 0
```
reupen: How about using QuerySet.exists() instead of QuerySet.count()?

datahub/core/permissions.py
```diff
@@ -22,3 +22,45 @@ def has_permission(self, request, view):
...
+        """Change ViewSet action to permissions method."""
+        return self._action2method.get(action, action)
+
+    def is_associated(self, request, view, obj):
+        """Check for connection."""
+        return True
+
+    def check_condition(self, request, view):
+        """Check if condition should be applied."""
+        return
```
reupen: Would go for `return False` here. It looks like something is missing (because you could take out the return and it wouldn't change anything).

datahub/core/permissions.py
```diff
@@ -22,3 +22,45 @@ def has_permission(self, request, view):
...
+    _action2method = {
+        'list': 'read',
+        'detail': 'read',
+        'retrieve': 'read',
+        'destroy': 'delete',
+        'destroy_all': 'delete',
+        'partial_update': 'change',
+    }
+
+    def action_2_method(self, action):
```
reupen: `action_to_method`? It's just one extra character :wink:

datahub/investment/views.py
```diff
@@ -117,7 +123,8 @@ def get_paginated_response(self, data):
 class IProjectModifiedSinceViewSet(IProjectViewSet):
     """View set for the modified-since endpoint (intended for use by Data Hub MI)."""
 
-    permission_classes = (IsAuthenticatedOrTokenHasScope,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,
+                          IsTeamAssociatedToInvestmentProjectPermission)
```
reupen: This view is a machine-to-machine one so won't have a user associated with requests (so I don't think we want IsTeamAssociatedToInvestmentProjectPermission here).

datahub/investment/permissions.py
reupen: Since the filtering is the same as in IsTeamAssociatedToInvestmentProjectFilter, would be nice to share that in a common function or similar.

Also we will need to show them projects that their team has created as well, as they are allowed to create projects (so need to look at created_by.dit_team as well).

PeterNociar: 1) I can change it to exists() if that is preferable 
2) UserTeamInvestmentProjectAssociationCheck is shared for both permission and for the filter class already 
3) Oh I didn't know that. I thought it was just for the team members. Ok I can add the creator of the iproject

datahub/core/permissions.py
PeterNociar: Yeah this is basically a base class.

datahub/core/permissions.py
PeterNociar: Sure :)

datahub/investment/views.py
PeterNociar: Ahh, ok. How can I know what is used only for machine-to-machine

reupen:
@marcofucci @elcct There's still a bit more to do, but I think there's a lot in this PR already.

These are still to be done:
* Tidy up how the permission tests create test users (it's a bit hacky)
* Add permissions for the team members endpoints
* Apply associated permissions to the search view (i.e. finish off https://github.com/uktrade/data-hub-leeloo/pull/652)

Will tackle those in separate PRs (so you can start looking at this one).

datahub/core/permissions.py
```diff
@@ -22,3 +37,36 @@ def has_permission(self, request, view):
         if not hasattr(view, 'permission_required'):
             raise ImproperlyConfigured()
         return request.user and request.user.has_perm(view.permission_required)
+
+
+class ObjectAssociationCheckerBase(ABC):
+    """Base association check class."""
```
marcofucci: Nothing wrong with the logic but I think the docs could be improved as it's not clear:
- how this can be used
- what the methods are and what they should return

I'm also not sure `is associated` is a good name for this as it's investment related whilst this logic seems quite abstract and applicable in other situations as well.

datahub/investment/permissions.py
```diff
@@ -0,0 +1,147 @@
...
+        view_should_be_filtered = view.action in self.actions_to_filter
+        restrictions_are_active = self.checker.should_apply_restrictions(
+            request=request, view=view
+        )
+
+        if view_should_be_filtered and restrictions_are_active:
+            query = Q(team_members__adviser__dit_team=request.user.dit_team)
+            for field in queryset.model.ASSOCIATED_ADVISER_FIELDS:
+                query |= Q(**{f'{field}__dit_team': request.user.dit_team})
+            return queryset.filter(query)
```
marcofucci: I know it might be tricky but I think splitting the logic into `ASSOCIATED_ADVISER_FIELDS` and `team_members` + having a similar logic in two places (here and in `investment.models.InvestmentProject.get_associated_advisers`) looks a bit risky.
Someone might add a new field (maybe m2m) and update the logic in one place and not the other for example).

I would try and centralise the logic somehow (maybe extend `ASSOCIATED_ADVISER_FIELDS` so that it might include `team_members.all` somehow?).

I'll leave it to you if you think it's too much trouble.

reupen:
@marcofucci I replaced CustomPermissionAdminMixin with a decorator. Avoids the messy inheritance.

marcofucci:
yeah, I noticed 👍

datahub/core/admin.py
```diff
@@ -97,41 +97,28 @@ def get_readonly_fields(self, request, obj=None):
         return readonly_fields
 
 
-class CustomPermissionAdminMixin:
+def custom_change_permission(permission_codename):
     """
-    Mixin for ModelAdmin that allows add, change and delete permissions with non-standard names
-    to be used.
+    Decorator that allows a custom change permission to be used with ModelAdmin subsclasses.
```
elcct: `subsclasses` - typo?

datahub/core/admin.py
```diff
@@ -97,41 +97,28 @@ def get_readonly_fields(self, request, obj=None):
...
 
-    To use, inherit from this class before ModelAdmin and override the add_permission,
-    change_permission and/or delete_permission class attributes as needed.
+    Usage example::
```
elcct: ::

reupen: It's reST/for Sphinx. Though we don't use it...

PR #641, opened 2017-11-14
 Add simple VCS tag-based versioning using setuptools_scm

reupen:
This adds versioning based on git tags. 

Since we don't have a setup.py, this only covers the case when the service is run from a git repo (which is how it is deployed anyway).

If a tag such as `v0.6.0` is made and that is checked out, 0.6.0 will be returned as the version number.

If further commits are made (without a new tag being created), something like `0.6.1.dev1+g3418f90f` will be returned as the version. That tells you how many commits there have been since the last release and what the current commit hash is.

If there are also uncommitted changes, you'll get something like `0.6.1.dev1+g3418f90f.d20171114`.

Initial versions will look like `0.1.dev2620+g40aedae` (until the first tag is made).

A public endpoint at `/v3/version` was added that returns the current version.

The ping subpackage was also renamed to status (it wasn't in INSTALLED_APPS before – haven't changed that).


reupen:
Closing for now as this seems a bit complicated due to how deployments are triggered at the moment.

PR #651, opened 2017-11-15
"and" operator

elcct:
This changes queries to use "and" operator, that means fewer results leaving only most relevant ones.

It uses `MultiMatch` query to achieve cross field matching with and operator.
~It introduces new "virtual" field `global_search` that enables "cross field" matching - like "company name country name". I was unable to use `MultiMatch` because it is difficult to deal with nested fields and effect is the same.~
~The old behaviour that is `SEARCH_FIELDS` is still partially available and used in `OMIS` to allow cost fields to match properly.~

Trigram filters in entity search also changed to use proper "and" operator. 
Previously query "abcd efg" would have matched "abcd efghi jklm", but query "abcd jkl" wouldn't. Now it will. 
I have checked and that change of behaviour is acceptable for OMIS.

I had to remove "The Advisories" from search results quality test, because it no longer matches. I tried to use fuzzy search, but it was giving too noisy results. I think current partial matching should be good MVP and improvements on using fuzzy matching etc. need more research.


reupen:
Comment as per Slack – would prefer to avoid the `copy_to`s as it will require a resync to change the global search fields. Though, if it's the only way to achieve the desired results, then that's fine.

elcct:
I have managed to use `MultiMatch` query instead, so the ugly `copy_to` and `global_search` field are gone.

datahub/search/omis/models.py
```diff
@@ -98,9 +98,13 @@ class Order(DocType, MapDBModelToDict):
     )
 
     SEARCH_FIELDS = (
+        'reference',
```
marcofucci: why do we need this if we have `reference_trigram`?

elcct: there is an issue with codes containing same characters like ABC-000000123. It could match ABC-0000000123 etc. which not necessarily gets exact match on the top. I was hoping that adding exact match as well, would boost that document up in the results, but that is not the case. I am looking now at specifically adding boost to those fields.

elcct: @marcofucci it seems to have no effect, so I removed those fields.

PR #654, opened 2017-11-16
[DH-1161] Always make company name and registered address editable

reupen:
Removes _CHPreferredField and its usage as this was causing various bugs. Instead the name and registered address of a company are normal editable fields.


marcofucci:
This is very nice but I'm not sure about the consequences.

At the moment the logic to get the registered address of a company is:
`company.companies_house_data.registered_address or company.registered_address`
and you are not allowed to change the registered address of the company.

If we allow people to change the registered address when there's a companies_house record linked, it's not obvious which registered_address is the most up to date and why.

elcct:
Maybe we could add a way to update company address in sync_ch

reupen:
@marcofucci Sorry, I meant to edit the OMIS stuff as well. company.registered_address is in fact labelled 'Primary address' in the front end. I think the root problem here is poor modelling of addresses.

Regardless, this doesn't change what the registered address Companies House has. Just what is labelled 'primary address' in the front end is editable (the fields were already being displayed in the FE, it just looked like changes weren't being saved).

So, I would probably pick company.registered_address if you have to pick one. Otherwise, let the user select.

marcofucci:
Yes but why would you change the registered address of a company if there’s already an official record from Companies House linked?
I’m not sure it makes sense.

marcofucci:
Reopening this as a cheap attempt to fix other major problems hopefully without introducing bigger ones.

PR #655, opened 2017-11-16
Escape markdown characters when genereting a quote

marcofucci:
To avoid all possible XSS or phishing attacks, the user generated content is made markdown safe when generating a quote.

### DIFF
--- a/datahub/omis/quote/utils.py
+++ b/datahub/omis/quote/utils.py
@@ -1,3 +1,5 @@
+import html
+import re
 from datetime import timedelta
 from pathlib import PurePath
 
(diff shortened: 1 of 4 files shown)


datahub/omis/quote/utils.py
```diff
@@ -16,6 +18,23 @@
 QUOTE_TEMPLATE = PurePath(__file__).parent / 'templates/content.md'
 
 
+def escape_markdown(content):
```
elcct: I would change the name to `escape_markdown_and_html`.

marcofucci:
I agree and I tested with the current frontend.

datahub/omis/quote/utils.py
marcofucci: You can use html in markdown so I think escape markdown implies html as well, no?

elcct: I see, that's true indeed. 
https://daringfireball.net/projects/markdown/syntax#html
Thanks

PR #660, opened 2017-11-16
Add ability to ES sync specify models

marcofucci:
This adds a new argument to the `sync_es` command.

When `--model=<name>` is used, only the model <name> is synced.
You can sync multiple models at once with `--model=<model1> --model=<model2>` etc.

If no --model is specified, all models are synced as usual.


datahub/search/management/commands/sync_es.py
```diff
@@ -74,10 +86,20 @@ def add_arguments(self, parser):
             default=600,
             help='Batch size - number of rows processed at a time',
         )
+        parser.add_argument(
+            '--model',
+            dest='model',
```
reupen: Can we set the choices here?

marcofucci: 👍  done

marcofucci:
@reupen very true, done both!

PR #662, opened 2017-11-17
Notify advisers when they get removed from order

marcofucci:
This sends a notification when an adviser is removed from an order.

### DIFF
--- a/datahub/omis/notification/signal_receivers.py
+++ b/datahub/omis/notification/signal_receivers.py
@@ -1,4 +1,6 @@
-from django.db.models.signals import post_save
+from functools import partial
+from django.db import transaction
+from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
 from datahub.omis.notification.client import notify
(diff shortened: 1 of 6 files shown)


datahub/omis/notification/signal_receivers.py
```diff
@@ -35,6 +35,18 @@ def notify_post_save_order_adviser(sender, instance, created, raw=False, **kwarg
...
+@receiver(post_delete, sender=OrderAssignee, dispatch_uid='notify_post_delete_assignee')
+@receiver(post_delete, sender=OrderSubscriber, dispatch_uid='notify_post_delete_subscriber')
+def notify_post_delete_order_adviser(sender, instance, **kwargs):
+    """
+    Notify people that they have been removed from the order.
+
+    Note that `instance` is no longer in the database at this point,
+    so be very careful what you do with it.
+    """
+    notify.adviser_removed(order=instance.order, adviser=instance.adviser)
```
reupen: The change hasn't been committed at this point though (there could still be an error on commit). So would use django.db.transaction.on_commit() like the Elasticsearch sync.

PR #664, opened 2017-11-17
Add filter contact by company id and archived, add sorting contact by archived_on.

elcct:
This adds what's in the title to contact search.

### DIFF
--- a/datahub/search/contact/serializers.py
+++ b/datahub/search/contact/serializers.py
@@ -10,6 +10,8 @@ from ..serializers import (
 class SearchContactSerializer(SearchSerializer):
     """Serialiser used to validate contact search POST bodies."""
 
+    name = serializers.CharField(required=False)
+    company = SingleOrListField(child=StringUUIDField(), required=False)
     company_name = serializers.CharField(required=False)
     company_sector = SingleOrListField(child=StringUUIDField(), required=False)
     company_uk_region = SingleOrListField(child=StringUUIDField(), required=False)
@@ -22,6 +24,7 @@ class SearchContactSerializer(SearchSerializer):
         'address_town',
         'adviser.name',
         'archived',
+        'archived_on',
         'archived_by.name',
         'company.name',
         'contactable_by_dit',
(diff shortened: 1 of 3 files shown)


datahub/search/contact/test/test_views.py
```diff
@@ -155,6 +155,24 @@ def test_search_contact_by_partial_name(self, setup_es):
         assert len(response.data['results']) == 1
         assert response.data['results'][0]['first_name'] == contact.first_name
 
+    def test_search_contact_by_company_id(self, setup_es):
+        """Tests filtering by company id."""
+        company = CompanyFactory()
+        ContactFactory(company=company)
```
reupen: Can we add another one with a different company, to check that it doesn't match?

datahub/search/contact/serializers.py
```diff
@@ -10,6 +10,8 @@
 class SearchContactSerializer(SearchSerializer):
     """Serialiser used to validate contact search POST bodies."""
 
+    name = serializers.CharField(required=False)
```
reupen: Was this not working before? Can we improve the test so that it doesn't regress?

elcct: It worked because without filters it returns all records. Test is indeed too weak and needs to have extra records here. Thanks!

datahub/search/contact/test/test_views.py
elcct: Done.

datahub/search/contact/serializers.py
```diff
@@ -10,6 +10,9 @@
 class SearchContactSerializer(SearchSerializer):
     """Serialiser used to validate contact search POST bodies."""
 
+    archived = serializers.BooleanField(required=False)
```
reupen: Is it meant to be archived yes/no or include archived yes/no? Have approved but would be good to check this.

reupen: Never mind actually, I guess omitting archived means all of them.

elcct: It appears that true/false is fine.

PR #674, opened 2017-11-24
Add archived documents URL path to test fixtures

feedmypixel:
Fixtures for documents AT tests


feedmypixel:
FYI Will get this rebased and CircleCi passing this weekend

feedmypixel:
this appears to be broken, will fix and re raise a PR

PR #675, opened 2017-11-25
Update metadata fixtures with correct values

marcofucci:
I've already made this change manually in production so this PR is just to align the metadata fixture to the current data. No need to rush releasing or merging this.

**What this is**
6 referralsourceactivity records where in the wrong place.
2 of them should have been in referralsourcemarketing and the other 4 in referralsourcewebsite.

This only adds those 6 records to the right metadata tables but keeps the same records in referralsourceactivity just in case they are used somewhere else.

**Post release**
Cleanup the old referralsourceactivity records making sure they aren't any records using them.

### DIFF
--- a/fixtures/metadata/referrals.yaml
+++ b/fixtures/metadata/referrals.yaml
@@ -29,19 +29,33 @@
 - model: metadata.referralsourceactivity
   pk: 812b2f62-fe62-4cc8-b69c-58f3e2ebac17
   fields: {disabled_on: null, name: "Website"}
-# The following values (up to EY EIM) were imported from CDMS for the data migration
+# The following values were needed for the data migration but they got moved to
+# other metadata tables at last minute.
+# TODO: check that no records link to these referralsourceactivity records and
+# if so, remove them from referralsourceactivity and only keep them in the
+# other tables.
 - model: metadata.referralsourceactivity
   pk: e0d48e78-b7d6-e311-8a2b-e4115bead28a
   fields: {disabled_on: null, name: LEP}
 - model: metadata.referralsourceactivity
-  pk: 500ae611-9c12-e611-9bdc-e4115bead28a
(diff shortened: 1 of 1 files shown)


fixtures/metadata/referrals.yaml
```diff
@@ -171,6 +176,19 @@
 - model: metadata.referralsourcewebsite
   pk: 072241d4-22d0-4657-8d8e-1848a0519b0a
   fields: {name: "Other website"}
+# The following values were imported from CDMS for the data migration
+- model: metadata.referralsourcewebsite
+  pk: 71bb9fc6-5f95-e211-a939-e4115bead28a
+  fields: {name: "Devolved Administration (DA)"}
```
reupen: Are you sure these are websites?

fixtures/metadata/referrals.yaml
```diff
@@ -195,3 +213,10 @@
 - model: metadata.referralsourcemarketing
   pk: c28d95c6-f094-4f97-b372-38815829a0f0
   fields: {name: "Television/radio"}
+# The following values were imported from CDMS for the data migration
+- model: metadata.referralsourcemarketing
+  pk: e0d48e78-b7d6-e311-8a2b-e4115bead28a
+  fields: {name: "LEP"}
```
reupen: What makes these marketing activities?

reupen:
CDMS doesn't have the hierarchy for referral source that we do (just a flat list). Are you sure the website/marketing categorisation for these ones is correct?

marcofucci:
@reupen No I'm not sure but Colin was quite sure that these values had to be here instead.
I suggest we merge this to have it aligned to the current prod db and then we fix it via code.
Without this the migration was failing.

marcofucci:
Closing this as it doesn't seem to be the right thing and also, records can be changed via django admin so not worth keeping in sync.

PR #676, opened 2017-11-27
This adds created_by field to contacts.

elcct:
It also adds created_by.id filter to enable to query data for dashboard.


datahub/search/contact/serializers.py
```diff
@@ -35,6 +36,7 @@ class SearchContactSerializer(SearchSerializer):
         'contactable_by_email',
         'contactable_by_phone',
         'created_on',
+        'created_by.name',
```
marcofucci: Do we need this?

elcct: We don't. Removed.

PR #678, opened 2017-11-27
Dashboard gets data via ES.

elcct:
As in the title.

On my laptop modified dashboard endpoint returns data @ ~65ms vs. ~2s when using the Database.

I checked it with latest Frontend and it appears to be working.


datahub/dashboard/serializers.py
```diff
@@ -1,31 +0,0 @@
...
-        model = Interaction
-        depth = 2
-        fields = '__all__'
-
-
-class IntelligentHomepageSerializer(serializers.Serializer):
-    """Intelligent homepage serializer."""
-
-    interactions = InteractionSerializer(many=True)
-    contacts = NestedContactSerializer(many=True)
```
reupen: Was the only usage of NestedContactSerializer, so can remove it now from datahub/company/serializers.py.

elcct: Done, thanks!

datahub/search/dashboard/views.py
```diff
@@ -0,0 +1,59 @@
...
+        serializer.is_valid(raise_exception=True)
+        limit = serializer.validated_data['limit']
+
+        interactions_query = get_search_by_entity_query(
+            term='',
+            entity=Interaction,
+            filters={
+                'dit_adviser.id': user.id,
+            },
+            field_order='created_on:desc',
```
reupen: Before there was logic to exclude records with a null `created_on` date. I can see this puts nulls last, but would suggest keeping the logic to exclude them (unless there's a reason for keeping them).

datahub/search/dashboard/views.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+        contacts = limit_search_query(
+            contacts_query,
+            offset=0,
+            limit=limit,
+        ).execute()
+
+        response = {
+            'interactions': [interaction.to_dict() for interaction in interactions.hits],
+            'contacts': [contact.to_dict() for contact in contacts],
```
reupen: Any reason for using .hits for interactions but not contacts?

datahub/search/dashboard/serializers.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework import serializers
+
+
+class LimitParamSerializer(serializers.Serializer):
```
reupen: Would suggest just calling this HomepageSerializer as it looks pretty specific to the home page (with default=5).

datahub/search/dashboard/views.py
elcct: Good catch :) Simply forgot about it. Fixed.

datahub/search/dashboard/views.py
elcct: No reason, somehow lost it. Fixed, thanks :)

datahub/search/dashboard/serializers.py
elcct: Good idea. Renamed.

PR #681, opened 2017-11-28
Improve name search.

elcct:
Added corresponding base field to most of `_trigram` fields in queries, so that if we have company name like `A & B` then it can be searchable.


reupen:
Looks like a bit of a pain, is this a temporary fix? I mean you could easily miss out one or the other of the fields and it would be hard to notice. Is it just a sign that trigram isn't doing what we need?

Also do we need trigram for project_code?

marcofucci:
Yeah I agree with @reupen, it looks like there should be another way as this seems a pain.

elcct:
@reupen @marcofucci I agree that this is painful and indeed there should be a better way to define which field are searched what way, therefore this is only temporary fix that addresses companies having names shorter than 3 characters not being searchable. I think additional query only complements trigram where trigram wouldn't match. 

I will remove project_code.

PR #683, opened 2017-11-29
[DH-1288] Fix bug with CH view and non-numeric company numbers

reupen:
The route was only allowing numbers, when company numbers can have letters in them (in our case particularly relevant for Northern Ireland and Scottish companies).

### DIFF
--- a/datahub/company/urls.py
+++ b/datahub/company/urls.py
@@ -88,6 +88,6 @@ company_urls = [
 ch_company_urls = [
     url(r'^ch-company$', ch_company_list,
         name='collection'),
-    url(r'^ch-company/(?P<company_number>[0-9]+)$', ch_company_item,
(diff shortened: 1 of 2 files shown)


datahub/company/urls.py
```diff
@@ -88,6 +88,6 @@
 ch_company_urls = [
     url(r'^ch-company$', ch_company_list,
         name='collection'),
-    url(r'^ch-company/(?P<company_number>[0-9]+)$', ch_company_item,
+    url(r'^ch-company/(?P<company_number>[\w]+)$', ch_company_item,
```
elcct: \w also contains `_`, should we allow that?

reupen: I don't see why not. TBH I don't see why it can't just accept any string – will be a lot simpler with Django 2.0 which moves away from using regular expressions for URL routes https://docs.djangoproject.com/en/dev/releases/2.0/#simplified-url-routing-syntax

PR #686, opened 2017-11-30
Fix bug when adding new metadata items via admin

reupen:
Was due to missing default value for id in BaseConstantModel.

Also makes the metadata admin forms more consistent, and adds export experience categories to admin.


marcofucci:
I think we have a huge problem with metadata.

We are using metadata in many ways but in particular:
- teams, team roles etc.: these are OK to be changed via django admin
- constants (e.g. business types etc.): which should not be allowed to be changed via the admin

At the moment we allow everyone with admin access to change anything which is very dangerous and we need to be careful as this might come back as a boomerang and cause much pain.

I don't think it's a good idea to have all metadata in the admin but only the ones we think it's safe to be changed.

reupen:
Isn't that a permissions problem though?

marcofucci:
No I don't think it's a permission problem, constants should not be in the admin area, they should only be in the code.
If you allow constants to be changed via the admin there will always be a misalignment between the fixtures and the actual database.

If you think it's a permission thing, we should change all the fixtures to be initial_fixtures instead of metadata and we need to solve the problem of our environments being completely different from production.

elcct:
But wouldn't that be a different PR?

marcofucci:
yes but why we don't remove things from the admin in the first place instead of fixing it? It seems a bit pointless, no?

elcct:
I think the requirement was to enable admin user to disable/enable referential data in the admin, if we remove that from the admin then we will remove that ability.

marcofucci:
Yeah that's exactly what I say, we allow some and disallow other :)

elcct:
We needed to enable that feature for all referential data in the admin.

reupen:
The specific problem encountered was that you can't add a team via admin. And it just happened to affect all the metadata models.

And even if you shouldn't be able to add, delete or edit some of them, I don't see why it follows that you should be able to see a list of them. And fixtures isn't a good way of maintaining data, those probably should've been loaded via a migration.

Anyway changing that would be a bigger piece of work, this is a simple fix for not being able to add a team.

reupen:
@marcofucci @elcct Updated this to make the read-only code for CH companies more reusable, and applied it to business types, investment project stages and investment types.

marcofucci:
💯

PR #696, opened 2017-12-06
Add `until` filter to Investment/from endpoint

r4vi:
This allows narrowing the results returned by the investments/from
endpoint.

The reason for this is that datahub-mi needs to do an initial load of
investment projects from the start of the financial year. Right now
trying to request this from the endpoint hits the 30 second PaaS request
timeout. This will allow datahub-mi to pull a day/week/month at a time
until they up to date.

### DIFF
--- a/datahub/investment/test/test_views.py
+++ b/datahub/investment/test/test_views.py
@@ -1044,6 +1044,14 @@ class TestPartialUpdateView(APITestMixin):
 class TestModifiedSinceView(APITestMixin):
     """Tests for the modified-since view."""
 
+    def _make_request(self, data=None):
+        url = reverse('api-v3:investment:investment-modified-since-collection')
+        client = self.create_api_client(
+            scope=Scope.mi,
+            grant_type=Application.GRANT_CLIENT_CREDENTIALS
+        )
+        return client.get(url, data=data)
+
     @pytest.mark.parametrize(
         'timestamp,num_results', (
             (datetime(2017, 12, 31), 5),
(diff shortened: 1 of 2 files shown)


r4vi:
@reupen do some swagger docs need to be updated too?

reupen:
Would you be able to add a test for that query param (see TestModifiedSinceView)?

API docs are here if you want to update those: https://github.com/uktrade/data-hub-api-spec/blob/master/generator/specs/paths/investment.yml#L70-L75

datahub/investment/test/test_views.py
```diff
@@ -1044,6 +1044,14 @@ def test_update_read_only_fields(self):
 class TestModifiedSinceView(APITestMixin):
     """Tests for the modified-since view."""
 
+    def make_request(self, data=None):
```
reupen: @r4vi CircleCI failed due to a Flake8 error. I'd just make this method internal (prefix with `_`) because it doesn't need to be public (and should get rid of the error).

r4vi:
@reupen thanks for the comments. Have fixed up. Just hoping the modified_on date on all the investments isn't the import date - in which case this wouldn't help :scream:

reupen:
> Just hoping the modified_on date on all the investments isn't the import date - in which case this wouldn't help
It's the date the record was last updated. Although, I'm not sure what it was set to for migrated records (can have a look to find out).

r4vi:
@reupen ok I'll hold on merging until you have a chance to look, no point adding code (even if trivial) if it doesn't help

PR #698, opened 2017-12-08
Add nested disabled_on filter to event search.

elcct:
This adds nested disabled_on filter that allows multiple or queries for that field.
In this case we need disabled_on_exists==False or disabled_on_after=date to filter events list.


datahub/search/elasticsearch.py
```diff
@@ -265,6 +265,11 @@ def get_filter_query(key, value):
     return must_filter
 
 
+def get_range_queries(ranges):
+    """Gets range queries."""
+    return [Q('range', **{k: v}) for k, v in ranges.items()]
+
+
```
marcofucci: The `get_search_by_entity_query` is getting very long and therefore a bit hard to read.
Maybe you can split it into different parts? E.g. a class with methods for the different blocks of logic.

datahub/search/event/serializers.py
```diff
@@ -8,10 +8,19 @@
 )
 
 
+class NestedDisabledFilterSerializer(serializers.Serializer):
+    """Serialiser used to validate disabled_on filter."""
+
+    exists = serializers.BooleanField(required=False)
+    after = RelaxedDateTimeField(required=False)
+    before = RelaxedDateTimeField(required=False)
```
reupen: Do we need the before filter?

Also, it feels a bit confusing when OR or AND applies. Is there a way we can design the API to make it clearer?

datahub/search/event/serializers.py
```diff
@@ -8,10 +8,19 @@
...
+
+
 class SearchEventSerializer(SearchSerializer):
     """Serialiser used to validate Event search POST bodies."""
 
     address_country = SingleOrListField(child=StringUUIDField(), required=False)
+    disabled_on = NestedDisabledFilterSerializer(required=False)
     disabled_on_exists = serializers.BooleanField(required=False)
```
reupen: Can we remove any of the old disabled_on filters?

datahub/search/elasticsearch.py
```diff
@@ -282,28 +287,38 @@ def get_search_by_entity_query(term=None,
...
-                # creates a "composite filter" - builds "or" query for given
-                # list of fields
-                should = list(
-                    chain.from_iterable(
-                        get_filter_query(ck, v) for ck in composite_filters[k]
-                    )
-                )
+                # process composite filters
+                should_filters = (get_filter_query(ck, v) for ck in composite_filters[k])
+            elif isinstance(filters[k], dict):
```
reupen: Can't this be v instead of filters[k]?

datahub/search/elasticsearch.py
reupen: I agree with splitting it up, also maybe some of the naming could be reviewed e.g. split_date_range_fields instead of date_range_fields.

elcct: Sure, good point that function has become too big.

datahub/search/event/serializers.py
elcct: We don't need before filter. I can remove it (added it for "completeness").

I was thinking about somehow say it with the filter name like `disabled_on_or_filter`, but doesn't seem right either. I'll think about it.

datahub/search/elasticsearch.py
elcct: Of course. Fixed.

datahub/search/elasticsearch.py
elcct: I did split that function into smaller ones.

datahub/search/event/serializers.py
elcct: I have removed unused filters and changed the name of serialiser class to point that the filters inside are "or".

datahub/search/event/serializers.py
elcct: Removed unused ones. `disabled_on_exists` seems to be used in the frontend.

datahub/search/elasticsearch.py
```diff
@@ -249,18 +249,82 @@ def apply_aggs_query(search, aggregates):
...
+            get_field_query(key, v) for v in value
+        ]
+        return Q('bool', should=should_filter, minimum_should_match=1)
+
+    return get_field_query(key, value)
+
+
+def get_filter_queries(filters):
+    """Gets filter queries."""
+    return [
```
reupen: What about flattening the list here?

datahub/search/elasticsearch.py
```diff
@@ -249,18 +249,82 @@ def apply_aggs_query(search, aggregates):
...
+        get_filter_query(k, v)
+        for k, v in filters.items()
+    ]
+
+
+def get_range_queries(ranges):
```
reupen: How about prefixing what functions are internal ones with _? It cuts down on the noise.

datahub/search/elasticsearch.py
```diff
@@ -249,18 +249,82 @@ def apply_aggs_query(search, aggregates):
...
+    if ranges:
+        should_filters.extend(get_range_queries(ranges))
+    return should_filters
+
+
+def get_must_filter_query(filters, composite_filters, ranges):
+    """Gets "and" filter query."""
+    must_filter = []
+
+    if filters:
```
reupen: Would be nice to get rid of one level of indentation by doing something like 
```
if filters is None:
    filters = {}
```

(and then getting rid of the `if filters:`)

datahub/search/elasticsearch.py
elcct: It is already flat, I changed that in 0581a1f commit.

reupen: Oh, cool :+1:

datahub/search/elasticsearch.py
elcct: Good idea, I have prefixed all internal functions with _.

datahub/search/elasticsearch.py
elcct: Also good idea. Changed, thanks :)

datahub/search/elasticsearch.py
```diff
@@ -327,7 +368,7 @@ def bulk(actions=None, chunk_size=None, **kwargs):
     return es_bulk(connections.get_connection(), actions=actions, chunk_size=chunk_size, **kwargs)
 
 
-def date_range_fields(fields):
+def _split_date_range_fields(fields):
```
reupen: This one's used in `datahub/search/views.py`. So either that module shouldn't be using it, or it shouldn't be prefixed with _.

elcct: Ah, nice find! I have removed it from the `views.py`.

PR #699, opened 2017-12-08
[DH-813] Refactor how permission-related tests create test users

reupen:
Stops such tests overwriting self._user (which was hacky and confusing).

### DIFF
--- a/datahub/core/test_utils.py
+++ b/datahub/core/test_utils.py
@@ -3,38 +3,50 @@ from secrets import token_hex
 
 import pytest
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
 from django.test.client import Client
 from django.utils.timezone import now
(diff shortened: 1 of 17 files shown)


datahub/core/test_utils.py
```diff
@@ -3,40 +3,52 @@
...
 
 
+def create_test_user(team=None, permissions=()):
+    """Return the test user."""
+    # Because AdviserFactory sets dit_team_id, passing dit_team to it doesn't work
+    dit_team_id = team.id if team else None
+    user = AdviserFactory(dit_team_id=dit_team_id)
+
+    for permission in permissions:
+        user.user_permissions.add(
```
elcct: Can't we add permissions in one go using `set` instead of looping through?

reupen: Yep, I've simplified that.

PR #702, opened 2017-12-11
dbmaintenance app and cmd to update OMIS regions

marcofucci:
This adds a new dbmaintenance app with a base class which can be subclassed to process a CSV stored in S3.

The command would be:
`./manage.py <comman-name> <bucket-name> <object-key>`

This also add an `update_omis_uk_regions` command which updates the `order.uk_region` field with the data stored in the CSV.


datahub/dbmaintenance/management/base.py
```diff
@@ -0,0 +1,87 @@
...
+
+    ./manage.py <command-name> <bucket> <object_key>
+    """
+
+    CSV_FIELDS = ()
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        parser.add_argument('bucket', help='S3 bucket where the CSV is stored.')
+        parser.add_argument('object_key', help='S3 key of the CSV.')
```
elcct: I would perhaps add `CSV file` or `CSV object.`.

datahub/dbmaintenance/management/base.py
```diff
@@ -0,0 +1,87 @@
...
+        response = s3_client.get_object(
+            Bucket=options['bucket'],
+            Key=options['object_key']
+        )['Body']
+
+        with closing(response):
+            csvfile = codecs.getreader('utf-8')(response)
+            reader = csv.DictReader(csvfile, fieldnames=self.CSV_FIELDS)
+
+            next(reader)  # ship header
```
elcct: Should be `skip` in the comment?

datahub/dbmaintenance/test/commands/test_update_omis_uk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+from datahub.core import constants
+from datahub.omis.order.test.factories import OrderFactory
+
+
+pytestmark = pytest.mark.django_db
+
+
+@mock.patch('datahub.dbmaintenance.management.base.get_s3_client')
+def test_run(mocked_get_s3_client):
+    """Test that the command updates the relevant records ignoring the ones in error."""
```
elcct: with error?

marcofucci: I think _in error_ is correct, asking @reupen to double-check maybe :)

elcct: I think "with" also will be incorrect. Maybe "the ones having an error". :)

datahub/dbmaintenance/management/base.py
```diff
@@ -0,0 +1,87 @@
...
+
+        s3_client = get_s3_client()
+        response = s3_client.get_object(
+            Bucket=options['bucket'],
+            Key=options['object_key']
+        )['Body']
+
+        with closing(response):
+            csvfile = codecs.getreader('utf-8')(response)
+            reader = csv.DictReader(csvfile, fieldnames=self.CSV_FIELDS)
```
reupen: Why don't we read the field names from the file/check that they're what we're expecting? Will just avoid any risk of columns being transposed.

datahub/dbmaintenance/management/base.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def process_row(self, record, **options):
+        """
+        Process one single row.
+
+        :returns: True if the record has been processed successfully, False otherwise.
+        """
+        try:
+            self._process_row(record, **options)
+        except Exception as e:
```
reupen: Seems very broad. I know we wanted to skip records where the old value didn't match the current value, but if something more fatal happens we might want to roll back?

datahub/dbmaintenance/test/commands/test_update_omis_uk_regions.py
```diff
@@ -0,0 +1,77 @@
...
+    ]
+
+    csv_content = f"""order_id,region_id
+00000000-0000-0000-0000-000000000000,{constants.UKRegion.guernsey.value.id}
+{orders[0].id},{constants.UKRegion.england.value.id}
+{orders[1].id},{constants.UKRegion.east_midlands.value.id}
+{orders[2].id},NULL
+{orders[4].id},00000000-0000-0000-0000-000000000000
+"""
+    mocked_get_s3_client().get_object.return_value = {
```
reupen: Could use the stubber in boto(core), there's an example here:
https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/documents/test/test_av_scan.py#L22-L47

datahub/dbmaintenance/test/commands/test_update_omis_uk_regions.py
reupen: I'm with @elcct here, I'd have taken 'in error' to mean 'by mistake' which doesn't make any sense here.

I'd probably have gone for something like 'ignoring ones with errors'.

datahub/dbmaintenance/management/base.py
marcofucci: I thought we decided that we should be able to run the same cmd several times and it would just continue from where it stopped, no?

elcct: I was also under such impression.

reupen: Oh yeah, that's true. I'm OK with it then.

datahub/dbmaintenance/management/base.py
marcofucci: I got rid of `CSV_FIELDS` and `Record` and use the built-in dict with keys as header values instead.

datahub/dbmaintenance/test/commands/test_update_omis_uk_regions.py
marcofucci: @reupen I'm now using the boto Stubber and it looks a lot nice, thanks for that. 
I've moved the `s3_stubber` fixture to the global `conftest` as it seems quite generic, are you OK with that?

reupen: Yep, cool :+1:

PR #705, opened 2017-12-13
Migration commands to fix investment projects.

elcct:
This adds 6 commands to fix various issues with investment projects.

To update descriptions, companies, referral sources, sectors and created dates.


datahub/dbmaintenance/management/commands/update_investment_project_comments.py
```diff
@@ -0,0 +1,30 @@
...
+            dest='simulate',
+            default=False,
+            help='If True it only simulates the command without saving the changes.',
+        )
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        # "Id" is not a typo
+        investment_project = InvestmentProject.objects.get(pk=row['Id'])
+        description = row['comments'].strip()
```
reupen: The field name 'comments' suggests there are comments from CDMS rather than descriptions.

If they're comments:

1. We discussed yesterday adding a new (possibly read-only) field to store comments.
2. Should it be appending to the description rather than replacing it?

datahub/dbmaintenance/management/commands/update_investment_project_company.py
```diff
@@ -0,0 +1,95 @@
...
+        Checks if Investment project should be updated.
+
+        :param investment_project: instance of InvestmentProject
+        :param investor_company: instance of Company or None
+        :param intermediate_company: instance of Company or None
+        :param uk_company: instance of Company or None
+        :param uk_company_decided: Boolean
+        :return: True if investment project needs to be updated
+        """
+        return (investment_project.investor_company != investor_company or
```
reupen: Weren't we also going to check that the old value is still what was previously loaded before doing updates?

datahub/dbmaintenance/management/commands/update_investment_project_company.py
```diff
@@ -0,0 +1,95 @@
...
+        translate = {
+            '0': False,
+            '1': True,
+        }
+        return translate[uk_company_decided.strip()]
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        investment_project = InvestmentProject.objects.get(pk=row['id'])
+        investor_company = self.get_company(row['investor_company_id'])
```
reupen: Are we able to just work with the IDs instead of loading the full company objects?

datahub/dbmaintenance/management/commands/update_investment_project_company.py
elcct: There was no old_value in the spreadsheet for that update. If it is going to be added then I can add that check as well.

jim68000: As there's no way for users to update the company this won't have been changed

datahub/dbmaintenance/management/commands/update_investment_project_comments.py
elcct: I see. That was a bit confusing. 

1. I could add new field, but what about schema freeze?
2. I thought this is to correct wrong values, so did replace.

datahub/dbmaintenance/management/commands/update_investment_project_company.py
reupen: uk_company can be updated by users. Not sure how many updates to investor company have been via support but I know at least some have.

datahub/dbmaintenance/management/commands/update_investment_project_company.py
marcofucci: Yeah I agree, there are lots of queries here, could be better to use IDs instead if possible. 
E.g. you could use `values_list` to get all the IDs you need or you could use `investment_project.investor_company_id` instead of `investment_project.investor_company` which lots the whole object if I'm not wrong.

marcofucci: Another optimisation you could do is: you could just use raw id strings to save `investment_project` company fields, e.g. `investment_project.investor_company_id = row['investor_company_id']`

elcct: So I updated it to only validate company ids and then just assign these if valid. It is querying all ids per line and just ids. I removed removed `lru_cache` as each row is quite unique, so no point having it.

datahub/dbmaintenance/management/commands/update_investment_project_comments.py
reupen: If these are comments then it might be best to remove this one from the PR and tackle it with the story in the next sprint.

elcct: Removed.

datahub/dbmaintenance/management/commands/update_investment_project_company.py
```diff
@@ -0,0 +1,112 @@
...
+        intermediate_company_id = self._parse_company_id(row['intermediate_company_id'])
+        uk_company_id = self._parse_company_id(row['uk_company_id'])
+        uk_company_decided = self.get_uk_company_decided(row['uk_company_decided'])
+
+        company_ids = [
+            company_id for company_id in
+            (investor_company_id, intermediate_company_id, uk_company_id,)
+            if company_id is not None
+        ]
+        if not self._can_update(company_ids):
```
marcofucci: you might not need this check, no? They are all foreign keys so if you set a value that doesn't exist, the update command should fail if I'm not missing anything, no?

datahub/dbmaintenance/management/commands/update_investment_project_company.py
```diff
@@ -0,0 +1,112 @@
...
+        :param investor_company: instance of Company or None
+        :param intermediate_company: instance of Company or None
+        :param uk_company: instance of Company or None
+        :param uk_company_decided: Boolean
+        :return: True if investment project needs to be updated
+        """
+        return (investment_project.investor_company != investor_company or
+                investment_project.intermediate_company != intermediate_company or
+                investment_project.uk_company != uk_company or
+                investment_project.uk_company_decided != uk_company_decided)
```
marcofucci: You have to use `_id` here otherwise django is making an extra query to load the company.
This should be `investment_project.uk_company_decided_id != uk_company_decided_id` etc.

datahub/dbmaintenance/management/commands/update_investment_project_company.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def _can_update(self, company_ids):
+        """Check if given list of company_ids exists.
+
+        :param: company_ids: list of company ids
+        :return: True if can update, otherwise False
+        """
+        return (
+            Company.objects.values_list('id').filter(id__in=company_ids).count() ==
```
marcofucci: I don't think `.values_list('id')` is needed if you use `count()`

datahub/dbmaintenance/management/commands/update_investment_project_company.py
elcct: That is true, we don't need that as these are foreign keys. I have removed it.

datahub/dbmaintenance/management/commands/update_investment_project_company.py
elcct: Oh, thanks for pointing that out. I thought I did that :( Fixed.

datahub/dbmaintenance/management/commands/update_investment_project_company.py
elcct: That is true. But method is now removed anyway. Thanks!

datahub/dbmaintenance/management/commands/update_investment_project_start_date.py
```diff
@@ -0,0 +1,30 @@
+from dateutil.parser import parse as dateutil_parse
+
+from datahub.investment.models import InvestmentProject
+from ..base import CSVBaseCommand
+
+
+class Command(CSVBaseCommand):
```
reupen: The command/module is called start date but the command updates created_on. Any particular reason the command is named as such? Also asking because start date is also a field in CDMS (and isn't the same thing as creation date).

reupen:
Just had a thought – we might want to create new reversion revisions when performing the save operations (otherwise it will mess up the audit history).

elcct:
@reupen I could add reversion to the commands, seems like a good idea.

datahub/dbmaintenance/management/commands/update_investment_project_referral_source_activity_marketing.py
```diff
@@ -0,0 +1,101 @@
...
+        :param referral_source_activity: instance of ReferralSourceActivity or None
+        :param referral_source_activity_marketing:
+                instance of ReferralSourceMarketing or None
+        :return: True if investment project needs to be updated
+        """
+        return (
+            investment_project.referral_source_activity !=
+            referral_source_activity or
+            investment_project.referral_source_activity_marketing !=
+            referral_source_activity_marketing
```
marcofucci: If you don't want to use raw ids everywhere you could just add a little optimization here with the check `investment_project.referral_source_activity_id != referral_source_activity.id` etc.
At least it doesn't make 2 extra queries per row.

datahub/dbmaintenance/management/commands/update_investment_project_referral_source_activity_website.py
```diff
@@ -0,0 +1,97 @@
...
+
+        :param investment_project: instance of InvestmentProject
+        :param referral_source_activity: instance of ReferralSourceActivity or None
+        :param referral_source_activity_website:
+                instance of ReferralSourceWebsite or None
+        :return: True if investment project needs to be updated
+        """
+        return (
+            investment_project.referral_source_activity != referral_source_activity or
+            investment_project.referral_source_activity_website != referral_source_activity_website
```
marcofucci: same here, might be worth optimising the check with `_id`

reupen:
> @reupen I could add reversion to the commands, seems like a good idea.

Yeah, I think so. It will be a little bit slower, but we aren't updating a huge number of rows here anyway.

datahub/dbmaintenance/management/commands/update_investment_project_referral_source_activity_marketing.py
elcct: Updated. Thanks.

elcct:
@reupen added reversion, so you can have a look.

PR #706, opened 2017-12-14
Fully enforce model permissions in search

reupen:
This makes various changes to get permissions fully working with search:

* Enforces read permissions in global search
* Enforces investment association permissions in global and investment project entity search
* Updates ES models so that created_by and various investment project fields include the adviser's team
* Adds signal receivers for investment project team member changes

The datahub.search.elasticsearch module really needs splitting up into multiple modules, hopefully we can look at that soon.

### DIFF
--- a/datahub/investment/models.py
+++ b/datahub/investment/models.py
@@ -15,10 +15,25 @@ from datahub.core.models import (
     BaseConstantModel,
     BaseModel,
 )
+from datahub.core.utils import StrEnum
 from datahub.documents.models import Document
-from datahub.investment.permissions import Permissions
 
 MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class Permissions(StrEnum):
+    """
+    Permission codename constants.
+
+    (Defined here rather than in permissions to avoid an import of that module.)
+    """
+
+    read_all = 'read_all_investmentproject'
+    read_associated = 'read_associated_investmentproject'
+    change_all = 'change_all_investmentproject'
+    change_associated = 'change_associated_investmentproject'
+    add = 'add_investmentproject'
+    delete = 'delete_investmentproject'
 
 
 class IProjectAbstract(models.Model):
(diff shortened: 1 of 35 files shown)


datahub/investment/models.py
```diff
@@ -270,6 +271,12 @@ def project_assurance_team(self):
         return None
 
 
+class _AssociatedToManyField(NamedTuple):
+    field_name: str
+    subfield_name: str
+    es_field_name: str
```
marcofucci: I'm not sure type annotation here adds anything and therefore subclassing the type `NamedTuple` is needed.

datahub/search/apps.py
```diff
@@ -66,6 +68,15 @@ def get_dataset(self):
 
         return DataSet(queryset, self.ESModel)
 
+    def get_permission_filters(self, request) -> Optional[dict]:
+        """
+        Gets filter arguments used to enforce permissions.
+
+        The returned dict containes rules in the form of field names and values. Results much
```
marcofucci: _much_ => _must_

datahub/investment/models.py
reupen: I'll change it to collections.namedtuple() for consistency, but if we're going to standardise I'd use typing.NamedTuple because it allows for docstrings and methods. The data classes coming in Python 3.7 use a similar syntax too https://www.python.org/dev/peps/pep-0557/.

reupen:
> I'm a bit worried that the logic behind permissions on an endpoint is in different places (IsAssociatedToInvestmentProjectFilter and InvestmentSearchApp).
>
> Not sure if we can do anything about it

Yes, that's a fair point. I think I can centralise a bit more of the logic so I'll give that a go.

datahub/investment/models.py
marcofucci: oh dear, type annotations are getting out of control 😢

reupen: I forgot, but typing.NamedTuple also supports default values :stuck_out_tongue:

reupen:
@marcofucci I centralised the generation of the filtering rules in datahub.investment.permissions.get_association_filters(). I also noticed and fixed a couple of problems around users with no teams.

I think we can get rid of InvestmentProject.get_associated_advisers() as well (and use get_association_filters()), but that might be something to look at another time.

elcct:
Looks good to me. Nice.

PR #707, opened 2017-12-15
Tidy up django admin for OMIS orders

marcofucci:
This makes the OMIS order read-only in the django admin and uses mode descriptive values for some of the fields.

### DIFF
--- a/datahub/omis/order/admin.py
+++ b/datahub/omis/order/admin.py
@@ -1,51 +1,111 @@
 from django.contrib import admin
+from django.template.defaultfilters import date as date_filter, time as time_filter
+from django.utils.html import format_html_join
+from django.utils.safestring import mark_safe
 
-from .models import Order, OrderAssignee, OrderSubscriber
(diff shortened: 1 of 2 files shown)


datahub/omis/order/admin.py
```diff
@@ -1,51 +1,105 @@
...
+        """
+        if not order.discount_value and not order.discount_label:
+            return ''
+        return f'{order.discount_value} pence - {order.discount_label}'
+
+    def uk_advisers(self, order):
+        """
+        :returns: descriptive list of advisers subscribed to the order
+        """
+        return '<br />'.join(f'{sub.adviser.name}' for sub in order.subscribers.all())
```
reupen: Don't we need to escape the field values? Maybe django.utils.html.format_html() and format_html_join()?

marcofucci: Very good point, thanks for spotting that! I've updated the code

PR #710, opened 2017-12-19
[DH-813] Use InvestmentProject permissions in team member endpoints

reupen:
Removes the InvestmentProjectTeamMember model permissions as they were an artefact of how team members were implemented.

Instead, the InvestmentProject model permissions are used in the team member endpoints and association permissions are enforced.


marcofucci:
@reupen just going through it quickly and wanted to double-check as this could be a quite dangerous assumption.
Does it mean that the Investment Project Team Member model gets the same permissions as the Investment Project one?
If so, what happens if someone is allowed to change the investment project (e.g. description etc.) but not its members? Or what if some users can only add team members whilst others can only read them? 

I can think of many cases where this logic wouldn't be quite right, unless I'm missing something?

reupen:
@marcofucci That would also apply in places we've used ManyToManyField. It's just in this case ManyToManyField wasn't used due to some complications around the role field. There aren't any requirements to restrict those actions at present. And some team members are in defined fields on InvestmentProject, so it's more complicated than what the default model permissions gave.

If we need restrictions on team members later on we can come up with a way to implement them then (with some understanding of what the requirements are :slightly_smiling_face: ).

marcofucci:
What I mean was, we should probably use the main object permissions + the specific model permissions at the same time => Investment Project Team Member AND Investment Project linked to it.

I'm not sure why we are getting completely rid of the Investment Project Team Member permissions altogether.

reupen:
We could do that, yes. It just seemed like a more complicated model, but it would mostly be manageable. The only complication I can see is the read permission on InvestmentProjectTeamMember, because the team members are also returned in the main investment project endpoint. So I would propose that we still get rid of that one.

marcofucci:
yeah it would look cleaner but it would probably add too much complexity, maybe you could keep it as it is.

reupen:
:+1: 

I'll add some more comments so that it's explained better.

datahub/investment/test/test_views.py
```diff
@@ -1554,14 +1711,60 @@ def test_patch_team_member_success(self):
...
+        team = TeamFactory()
+        _, api_client = _create_user_and_api_client(self, team, [Permissions.change_associated])
+        response = api_client.patch(url, format='json', data=request_data)
+
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_restricted_user_can_patch_team_member_of_associated_project(self):
+        """Test that a restricted user can update a team member of an associated project."""
+        creator = AdviserFactory()
+        team_member = InvestmentProjectTeamMemberFactory()
```
elcct: It is a little bit difficult to guess why would project be associated. That would be because `AdviserFactory` assigns the same team if not otherwise specified. We have some fields that are changing each time and some are constant like team in this case. Maybe it would be good to either comment about team member or assign it explicitly so it is more clear what is happening.

reupen: Thanks, that was an oversight. I've updated the test and amended AdviserFactory so that it uses a sub-factory for the team to avoid similar things happening in future.

datahub/investment/admin.py
```diff
@@ -40,6 +43,9 @@ class InvestmentProjectAdmin(BaseModelVersionAdmin):
 
 
 @admin.register(InvestmentProjectTeamMember)
+@custom_add_permission(Permissions.change_all)
+@custom_change_permission(Permissions.change_all)
+@custom_delete_permission(Permissions.change_all)
```
marcofucci: The names of these constants e.g. `Permissions.change_all` etc. don't really seem to express what they related to, that is, investment projects. Would it make sense to rename them?

reupen: Yes, if it makes it more obvious what they are. Could rename the class to InvestmentProjectPermission?

marcofucci: Yes that would be enough 👍

PR #711, opened 2017-12-19
Omis admin cancel order

marcofucci:
At the moment, advisers can only cancel an order if in status `draft` or `quote_awaiting_acceptance` using the datahub-frontend.

This adds the ability for DIT staff to cancel orders in status `quote_accepted` or `paid` as well. 
It belongs to the admin area as it should be considered an exceptional case.

It also refactors datahub.core.test_utils as there were a couple of functions which were confusingly similar.

### DIFF
--- a/datahub/omis/order/admin.py
+++ b/datahub/omis/order/admin.py
@@ -1,11 +1,68 @@
+from functools import update_wrapper
+from django import forms
+from django.conf.urls import url
 from django.contrib import admin
+from django.contrib import messages
+from django.contrib.admin.exceptions import SuspiciousOperation
+from django.contrib.admin.options import IS_POPUP_VAR
+from django.contrib.admin.templatetags.admin_urls import add_preserved_filters
+from django.contrib.admin.utils import quote, unquote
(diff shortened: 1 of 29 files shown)


datahub/omis/order/admin.py
```diff
@@ -109,3 +165,94 @@ def post_advisers(self, order):
             ) for assignee in order.assignees.all())
         )
     post_advisers.allow_tags = True
+
+    def get_urls(self):
+        """Extend the standard get_urls by adding extra urls."""
+        from django.conf.urls import url
```
reupen: I know this is the same in Django, but is there any reason for the local import?

marcofucci: it seems happy without that local import as well so moved it out, thanks!

marcofucci:
> Something we we might want to bear in mind is that Django 2.0 has some changes in the admin interface (mainly making it responsive). Not sure what the impact on customisations is, but it might be something to bear in mind before we make too many of them.

@reupen yeah I know :/ hope we can upgrade sooner rather than later...

PR #713, opened 2017-12-20
[DH-1000] Add interaction investment project association permissions

reupen:
Most of the changes are in the tests.

@feedmypixel FYI.

It's not really possible to add the new permissions to the existing LEP and DA groups via a migration, because Django creates the permissions using the post_migrate signal so they don't exist until the migrations and all been applied.


marcofucci:
As we discussed this solution has the following limitations:

1. It's hard to understand, configure, maintain. If a user can only access associated investment projects (perm `read_associated_investmentproject`) you need to remember to add and maintain the related `read_associated_investmentproject_interaction` as they have to go together in order for this to work properly.

2. Slightly related to 1), it can add security issues if you misconfigure the permissions and a user has `read_associated_investmentproject` but `read_interaction`).

3. It doesn't scale well. If we need to limit access to another related object or type of interaction, we need to add other specific permissions instead of having an abstraction layer on the top of it (e.g. `read_asociated_interaction` + an object type on the interaction model).

4. It makes the code quite complex to follow and understand.

However, other solutions have some limitations as well so I'm OK with this and we can refactor later on if it becomes a problem.

One thing I would do is renaming `read_interaction` to `read_all_interaction` as we are really using an OR here so it might make it clearer.

datahub/interaction/test/factories.py
```diff
@@ -5,10 +5,11 @@
 from datahub.core import constants
 from datahub.event.test.factories import EventFactory
 from datahub.interaction.models import Interaction
+from datahub.investment.test.factories import InvestmentProjectFactory
 
 
 class InteractionFactory(factory.django.DjangoModelFactory):
-    """Interaction factory."""
+    """Company interaction factory."""
```
elcct: That should be Interfaction factory?

reupen: It's an interaction for a company (as opposed to an interaction for an investment project). So you could rephrase the docstring as 'Factory that creates an interaction for a company'.

I was going to rename the class, but it would add extra noise here, so it could be done as a separate PR.

elcct: Right, I see. If that really creates company interaction, then I think changing class name to `CompanyInteractionFactory` could help better understand what's going on where factory is being used.

reupen: Yep. I'll do it in another PR to avoid noise here.

reupen:
@marcofucci Yep, we discussed it, and it's hard to find an optimal solution here. The problem with alternatives is that they end up with even more complex logic, and the requirements we have here are very specific. As the logic gets more complicated, it increases the scope for things to go wrong, and there is a knock-on effect as the front end has to grey out appropriate links etc. It's also hard to anticipate what future requirements will be, so while alternatives might look more flexible, that might not be true in reality.

I've gone ahead and renamed the normal read/change/add permissions.

PR #715, opened 2017-12-21
Refactor based on deprecation warnings

marcofucci:


datahub/omis/order/admin.py
```diff
@@ -131,8 +131,7 @@ def public_facing_url(self, order):
         :returns: read-only and clickable URL to the public facing OMIS page
         """
         url = order.get_public_facing_url()
-        return f'<a href="{url}">{url}<a>'
-    public_facing_url.allow_tags = True
+        return mark_safe(f'<a href="{url}">{url}<a>')
```
reupen: What about something like:
```
format_html('<a href="{href}">{text}<a>', href=mark_safe(url), text=url)
```
?

marcofucci: Well the `url` is not user-generated so `safe_url` was enough but I changed it in `format_html`.

PR #716, opened 2017-12-21
[DH-819] [DH-1000] Documents access permissions for Company and Contact.

elcct:
This adds two permissions `company.read_contact_document` and `company.read_company_document`. If user doesn't have them, then she/he will be unable to see `archived_documents_url_path` in corresponding entities.


datahub/company/serializers.py
```diff
@@ -113,6 +114,16 @@ class Meta:
             'archived_documents_url_path',
         )
 
+    def __init__(self, *args, **kwargs):
+        """Initialise Contact serialiser and check permissions."""
+        super().__init__(*args, **kwargs)
+        context = kwargs.get('context')
+        if context and 'request' in context:
```
reupen: Are we able to separate the logic for this so that we can reuse it for both companies and contacts?

I mean define something on Meta saying that `archived_documents_url_path` needs `f'company.{ContactPermission.read_contact_document}'`. And use e.g. a mixin to enforce it.

It might be nicer to override get_fields() and put the logic in there.

elcct: That's actually a very good idea. I am going to look into it. Thanks!

elcct: @reupen I have updated the serializers following your suggestions.

datahub/company/serializers.py
```diff
@@ -14,6 +15,28 @@
 MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
 
 
+class PermittedFieldsSerializer(serializers.ModelSerializer):
+    """Lets you get permitted fields only.
+
+    Needs 'permissions' key in extra_kwargs Meta class in following format:
+    extra_kwargs = {
+        'app_name.permission': 'field'
+    }
```
marcofucci: I think you meant this ;-)
```
extra_kwargs = {
    'permissions': {
        'app_name.permission': 'field'
    }
}
```

datahub/company/serializers.py
```diff
@@ -14,6 +15,30 @@
...
+
+    def get_fields(self):
+        """Gets filtered dictionary of fields based on permissions."""
+        fields = super().get_fields()
+        request = self.context.get('request', None)
+        if request:
+            permissions = self.get_extra_kwargs().get('permissions', {})
+            for permission, field in permissions.items():
+                if not request.user.has_perm(permission):
+                    fields.pop(field)
```
reupen: Would go with `del fields[field]` since the return value from `dict.pop()` isn't used.

elcct: Oh, I merged before seeing your comment. Can do a PR with an update.

datahub/company/serializers.py
```diff
@@ -14,6 +15,30 @@
 MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
 
 
+class PermittedFieldsSerializer(serializers.ModelSerializer):
+    """Lets you get permitted fields only.
+
+    Needs 'permissions' key in extra_kwargs Meta class in following format:
+    extra_kwargs = {
+        'permissions': {
+            'app_name.permission': 'field'
```
reupen: Just to note that this structure doesn't let you use the same permission for multiple fields. (It's fine though as we can change it later if needed.)

PR #719, opened 2017-12-22
Deprecate OMIS billing phone/email/contact name

marcofucci:
This makes `billing_contact_name`, `billing_phone` and `billing_email` readonly with its values set only in previous orders.


reupen:
editable=False removes them from admin, did you intend for that to happen? (They're still in the admin fields.)

marcofucci:
@reupen yep, all the fields in the admin are readonly.

reupen:
Yeah but they're not displayed in admin at all now, so it doesn't make sense to list them in OrderAdmin.

marcofucci:
oh they still show up in the admin, that's weird that the docs say that.

reupen:
Oh yes, they are showing up. I guess it's only if the admins fields aren't explicitly defined.

PR #720, opened 2017-12-22
[DH-819] [DH-1000] Fix migration.

elcct:
This makes `makemigrations` work correctly.


elcct:
That is correct. I have updated the PR.

PR #721, opened 2017-12-22
Optimise OMIS order migration

marcofucci:
This optimises one of the squashed omis migrations to make it quicker to run.

@feedmypixel this should speed things up a little.


marcofucci:
@reupen very true, reorganised the dependencies and it's now fully flat!

PR #723, opened 2017-12-27
Add management command for updating company headquarter types

reupen:
DH-1401

### DIFF
--- a/datahub/dbmaintenance/test/commands/test_update_company_headquarter_type.py
+++ b/datahub/dbmaintenance/test/commands/test_update_company_headquarter_type.py
@@ -1,0 +1,160 @@
+from io import BytesIO
+from uuid import UUID
+
+import pytest
+from django.core.management import call_command
+from reversion.models import Version
+
+from datahub.company.test.factories import CompanyFactory
+from datahub.core.constants import HeadquarterType
+
+pytestmark = pytest.mark.django_db
+
+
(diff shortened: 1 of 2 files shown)


datahub/dbmaintenance/test/commands/test_update_company_headquarter_type.py
```diff
@@ -0,0 +1,163 @@
...
+    assert 'Company matching query does not exist' in err
+    assert err.count('Failed') == 1
+
+    assert company_no_hq_type.headquarter_type_id is None
+    assert company_ghq.headquarter_type_id == UUID(HeadquarterType.ghq.value.id)
+    assert company_ehq.headquarter_type_id == UUID(HeadquarterType.ehq.value.id)
+    assert company_ukhq.headquarter_type_id == UUID(HeadquarterType.ukhq.value.id)
+
+
+AUDIT_LOG_MESSAGE = 'Headquarter type data migration correction.'
```
elcct: Just thinking, if using constant like that, maybe it would be better to declare it in the command itself and use there and then perhaps import in the test for comparison.

reupen: We need to change all of the commands then :)

I'll just get rid of the constant. (I think I was going to reuse it, but looks like I didn't.)

PR #728, opened 2017-12-28
Add disabled_on to team metadata serializer.

elcct:
As in the title.

### DIFF
--- a/datahub/metadata/serializers.py
+++ b/datahub/metadata/serializers.py
@@ -13,6 +13,7 @@ class TeamSerializer(serializers.Serializer):
     role = NestedRelatedField(TeamRole, read_only=True)
     uk_region = NestedRelatedField(UKRegion, read_only=True)
     country = NestedRelatedField(Country, read_only=True)
+    disabled_on = serializers.DateTimeField(allow_null=True, read_only=True)
 
     class Meta:
         fields = '__all__'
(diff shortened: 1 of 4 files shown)


datahub/metadata/serializers.py
```diff
@@ -13,6 +13,7 @@ class TeamSerializer(serializers.Serializer):
     role = NestedRelatedField(TeamRole, read_only=True)
     uk_region = NestedRelatedField(UKRegion, read_only=True)
     country = NestedRelatedField(Country, read_only=True)
+    disabled_on = serializers.DateTimeField(allow_null=True, read_only=True)
```
reupen: Why don't we inherit from ConstantModelSerializer?

datahub/metadata/serializers.py
```diff
@@ -13,6 +13,7 @@ class TeamSerializer(serializers.Serializer):
...
 
     class Meta:
         fields = '__all__'
```
reupen: Weird this is here since it doesn't use ModelSerializer, can we remove? Looks like ConstantModelSerializer has the same problem too.

elcct: Indeed :) Removed. Thanks!

datahub/metadata/serializers.py
elcct: Good idea, done.

PR #730, opened 2017-12-29
 Move company trading address validation to serialiser

reupen:
Part of DH-1285

Also moves the AddressValidator to the core app.


datahub/company/models/company.py
```diff
@@ -174,42 +176,12 @@ def companies_house_data(self):
 
     def has_valid_trading_address(self):
         """Tells if Company has all required trading address fields defined."""
+        field_mapping = self.TRADING_ADDRESS_VALIDATION_MAPPING
+
         return all(
-            getattr(self, field) for field in self.REQUIRED_TRADING_ADDRESS_FIELDS
+            getattr(self, field) for field, rules in field_mapping.items() if rules['required']
```
elcct: Maybe would be good to support default choice not requiring to have 'required' key?

datahub/core/validators.py
```diff
@@ -110,6 +110,83 @@ def __repr__(self):
...
+
+    message = 'This field is required.'
+
+    DEFAULT_FIELDS_MAPPING = {
+        'address_1': {'required': True},
+        'address_2': {'required': False},
+        'address_town': {'required': True},
+        'address_county': {'required': False},
+        'address_postcode': {'required': False},
+        'address_country': {'required': True},
```
elcct: Also here, maybe we could assume some default value for `required`, so we wouldn't have to type it in for each field or maybe we don't need to include fields, that are not required at all and then no need to have `required` key?

reupen: Yeah, I would probably just have had it as `'trading_address_1': True` etc. But AddressValidator wants it like this.

We need all the fields here because the validation kicks in if any of the fields have a value.

marcofucci: We could use `'trading_address_1': True` etc. but we need to change the name as _mapping_ assumes a more complex config object in my opinion and not just a True or False value.
I was designing it with the idea that we could have other settings like default values etc.

Although using bool values and renaming all those references makes it less extendable/flexible, I'm easy on it so feel free to change it. We can always refactor it back to a dict if needed.

reupen: I'm inclined to leave it – we made a decision before, and I don't think changing it is necessarily improving things.

PR #734, opened 2018-01-03
Add billing fields to OMIS invoices

marcofucci:
As we are going to allow people to change billing fields in the future, `order.billing_*` fields have been added to the invoice model as well and get populated when an invoice is created.

This way, we record those value against the invoice in case they change later on.


datahub/omis/invoice/test/factories.py
```diff
@@ -9,6 +11,16 @@ class InvoiceFactory(factory.django.DjangoModelFactory):
...
     id = factory.LazyFunction(uuid.uuid4)
     invoice_number = factory.Faker('pystr')
+
+    billing_company_name = factory.Faker('text')
+    billing_contact_name = factory.Faker('name')
+    billing_address_1 = factory.Sequence(lambda n: f'Apt {n}.')
+    billing_address_2 = factory.Sequence(lambda n: f'{n} Foo st.')
+    billing_address_country_id = Country.united_kingdom.value.id
+    billing_address_county = factory.Faker('text')
+    billing_address_postcode = 'SW1A1AA'
```
reupen: You can generate a postcode and something town-like using Faker using factory.Faker('postcode') and factory.Faker('city') if you want.

PR #736, opened 2018-01-03
Add contact_email to OMIS invoices

marcofucci:
As we are going to allow editing contacts at later stages, we need to keep track of the contact email at the time when the invoice was created.
This adds an invoice.contact_email field populated with the email address of the contact without exposing it just yet.


datahub/omis/invoice/models.py
```diff
@@ -49,6 +49,11 @@ class Invoice(BaseModel):
     invoice_vat_number = models.CharField(max_length=100, blank=True)
     payment_due_date = models.DateField()
 
+    contact_email = models.EmailField(
+        max_length=MAX_LENGTH, blank=True,
+        help_text='Email address of the contact at the time of creation.'
```
reupen: 'time of invoice creation' to reduce ambiguity?

marcofucci: 👍

PR #737, opened 2018-01-03
Add UK establishment business type

reupen:
DH-1403

Includes related validation for company numbers, as users are required to enter these in manually for UK establishments at present.

(Unfortunately UK establishments aren't included in the Companies House data file we use, but they are available via the Companies House API.)

The company tests were also reorganised (you can look at the two commits separately to make things clearer).


fixtures/metadata/companies.yaml
```diff
@@ -32,7 +32,9 @@
   pk: 43281c5e-92a4-4794-867b-b4d5f801e6f3
   fields: {name: ghq}
 
-# business type
+# Initial business types
+# Note: These are the business types that were present when Data Hub went live.
+# Business types are now added and updated using migrations
```
marcofucci: I'm not convinced this is the right approach.
I would rather be able to load some fixtures in production every time to keep things in sync instead (and/or add a migration to reload the fixture when it changes). That's why I mentioned we should isolate initial fixtures (changeable from the django admin) from the system fixtures which cannot be changed from the django admin and could be loaded as many times as we want using `manage.py loaddata`.

I know that this could take time and could be risky at this point so how about:
- we add this new business type to the fixture
- the migration command gets or creates the record instead of creating it

This way new development environments will have that value as well.

elcct: I like @marcofucci idea, it seems it will be less messy if we have that data in one place.

reupen: I agree we need to change something, but I think keeping the fixture is confusing and implies that it should be manually loaded. I also think having migrations reference a constantly changing file is potentially problematic. And having it automatically implied to environments is important, as the app depends on it.

So I would propose that we either:

1. Remove the business types from the fixtures, move the business type constants to the company app, and add a migration before this one that makes sure that they all exist. Future migrations only make the business type changes needed.

1. Remove the business types from the fixtures, move the business type constants to the company app, and use a post_migrate signal to make sure that all the business types exist.

elcct: I thought we are going to get rid of constants. Re fixtures, I still think @marcofucci proposal is better, but maybe it would help to move those fixtures to the company app and split them into individual files corresponding to the type, that way it will not be one constantly changing file. I would assume fixtures are to be manually loaded and that is expected.

reupen: > I thought we are going to get rid of constants. 

I think the main point was that we avoid putting them in two places. We said that business types were different because there's system logic around them. There's no way to avoid having constants in that case.

> I would assume fixtures are to be manually loaded and that is expected.

Yeah, and that's why they shouldn't be in a fixture. Because there's system logic around them, so they need to be loaded automatically.

elcct: I see your point. I would like 2. then.

marcofucci: Yeah if there's system logic behind them they are tricky ones.
To be honest I'm still not happy with the constants, I'd rather have a util that gets the row from the db and caches it instead of having a fixed ID in the code.

Keeping code and db in sync is still hard in your 2 cases as you need to make sure you write a migration script if you change those constants and it's not very intuitive.

Or I'm missing something?

reupen: How will you get it from the database without a constant? :)

> ...in your 2 cases as you need to make sure you write a migration script if you change those constants...

Not sure why that'd be true for post_migrate. The docs say 'Sent at the end of the migrate (even if no migrations are run)'. Django uses post_migrate for permissions and some other things, that's where that idea came from.

In any case, you're always going to have to do something to get the constant in the database. Leaving it for a manual action is just going to leave you liable to someone forgetting to do it.

marcofucci: > How will you get it from the database without a constant? :)

For example, we could use friendly codes (new field) instead of IDs and getting the rows from the code instead. We wouldn't need to keep all the records in constants either, just the ones with a logic behind it.

Re. the other point, I meant that in theory you would need to write something like migrations checking that if you change the constants you make sure there's a migration script. I'm not sure what `post_migrate` has to do with what I meant.

Anyway it's easier to discuss some code so why don't you write something and we discuss it?

reupen: @marcofucci I've pushed an implementation of 2. from above. Let me know what you think.

marcofucci: I think what you have implemented is a slightly simplified version of _django fixtures_. 
You get the same result by using a fixture file and load it with a post_migrate signal, no?
I understand from the django docs that maybe the idea around fixtures is that they only provide initial data though so yeah, this is OK for me.

I think everything could be simplified by not having a separate table for this data, I'm not sure what it adds a part from unnecessary complexity.

However, if you and @elcct are happy with this, I'll press approve.

reupen: @marcofucci Yes, you're right, it's a similar concept. The difference is that this avoids having the constants in two places (the business type ones in the code actually did not match the ones in the fixture before, which is really not good).

The Django docs do actually say:

> If you want to automatically load initial data for an app, don’t use fixtures. Instead, create a migration for your application with RunPython or RunSQL operations.

https://docs.djangoproject.com/en/2.0/howto/initial-data/#providing-initial-data-with-migrations

Unfortunately, they don't really elaborate on that. It's difficult to find a perfect solution here...

This one's gone down in priority, but I'll need to add a new metadata model soon with initial values (not one that I need constants for). So that one may well be a YAML file with a migration to load it :/

marcofucci: Yeah I don't want to use fixtures if it's not the recommended way but having constants in the code and the database doesn't seem like a good idea.
Can't they just be code enums without a table?

reupen: We tried using enums in investment project for the project status and a few other places, but they weren't liked because (without a metadata endpoint) it pushed more constants into the front end. You'd also need to come up with a different way of disabling them.

It'd also be quite a difficult change (for which I'm not sure there's a clear net benefit).

marcofucci: This solution also has the disabled_on problem.
You would have to add disabled_on to the constants and sync it all the time.

It also has the problem what if you need extra attribute on the object, you still need to keep them in the code as well and sync them as well.

Also, the file could get pretty big depending on the type of metadata with possibly hundreds of items.

Plus, in theory you should never delete a constant item otherwise the could introduce some obscure bugs as the record is still in the database but not in the code (this is true with fixtures as well).

reupen: @jim68000 do you want to want to change business type to be an enum instead of a foreign key with metadata model etc.?

> Also, the file could get pretty big depending on the type of metadata with possibly hundreds of items.

I don't think so, using constants would only be for system-managed metadata models that we've made read-only in admin.

PR #738, opened 2018-01-04
Investment Project search created_on filter.

elcct:
This adds `created_on_after` and `created_on_before` filter to the Investment Project search.

It also reorganises tests a little, so that permission tests are in its own place.

### DIFF
--- a/datahub/search/investment/test/test_views.py
+++ b/datahub/search/investment/test/test_views.py
@@ -1,5 +1,7 @@
 import datetime
+
 import pytest
+from freezegun import freeze_time
 from rest_framework import status
 from rest_framework.reverse import reverse
 
@@ -68,16 +70,27 @@ def setup_data(setup_es):
     yield investment_projects
 
 
+@pytest.fixture
+def created_on_data(setup_es):
+    """Setup data for created_on date filter test."""
+    investment_projects = []
+    dates = (
+        '2015-01-01', '2016-09-12', '2017-09-12', '2048-02-04', '2048-01-24',
+    )
+
+    for date in dates:
+        with freeze_time(date):
+            investment_projects.append(
+                InvestmentProjectFactory()
+            )
+
+    setup_es.indices.refresh()
+
+    yield investment_projects
+
+
 class TestSearch(APITestMixin):
     """Tests search views."""
-
-    def test_investment_project_no_permissions(self):
-        """Should return 403"""
-        user = create_test_user(dit_team=TeamFactory())
-        api_client = self.create_api_client(user=user)
-        url = reverse('api-v3:search:investment_project')
-        response = api_client.get(url)
-        assert response.status_code == status.HTTP_403_FORBIDDEN
 
     def test_search_investment_project_json(self, setup_data):
         """Tests detailed investment project search."""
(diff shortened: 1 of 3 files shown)


datahub/search/investment/test/test_views.py
```diff
@@ -92,17 +105,47 @@ def test_search_investment_project_json(self, setup_data):
...
+        response = self.api_client.post(url, filter, format='json')
 
         assert response.status_code == status.HTTP_200_OK
-        assert response.data['count'] == 1
-        assert len(response.data['results']) == 1
+        assert response.data['count'] == results
+        assert len(response.data['results']) == results
+
+    @pytest.mark.parametrize('filter,results', (
+        ({'created_on_before': datetime.datetime(2017, 6, 13, 9, 44, 31, 62870)}, 2, ),
```
reupen: I found this a bit hard to read, maybe make the formatting a bit more consistent? Could also do `import datetime from datetime` to avoid having to write datetime.datetime (although it might make more sense to use strings since they're part of the request – it's not really clear what format they're being serialised to).

datahub/search/investment/test/test_views.py
```diff
@@ -92,17 +105,47 @@ def test_search_investment_project_json(self, setup_data):
         assert len(response.data['results']) == 1
         assert response.data['results'][0]['name'] == 'abc defg'
 
-    def test_search_investment_project_date_json(self, setup_data):
+    @pytest.mark.parametrize('filter,results', (
+        ({'estimated_land_date_before': datetime.datetime(2017, 6, 13, 9, 44, 31, 62870)}, 1, ),
```
reupen: `estimated_land_date` is a date, any reason for using datetimes?

Might be nice to test some edge cases e.g. when the actual estimated_land_date value is equal to one of filter values.

What happens when you pass a estimated_land_date_after value greater than the estimated_land_date_before value? :)

datahub/search/investment/test/test_views.py
```diff
@@ -92,17 +105,47 @@ def test_search_investment_project_json(self, setup_data):
...
+    ))
+    def test_search_investment_project_created_on_json(self, created_on_data, filter, results):
+        """Tests detailed investment project search."""
+        url = reverse('api-v3:search:investment_project')
+
+        response = self.api_client.post(url, filter, format='json')
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == results
+        assert len(response.data['results']) == results
```
reupen: Would be nice to check that they're the expected results (rather than just the count).

datahub/search/investment/test/test_views.py
```diff
@@ -92,17 +105,47 @@ def test_search_investment_project_json(self, setup_data):
...
+        ({'estimated_land_date_after': datetime.datetime(2017, 6, 13, 9, 44, 31, 62870)}, 2,),
+        (
+            {
+                'estimated_land_date_after': datetime.datetime(2017, 6, 13, 9, 44, 31, 62870),
+                'estimated_land_date_before': datetime.datetime(2030, 6, 13, 9, 44, 31, 62870),
+            },
+            1,
+        ),
+    ))
+    def test_search_investment_project_estimated_land_date_json(self, setup_data, filter, results):
```
marcofucci: Can you use a name different from `filter` so that it doesn't shadow the built-in `filter`

elcct: Oops. Sure I will fix that, thanks :blush:

datahub/search/investment/test/test_views.py
elcct: I have improved formatting and dropped datetime.datetime in favour of string.

datahub/search/investment/test/test_views.py
elcct: Removed datetimes, added some edge cases.

datahub/search/investment/test/test_views.py
elcct: Also added checking the results. Thanks :)

elcct:
@reupen no particular reason. I have removed it now. Thanks.

PR #741, opened 2018-01-04
Add Investment Project comments.

elcct:
This adds `comments` field to Investment Project and migration command to migrate missing comments.


datahub/investment/serializers.py
```diff
@@ -23,6 +23,7 @@ class IProjectSummarySerializer(serializers.ModelSerializer):
     """Serialiser for investment project endpoints."""
 
     incomplete_fields = serializers.SerializerMethodField()
+    comments = serializers.CharField(read_only=True)
```
reupen: How about using read_only_fields on Meta? Most of the other read-only fields are specified as read-only there.

datahub/investment/test/test_views.py
```diff
@@ -68,6 +68,7 @@ def test_list_projects_success(self):
             'project_code',
             'investment_type',
             'description',
+            'comments',
```
reupen: Could test updating of comments in test_update_read_only_fields() as well.

datahub/investment/serializers.py
reupen: Also, maybe it should be read-only in admin as well.

datahub/dbmaintenance/test/commands/test_update_investment_project_comments.py
```diff
@@ -0,0 +1,110 @@
...
+        # should be ignored
+        InvestmentProjectFactory(),
+    ]
+
+    old_comments = investment_projects[1].comments
+
+    bucket = 'test_bucket'
+    object_key = 'test_key'
+    csv_content = f"""Id,comments
+00000000-0000-0000-0000-000000000000,Comment 1
```
reupen: Might be nice to put a muiti-line comment in there (quoted).

datahub/investment/serializers.py
elcct: Good points. Done.

datahub/investment/test/test_views.py
elcct: Good idea. Done.

datahub/dbmaintenance/test/commands/test_update_investment_project_comments.py
elcct: Also good idea. Done.

elcct:
Found a bug in `test_simulate` for another migration command and couldn't resist to fix it.

PR #742, opened 2018-01-04
Fix permissions enforcement in global search for users with no permissions

reupen:


datahub/search/elasticsearch.py
```diff
@@ -167,10 +167,12 @@ def _get_global_permission_query(permission_filters_by_entity=None):
     if permission_filters_by_entity is None:
         return None
```
marcofucci: I find it a bit confusing that `None` means do not apply any filtering whilst anything else falsy means you don't have any permissions. Maybe it's just me though :/

Could you add some docs here and in the `_get_entity_permission_query` if we want to keep it this way?

datahub/search/test/test_elasticsearch.py
```diff
@@ -114,3 +114,17 @@ def test_configure_connection(connections, settings):
         'hosts': [settings.ES_URL],
         'verify_certs': settings.ES_VERIFY_CERTS
     })
+
+
+def test_get_entity_permission_query_no_conditions():
+    """
+    Test that _get_entity_permission_query_no_conditions() correctly handles an empty dict
```
marcofucci: typo there. should be __get_entity_permission_query_

datahub/search/test/test_elasticsearch.py
```diff
@@ -114,3 +114,17 @@ def test_configure_connection(connections, settings):
...
+    of conditions.
+
+    (An empty dict of conditions should mean that there are no conditions that would permit
+    access.)
+    """
+    query = elasticsearch._get_entity_permission_query({})
```
marcofucci: I would use `permission_filters={}` instead of just `{}` to make clear what is that empty dict about

PR #746, opened 2018-01-05
Add ability to edit billing info in quote awaiting acceptance

marcofucci:
When the order is in status `quote_awaiting_acceptance`, the billing fields can still be edited as they are only used when an invoice is generated.

### DIFF
--- a/datahub/omis/order/validators.py
+++ b/datahub/omis/order/validators.py
@@ -292,12 +292,7 @@ class OrderInStatusValidator:
         if not self.instance and not self.order_required:
             return  # all fine
 
-        allowed = any(
-            self.instance.status == status
-            for status in self.allowed_statuses
-        )
-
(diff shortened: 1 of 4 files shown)


datahub/omis/order/validators.py
```diff
@@ -376,3 +377,28 @@ def __init__(self, field, value):
     def __call__(self, combiner):
         """Check that the force_delete flag has the expected value."""
         return combiner.serializer.context.get('force_delete', False) == self.value
+
+
+class EditableFieldsRule(AbstractRule):
+    """Rule for checking that only editable fields can be changed."""
```
reupen: I found this a little bit confusing, maybe something like 'Rule that checks that only certain fields have been modified.'?

marcofucci: 👍  much better

PR #748, opened 2018-01-05
Release 1.7.1

reupen:


reupen:
Just noticed that the comments field is in this, so it should be 1.8.0.

PR #751, opened 2018-01-07
Add pricing fields to OMIS invoices

marcofucci:
As we are going to allow people to change VAT fields in the future, pricing fields that depend on them need to be stored against invoices to keep track of their values at the time of invoice creation.

Note that quotes only include the net cost (without VAT) so it's safe to change the VAT status after a quote is accepted as its net value doesn't change, only the gross one could.


marcofucci:
@reupen are you happy for me to merge this?

datahub/omis/order/test/factories.py
```diff
@@ -96,9 +96,15 @@ class OrderWithAcceptedQuoteFactory(OrderFactory):
     """Order factory with an accepted quote."""
 
     quote = factory.SubFactory(AcceptedQuoteFactory)
-    invoice = factory.SubFactory(InvoiceFactory)
     status = OrderStatus.quote_accepted
 
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
```
reupen: Could use http://factoryboy.readthedocs.io/en/latest/reference.html#factory.PostGeneration as well I think.

marcofucci: 💯  I had already refactored using PostGeneration in the PR I'm working on which depends on it, can't be bothered changing this :-), following soon...

PR #754, opened 2018-01-08
Create new OMIS invoice when billing fields change

marcofucci:
This allow advisers to edit invoice/billing details when an order is in 'Quote Accepted' and creating a new invoice as a consequence.

The net cost (the one shown in the quote) cannot change after the quote is accepted but the VAT details can and therefore the gross cost on the invoice as well.


datahub/omis/order/serializers.py
```diff
@@ -251,6 +263,22 @@ def create(self, validated_data):
...
+        """
+        Update invoice details if any of the invoice related fields has changed.
+        """
+        with transaction.atomic():
+            instance = super().update(instance, validated_data)
+
+            # update invoice details if necessary
+            if (
+                instance.status == OrderStatus.quote_accepted
+                and (set(ORDER_FIELDS_INVOICE_RELATED) & set(validated_data))
```
reupen: Could make `ORDER_FIELDS_INVOICE_RELATED` a set if you don't need the order?

Also, the return value of dict.keys() is set-like so validated_data.keys() should work.

datahub/omis/order/test/factories.py
```diff
@@ -92,18 +92,17 @@ class OrderWithCancelledQuoteFactory(OrderFactory):
     quote = factory.SubFactory(CancelledQuoteFactory)
 
 
+def _make_invoice(obj, *args, **kwargs):
+    obj.invoice = Invoice.objects.create_from_order(obj)
```
reupen: Should we check `create` here? I guess it's just the order reference that will be missing if create == False though.

datahub/omis/order/serializers.py
marcofucci: yes and yes, very good points, thanks!

datahub/omis/order/test/factories.py
marcofucci: yep true, I've refactored it as well

PR #757, opened 2018-01-10
use AWS ES with AWS request signing

nusnewob:


config/settings/common.py
```diff
@@ -184,7 +184,11 @@
 APPEND_SLASH = False
 
 # Leeloo stuff
+AWS_ELASTICSEARCH_REGION = env('AWS_ELASTICSEARCH_REGION')
```
reupen: Don't we need a ES_USE_AWS_AUTH setting as we won't use AWS auth locally etc. and we'll need to be able to manage the switchover as well?

datahub/search/elasticsearch.py
```diff
@@ -80,10 +83,28 @@ class MatchNone(Query):
...
+        aws_secret_access_key = settings.AWS_ELASTICSEARCH_SECRET,
+        aws_host = es_host.netloc,
+        aws_region = settings.AWS_ELASTICSEARCH_REGION,
+        aws_service = 'es'
+    )
     connections.configure(
         default={
-            'hosts': [settings.ES_URL],
-            'verify_certs': settings.ES_VERIFY_CERTS
+            'hosts': [es_host.netloc],
```
reupen: Doesn't passing the URL work with AWS auth?

datahub/search/elasticsearch.py
```diff
@@ -80,10 +83,28 @@ class MatchNone(Query):
...
+            'port': es_port,
+            'use_ssl': settings.ES_USE_SSL,
```
reupen: Can't we get this from http vs https in the URL?

datahub/search/elasticsearch.py
nusnewob: AWS auth requires the signing URL and request URL to be exact match (without http/https and port)

datahub/search/elasticsearch.py
nusnewob: `urlparse` doesn't return port if only http/https specified.

datahub/search/elasticsearch.py
nusnewob: `ES_USE_AWS_AUTH` switch added in new commit.

config/settings/common.py
nusnewob: `ES_USE_AWS_AUTH` switch added in new commit.

reupen: So can we only load AWS_ELASTICSEARCH_REGION/AWS_ELASTICSEARCH_KEY/AWS_ELASTICSEARCH_SECRET if ES_USE_AWS_AUTH is True?

datahub/search/elasticsearch.py
reupen: Yes but can't we determine USE_SSL by checking if the URL scheme is https?

config/settings/common.py
nusnewob: yep, updated it

datahub/search/elasticsearch.py
nusnewob: not set by default now, only loaded when `ES_USE_AWS_AUTH` is True, it's default protocol on AWS ES.

reupen:
@nusnewob There's still a failing test.

datahub/search/elasticsearch.py
```diff
@@ -1,7 +1,11 @@
 from collections import defaultdict
 from itertools import chain
 
+from urllib.parse import urlparse
+from aws_requests_auth.aws_auth import AWSRequestsAuth
```
reupen: The urllib import should be in the first group (standard library imports) and the aws one in the next group (third-party imports). So it should look like:

```
from collections import defaultdict
from itertools import chain
from urllib.parse import urlparse

from aws_requests_auth.aws_auth import AWSRequestsAuth
from elasticsearch import RequestsHttpConnection
...
```

PR #758, opened 2018-01-10
Fix the migration dependencies in OMIS invoice

marcofucci:
This fixes a misconfigured dependency in the populate billing field data migration for the OMIS invoice app.


marcofucci:
Don’t we still need those dependencies in case we squash this one?

PR #759, opened 2018-01-11
Omis change contact

marcofucci:
This allows the contact on an order to be changed if the order is in status != complete or cancelled.

This also refactors the order validators trying to simplify things. 
Instead of several, sometimes abstract validators, now the view only uses one `OrderEditableFieldsValidator` which defines clearly what can be edited and what can't at any particular stage.

### DIFF
--- a/datahub/omis/order/validators.py
+++ b/datahub/omis/order/validators.py
@@ -76,6 +76,43 @@ class ReadonlyAfterCreationValidator:
                     })
 
 
+class OrderEditableFieldsValidator:
+    """
+    Validator that makes sure that only certain fields have been modified
+    depending on the order status.
+    """
+
+    message = 'This field cannot be changed at this stage.'
+
+    def __init__(self, mapping=None):
+        """
+        Set the mapping.
+
+        :param mapping: dict of <order status, editable fields>
+        """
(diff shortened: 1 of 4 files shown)


datahub/omis/order/validators.py
```diff
@@ -61,19 +65,34 @@ def set_context(self, serializer):
...
 
-                if value != getattr(self.instance, field):
-                    field_name = self.instance._meta.get_field(field).verbose_name
-                    raise ValidationError({
-                        field: self.message.format(field_name)
-                    })
+        editable_fields = self.mapping[self.instance.status]
+        for field in combiner.data:
+            if field not in editable_fields and self._has_changed(field, combiner):
+                raise ValidationError({field: self.message})
```
reupen: It can be nice to collect the errors and then raise them together to give more feedback in one go, but I assume it's not that important in this case if the front end won't normally allow the user to attempt to edit the fields.

marcofucci: Yes, good point but yes as you said it shouldn't really happen even with one if the frontend is doing what it's supposed to do. I might leave it as it is and come back to it if I have time.
Thanks!

PR #782, opened 2018-01-19
Replace empty town with comma during CH import.

elcct:
This puts comma for address town if it is empty in the imported record. This is a workaround for the data quality issue and enables user to create company that would have had address town missing.

It also adds simulate option, so that when it is added it will not save the results to the database.

### DIFF
--- a/datahub/company/management/commands/sync_ch.py
+++ b/datahub/company/management/commands/sync_ch.py
@@ -102,7 +102,7 @@ def transform_ch_row(row):
     for name in settings.CH_RELEVANT_FIELDS:
         ret[name] = row.get(name, '')
 
-    if 'registered_address_town' in row and row['registered_address_town'] == '':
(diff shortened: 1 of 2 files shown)


datahub/company/management/commands/sync_ch.py
```diff
@@ -102,6 +102,15 @@ def transform_ch_row(row):
     for name in settings.CH_RELEVANT_FIELDS:
         ret[name] = row.get(name, '')
 
+    if 'registered_address_town' in row and row['registered_address_town'] == '':
```
reupen: You could check `ret` instead here since that should always have `registered_address_town`.

Also I'd only do the substitution if `registered_address_1` is set – if there is no address at all, better not to do the substitution.

elcct: Good points. I have updated the code. Thanks!

PR #805, opened 2018-01-31
Improve various admin lists

reupen:
Issue number: N/A

### Description of change

Modifies the search fields and list display fields for various models. Also appends '(disabled)' in __str__() for metadata objects, so that it shows in admin when selecting options for related fields.

### Checklist

### DIFF
--- a/datahub/core/models.py
+++ b/datahub/core/models.py
@@ -85,7 +85,10 @@ class BaseConstantModel(DisableableModel):
 
     def __str__(self):
         """Human readable admin name."""
-        return self.name
+        parts = [self.name]
+        if self.disabled_on:
+            parts.append('(disabled)')
+        return ' '.join(parts)
 
 
 class BaseOrderedConstantModel(BaseConstantModel):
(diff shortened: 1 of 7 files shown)


datahub/core/models.py
```diff
@@ -85,7 +85,10 @@ class Meta:
 
     def __str__(self):
         """Human readable admin name."""
-        return self.name
+        parts = [self.name]
+        if self.disabled_on:
+            parts.append('(disabled)')
+        return ' '.join(parts)
```
elcct: Just thinking if that could be something like `f'{self.name}{" (disabled)" if self.disabled_on else ""}'`. On one hand it is short, but on another is less obvious.

reupen: Another option would be something like:

```
parts = [self.name, '(disabled)' if self.disabled_on else None]
return ' '.join(filter(None, parts))
```

But I think I'll leave it for now.

reupen: Added a helper function for this as it's a recurring problem.

PR #816, opened 2018-02-05
Add grant amount offered field to service deliveries

reupen:
Issue number: DH-1478

### Description of change

Adds a 'grant amount offered' field to interactions. It's only valid for interactions with kind 'service_delivery'.

### Checklist


reupen:
Decimal was originally used in investment projects, so have stuck with that here. @yeahfro can comment from a front-end perspective, but I reckon treating it as a string is simpler than converting an integer in pence to pounds and formatting to two decimal places.

marcofucci:
👍  I agree, unless the frontend has to act on it, e.g. validate the value, do some maths etc. In those cases numbers are always better than strings.
Formatting to/from pounds is straightforward as well because it's always /100 and *100 in case of integers, it's really a no-brainer.

reupen:
Not sure it has much impact on validation as it doesn't affect what the user enters (which is ultimately a string, anyway). No arithmetic involved here, but if it turned out e.g. the total for a company was required, the API would be better placed to provide that.

elcct:
It feels like we should agree to have just one way of representing money, so that we can avoid potential problems in the future. I am not a fan of having it in two ways.

PR #817, opened 2018-02-06
Allow OMIS T&Cs to be changed via admin

marcofucci:
This allows TermsAndConditions records to be added via the django admin. 
After creation records cannot be changed or deleted as they might have been used in generated quotes.

t cs admin

### DIFF
--- a/datahub/omis/quote/admin.py
+++ b/datahub/omis/quote/admin.py
@@ -1,0 +1,34 @@
+from django.contrib import admin
+
+from .models import TermsAndConditions
+
+
+@admin.register(TermsAndConditions)
+class TermsAndConditionsAdmin(admin.ModelAdmin):
+    """Admin for TermsAndConditions."""
+
+    fields = (
+        'id',
+        'name',
+        'created_on',
+        'content',
+    )
+    list_display = (
+        'name',
+        'created_on',
+    )
+    readonly_fields = (
+        'id',
+        'created_on',
+    )
+    actions = None
+
+    def has_delete_permission(self, request, obj=None):
+        """Records cannot be deleted."""
+        return False
+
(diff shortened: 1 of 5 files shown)


marcofucci:
> it might be nice to apply restrictions based on whether the terms and conditions have actually been used, but I'm not sure where that logic could be put.
@reupen 

Yeah I thought about it but I think it's slightly less easy to understand for staff users and it's too much effort for what you get back.

> Looks OK, but it doesn't seem ideal as there's still a lot of messaging etc. from Django about editing like 'You may edit it again below.'

What do you suggest? I could get rid of that message but I'm not sure if there are other occurrences.

marcofucci:
@reupen 
Improved it based on your feedback

**Add** 
Without save and continue editing button.

add

**Change**
Without save buttons.

change

**List**
Without actions and with correct success message.

list

reupen:
> Yeah I thought about it but I think it's slightly less easy to understand for staff users and it's too much effort for what you get back.

I was more coming from the angle that if you create one, notice a mistake after saving it, then it's quite difficult to fix. Another option might be to emphasise that new terms and conditions should be tested in another environment first? Or another option might be to have new terms and conditions in a draft status initially (still editable) and then have to make them active to put it into use.

Also not sure if you've seen the 'Save as new' admin functionality, not sure how you feel about that https://docs.djangoproject.com/en/2.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.save_as

Just a few ideas, not critical (have already approved the PR 🙂).

marcofucci:
Yes, you might be right but I can't spend too much time on this so let's see how they use it and I'll keep your ideas in mind in case it creates too many problems.

Thanks for the feedback!

BTW, I'm not a big fan on the new 'Save as new' button, it's not very intuitive :/

PR #818, opened 2018-02-07
Add OMIS UK Regional Settings

marcofucci:
This adds a OneToOne model UKRegionalSettings for storing UK manager emails used by OMIS notifications.

I'm not a big fan of the name `UKRegionalSettings` and `ukregion.omis_settings` but I couldn't find a better one.

This also adds the `multi_email_field` dependency defining a field with multiple email addresses separated by new lines. I'm not a big fan of this extra dependency either but anything else seems messier.


reupen:
Did you consider ArrayField from django.contrib.postgres.fields? I think that ends up being comma-separated in admin, so not quite the same, but that's not much different from writing an email anyway.

Not much of a fan of that package as it doesn't claim compatibility with anything newer than Python 3.5 and Django 1.10.

datahub/omis/region/models.py
```diff
@@ -0,0 +1,21 @@
...
+
+from multi_email_field.fields import MultiEmailField
+
+from datahub.metadata.models import UKRegion
+
+
+class UKRegionalSettings(models.Model):
+    """OMIS settings for UK Regions."""
+
+    ukregion = models.OneToOneField(
```
reupen: Why `ukregion` rather than `uk_region`?

datahub/omis/region/models.py
```diff
@@ -0,0 +1,21 @@
...
+        UKRegion,
+        primary_key=True,
+        on_delete=models.CASCADE,
+        related_name='omis_settings'
+    )
+    manager_emails = MultiEmailField(blank=True)
+
+    class Meta:
+        verbose_name = 'OMIS UK Regional Settings'
```
reupen: Verbose name should be in lower case (except for initialisms, proper nouns etc.) So I would expect `OMIS UK regional settings`.

marcofucci: 👍

datahub/omis/region/models.py
marcofucci: 👍

marcofucci:
> Did you consider ArrayField from django.contrib.postgres.fields?
Wow, I didn't notice that it had a base_field, it looks just perfect.

marcofucci:
@reupen much better now, thanks!

elcct:
Indeed much nicer with `ArrayField`.

PR #820, opened 2018-02-07
Notify OMIS regional managers when an order is created

marcofucci:
This sends an email to the regional managers of the UK region related to an order when this gets created.


datahub/omis/notification/constants.py
```diff
@@ -7,6 +7,7 @@ class Template(Enum):
     generic_order_info = '53055338-6acd-4dd6-b315-d1bd74015eb2'
 
     order_created_for_post_manager = '335f5402-f610-43ef-bfcf-e196215b3cf1'
+    order_created_for_reginal_manager = '15a7824a-b179-4ffb-a8a1-74ba89285e20'
```
reupen: Typo here ('reginal').

PR #827, opened 2018-02-13
Squash interaction and metadata migrations

reupen:
Issue number: N/A

### Description of change

Squashes interaction migrations up to 0019, and metadata migrations up to 0011.

There were circular dependencies again arising from the original migrations (which were difficult to resolve). I've resolved them in a similar way to the previous PR. Will leave the squashing at this for now so that we can deploy these to all environments.

### Checklist


marcofucci:
@reupen shouldn't you keep the old migrations untouched?

reupen:
@marcofucci Yes, but I don't have another way to resolve the circular dependencies because they're (weirdly) coming from the old migrations. Do you have any other suggestions? I don't think it's a big deal anyway, they aren't recent and we're just going to delete them after the squashing is complete.

reupen:
Not sure why but the UAT environment was unhappy with this. Going to park this until #825 has been transitioned to a normal migration.

reupen:
Ah, nope, it failed because the the wrong commit was deployed. So all looks fine.

PR #832, opened 2018-02-16
Add GOV.UK Pay client library

marcofucci:
This adds a client library which can be used to interface with GOV.UK Pay.


datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        :raises GOVUKPayAPIException if status code >= 400
+
+        :param response: Response instance to check
+        """
+        if response.ok:
+            return
+
+        error_msg = f'{response.status_code} Error for url: {response.url}'
+        raise GOVUKPayAPIException(error_msg, response=response)
```
reupen: Might be worth checking that the content type of the response is 'application/json' before trying to decode it as JSON. You can use `cgi.parse_header()` to parse the Content-Type header.

datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,112 @@
...
+            'Authorization': 'Bearer %s' % settings.GOVUK_PAY_AUTH_TOKEN
+        }
+
+    def raise_for_status(self, response):
+        """
+        :raises GOVUKPayAPIException if status code >= 400
+
+        :param response: Response instance to check
+        """
+        if response.ok:
```
reupen: What about calling `response.raise_for_status()`, and re-raising `HTTPError`s from that as `GOVUKPayAPIException`s? It generates nice error messages for you (which include the status code, reason and URL).

datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,112 @@
...
+
+    TIMEOUT = 15
+
+    @cached_property
+    def _headers(self):
+        """GOV.UK common headers including the Authorization one."""
+        return {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer %s' % settings.GOVUK_PAY_AUTH_TOKEN
```
reupen: Did you mean `f'Bearer {settings.GOVUK_PAY_AUTH_TOKEN}'` 😛

datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,112 @@
...
+    """Client used to interface with GOV.UK Pay."""
+
+    TIMEOUT = 15
+
+    @cached_property
+    def _headers(self):
+        """GOV.UK common headers including the Authorization one."""
+        return {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json',
```
reupen: Doesn't requests set the Content-Type header automatically if you use `json=`?

datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,112 @@
...
+
+    :param path: path without leading `/` e.g. `payments`
+    """
+    return urllib.parse.urljoin(settings.GOVUK_PAY_URL, path)
+
+
+class PayClient:
+    """Client used to interface with GOV.UK Pay."""
+
+    TIMEOUT = 15
```
reupen: Could go in settings?

datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,112 @@
...
+        response = requests.post(
+            govuk_url('payments'),
+            headers=self._headers,
+            json={
+                'amount': amount,
+                'reference': reference,
+                'description': description,
+                'return_url': return_url,
+            },
+            timeout=self.TIMEOUT
```
reupen: Could perhaps use a `self._request()` method that uses `requests.requests()` and sets the timeout and headers automatically. Would have to pass it the HTTP method as an argument then.

datahub/omis/payment/govukpay.py
marcofucci: Could do but it doesn't do much more than what I'm doing here, I got rid of the reason as it was often empty so useless.

datahub/omis/payment/govukpay.py
marcofucci: True eheh

datahub/omis/payment/govukpay.py
marcofucci: 👍

datahub/omis/payment/govukpay.py
marcofucci: Yep, good idea, that's what I usually do but didn't want to make it complex/less readable than necessary here. I'll refactor it

datahub/omis/payment/govukpay.py
marcofucci: True, wouldn't it better to be explicit though? Do you think it's better to remove it?

datahub/omis/payment/govukpay.py
marcofucci: Why do you think we need this? 
If the response is not a valid JSON the `.json()` call would raise a `JSONDecodeError` which is what we need.
I could add a try/catch but we can't guarantee everything was OK if we were expecting a JSON response.
It looks pythonic enough to me, _Easier to ask for forgiveness than permission_, no?

reupen: Yeah I think it's fine. The main context in my head was decoding JSON error responses, which we aren't doing here.

datahub/omis/payment/govukpay.py
reupen: Yeah, I think it can go. It doesn't make sense for requests without a body either.

datahub/omis/payment/govukpay.py
```diff
@@ -0,0 +1,115 @@
...
+            exactly as they are.
+            The headers and timeout values are set automatically but can be
+            overridden by kwargs if needed.
+        """
+        request_kwargs = {
+            'headers': self._headers,
+            'timeout': settings.GOVUK_PAY_TIMEOUT,
+            **kwargs
+        }
+        response = getattr(requests, verb)(govuk_url(path), **request_kwargs)
```
reupen: Why don't you use `requests.requests()` here? It has a (case-insensitive) method parameter, and other methods just call that.

Also would be nice to log that a request is being made.

marcofucci: oh you mean `requests.request`, sorry misunderstood what you meant... 👍

datahub/omis/payment/govukpay.py
marcofucci: I've refactored this as well, didn't change much but maybe it looks nicer now.

marcofucci:
@elcct do you have any comments before I merge this?

PR #833, opened 2018-02-16
Add linking to a Global Headquarter for Company.

elcct:
Issue number: DH-1469, DH-1567

### Description of change

This change adds `global_headquarter` to the `Company` model which lets user link Company to its Global Headquarter. It also adds reverse relation `subsidiaries` that allows to list all linked companies.

This need has been discovered through user research, that we need to show a link from any level in the company hierarchy to the global headquarter.

I have also extended `OperatorRule` with `ForeignKeyOperatorRule` that gets value of given field, so that the object linked by foreign key can be passed to the operator, not only the id. I am not sure if the name of this new class is correct, but couldn't come up with anything better.

### Checklist


datahub/company/models/company.py
```diff
@@ -130,6 +130,10 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
         'self', blank=True, null=True, on_delete=models.SET_NULL,
         related_name='children'
     )
+    global_headquarter = models.ForeignKey(
```
reupen: Should be headquarters here (the noun is always headquarters).

datahub/company/serializers.py
```diff
@@ -200,7 +202,10 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         ),
         'uk_establishment_not_in_uk': ugettext_lazy(
             'A UK establishment (branch of non-UK company) must be in the UK.'
-        )
+        ),
+        'global_headquarter_company_is_not_a_global_headquarter': ugettext_lazy(
+            'Company to be set as Global Headquarter must be a Global Headquarter.'
```
reupen: The message is a bit confusing :) Maybe it could be a bit more explicit? Also no need for the capital G and H.

elcct:
@reupen 

> Do we need parent/children and global_headquarters/subsidiaries? It's a bit confusing...

Yes, as parent/children would be used to build multi-level hierarchies (if there is going to be a need for that - we are working on this).

datahub/company/models/company.py
elcct: I should have checked in the dictionary as I wasn't sure about it. Thanks! Renamed now.

datahub/company/serializers.py
elcct: I have edited the message a bit, but I am not sure how could I write it simpler. This message shouldn't really be shown to the user anyway, because user will be able only to select the link from companies that are global headquarters. Possible edge case is when someone else remove the global headquarters type at the time the other user selects that company to link it.

reupen:
> Yes, as parent/children would be used to build multi-level hierarchies (if there is going to be a need for that - we are working on this).

Yeah, but there's inherent inconsistency in that. Why not get rid of parent until it's actually needed? Then it can be modelled properly.

datahub/company/migrations/0025_company_global_headquarter.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.0.2 on 2018-02-12 11:58
```
marcofucci: Change name as well?

elcct: Nice one! Fixed.

reupen:
> Parent models link to immediate parent company whereas global headquarters models direct link to the company's headquarters. These are two different things. I could deal with parent in another PR.

Isn't the global HQ just the top of the hierarchy? That was my point about consistency. We also now have children and subsidiaries and it's not obvious what the difference is. At the very least the unused stuff could be removed from the API response.

datahub/company/serializers.py
```diff
@@ -354,6 +365,14 @@ class Meta:
                     when=EqualsRule('business_type',
                                     BusinessTypeConstant.uk_establishment.value.id),
                 ),
+                ValidationRule(
+                    'global_headquarters_company_is_not_a_global_headquarters',
```
reupen: Should setting a company as its own global HQ be allowed?

elcct: It shouldn't be allowed. Added additional validation. Thanks

elcct:
> Isn't the global HQ just the top of the hierarchy?

In many cases it would be, but there are exceptions.

> We also now have children and subsidiaries and it's not obvious what the difference is.

Parent/children is about immediate relationship, maybe this would need either better naming or documentation.

>  At the very least the unused stuff could be removed from the API response.

I have removed parent/children from the API for now.

datahub/company/serializers.py
```diff
@@ -354,6 +363,19 @@ class Meta:
                     when=EqualsRule('business_type',
                                     BusinessTypeConstant.uk_establishment.value.id),
                 ),
+                ValidationRule(
```
reupen: Last comment 🙂 

I think both of these could be validators on the fields instead of the class as they don't involve more than one field. That's better as the errors are returned at an earlier stage.

(For the first one, you can get the current instance for update operations via set_context() on your validator.)

marcofucci:
What happens in this scenario:
- A is a GHQ company
- user changes company B with GHQ fk pointing to A
- user change A not to be a GHQ any more

After this change company B still has GHQ pointing to A which is not a GHQ company any longer.

elcct:
@marcofucci we have not decided yet. There are two solutions I have pointed out:

- when user changes a company from GHQ to non GHQ, then all links are being removed.
- when user changes a company from GHQ an error message appears that GHQ has subsidiaries and asks user to remove links manually.
There might be also a need of moving the links from one company to another.

Since we don't know what would be the best solution, we agreed to address this later. In case such event happens before we address it, it could be fixed manually by the user or by a script.

marcofucci:
Hm OK. In the meantime it would be good to notify something when that happens so that we can get an idea of the problem without forgetting it.
It's a bit of a misuse but we could maybe send a warning to sentry so that we get a notification and we know how often it happens?

datahub/company/serializers.py
```diff
@@ -215,7 +224,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
     business_type = NestedRelatedField(
         meta_models.BusinessType, required=False, allow_null=True
     )
-    children = NestedRelatedField('company.Company', many=True, required=False)
+    subsidiaries = NestedRelatedField('company.Company', many=True, required=False)
```
reupen: One more comment actually 😬

I'm not sure about embedding all subsidiaries in the response as there could be a large number in cases. I know limited information is returned here, but then how useful is it?  It seems like something better suited to a separate endpoint with pagination.

marcofucci: +1 good point

elcct: That makes sense, it should be removed as there will be another endpoint to obtain this data. Fixed.

datahub/company/serializers.py
elcct: You're right - it looks like doing it that way was a bit of an overkill :) I have replaced it with just the field validation which does the job.

datahub/company/serializers.py
```diff
@@ -264,6 +270,28 @@ class CompanySerializer(PermittedFieldsModelSerializer):
...
+            if (
+                global_headquarters.headquarter_type
+                and str(global_headquarters.headquarter_type.id) != HeadquarterType.ghq.value.id
+            ) or not global_headquarters.headquarter_type:
+                raise serializers.ValidationError(
+                    self.error_messages['global_headquarters_company_is_not_a_global_headquarters']
+                )
+
+            # check if global_headquarters is not pointing to an instance of the model
+            if self.instance and self.instance.id == global_headquarters.id:
```
reupen: Could be just `if self.instance == global_headquarters`?

datahub/company/serializers.py
```diff
@@ -264,6 +270,28 @@ class CompanySerializer(PermittedFieldsModelSerializer):
...
+    def validate_global_headquarters(self, global_headquarters):
+        """Ensure that global headquarters is global headquarters and it is not pointing
+        at the model itself.
+        """
+        if global_headquarters:
+            # checks if global_headquarters is global_headquarters
+            if (
+                global_headquarters.headquarter_type
+                and str(global_headquarters.headquarter_type.id) != HeadquarterType.ghq.value.id
+            ) or not global_headquarters.headquarter_type:
```
reupen: What about just `global_headquarters.headquarter_type_id != UUID(HeadquarterType.ghq.value.id)` instead of both conditions?

reupen:
> Hm OK. In the meantime it would be good to notify something when that happens so that we can get an idea of the problem without forgetting it.

I agree with @marcofucci here, but I'd say it should just be an error until we decide to do something else. Otherwise we'll end up with objects in an inconsistent state.

elcct:
@reupen @marcofucci so I talked with the team today and we decided to go with an error message when user attempts to change global hq to non global hq when it has subsidiaries. Validation is in the last commit.

datahub/company/serializers.py
elcct: Good idea. Fixed.

datahub/company/serializers.py
elcct: Also good idea. Fixed :)

datahub/core/validators.py
```diff
@@ -3,7 +3,6 @@
 from functools import partial
 from operator import eq
 from typing import Any, Callable, Sequence
-
```
reupen: Why was this removed? :slightly_smiling_face:

datahub/company/views.py
```diff
@@ -48,7 +48,7 @@ class CompanyViewSet(ArchivableViewSetMixin, CoreViewSetV3):
     ).prefetch_related(
         Prefetch('contacts', queryset=get_contact_queryset()),
         Prefetch('investor_investment_projects', queryset=get_slim_investment_project_queryset()),
-        'children',
+        'subsidiaries',
```
reupen: I think we can remove this now since it's not in the response?

datahub/core/validators.py
elcct: That is weird :) Maybe I had focus on wrong window for a moment when deleting. Fixed.

datahub/company/views.py
elcct: Of course. Thanks for spotting that! Fixed.

PR #838, opened 2018-02-22
Register models with reversion via decorator

marcofucci:
Before we were relying on the `reversion.admin.VersionAdmin` logic which auto-registers models automatically.
This was not very safe as we are using reversion not only in the admin but in many other places (e.g. endpoints) as well.

This removes the registration via admin and replaces it with explicitly registering the models with reversion.

As a result, `BaseModelVersionAdmin` and `ConfigurableVersionAdmin` are not needed any more.

### DIFF
--- a/datahub/core/reversion.py
+++ b/datahub/core/reversion.py
@@ -1,0 +1,34 @@
+import reversion
+
+EXCLUDED_BASE_MODEL_FIELDS = ('created_on', 'created_by', 'modified_on', 'modified_by')
+
+
+def register_base_model(extra_exclude=None, **kwargs):
+    """
+    Shortcut to reversion.register() which excludes some fields defined on BaseModel
+    by default.
+
+    These aren't particularly useful to save in django-reversion versions because
+    created_by/created_on will not change between versions, and modified_on/modified_by
+    is tracked by django-reversion separately in revisions.
+
+    :param extra_exclude: list of "extra" fields to exclude along with the default ones
+    :param kwargs: same as the reversion.register ones
+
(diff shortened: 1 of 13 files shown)


datahub/core/reversion.py
```diff
@@ -0,0 +1,29 @@
...
+    Note: if you pass `exclude` you override the default excluded fields. You cannot use
+        exclude and extra_exclude at the same time as they are mutual exclusive.
+    """
+    assert 'exclude' not in kwargs or not extra_exclude, (
+        "You can't pass in extra_exclude and exclude at the same time."
+    )
+
+    if 'exclude' not in kwargs:
+        kwargs['exclude'] = (
+            'created_on', 'created_by', 'modified_on', 'modified_by',
```
reupen: Could put these in a named constant?

datahub/core/reversion.py
```diff
@@ -0,0 +1,29 @@
...
+    Shortcut to reversion.register() which excludes some fields defined on BaseModel
+    by default.
+
+    These aren't particularly useful to save in django-reversion versions because
+    created_by/created_on will not change between versions, and modified_on/modified_by
+    is tracked by django-reversion separately in revisions.
+
+    :param extra_exclude: list of "extra" fields to exclude along with the default ones
+    :param kwargs: same as the reversion.register ones
+
```
reupen: Probably worth adding a note that saying that if this is removed from a model, VersionAdmin must be removed as well otherwise the model will still get registered?

marcofucci: done below

PR #840, opened 2018-02-23
[MI-1413] Add support for Policy Interaction Kind

r4vi:
Issue number: MI-1413

### Description of change

Adds a new Interaction kind called 'policy', the frontend will be updated to add support for creating these policy interactions.

### Checklist

* [x] ~Have any relevant search models been updated?~
* [x] ~Have any relevant select-/prefetch-related field lists in the views and search apps been updated?~
* [x] ~Has the admin site been updated (for new models, fields etc.)?~
* [x] ~Has the README been updated (if needed)?~

### DIFF
--- a/datahub/core/test/test_validators.py
+++ b/datahub/core/test/test_validators.py
@@ -12,6 +12,7 @@ from datahub.core.validators import (
     ConditionalRule,
     EqualsRule,
     FieldAndError,
+    InRule,
     OperatorRule,
     RequiredUnlessAlreadyBlankValidator,
     RulesBasedValidator,
@@ -191,6 +192,17 @@ def test_equals_rule(data, field, test_value, res):
     """Tests ValidationCondition for various cases."""
     combiner = Mock(spec_set=DataCombiner, __getitem__=lambda self, field_: data[field_])
     condition = EqualsRule(field, test_value)
+    assert condition(combiner) == res
+
+
+@pytest.mark.parametrize('data,field,test_value,res', (
+    ({'colour': 'red'}, 'colour', ['red', 'green'], True),
+    ({'colour': 'red'}, 'colour', ['blue', 'green'], False),
+))
+def test_in_rule(data, field, test_value, res):
+    """Tests InRule for various cases."""
+    combiner = Mock(spec_set=DataCombiner, __getitem__=lambda self, field_: data[field_])
+    condition = InRule(field, test_value)
     assert condition(combiner) == res
 
 
--- a/datahub/interaction/migrations/0008_add_interaction_kind.py
+++ b/datahub/interaction/migrations/0008_add_interaction_kind.py
@@ -15,7 +15,11 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='interaction',
             name='kind',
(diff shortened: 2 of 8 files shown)


r4vi:
@jim68000 @reupen I think this is all we need to do on the backend

datahub/interaction/models.py
```diff
@@ -74,6 +74,7 @@ class Interaction(BaseModel):
     KINDS = Choices(
         ('interaction', 'Interaction'),
         ('service_delivery', 'Service delivery'),
+        ('policy', 'Policy Interaction'),
```
reupen: Minor thing but we generally use sentence case for titles etc. (so it would be `Policy interaction` in this case).

r4vi:
@reupen thanks for the comments, I've made a fixup commit if you think it's fine I'll rebase

datahub/interaction/migrations/0008_add_interaction_kind.py
```diff
@@ -15,7 +15,11 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='interaction',
             name='kind',
-            field=models.CharField(choices=[('interaction', 'Interaction'), ('service_delivery', 'Service delivery')], default='interaction', max_length=255),
+            field=models.CharField(choices=[
```
reupen: Sorry, this migration was squashed into `0001_squashed_0019_rename_default_permissions` but the old migrations haven't been removed yet. So it'll need to be changed in `0001_squashed_0019_rename_default_permissions`. If you run `./manage.py makemigrations` afterwards, it should say there are no changes.

reupen:
You might want to review the validation rules on InteractionSerializer as well as they only handle the two existing interaction kinds. But you could do that separately if you want.

r4vi:
@reupen thanks for the advice, I'll do the changes in the serializer in this PR too

r4vi:
@reupen 
* we want to treat `policy` interactions the same as `interaction` for now so I've updated the serializer to match `interaction`. Added a new InRule and tests for it
* I've updated the right migration file now :ok_hand:

datahub/core/test/test_validators.py
```diff
@@ -194,6 +195,17 @@ def test_equals_rule(data, field, test_value, res):
     assert condition(combiner) == res
 
 
+@pytest.mark.parametrize('data,field,test_value,res', (
+    ({'colour': 'red'}, 'colour', ['red', 'green'], True),
+    ({'colour': 'red'}, 'colour', ['blue', 'green'], False),
+))
+def test_in_rule(data, field, test_value, res):
+    """Tests ValidationCondition for various cases."""
```
reupen: The docstring is wrong here (but so are the ones in the preceding functions :grimacing: )

reupen:
@r4vi That looks good. But would suggest adding some tests using that kind with the views and/or paramet(e)rising some of the existing tests.

r4vi:
@reupen https://github.com/uktrade/data-hub-leeloo/pull/840/commits/9f3815901fd97473e86f21fdd13e4246a6408aef paramertises some of the existing tests

Issue #770, opened 2018-01-16
Update flake8-import-order to 0.16


reupen:
The new version is stricter about where line breaks are in imports, and generates errors mostly in OMIS and search code.

@marcofucci @elcct will leave this for you both to look at.

reupen:
Fixed by https://github.com/uktrade/data-hub-leeloo/pull/843.

PR #841, opened 2018-02-23
Re-group imports

marcofucci:
This re-groups imports for OMIS modules to make them conform to the cryptography style.

Issue number: #770


marcofucci:
@elcct I haven't touched the search/investment ones, they are the only ones to fix before upgrading.

PR #842, opened 2018-02-23
Add versioning tests

marcofucci:
This adds tests for reversion versions created when interacting with the endpoints dealing with the following models:
- company
- contact
- adviser
- interaction
- event
- investment project
- investment project team member

### DIFF
--- a/datahub/company/test/test_contact_views.py
+++ b/datahub/company/test/test_contact_views.py
@@ -8,7 +8,9 @@ from rest_framework import status
 from rest_framework.reverse import reverse
 from reversion.models import Version
 
+from datahub.company.models import Contact
 from datahub.core import constants
+from datahub.core.reversion import EXCLUDED_BASE_MODEL_FIELDS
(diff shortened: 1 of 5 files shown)


datahub/company/test/test_contact_views.py
```diff
@@ -681,6 +683,149 @@ def test_filter_by_company(self):
...
+        )
+
+        assert response.status_code == status.HTTP_201_CREATED
+        assert response.data['first_name'] == 'Oratio'
+        assert response.data['last_name'] == 'Nelson'
+
+        contact = Contact.objects.get(pk=response.data['id'])
+
+        # check version created
+        assert Version.objects.get_for_object(contact).count() == 1
```
reupen: Version.objects.get_for_object(contact) isn't cached is it? So you might want to call it just once to avoid two queries.

reupen: Ah, of course it doesn't make a difference to the number of queries. Still it might be nice to assign the queryset to a variable.

PR #844, opened 2018-02-26
Switch to datahub.core.reversion.NonAtomicRevisionMiddleware

marcofucci:
This replaces `reversion.middleware.RevisionMiddleware` with `datahub.core.reversion.NonAtomicRevisionMiddleware` which sets `atomic = False` and lets the atomicity of the transaction be determined by the django logic instead.


marcofucci:
I do prefer the mixin or the context manager to be honest, this middleware is way too magic, with lots of assumptions and unnecessary logic for requests that don't need versioning.

reupen:
Could switch to those if you think it's straightforward enough.

PR #845, opened 2018-02-27
Add headquarter_type and global_headquarters filters.

elcct:
Issue number: DH-1569 DH-1574

### Description of change

This adds two search filters for company collection - headquarter_type and global_headquarters.

It introduces new serializer type - `NullStringUUIDField` that allows "null-ish" value to be passed in place of UUID. Elasticsearch query generation is modified so that when value is null, then "exists" query is being used (as Elasticsearch don't keep null columns). This is to simplify the filter query for `headquarters_type` as we sometimes want to find companies that have None, ehq, ukhq types assigned.

### Checklist


datahub/search/serializers.py
```diff
@@ -53,6 +55,34 @@ def to_internal_value(self, data):
...
+
+    We can't use UUID in ES queries, that's why we need to convert them back to string.
+    If input value is null-ish, we return None
+    """
+
+    NULL_VALUES = {'n', 'N', 'null', 'Null', 'NULL', '', None, 'None'}
+
+    def to_representation(self, value):
+        """Converts string to UUID or returns None if null-ish."""
+        if value in self.NULL_VALUES:
```
reupen: Shouldn't to_representation() be returning a string (or None)?

elcct: Good point. What I was thinking? :) Fixed.

PR #847, opened 2018-02-28
Add OMIS create payment session endpoint

marcofucci:
This adds a new endpoint to create a payment gateway session for an OMIS order.
Behind the scenes, it calls GOV.UK Pay and stores the related metadata in django.

### DIFF
--- a/datahub/omis/payment/models.py
+++ b/datahub/omis/payment/models.py
@@ -50,6 +50,15 @@ class PaymentGatewaySession(BaseModel):
         """
         return PayClient().get_payment_by_id(self.govuk_payment_id)
 
+    def get_payment_url(self):
+        """
+        :returns: the GOV.UK Pay payment url to redirect the users to complete the payment
+
+        :raises GOVUKPayAPIException: if there is a problem with GOV.UK Pay
+        """
+        next_url = self._get_payment_from_govuk_pay()['_links']['next_url'] or {}
+        return next_url.get('href', '')
+
     def is_finished(self):
         """
         :returns: True if this payment gateway session is in a finished status
--- a/datahub/omis/payment/views.py
+++ b/datahub/omis/payment/views.py
@@ -3,10 +3,12 @@ from rest_framework import status
 from rest_framework.response import Response
 
 from datahub.oauth.scopes import Scope
+from datahub.omis.core.exceptions import Conflict
+from datahub.omis.order.constants import OrderStatus
 from datahub.omis.order.models import Order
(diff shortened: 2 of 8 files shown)


datahub/omis/payment/models.py
```diff
@@ -50,6 +50,15 @@ def _get_payment_from_govuk_pay(self):
         """
         return PayClient().get_payment_by_id(self.govuk_payment_id)
 
+    def get_payment_url(self):
+        """
+        :returns: the GOV.UK Pay payment url to redirect the users to complete the payment
+
+        :raises GOVUKPayAPIException: if there is a problem with GOV.UK Pay
+        """
+        next_url = self._get_payment_from_govuk_pay()['_links']['next_url'] or {}
```
elcct: It seems like it is possible to receive data from `_get_payment_from_govuk_pay` so that `_links` key or `next_url` wouldn't exist and therefore giving `KeyError` instead of `GOVUKPayAPIException`.

marcofucci: At the moment, the response data always returns `next_url` with value:
- `None` or
- a dict of type
```
    "next_url": {
      "href": "https://www-integration-2.pymnt.uk/secure/bb0a272c-8eaf-468d-b3xf-ae5e000d2231",
      "method": "GET"
    },
```

If `_links` or `next_url` doesn't exist in the response than that should be treated as an uncaught exception and it would be useful to see the stack trace or the `KeyError`.

I think this is what we want.

datahub/omis/payment/views.py
```diff
@@ -50,3 +52,43 @@ class PublicPaymentViewSet(BasePaymentViewSet):
...
+    )
+
+    queryset = PaymentGatewaySession.objects.all()
+    serializer_class = PaymentGatewaySessionSerializer
+
+    def get_queryset(self):
+        """
+        :returns: the queryset with session gateway payments for the order.
+        """
+        return super().get_queryset().filter(order=self.get_order())
```
reupen: Just wondering how the payment status will get updated. Are you going to rely on Celery for that?

datahub/omis/payment/models.py
elcct: Ok fair enough

datahub/omis/payment/views.py
marcofucci: In two ways:
1. when the frontend calls an extra endpoint to check the status on the session (coming soon)
2. celery checking the ongoing payments that didn't manage to get refreshed by point 1 (eg. session abandoned or exception in our code)

GOV.UK Pay will always be the source of truth.

PR #851, opened 2018-03-01
Add script to startup for UAT

web-bert:
Add a script to help FE devs start leeloo for automated UAT

### DIFF
--- a/start-uat.sh
+++ b/start-uat.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+clear;
+docker-compose stop;
+docker-compose rm -f;
+docker-compose build \
+  && docker-compose run --publish 8000:8000 leeloo bash setup-uat.sh;


web-bert:
Just having a chat with the FE devs and we should have a script to easily start leeloo for UAT as we currently all do it our own way

start-uat.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+clear;
+docker-compose stop;
+docker-compose rm -f;
```
reupen: Are those two commands better than just `docker-compose down`?

start-uat.sh
```diff
@@ -0,0 +1,7 @@
...
+docker-compose build \
+  && docker-compose run --publish 8000:8000 leeloo bash setup-uat.sh;
```
reupen: Not sure why these are joined with && (as opposed to being on separate lines) – if it's about exit codes you could use `set -e` (or even `set -xe`) at the start of the script.

start-uat.sh
web-bert: Nope, nice improvement!

start-uat.sh
web-bert: I just took these from some bash commands I had, updating...

web-bert:
Updated

PR #852, opened 2018-03-02
Additional validation of headquarter_type field.

elcct:
Issue number: DH-1569

### Description of change

This adds additional validation of `headquarter_type`, so that when a company is linked to a global headquarters it can't become global headquarters itself.

### Checklist

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -213,7 +213,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
             'Subsidiaries have to be unlinked before changing headquarter type.',
         ),
         'subsidiary_cannot_be_a_global_headquarters': ugettext_lazy(
-            'Global headquarters cannot have assigned global headquarters.',
+            'A company cannot both be and have a global headquarters.',
         )
     }
 
@@ -281,7 +281,10 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         """Ensure that global headquarters doesn't have any subsidiaries before changing
         headquarter type.
         """
-        if self.instance and self.instance.headquarter_type != headquarter_type:
+        if (
+            self.instance
+            and self.instance.headquarter_type_id != getattr(headquarter_type, 'id', None)
+        ):
             if (
                 self.instance.headquarter_type_id == UUID(HeadquarterType.ghq.value.id)
                 and self.instance.subsidiaries.count() > 0
@@ -290,7 +293,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
                     self.error_messages['global_headquarters_has_subsidiaries']
                 )
             if (
-                str(headquarter_type) == HeadquarterType.ghq.value.name
+                getattr(headquarter_type, 'id', None) == UUID(HeadquarterType.ghq.value.id)
                 and self.instance.global_headquarters is not None
             ):
                 raise serializers.ValidationError(
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -211,6 +211,9 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         ),
         'global_headquarters_has_subsidiaries': ugettext_lazy(
             'Subsidiaries have to be unlinked before changing headquarter type.',
+        ),
+        'subsidiary_cannot_be_a_global_headquarters': ugettext_lazy(
+            'Global headquarters cannot have assigned global headquarters.',
         )
     }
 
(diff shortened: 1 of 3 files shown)


datahub/company/serializers.py
```diff
@@ -211,6 +211,9 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         ),
         'global_headquarters_has_subsidiaries': ugettext_lazy(
             'Subsidiaries have to be unlinked before changing headquarter type.',
+        ),
+        'subsidiary_cannot_be_a_global_headquarters': ugettext_lazy(
+            'Global headquarters cannot have assigned global headquarters.',
```
reupen: Maybe something like 'A company cannot both be and have a global headquarters.'? Also would be nice for it to give some hint about how to overcome the error e.g. 'The company and its global headquarters must be unlinked first.' (I mean otherwise if you need to change which company is the GHQ it may feel like its impossible, when it just needs to be done in steps.)

datahub/company/test/test_company_views.py
```diff
@@ -559,6 +559,28 @@ def test_company_pointing_itself_as_global_headquarters(self):
         error = ['Global headquarters cannot point to itself.']
         assert response.data['global_headquarters'] == error
 
+    def test_subsidiary_becomes_a_global_headquarters(self):
+        """Tests if subsidiary can become a global headquarter."""
```
marcofucci: Instead of being an open statement, I'd rather state what you are expecting instead so that it's clear just by reading the test name and/or the docs.

Eg. instead of _Test if subsidiary can become a global headquarters)_ 
I'd write _"Test that a subsidiary cannot become a global headquarters."_

datahub/company/test/test_company_views.py
```diff
@@ -559,6 +559,28 @@ def test_company_pointing_itself_as_global_headquarters(self):
...
+        global_headquarters = CompanyFactory(
+            headquarter_type_id=HeadquarterType.ghq.value.id
+        )
+        company = CompanyFactory(
+            headquarter_type=None,
```
marcofucci: I would use `headquarter_type_id` as this might hide problems if we add `headquarter_type_id` to the factory later on.

datahub/company/serializers.py
```diff
@@ -286,6 +289,13 @@ def validate_headquarter_type(self, headquarter_type):
                 raise serializers.ValidationError(
                     self.error_messages['global_headquarters_has_subsidiaries']
                 )
+            if (
+                str(headquarter_type) == HeadquarterType.ghq.value.name
```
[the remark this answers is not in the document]
elcct: Good point. I have changed it now.

datahub/company/serializers.py
elcct: The wording for this validation isn't great. I like your idea so I changed to that.

datahub/company/test/test_company_views.py
elcct: Good point. I have renamed the tests and updated comments where applicable.

datahub/company/test/test_company_views.py
elcct: Good catch. Fixed.

datahub/company/serializers.py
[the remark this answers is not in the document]
elcct: Changed HeadquarterType to be read only in admin.

datahub/company/serializers.py
```diff
@@ -278,14 +281,24 @@ def validate_headquarter_type(self, headquarter_type):
         """Ensure that global headquarters doesn't have any subsidiaries before changing
         headquarter type.
         """
-        if self.instance and self.instance.headquarter_type != headquarter_type:
+        if (
```
reupen: I'm finding this function a bit hard on the :eyes: . How about splitting it up into multiple functions? Also might benefit from using a few variables to make the if statements a bit nicer.

datahub/company/test/test_company_views.py
```diff
@@ -543,8 +543,8 @@ def test_remove_global_headquarters_link(self):
 
         assert global_headquarters.subsidiaries.count() == 0
 
-    def test_company_pointing_itself_as_global_headquarters(self):
-        """Test if you can point company as its own global headquarters."""
+    def test_cannot_point_company_to_itself_as_global_headquarters(self):
+        """Test if you cannot point company as its own global headquarters."""
```
reupen: I think 'Test that' would be clearer than 'Test if' now.

elcct: Updated.

datahub/company/serializers.py
elcct: It is indeed hard to read. I have split it into couple of functions.

PR #853, opened 2018-03-02
Throttle OMIS create gateway session endpoint

marcofucci:
As the create gateway session endpoint can be triggered by a user action (e.g. user clicks on a pay button), we want to disallow clients from hitting the API often enough to create issues so we limit the number to max x per minute.

### DIFF
--- a/datahub/core/test/test_throttling.py
+++ b/datahub/core/test/test_throttling.py
@@ -1,0 +1,99 @@
+import time
+
+import pytest
+from freezegun import freeze_time
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.test import APIRequestFactory
+from rest_framework.views import APIView
+
+from ..throttling import PathRateThrottle
+
+
+class Path3SecRateThrottle(PathRateThrottle):
+    """PathRateThrottle with a specific rate."""
+
+    scope = 'path'
+    rate = '3/sec'
+
+    def __init__(self, *args, **kwargs):
+        """
+        Re-define self.timer to allow classes to mock time.time if needed.
+        This is because DRF sets this property on the class and it doesn't give
+        opportunities for tests to override it.
+        """
+        super(Path3SecRateThrottle, self).__init__(*args, **kwargs)
+        self.timer = time.time
+
+
+class MockView(APIView):
+    """Simple APIView to test PathRateThrottle."""
+
+    authentication_classes = ()
+    permission_classes = ()
+    throttle_classes = (Path3SecRateThrottle,)
+
+    def get(self, request):
+        """Simple implementatin of the GET method."""
+        return Response('foo')
+
+
+@pytest.mark.usefixtures('local_memory_cache')
+class TestPathRateThrottle:
+    """Tests for the PathRateThrottle class."""
+
+    @freeze_time('2018-03-01 00:00:00')
+    def test_requests_are_throttled(self):
+        """Test that the requests are throttled."""
+        factory = APIRequestFactory()
+
+        for dummy in range(4):
+            request = factory.get('/some-path/')
+            response = MockView.as_view()(request)
+        assert response.status_code == status.HTTP_429_TOO_MANY_REQUESTS
+
+    @freeze_time('2018-03-01 00:00:00')
+    def test_query_params_do_not_count(self):
+        """
+        Test that query params are ignored.
+        Eg. GET /some-path/ and GET /some-path/?param=value
+        are treated as the same.
+        """
+        factory = APIRequestFactory()
+
+        for index in range(4):
+            request = factory.get(f'/some-path/?param={index}')
+            response = MockView.as_view()(request)
+        assert response.status_code == status.HTTP_429_TOO_MANY_REQUESTS
+
+    @freeze_time('2018-03-01 00:00:00')
+    def test_case_does_not_count(self):
+        """
+        Test that the match is case insensitive..
+        Eg. GET /some-path/ and GET /some-Path/
+        are treated as the same.
+        """
+        factory = APIRequestFactory()
+
+        for path in ['path', 'Path', 'pAth', 'paTh']:
+            request = factory.get(f'/some-{path}/')
+            response = MockView.as_view()(request)
+        assert response.status_code == status.HTTP_429_TOO_MANY_REQUESTS
+
+    @freeze_time('2018-03-01 00:00:00')
+    def test_throttling_is_per_path(self):
+        """
+        Test that throttling is per path.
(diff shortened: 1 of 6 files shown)


datahub/core/throttling.py
```diff
@@ -0,0 +1,28 @@
...
+    E.g. /user/<user-token>/forgot-password/
+
+    Note: query params are ignored.
+    """
+
+    def get_cache_key(self, request, view):
+        """
+        :returns: cache key constructed from the request path
+        """
+        ident_hash = hashlib.md5(request.path.lower().encode('utf-8'))
```
reupen: It's not obvious to me why the hashing is needed, so you might want to add a comment if you think it's needed.

datahub/core/test/test_throttling.py
```diff
@@ -0,0 +1,81 @@
...
+    authentication_classes = ()
+    permission_classes = ()
+    throttle_classes = (Path3SecRateThrottle,)
+
+    def get(self, request):
+        """Simple implementatin of the GET method."""
+        return Response('foo')
+
+
+class TestPathRateThrottle:
```
reupen: Shouldn't we use freeze_time (or similar) for these?

datahub/omis/payment/test/views/test_public_payment_gateway_sessions.py
```diff
@@ -17,6 +17,12 @@
 class TestPublicCreatePaymentGatewaySession(APITestMixin):
     """Public create payment gateway session test case."""
 
+    @pytest.fixture(autouse=True)
+    def disable_payment_throttle_rate(self, settings):
+        """Disable the throttling for all the tests in this class."""
+        drf_throttle_rates = settings.REST_FRAMEWORK['DEFAULT_THROTTLE_RATES']
+        drf_throttle_rates['payment_gateway_session.create'] = '100/hour'
```
reupen: Why don't we patch the dict instead? Also '100/hour' isn't quite the same thing as disabling it, is it?

datahub/omis/payment/test/views/test_public_payment_gateway_sessions.py
```diff
@@ -447,3 +453,54 @@ def test_403_if_scope_not_allowed(self, scope):
         )
         response = client.post(url, format='json')
         assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_429_if_too_many_requests_made(self, requests_stubber, settings):
+        """Test that the throttling for the create endpoint works if its rate is set."""
+        drf_throttle_rates = settings.REST_FRAMEWORK['DEFAULT_THROTTLE_RATES']
+        drf_throttle_rates['payment_gateway_session.create'] = '3/sec'
```
reupen: Same comments as earlier about freeze_time and patching the dict. (monkeypatch has setitem, there's patch.dict() as well.)

datahub/core/test/test_throttling.py
marcofucci: I'm trying to be consistent with the DRF code https://github.com/encode/django-rest-framework/blob/master/tests/test_throttling.py#L71

Also, `freeze_time` freezes the time so the DRF timer wouldn't work.

reupen: It looks like it's relying on the time advancing at a certain pace, right? That seems wrong to me and it won't work when e.g. stepping through the test in a debugger.

marcofucci: True, we can do better than the DRF

datahub/core/throttling.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Note: query params are ignored.
+    """
+
+    def get_cache_key(self, request, view):
+        """
+        :returns: cache key constructed from the request path
+        """
+        # we hash the path to have a deterministic length (32 chars)
+        ident_hash = hashlib.md5(request.path.lower().encode('utf-8'))
```
reupen: I know it shouldn't make much difference for our purposes, but why don't we just go for SHA-256 unless we have a good reason to use MD5?

datahub/core/test/test_throttling.py
marcofucci: I've added `freeze_time` to those tests. In theory I should really increase the time before each requests but I don't want to overcomplicate the tests without any added value.

What do you think?

reupen: It seems reasonable to me :+1:

datahub/core/throttling.py
marcofucci: Well, smaller keys are nicer in this case, don't you think? We don't need 256, 32 is enough

reupen: Don't think length is a reason as you can just truncate SHA-256? And it's 64 chars vs 32 (rather than 256 vs 32).

marcofucci: oh damn, did I really write 256? 🤦‍♂️ 
I'm not sure about truncating the hash but I'm easy as it's not really a bit deal...
SHA256 it is

elcct:
Just thinking if that shouldn't be a responsibility of a load balancer or gateway in front of the backend?

datahub/core/throttling.py
reupen: Truncating is OK. if you fancy a read there's https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf (section 7) and https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-107r1.pdf (section 5.1) or https://crypto.stackexchange.com/questions/9435/is-truncating-a-sha512-hash-to-the-first-160-bits-as-secure-as-using-sha1.

But yeah I'm sure we can cope with 64-char keys in Redis.

marcofucci: Yes I know but really at that point we'd rather use MD5 instead.

datahub/core/throttling.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Note: query params are ignored.
+    """
+
+    def get_cache_key(self, request, view):
+        """
+        :returns: cache key constructed from the request path
+        """
+        # we hash the path to have a deterministic length (32 chars)
+        ident_hash = hashlib.sha256(request.path.lower().encode('utf-8'))
```
reupen: > request.path.lower()

Aren't our paths case sensitive?

marcofucci: yeah but this class is generic and I'm not sure I like the idea of being case sensitive, it could introduce weaknesses.

marcofucci:
As discussed offline, in this instance keeping this particular logic here is okay as it's implementation specific and does not depend on the underlaying infrastructure.

PR #859, opened 2018-03-06
Configure redis and cache

marcofucci:
This adds redis as optional dependency and sets up the cache if the REDIS_URL env var is set.


docker-compose.yml
```diff
@@ -45,3 +46,9 @@ services:
     restart: always
     ports:
       - "9200:9200"
+
+  redis:
+    image: redis:3.2
+    restart: always
+    ports:
+      - "6379:6379"
```
reupen: Do we need to bind the port on the host? We don't do it for postgres above, but we do for ES :/

config/settings/test.py
```diff
@@ -22,6 +22,8 @@
 DOCUMENT_BUCKET='test-bucket'
 AV_SERVICE_URL='http://av-service/'
 
+REDIS_URL=None
```
reupen: Do we need to set CACHES here as well?

docker-compose.yml
marcofucci: We don't need to, I did it just because we are doing it for ES but I'm going to remove this line

config/settings/test.py
marcofucci: yes very true, shall we use `DummyCache` in tests?

reupen: Not sure. Makes sense generally, but don't the tests for the rate-limiting stuff need a working cache?

marcofucci: We have two options:
1. we use `DummyCache` and configure the cache to use `LocMemCache` only when testing caching stuff (e.g. throttling)
2. we use `LocMemCache` for all tests

reupen: 1. sounds better to me if it's straightforward.

marcofucci: I agree, I'll see how painful it is

PR #862, opened 2018-03-08
Enhance references in OMIS payments

marcofucci:
This:
- adds the order reference to the description of a GOV.UK payment
- uses the order reference and the first 8 chars of the payment gateway session id as GOV.UK payment reference. This is because it's shown to the user at the end of the transaction and it could be used to get to the record either from the GOV.UK Pay admin area and/or in our system
- stores the GOV.UK payment reference on the OMIS payment record as `transaction_reference`

### DIFF
--- a/datahub/omis/payment/utils.py
+++ b/datahub/omis/payment/utils.py
@@ -20,8 +20,9 @@ def trasform_govuk_payment_to_omis_payment_data(govuk_payment):
     return {
         'amount': govuk_payment['amount'],
         'method': PaymentMethod.card,
+        'transaction_reference': govuk_payment['reference'],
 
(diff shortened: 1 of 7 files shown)


datahub/omis/payment/utils.py
```diff
@@ -20,6 +20,7 @@ def trasform_govuk_payment_to_omis_payment_data(govuk_payment):
     return {
         'amount': govuk_payment['amount'],
         'method': PaymentMethod.card,
+        'transaction_reference': govuk_payment['reference'],
 
         # unfortunatelly GOV.UK Pay doesn't tell us when the payment actually happened
```
reupen: Just spotted a little typo here in the comment :slightly_smiling_face:

marcofucci: 🙄

PR #864, opened 2018-03-09
Add sector hierarchy

reupen:
Issue number: DH-1585

### Description of change

Adds a hierarchy to sectors using django-mptt.

The migration reloads the sectors with the hierarchy in place. (Note that sectors themselves won't be modified in production while this work is ongoing.) The name column has been removed from the model, but remains in the database for now so that we can deploy without any downtime.

The main complexity here is around generating the full name/path of each sector. In theory, it's possible to do so in the same query using StringAgg (string_agg()), but getting that working with the ORM and DRF is not straightforward. So a relatively efficient approach (taking into account pre-fetching) of walking up parent attributes was used instead. (Another option might be to save the generated names in the database, but we'd have to regenerate them when part of the tree changes.)

The metadata endpoint for sectors has been extended to include a few more fields. Otherwise, the API remains unchanged (i.e. these changes are backwards compatible).

There will be a follow-up PR to add a new filter to search making use of the hierarchy.

### Checklist


datahub/metadata/models.py
```diff
@@ -11,10 +16,47 @@ class BusinessType(BaseConstantModel):
...
 
-class Sector(BaseConstantModel):
+class Sector(MPTTModel, DisableableModel):
     """Company sector."""
 
-    pass
+    PATH_SEPARATOR = ' : '
+
+    id = models.UUIDField(primary_key=True, default=uuid4)
+    node_name = models.TextField(verbose_name='name')
```
marcofucci: why `verbose_name='name`? I find it a bit confusing as you already have `name` as prop.

datahub/metadata/models.py
```diff
@@ -11,10 +16,47 @@ class BusinessType(BaseConstantModel):
...
+        ancestors = self._get_ancestors_using_parent(include_self=True)
+        return self.PATH_SEPARATOR.join(ancestor.node_name for ancestor in ancestors)
+
+    def _get_ancestors_using_parent(self, include_self=False):
+        ancestors = [self] if include_self else []
+        obj = self
+        while obj.parent:
+            obj = obj.parent
+            if obj in ancestors:
+                raise ObjectStateError('Recursive sector hierarchy encountered.')
```
marcofucci: Why do we need a specific error class? You are not catching it anywhere and I can't think of a reason why you would want it. I'd use the generic datahub one instead.

datahub/metadata/migrations/0002_add_sector_hierarchy.py
```diff
@@ -0,0 +1,100 @@
...
+import mptt.managers
+from django.db import migrations, models
+from django.core.management import call_command
+
+
+def load_sectors(apps, schema_editor):
+    call_command(
+        'loaddata',
+        PurePath(__file__).parent / '0002_add_sector_hierarchy.yaml'
+    )
```
marcofucci: what about potential objects created using the django admin? Should we check and raise an exception if we find an object in the db that's not in the fixture before applying the migration?

datahub/metadata/migrations/0002_add_sector_hierarchy.py
```diff
@@ -0,0 +1,100 @@
...
+    dependencies = [
+        ('metadata', '0001_squashed_0011_add_default_id_for_metadata'),
+    ]
+
+    # We're temporarily leaving the name column in the database to avoid downtime and allow the
+    # migration to be reversed.
+    #
+    # Once these changes have been released, the column can be removed from the database using
+    # another migration.
+    state_operations = [
```
marcofucci: This is OK as long as we delete the column in another migration immediately after this release so that we don't forget and leave it system in this state for too long.
Another solution is to just rename the field to something like `legacy_name` and keep it in the db and the model at the same time if you're planning to keep it for a bit longer.

requirements.txt
```diff
@@ -5,36 +5,37 @@
...
-appnope==0.1.0            # via ipython
 attrs==17.4.0             # via pytest
 aws-requests-auth==0.4.1
 boto3==1.6.3
-botocore==1.9.3           # via boto3, s3transfer
+botocore==1.9.6           # via boto3, s3transfer
 certifi==2018.1.18        # via requests
-chardet==3.0.4            # via requests
-click==6.7                # via pip-tools
+chardet==3.0.4            # via chardet, requests
```
marcofucci: why your pip-compile adds the same library to the `via` docs?

reupen: Are you suggesting that we switch to pipenv? :slightly_smiling_face: 

Not sure though, it was bugging me before but it still happens in a clean virtual environment. It must be a bug.

datahub/metadata/migrations/0002_add_sector_hierarchy.py
reupen: It was more because there'll be a short period of time when we deploy when both the old and new version of the app are running (just before the routes are switched). That old version of the app might try and access the column, so this was just to minimise risk. (Unfortunately, Django doesn't seem to have a way to defer a column by default, in which case I could've left it in the model and just deferred it.)

datahub/metadata/migrations/0002_add_sector_hierarchy.py
reupen: Could do. What would you be trying to avoid?

datahub/metadata/models.py
reupen: I suppose I thought it would be better for Sentry and also make the test more specific. But I can get rid of it :+1:

datahub/metadata/models.py
reupen: It's displayed in admin, and I don't think 'node name' will mean much to everyone using the admin site.

reupen:
> I'm not convinced by:
> - the name `node_name`, why a node? I'm not sure it should reveal the underlying tree implementation as well
> - `name` for the full path. Shouldn't name be `node_name` and something like `full_name` or `path` being thing `name` instead?

I did start off with path for name, but the problem is that we don't really want to break API compatibility and sectors are referenced in various places in the API (so it's not just the case of having a new version of the sectors endpoint). I also don't think using a pair of (id, path) for sectors when we use (id, name) for everything else would be particularly helpful.

We could call it path on the model and name in the API, but I'm not sure that would be particularly helpful.

As for node_name, I couldn't think of anything else :slightly_smiling_face:

Do you have any suggestions for how to improve the naming side of thing?

elcct:
I am happy with `node_name`. Perhaps it reveals what is the implementation beneath it, as mptt uses node terminology or that we are using tree structure to store it, but I am curious what could be alternative as well?

datahub/metadata/migrations/0002_add_sector_hierarchy.py
marcofucci: Confusion and potential unwanted bugs because of pre-existing objects. 
E.g. some existing records might exists and in logical conflict with the ones in the fixture (e.g. same name). 

The default values would be applied when migrating existing records so if you're sure it doesn't add too much confusion you can skip it I guess.

datahub/metadata/migrations/0002_add_sector_hierarchy.py
marcofucci: oh I see 👍

marcofucci:
> I also don't think using a pair of (id, path) for sectors when we use (id, name) for everything else would be particularly helpful.

Yeah that would be messy, let's keep `name` and change `node_name`.
How about, `local_name` or `part_name`?

PR #867, opened 2018-03-14
Add additional Investment Project fields and Stage log.

elcct:
Issue number: DH-1569 DH-1593 DH-1595

### Description of change

This adds additional fields for Investment Project and logging of `stage` field changes.
We don't need information about past changes. 

### Checklist


datahub/investment/models.py
```diff
@@ -337,6 +337,16 @@ def project_assurance_team(self):
         return None
 
 
+class IProjectSPIAbstract(models.Model):
+    """The SPI part of an investment project."""
+
+    class Meta:
+        abstract = True
+
+    project_arrived_in_triage = models.DateField(blank=True, null=True)
```
reupen: Perhaps this should end in `_on` or start with `date_`? It sounds like a boolean at the moment.

datahub/investment/signals.py
```diff
@@ -18,3 +20,26 @@ def project_post_save(sender, **kwargs):
...
+    """Creates a log of changes to stage field."""
+    instance = kwargs['instance']
+    created = kwargs['created']
+
+    created_on = None
+
+    if created:
+        created_on = instance.created_on
+    else:
+        last_stage = InvestmentProjectStageLog.objects.order_by('-created_on').first()
```
reupen: Is this missing a filter on `investment_project=instance`?

datahub/investment/signals.py
```diff
@@ -18,3 +20,26 @@ def project_post_save(sender, **kwargs):
...
     if created and not instance.cdms_project_code:
         InvestmentProjectCode.objects.create(project=instance)
+
+
+@receiver(post_save, sender=InvestmentProject, dispatch_uid='stage_log_project_post_save')
+def stage_log_project_post_save(sender, **kwargs):
+    """Creates a log of changes to stage field."""
+    instance = kwargs['instance']
+    created = kwargs['created']
+
```
reupen: We should probably check `kwargs['raw']` as well (do nothing if raw == True).

datahub/investment/test/test_views.py
```diff
@@ -1012,6 +1015,22 @@ def test_patch_project_success(self):
         assert len(response_data['client_contacts']) == 1
         assert response_data['client_contacts'][0]['id'] == str(new_contact.id)
 
+    def test_patch_spi_fields(self):
+        """Test updating a project with SPI fields."""
+        project = InvestmentProjectFactory()
+        url = reverse('api-v3:investment:investment-item', kwargs={'pk': project.pk})
+        request_data = {
+            'project_arrived_in_triage': '2017-04-18T13:25:30.986208Z',
```
reupen: If they're `DateField`s why does this include the time?

datahub/investment/test/test_views.py
```diff
@@ -1239,6 +1323,15 @@ def test_revert_stage_to_verify_win(self):
         }
         assert response_data['status'] == 'ongoing'
 
+        project.refresh_from_db()
```
reupen: It looks a bit repetitive checking the stage log in all these tests.  What about some separate tests for it?

datahub/investment/serializers.py
```diff
@@ -213,6 +231,11 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     team_members = NestedIProjectTeamMemberSerializer(many=True, read_only=True)
     team_complete = serializers.SerializerMethodField()
 
+    # SPI fields
+    project_arrived_in_triage = serializers.DateTimeField(required=False, allow_null=True)
+    proposal_deadline = serializers.DateTimeField(required=False, allow_null=True)
+    stage_log = NestedInvestmentProjectStageLogSerializer(many=True, read_only=True)
```
reupen: I guess it should be small enough to include it here (rather than in a separate endpoint). Although, I don't see where we're sorting it by `-created_on`?

datahub/investment/signals.py
```diff
@@ -18,3 +20,26 @@ def project_post_save(sender, **kwargs):
     created = kwargs['created']
     if created and not instance.cdms_project_code:
         InvestmentProjectCode.objects.create(project=instance)
+
+
+@receiver(post_save, sender=InvestmentProject, dispatch_uid='stage_log_project_post_save')
+def stage_log_project_post_save(sender, **kwargs):
```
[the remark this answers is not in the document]
marcofucci: In any case, in general I tend to use signals only for unrelated, non-dependent logic. In your case stage logs are part of the app and the logic (you return them in your API as well as part of the usual response) so I wouldn't use signals for this.
I'm easy on this though as it's not my call to make.

datahub/investment/test/test_views.py
marcofucci: yeah one test (maybe parametrize) would be better

datahub/investment/signals.py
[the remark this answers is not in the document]
reupen: The stage can be changed from admin as well. So, however it's done, that needs to be covered off as well.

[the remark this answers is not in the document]
marcofucci: Should it be? That's potentially dangerous, isn't it? It's still possible to build the admin to allow that even if you wanted to keep this functionality.
Also, I can't believe that progressing a project doesn't trigger some other logic (e.g. sending notifications, changing the state of other models). 
I'd design the system as objects and messages instead of abstract records that get generally saved, changed and deleted.
That's a design choice though and if you want to keep it this way I won't disapprove it.

I just wanted to say that this kind of things make the code hard to follow and less obvious to other developers. I have no idea what's happening there.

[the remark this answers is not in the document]
elcct: I think the purpose of this is only to log any change to the stage field regardless of what's logic behind it. We have a timeline that shows progress of the project and the requirement is to show date alongside each stage.  
I could expand the comment to explain what it is for in more detail.

datahub/investment/models.py
```diff
@@ -337,6 +337,16 @@ def project_assurance_team(self):
         return None
 
 
+class IProjectSPIAbstract(models.Model):
+    """The SPI part of an investment project."""
```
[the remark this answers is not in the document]
elcct: SPI stands for Service Performance Indicators. I created this abstract class to follow the convention we have in the investment projects.

datahub/investment/serializers.py
elcct: I think sorting/transforming the data should be done on the client (for example getting latest date for each stage).

reupen: Since it's a log doesn't it make sense to have them in date order (whether ascending or descending)?

datahub/investment/models.py
elcct: Appened `_on` to that field. Thanks!

datahub/investment/signals.py
elcct: Oh yes, how could I forgot that... Fixed.

datahub/investment/signals.py
elcct: I have added that check now.

datahub/investment/test/test_views.py
elcct: True. I have updated the serialiser as well.

datahub/investment/test/test_views.py
elcct: I have removed that from those tests and added separate test to see if the log is being created.

datahub/investment/serializers.py
elcct: I think it won't hurt to add that, so I updated the serializer so that the log is being sorted by `created_on` in ascending order.

elcct:
@reupen added some test data.

datahub/investment/models.py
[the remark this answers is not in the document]
marcofucci: Can you write _Service Performance Indicator_ in the docs and explain briefly what that is?

datahub/investment/signals.py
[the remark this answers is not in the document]
marcofucci: That's exactly my point, it's dangerous to think about requirements individually and separately without having an overall code architecture in mind.

Let's stop here as I don't want to force things, I just wanted to share my opinions and concern.

datahub/investment/models.py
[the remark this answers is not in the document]
elcct: I have updated the comments.

datahub/investment/signals.py
```diff
@@ -18,3 +20,36 @@ def project_post_save(sender, **kwargs):
...
+    created_on = None
+
+    if created:
+        created_on = instance.created_on
+    else:
+        last_stage = InvestmentProjectStageLog.objects.filter(
+            investment_project_id=instance.pk
+        ).order_by('-created_on').first()
+
+        if str(last_stage.stage_id) != instance.stage_id:
```
reupen: This seems to assume that `last_stage` is not `None`, but for existing projects it will be None, won't it?

elcct: Yes indeed! Thanks, I will fix soon.

marcofucci:
@elcct didn't we move away from adding ticket numbers to commit titles? 
Ticket numbers in PRs / commit body should be enough.

datahub/investment/test/test_views.py
```diff
@@ -1239,6 +1258,41 @@ def test_revert_stage_to_verify_win(self):
         }
         assert response_data['status'] == 'ongoing'
 
+    def test_change_stage_log(self):
+        """Tests change of the project stage is being logged."""
```
reupen: I think there are more cases to test. I can think of three cases:

1. New project
1. Existing project, no existing log
1. Existing project, with existing log

elcct: I have added tests that cover those cases.

datahub/investment/signals.py
elcct: Fixed.

datahub/investment/signals.py
```diff
@@ -18,3 +20,36 @@ def project_post_save(sender, **kwargs):
...
+    created_on = None
+
+    if created:
+        created_on = instance.created_on
+    else:
+        last_stage = InvestmentProjectStageLog.objects.filter(
+            investment_project_id=instance.pk
+        ).order_by('-created_on').first()
+
+        if last_stage is None or str(last_stage.stage_id) != instance.stage_id:
```
reupen: It might be acceptable but for existing projects the first entry created will be wrong (as in the date won't be when it actually moved to that project in most cases). Do we know if that is OK? Can we flag the record somehow so that we know the date is wrong?

datahub/investment/test/test_views.py
```diff
@@ -1239,6 +1258,62 @@ def test_revert_stage_to_verify_win(self):
...
+
+        assert [
+            entry.stage.id for entry in project2.stage_log.order_by('created_on')
+        ] == [
+            uuid.UUID(constants.InvestmentProjectStage.assign_pm.value.id),
+            uuid.UUID(constants.InvestmentProjectStage.active.value.id),
+        ]
+
+    def test_change_stage_log_when_log_is_empty(self):
+        """Tests that stage is being logged for Investment Projects with empty log."""
```
reupen: This and the next test aren't testing any views. So they should go in a different module (rather than `test_views.py`).

datahub/investment/serializers.py
```diff
@@ -258,6 +281,15 @@ def get_team_complete(self, instance):
             instance=instance, fields=TEAM_FIELDS, next_stage=True
         )
 
+    def get_stage_log(self, instance):
```
reupen: :+1:

I suppose we could've set ordering on the model's Meta class as well, but that's perhaps a bit indirect.

datahub/investment/signals.py
elcct: I am not sure it will be wrong, as it will just state when new stage was set. Date for previous stages just will be unavailable. My understanding is that we don't have to display the data if it is unavailable and it is only important to be recorded for new projects.

datahub/investment/test/test_views.py
elcct: That is true. I have moved some tests to test_models.py and added another one here. Thanks!

datahub/investment/models.py
```diff
@@ -453,6 +467,26 @@ class Meta:
...
+    """
+
+    investment_project = models.ForeignKey(
+        InvestmentProject, on_delete=models.CASCADE, related_name='stage_log'
+    )
+    stage = models.ForeignKey(
+        'metadata.InvestmentProjectStage', on_delete=models.PROTECT,
+        related_name='+',
+    )
+    created_on = models.DateTimeField(null=True, blank=True)
```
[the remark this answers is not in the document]
elcct: I think we don't need both, so I removed them. Thanks!

datahub/investment/signals.py
reupen: It is potentially wrong if `last_stage` is None, because we don't know that the stage changed, no?

elcct: I understand now. I'll think about that.

elcct: I have removed the signal and update the log on save in the model instead. That way I could make a check for this case also created a test for that.

datahub/investment/signals.py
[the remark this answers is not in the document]
elcct: I have removed the signal and now do update on save in the model instead. That way it is easier to see if stage field has been changed and create the log accordingly.

datahub/investment/models.py
```diff
@@ -363,6 +377,39 @@ class InvestmentProject(ArchivableModel, IProjectAbstract,
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4)
 
+    def __init__(self, *args, **kwargs):
+        """Keep the original stage value so that we can see if it changes when saving."""
+        super().__init__(*args, **kwargs)
+        self.__stage_id = self.stage_id
```
reupen: I wonder if using the post_init signal will be cleaner than overriding the methods and using super calls. What do you think?

datahub/investment/serializers.py
marcofucci: yeah ordering on Meta class looks clearer.

datahub/investment/models.py
marcofucci: why don't you use a specific method like `def move_to_stage(self, new_stage):` and call that instead?

elcct: @reupen I think we use this pattern elsewhere. I think moving that to post_init signal would increase complexity.

@marcofucci how would that work with DRF?

marcofucci: oh you have just one save endpoint for everything 😞

datahub/investment/test/test_models.py
```diff
@@ -129,3 +131,48 @@ def test_associated_advisers_no_none():
...
     project = InvestmentProjectFactory(client_relationship_manager_id=None)
     assert None not in tuple(project.get_associated_advisers())
+
+
+def test_change_stage_log():
+    """Tests that stage is being logged for Investment Projects with empty log."""
+    project = InvestmentProjectFactory()
+    project.stage_id = constants.InvestmentProjectStage.assign_pm.value.id
+    project.save()
+    assert [
```
reupen: Why don't we use freeze_time and check that the created_on values are correct as well in these tests? It will also make it clearer why the stage_id is set in a separate step.

datahub/investment/test/test_views.py
```diff
@@ -1239,6 +1258,75 @@ def test_revert_stage_to_verify_win(self):
...
+                'id': constants.InvestmentProjectStage.active.value.id
+            }
+        }
+
+        response = self.api_client.patch(url, data=request_data, format='json')
+        assert response.status_code == status.HTTP_200_OK
+
+        response_data = response.json()
+        assert len(response_data['stage_log']) == 2
+        assert [
```
reupen: How about asserting on the entire contents of `response_data['stage_log']`?

datahub/investment/test/test_models.py
elcct: I have added checking for the date as well.

datahub/investment/test/test_views.py
elcct: Good idea. I have added a check of entire response.

datahub/investment/serializers.py
elcct: I have moved ordering to the Meta class.

datahub/investment/models.py
```diff
@@ -363,6 +377,39 @@ class InvestmentProject(ArchivableModel, IProjectAbstract,
...
+
+        This allows us to construct the timeline of changes to the stage field as
+        required for Service Performance Indicators (SPI).
+        """
+        created_on = None
+
+        if adding:
+            created_on = self.created_on
+        else:
+            # if there is no stage log, check if stage has been updated
```
marcofucci: > if there is no stage log
Was this an old comment? Doesn't seem true any longer, right?

datahub/investment/models.py
```diff
@@ -363,6 +377,39 @@ class InvestmentProject(ArchivableModel, IProjectAbstract,
...
+        super().save(*args, **kwargs)
+        self._update_stage_log(adding)
+
+    def _update_stage_log(self, adding):
+        """Creates a log of changes to stage field.
+
+        This allows us to construct the timeline of changes to the stage field as
+        required for Service Performance Indicators (SPI).
+        """
+        created_on = None
```
marcofucci: can you call it something more like `saved_on` or similar? I can see that `created_on` does not always mean `created on` in your code but also `modified on`, right?

datahub/investment/test/test_models.py
```diff
@@ -129,3 +134,72 @@ def test_associated_advisers_no_none():
     """Tests that get_associated_advisers() does not include None."""
     project = InvestmentProjectFactory(client_relationship_manager_id=None)
     assert None not in tuple(project.get_associated_advisers())
+
+
+def test_change_stage_log():
+    """Tests that stage is being logged for Investment Projects with empty log."""
+    dates = (
+        datetime(2017, 4, 28, 17, 35, tzinfo=utc),
+        datetime(2017, 4, 28, 17, 37, tzinfo=utc),
```
marcofucci: you can just use strings, e.g. `'2017-04-28 17:37'` as `freeze_time` understands that.

elcct: I am later on comparing that with model's `created_on` field which is datetime. That way I don't have to convert `created_on` to string, no?

datahub/investment/test/test_models.py
```diff
@@ -129,3 +134,72 @@ def test_associated_advisers_no_none():
...
+    assert [
+        (entry.stage.id, entry.created_on,) for entry in project.stage_log.order_by('created_on')
+    ] == [
+        (UUID(constants.InvestmentProjectStage.prospect.value.id), next(date_iter),),
+        (UUID(constants.InvestmentProjectStage.assign_pm.value.id), next(date_iter),)
+    ]
+
+
+def test_change_stage_log_when_log_is_empty():
+    """Tests that stage is being logged for Investment Projects with empty log."""
```
marcofucci: Not sure why you care about the log being empty?

datahub/investment/test/test_models.py
```diff
@@ -129,3 +134,72 @@ def test_associated_advisers_no_none():
...
+    """Tests that stage log is not created when there is no change to stage on save."""
+    project = InvestmentProjectFactory()
+    project.stage_log.all().delete()
+    # no change to the stage
+    project.save()
+    assert project.stage_log.count() == 0
+
+
+@freeze_time(datetime(2017, 4, 28, 17, 35, tzinfo=utc))
+def test_stage_log_added_when_investment_project_is_created():
```
marcofucci: Same here, not sure why you care about the log being empty?

datahub/investment/test/test_models.py
```diff
@@ -129,3 +134,72 @@ def test_associated_advisers_no_none():
...
+    date_iter = iter(dates)
+    next(date_iter)
+    assert [
+        (entry.stage.id, entry.created_on,) for entry in project.stage_log.all()
+    ] == [
+        (UUID(constants.InvestmentProjectStage.assign_pm.value.id), next(date_iter),)
+    ]
+
+
+def test_no_stage_log_when_log_is_empty_and_no_change_on_save():
```
marcofucci: Not sure why you care about the log being empty?

datahub/investment/models.py
elcct: I think in the context of `InvestmentProjectStageLog` it represents the date when that particular record has been created. We are not supposed to modify it. In my view it always means the same thing.

datahub/investment/test/test_models.py
elcct: To avoid incorrect data. So this one just checks that if stage has not changed then we also won't get stage log record. Since log is empty, one could think that the stage we are saving is a new state and create a log, but that would have been incorrect.

datahub/investment/models.py
marcofucci: OK, but if `created_on` refers to the log, the `adding` variable in my mind should refer to that log as well whilst it refers to the `project` instead.
So IMHO, `stage_changed_on` and `adding` are clearer or `created_on` and `project_added` are clear as well.
It's OK if you want to keep it this way though, it's not hard to understand what you mean after all.

datahub/investment/test/test_models.py
elcct: Because existing projects don't have stage log and I want to check if the feature will behave correctly when applied to such project. If there is no log and the stage actually has been modified then that should result in new log record.

marcofucci: That was your old logic, now you don't check for the existence of logs any more if I'm not wrong. this means that you always gonna create a log record when stage_id changes so you can test that no matter if logs are empty or not.

datahub/investment/models.py
elcct: Oh I thought you are referring to the column name. It makes sense to me now to change it. Sorry and thanks

datahub/investment/models.py
elcct: Removed. Good catch!

datahub/investment/test/test_models.py
elcct: Those tests lead to change in the logic, so that's why I left them. I am happy to remove them, but then what if for some reason change in the logic in the future wouldn't not acknowledge those edge cases and we will be getting incorrect data?

marcofucci: correct me if I'm wrong but:
1. be there's no log and the stage has been modified => a new log record is created
2. if there's a log and the stage has been modified => a new log record is created

Am I missing something?

elcct: Nope, you are not missing anything.

marcofucci: OK I still not follow but I'll hit approve and maybe if you want you can update the docs to expand a bit on it as all those tests now say the same _ Tests that stage is being logged for Investment Projects with empty log._

elcct: I see the naming and comment in this thing is not too great. I'll try to come up with something more readable. Thanks

elcct: I have removed that test and modified the other, so that it only checks if we are not creating the log if there is no change.

PR #868, opened 2018-03-14
Set up celery

marcofucci:
This:
- adds the celery library dependency
- sets up celery in the project
- updates the Procfile with production settings
- sets up sentry for celery
- updates docker-compose.yml to start celery automatically
- updates the README

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -1,1 +1,5 @@
+# This will make sure the app is always imported when
+# Django starts so that shared_task will use this app.
+from .celery import app as celery_app
 
+__all__ = ['celery_app']
--- a/config/celery.py
+++ b/config/celery.py
@@ -1,0 +1,33 @@
+import logging
+import os
+
+from celery import Celery
+from raven import Client
+from raven.contrib.celery import register_logger_signal, register_signal
+
+
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'config.settings.local')
+
+app = Celery('datahub')
+
+# Using a string here means the worker doesn't have to serialize
+# the configuration object to child processes.
+# - namespace='CELERY' means all celery-related configuration keys
+#   should have a `CELERY_` prefix.
+app.config_from_object('django.conf:settings', namespace='CELERY')
+
+# Load task modules from all registered Django app configs.
+app.autodiscover_tasks()
+
+# config sentry
+client = Client(
+    dsn=os.environ.get('DJANGO_SENTRY_DSN'),
+    environment=os.environ.get('SENTRY_ENVIRONMENT')
+)
+
+# register a custom filter to filter out duplicate logs
+register_logger_signal(client)
+
+# hook into the Celery error handler
+register_signal(client, ignore_expected=True)
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -8,6 +8,7 @@ https://docs.djangoproject.com/en/dev/ref/settings/
 """
 
 import environ
+import ssl
 
 from .companieshouse import *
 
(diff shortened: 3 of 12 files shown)


config/__init__.py
```diff
@@ -1 +1,7 @@
+from __future__ import absolute_import, unicode_literals
```
reupen: Do we need these?

config/celery.py
```diff
@@ -0,0 +1,45 @@
...
+    environment=os.environ.get('SENTRY_ENVIRONMENT')
+)
+
+# register a custom filter to filter out duplicate logs
+register_logger_signal(client)
+
+# The register_logger_signal function can also take an optional argument
+# `loglevel` which is the level used for the handler created.
+# Defaults to `logging.ERROR`
+register_logger_signal(client, loglevel=logging.ERROR)
```
reupen: Are we supposed to be calling register_logger_signal() and register_signal() twice?

docker-compose.yml
```diff
@@ -46,7 +45,11 @@ services:
     restart: always
     ports:
       - "9200:9200"
+    environment:
+      - ES_JAVA_OPTS=-Xms750m -Xmx750m
```
reupen: Can we use `ES_HEAP_SIZE`? https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html

config/celery.py
marcofucci: oops, thanks for spotting that, I was trying different things :/

config/__init__.py
marcofucci: no we don't

docker-compose.yml
marcofucci: it's apparently no longer supported:
```
es_1        | Error: encountered environment variables that are no longer supported
es_1        | Use jvm.options or ES_JAVA_OPTS to configure the JVM
es_1        | ES_HEAP_SIZE=750m: set -Xms750m and -Xmx750m in jvm.options or add "-Xms750m -Xmx750m" to ES_JAVA_OPTS
```

config/settings/common.py
```diff
@@ -227,6 +228,17 @@
         }
     }
 
+# CELERY (it does not understand rediss:// yet so extra work needed)
+if REDIS_URL:
+    CELERY_BROKER_URL = REDIS_URL.replace('rediss://', 'redis://')
+    CELERY_RESULT_BACKEND = CELERY_BROKER_URL
```
reupen: Should we be using the same Redis DB number for caching, the Celery broker and the Celery results back end?

marcofucci: Do you think we shouldn't? Are you thinking about key collision?

config/settings/common.py
```diff
@@ -227,6 +228,17 @@
...
+    if 'rediss://' in REDIS_URL:
+        CELERY_REDIS_BACKEND_USE_SSL = {
+            'ssl_cert_reqs': ssl.CERT_NONE
```
reupen: `ssl.CERT_NONE` seems a bit unfortunate, but looks like it wants CA certificates passed to it otherwise :disappointed:

config/settings/common.py
reupen: I guess the risk of key collision is low, but since Redis has 16 databases configured by default I'm not sure why we shouldn't use them. If we ever need to look at what's inside Redis won't it be clearer what we're looking at?

marcofucci: yep

config/settings/common.py
marcofucci: yeah I know :(

config/settings/common.py
elcct: I think this is a good point, I would prefer to use separate Redis databases.

PR #870, opened 2018-03-16
Add level__lte filter to sector view

reupen:
Issue number: DH-1585

### Description of change

Add the ability to filter sectors to those with a level less than or equal to a value . Will be used by the front end to get sectors up to a certain level (e.g. only root sectors).

### Checklist


marcofucci:
@reupen I don't think removing the registration and adding the item in `datahub/metadata/urls.py` is a good idea, can you expand the `register` method to support an explicit view instead?

The idea behind the registry is to allow distributed metadata, this is bringing back the old problem.

PR #871, opened 2018-03-19
Omis refresh task

marcofucci:
This adds a periodic celery task that gets old ongoing payment gateway sessions and refreshes them against GOV.UK Pay.
This is because some users might abandon payment sessions and to keep the systems in sync in case of any communication problems.


datahub/omis/payment/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+
+    :param age_check: minutes since the session was last modified to be
+        included in the query. E.g. age_check=60 means that only ongoing
+        sessions 1-hour old are refreshed.
+        This is to give time to the user to complete the journey normally.
+    """
+    with transaction.atomic():
+        dt_check = now() - timedelta(minutes=age_check)
+        qs = PaymentGatewaySession.objects.ongoing()
+        qs.filter(modified_on__lte=dt_check).refresh_from_govuk_pay()
```
reupen: Does this leave some scope for the number of sessions needing checking to build up over time?

datahub/omis/payment/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+    Celery task that refreshes old ongoing payments in case something
+    happens during the payment journey or the user abandons the payment
+    session.
+
+    :param age_check: minutes since the session was last modified to be
+        included in the query. E.g. age_check=60 means that only ongoing
+        sessions 1-hour old are refreshed.
+        This is to give time to the user to complete the journey normally.
+    """
+    with transaction.atomic():
```
reupen: Not clear to me why the whole thing should be atomic, won't there be unrelated sessions in there? Could you add something to explain it if you think it's needed?

datahub/omis/payment/managers.py
```diff
@@ -85,7 +85,7 @@ def refresh_from_govuk_pay(self):
         Refresh the objects in the queryset by re-synching the status with the
         GOV.UK payment.
         """
-        for session in self:
+        for session in self.select_for_update():
```
reupen: Not clear to me why all the sessions need to be locked at the same time. Could we lock them one at a time instead? Otherwise, could you add a comment to explain why if it's needed? It seems a bit undesirable to me to lock them all together since there will be HTTP calls to GOV.UK Pay in there as well.

datahub/omis/payment/tasks.py
marcofucci: GOV.UK Pay times out the sessions eventually so they should not build up. It could only happens if there's a bug in the code and the refresh doesn't work.
I guess it should be OK? What are you thinking about?

datahub/omis/payment/tasks.py
marcofucci: Unfortunately the same piece of code can and is used in other places. E.g. when creating a new session, the system refreshes the potential existing ones before doing anything.
This means that those records do have to be locked otherwise you run into race conditions.

In this case the records might not be _all_ related (some could) but if I start grouping things together the code gets unnecessarily complex.

I need a transaction for the _select for update_ bit :/

Ongoing old pending sessions should be minimal and updating them can be slow so it shouldn't be a problem. I could increase the age_check number to like 6 hours to make it even safer?

datahub/omis/payment/managers.py
marcofucci: Yes I guess I could execute this queryset without locks and then make n extra single queries to lock the records individually. 👍

datahub/omis/payment/tasks.py
elcct: I think if for some reason one payment session will have wrong data or other problem and cause method to fail, then the remaining sessions will not be processed and that will lead to build up, because processing would always stop on that record. Then you would have to delete/modify the record to "unblock" the task.

reupen: > GOV.UK Pay times out the sessions eventually so they should not build up. It could only happens if there's a bug in the code and the refresh doesn't work.
I guess it should be OK? What are you thinking about?

Yeah or if GOV.UK Pay doesn't behave like we expect. I was just thinking about sessions building up silently and going unnoticed (until that causes some other problem). Perhaps we could send a warning to Sentry if we encounter something old that should've expired? Also, should we stagger the requests just to make sure we don't hit Pay too hard?

datahub/omis/payment/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+        This is to give time to the user to complete the journey normally.
+    :param refresh_rate: delay in seconds between each refresh subtasks
+        to avoid hitting GOV.UK Pay too hard.
+    """
+    dt_check = now() - timedelta(minutes=age_check)
+    qs = PaymentGatewaySession.objects.ongoing()
+    session_ids = qs.filter(modified_on__lte=dt_check).values_list('id', flat=True)
+
+    for index, session_id in enumerate(session_ids):
+        refresh_payment_gateway_session.apply_async(
```
elcct: If for some reason not all of the sessions get refreshed within an hour, is there a chance of having them added again as tasks to be processed? Could this lead to infinite build up of tasks?

marcofucci: yeah, I thought about it but the number of payments is so low that this will never happen.
Also, the default logic is to cancel other session when a new one is created so this is just to clean up edge cases and for housekeeping. Most of the times there will be just 0 or 1 pending sessions.

Just to give you an idea, at the moment there are something like 200 new orders per month.
The only case where your scenario might happen is if there's a problem with the code and/or the integration and developers ignore the sentry errors for a few weeks which would be un acceptable outcome IMO.

elcct: Hmm maybe in the task itself would be good to check if the session is still ongoing before proceeding to refresh? This could help to clear up the build it in case it will ever happen (probably never as you say, but...)

marcofucci: the `PaymentGatewaySession.refresh_from_govuk_payment` does that check already :)

PR #874, opened 2018-03-20
Remove incorrect column attributes in the SPI migration.

elcct:
As in the title.


reupen:
Not sure about changing the existing migration for that as some environments might already have that column as nullable. Can we add a new migration to correct it instead?

elcct:
@reupen Added new migration also it looks like previous one has missing the ordering. Thanks!

PR #876, opened 2018-03-20
Add new service types for SPI.

elcct:
Issue number: DH-1594 DH-1597

### Description of change

This adds two new Service Types for Service Performance Indicators (SPIs).

### Checklist


reupen:
Isn't this something that the Live Services team should do via the admin site?

elcct:
@reupen my next task is to create a command that builds a report using one of the service types in this PR.  It would be difficult to manage possibily different Ids or names of the services if that was created via the admin site.

reupen:
Not sure using the service IDs is the best way to do that. We might want a new column in the service model so that we can flag the relevant services as being relevant to the report.

elcct:
I see. Adding a column just for one particular record doesn't sound good to me, could end up having a flag for each service? ID could be set as env variable and read by the command. Then indeed this PR wouldn't be needed.

PR #877, opened 2018-03-20
Add sector descends filter to interactions search

reupen:
Issue number: DH-1620

### Description of change

Adds the company and investment project sectors to the interaction search model. Also adds a sector_descends filter to interaction search.

### Checklist

### DIFF
--- a/datahub/search/interaction/apps.py
+++ b/datahub/search/interaction/apps.py
@@ -15,11 +15,17 @@ class InteractionSearchApp(SearchApp):
     permission_required = (f'interaction.{InteractionPermission.read_all}',)
     queryset = DBInteraction.objects.prefetch_related(
         'company',
+        'company__sector',
+        'company__sector__parent',
+        'company__sector__parent__parent',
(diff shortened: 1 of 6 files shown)


reupen:
@marcofucci Not sure how that relates to the PR (or the front end)? There is a requirement that we add a sector filter to the interactions search, and it is meant to work for both company and investment project interactions.

marcofucci:
I mean, we make the frontend more tight to this API endpoint which is obviously wrong as investment should not be part of an interaction. This makes life tricker for whoever is going to refactor this API in the future because we create more dependencies.

reupen:
Not sure what you mean. This adds a single sector_descends filter to interactions search. As long as you need to filter interactions by sector, you'll need a sector_descends filter. That doesn't have much relation to how the data is stored internally.

> which is obviously wrong

Why is it obviously wrong? When we have a new type of interaction to add that is nothing like the existing types then it would be a good time to review the design (as we would actually have some requirements to work from).

PR #880, opened 2018-03-21
Drop name column from metadata_sector table

reupen:
Issue number: DH-1585

### Description of change

We left it in the database for one release. Now, it can be removed.

### Checklist


datahub/metadata/migrations/0003_remove_sector_name_from_db.py
```diff
@@ -0,0 +1,23 @@
...
+        migrations.SeparateDatabaseAndState(
+            state_operations=[
+                migrations.AddField(
+                    model_name='sector',
+                    name='name',
+                    field=models.TextField(blank=True, null=True),
+                ),
+            ],
+        ),
+        migrations.RemoveField('sector', 'name'),
```
marcofucci: Is this the correct way of doing this? Add the field logically and they remove the field altogether?

reupen: Just doing a database operation to remove the field didn't work. Hence the route of adding the field back to the state, and then removing it normally. It makes sense as we removed the field from the state previously. Do you have another suggestion?

The other option was in the first commit (RunSQL).

marcofucci: 👍 was just curious, I trust you tried other options.

PR #883, opened 2018-03-22
Add Policy Feedback specific fields to Interaction model

r4vi:
Issue number: 
MI-1487

### Description of change

To support more granular analysis of feedback this commit is adding two fields to the
Interaction Model for use only when the KIND is policy_feedback.

The Serializer supports these new fields and Serializer validation requires them
when an Interaction's type is policy_feedback, it will disallow them if the
Interaction is of any other kind.

These fields don't need indexing by ES so search.interaction.models.Interaction
will not be updated.

### Checklist

* [x] ~Has the admin site been updated (for new models, fields etc.)~?
* [x] ~Has the README been updated (if needed)~?


r4vi:
TODO: need to update test_data.yaml

r4vi:
I guess I do need to update the search model...

datahub/interaction/migrations/0024_issue_type.yaml
```diff
@@ -0,0 +1,15 @@
...
+  fields: {disabled_on: null, name: Non-EU trade priority, order: 300.0}
+- model: interaction.policyissuetype
+  pk: c1543164-7612-42df-b556-f59ae5173042
+  fields: {disabled_on: null, name: Economic risk, order: 500.0}
+- model: interaction.policyissuetype
+  pk: c1f1e29b-17ba-402f-ac98-aa23f9dc9bcb
+  fields: {disabled_on: null, name: EU exit, order: 200.0}
+- model: interaction.policyissuetype
+  pk: fa35595d-780d-4faa-b575-35f1c319bfee
+  fields: {disabled_on: null, name: Economic oppertunity, order: 400.0}
```
reupen: Typo here (`oppertunity`).

datahub/interaction/test/fixtures.py
```diff
@@ -0,0 +1,40 @@
...
+            'response_data': {
+                'policy_area': {
+                    'id': str(policy_area.pk), 'name': policy_area.name
+                },
+                'policy_issue_type': {
+                    'id': str(policy_issue_type.pk), 'name': policy_issue_type.name
+                },
+            }
+        }
+    return defaultdict(dict)
```
reupen: Would suggest returning a normal dict with appropriate contents instead of a defaultdict. The latter makes it easy to make typos unnoticed etc.

datahub/interaction/test/fixtures.py
```diff
@@ -0,0 +1,40 @@
...
+    Interaction.KINDS.policy_feedback,
+    Interaction.KINDS.interaction,
+    Interaction.KINDS.service_delivery
+])
+def defaults_for_kind(kind):
+    """
+    Returns additional required defaults for each KIND.
+    This can be used to keep the tests like add_interaction generic instead of
+    copy & pasting 99% of the same thing to test KINDS.Interaction and KINDS.policy_feedback.
+    To use this just add a 'defaults_for_kind' param to any test that is parametertized on
```
reupen: Only one t in `parametertized` :slightly_smiling_face:

datahub/interaction/migrations/0024_issue_type.yaml
r4vi: I copy & pasted from the system we're going to interop with. But I'll fix it here and when they try to import our records they'll realise and fix their system (hopefully)

datahub/interaction/test/conftest.py
```diff
@@ -0,0 +1 @@
+from .fixtures import defaults_for_kind  # noqa: F401
```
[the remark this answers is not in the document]
r4vi: I think I may have gone too far with moving it into its own file. I'll move it to test_view.py

marcofucci:
😢

datahub/interaction/test/test_views.py
```diff
@@ -697,7 +713,7 @@ def test_add_interaction_project_missing_fields(self):
         'kind',
         (Interaction.KINDS.interaction, Interaction.KINDS.policy_feedback,)
     )
-    def test_add_interaction_missing_interaction_only_fields(self, kind):
+    def test_add_interaction_missing_interaction_only_fields(self, kind, defaults_for_kind):
```
elcct: Why those tests test kind `interaction` and `policy_feedback`, but not `service_delivery`? Then the name suggests as if it should test `interaction` only, so why there is `policy_feedback` in the mix? This is a bit confusing.

r4vi: I guess because `policy_feedback` is up until now exactly the same as `interaction` in terms of data. So this is testing `interaction` only fields which are also `policy_feedback` only fields.

datahub/interaction/test/fixtures.py
```diff
@@ -0,0 +1,40 @@
...
+from datahub.core.test_utils import random_obj_for_model
+from datahub.interaction.models import Interaction, PolicyArea, PolicyIssueType
+
+
+@pytest.fixture(params=[
+    Interaction.KINDS.policy_feedback,
+    Interaction.KINDS.interaction,
+    Interaction.KINDS.service_delivery
+])
+def defaults_for_kind(kind):
```
elcct: I am not convinced by this name as it really gives defaults for `policy_feedback`. The name suggests as if it was giving something for other kinds and that is not the case. This also removes visibility of what is being tested in the tests. I'd prefer more explicit way of setting the data for `policy_feedback`.

r4vi: @elcct The name is only bad because we're not using it to provide the defaults for the other types. 

I could refactor the tests and move the other defaults into this and that would make the tests much shorter. Happy to do this now or we can leave it as an extension point for the future?

r4vi: @elcct I could just split the tests for add_interaction and add_policy_feedback and get rid of this totally - there is an argument for simplicity and copy & paste is not as bad if in tests

datahub/interaction/admin.py
```diff
@@ -17,6 +17,17 @@ class MetadataAdmin(admin.ModelAdmin):
...
 
+@admin.register(
+    PolicyArea,
+    PolicyIssueType
+)
+class OrderedMetadataAdmin(MetadataAdmin):
+    """Ordered Metadata model admin."""
+
+    fields = ('id', 'name', 'disabled_on', 'order')
+    list_display = ('name', 'disabled_on', 'order')
```
reupen: Shouldn't we keep the

```python
    readonly_fields = ('id',)
    search_fields = ('name', 'pk')
    list_filter = (DisabledOnFilter,)
```

we normally use?

r4vi: @reupen it's inheriting them from MetadataModel :point_up:

reupen: Ah, cool, I missed that :+1:  I think we need to tidy up some of the metadata admin classes as there's a fair bit of repetition around (obviously this one is the same as `datahub.metadata.admin.OrderedMetadataAdmin`), but that's something for another day.

datahub/interaction/test/fixtures.py
r4vi: @elcct I've split them up and it's definitely more understandable

datahub/interaction/test/test_views.py
```diff
@@ -721,18 +816,40 @@ def test_add_interaction_missing_interaction_only_fields(self, kind):
             'communication_channel': ['This field is required.'],
         }
 
-    @pytest.mark.parametrize(
-        'kind',
-        (Interaction.KINDS.interaction, Interaction.KINDS.policy_feedback,)
-    )
-    def test_add_interaction_with_service_delivery_fields(self, kind):
+    def test_add_policy_feedback_missing_policy_feedback_only_fields(self):
+        """Test add new interaction without required interaction-only fields."""
```
reupen: The docstring and function name don't quite match here.

datahub/interaction/test/test_views.py
```diff
@@ -817,26 +994,53 @@ def test_add_interaction_project(self, kind):
...
         assert response_data['created_on'] == '2017-04-18T13:25:30.986208Z'
 
-    @pytest.mark.parametrize(
-        'kind',
-        (Interaction.KINDS.interaction, Interaction.KINDS.policy_feedback,)
-    )
-    def test_add_interaction_no_entity(self, kind):
+    @freeze_time('2017-04-18 13:25:30.986208')
+    def test_add_policy_feedback_project(self):
+        """Test add new policy feedback interaction for an investment project."""
```
reupen: Are we adding policy feedback interactions to investment projects as well?

r4vi: @reupen initial brief for this was to make policy interactions exactly the same as interactions. Now that they've diverged in terms of fields I've checked again with Adrian and he still wants this to be the case

datahub/interaction/test/test_views.py
r4vi: will update

datahub/interaction/test/test_views.py
reupen: Investment project interactions are different though, those are added through an investment project. Is the kind selection page being added there (it doesn't exist at present)?

r4vi: @reupen we haven't done a selection page after the 'add interaction' button on an Investment, I'll talk to Adrian about this, not sure how I feel about putting an extra click between adding an interaction to an Investment when 99.9% of the time they just need normal interaction vs policy feedback

r4vi: @reupen good point, just talked to @web-bert and Adrian. None of us even knew that journey existed. I don't know who should make the call but our feeling is to allow it in the BE and we'll add the FE when we get the chance. @jim68000 do you have any thoughts?

jim68000: So no, we don't need policy interactions in investment projects - investment interactions should be a subset of interactions that are only related to the management of the project. If a policy note comes up as part of a project it should be stored against the company
, so no need for a 'kind' page in investments

r4vi: @reupen I've updated the test as you've suggested

PR #884, opened 2018-03-23
Return business type in Companies House view

reupen:
Issue number: DH-1607

### Description of change

This is part of the work to move the company category to business type mapping from the front-end code base to the back end.

This is so that there does not need to be a one-to-one mapping between company categories and business types, and to allow companies limited by guarantee to be added to the system.

### Checklist

### DIFF
--- a/datahub/company/ch_constants.py
+++ b/datahub/company/ch_constants.py
@@ -1,0 +1,51 @@
+from datahub.company.constants import BusinessTypeConstant
+
+# This mapping determines which Companies House company categories we are interested in,
+# and what their corresponding business type should be in Data Hub.
+#
+# Any company categories that are mapped to a falsey value are not loaded to the database by the
+# sync_ch command.
+#
+# Additionally, the Companies House company view returns a business type using this mapping. That
+# business type is then used by the front end when POSTing to /v3/company.
+COMPANY_CATEGORY_TO_BUSINESS_TYPE_MAPPING = {
+    # Address not available/main registration not with Companies House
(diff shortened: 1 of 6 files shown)


datahub/company/ch_constants.py
```diff
@@ -0,0 +1,51 @@
...
+#
+# Any company categories that are mapped to a falsey value are not loaded to the database by the
+# sync_ch command.
+#
+# Additionally, the Companies House company view returns a business type using this mapping. That
+# business type is then used by the front end when POSTing to /v3/company.
+COMPANY_CATEGORY_TO_BUSINESS_TYPE_MAPPING = {
+    # Address not available/main registration not with Companies House
+    'charitable incorporated organisation': None,
+    'community interest company': BusinessTypeConstant.community_interest_company,
```
elcct: Just out of curiosity what document can you refer to, to confirm these mappings are correct?

reupen: There's no simple document available. You (or someone else) would need to look at what the legal definitions are, what's in the data coming from Companies House, their documentation etc.

The latest changes for limited by guarantee are what I agreed with @jim68000. Unfortunately, we made a mistake trying to make our business types highly granular and match the CH categories one to one (rather than using the existing generic types like 'company'). In some cases things like the limited exemption can change in a company's history, so it didn't make any sense at all.

If we need the CH company category for a company, we can get it from Companies House. (In fact the front end displays the company category from CH in place of the business type when the company is linked to a CH record.)

PR #888, opened 2018-03-27
Refactor signal receivers in search apps

reupen:
Issue number: N/A

### Description of change

Reduces repetition by replacing the connect_signals() and disconnect_signals() functions with objects.

Also reduces the risk of the connect_signals() and disconnect_signals() functions containing mismatched operations.

### Checklist

### DIFF
--- a/datahub/search/apps.py
+++ b/datahub/search/apps.py
@@ -53,12 +53,14 @@ class SearchApp:
         Connects all signal handlers so DB models can be synced with Elasticsearch on save.
         """
         signals_mod = import_module(self.mod_signals)
-        getattr(signals_mod, 'connect_signals')()
+        for receiver in signals_mod.receivers:
+            receiver.connect()
 
     def disconnect_signals(self):
         """Disconnects all signal handlers."""
         signals_mod = import_module(self.mod_signals)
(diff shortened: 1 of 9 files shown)


reupen:
> 1. SignalReceiver might read better if you move the sender param to 2nd position? E.g. SignalReceiver(post_save, DBContact, comntact_sync_es). We might make sender optional in the future though so that could be a reason to keep it this way instead.

Yes, I thought the same when self-reviewing before. I left it as it was as it matches the order when calling Signal.connect(). I'll change it though as it would be more readable, and if we ever need sender to be optional (which seems unlikely) we can just change it back.

> 2. This is not related to your refactor but I don't know why we ended up using the signals.py module for these. In theory that module is used to define signals and there should be a receivers.py module (or similar) for the actual receivers instead.

We were probably going off the Django docs which refers to defining receivers in a signals submodule:

https://docs.djangoproject.com/en/2.0/topics/signals/#connecting-receiver-functions

marcofucci:
> We were probably going off the Django docs which refers to defining receivers in a signals submodule:

Yeah I know :/ they decide to over-simply things now and use the same module for both defining custom signals and signal handlers which I find very messy :/
Let's leave it like this to keep it aligned with the django docs.

PR #890, opened 2018-03-27
Add new `metadata.Service` for Policy feedback

r4vi:
Issue number: 
MI-1438

### Description of change
Adds a new service which will be the default one for Policy feedback
interactions. This is because making `Service` a nullable field on interaction
will cause problems with filters.

### Checklist

* [ ] ~Have any relevant search models been updated~?
* [ ] ~Have any relevant select-/prefetch-related field lists in the views and search apps been updated~?
* [ ] ~Has the admin site been updated (for new models, fields etc.)~?
* [ ] ~Has the README been updated (if needed)~?


r4vi:
Should I add to initial metadata too or is the migration enough?

reupen:
Won't this make the new service visible for all kinds of interactions? Services are managed by admin users so I don't think hard-coding is desirable. I think we need a better way of solving it e.g. giving services attributes or a type so we know what they should and shouldn't be used for.

PR #891, opened 2018-03-27
Add OMIS refunds to django admin

marcofucci:
This allows refund records to be managed via the django admin.

- Validation is kept to a minimum
- Only approved refunds can be added
- Existing refunds cannot have their status changed
- Required fields are configured depending on the refund status

Note:
- There are no refund records in status _Requested_ in production
- Things can be improved, e.g. the amounts are still in pence. This is the first iteration and others will follow if time allows.

### DIFF
--- a/datahub/omis/payment/admin.py
+++ b/datahub/omis/payment/admin.py
@@ -1,0 +1,226 @@
+from django import forms
+from django.conf import settings
+from django.contrib import admin
+from django.core.exceptions import ValidationError
+from django.template.defaultfilters import date as date_formatter
+from django.utils.translation import ugettext_lazy as _
+
+from datahub.omis.order.constants import OrderStatus
+from .constants import RefundStatus
+from .models import Refund
+
+
+class RefundForm(forms.ModelForm):
+    """Form for adding/changing refund records via the django admin."""
+
+    APPROVED_MANDATORY_FIELDS = [
+        'level1_approved_on',
+        'level1_approved_by',
+        'level2_approved_on',
+        'level2_approved_by',
+        'method',
+        'net_amount',
+        'vat_amount',
+    ]
+
+    class Meta:
+        model = Refund
+        fields = (
+            'id',
+            'order',
+            'reference',
+            'status',
+            'requested_on',
+            'requested_by',
+            'requested_amount',
+            'refund_reason',
+            'level1_approved_on',
+            'level1_approved_by',
+            'level1_approval_notes',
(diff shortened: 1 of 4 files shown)


datahub/omis/payment/admin.py
```diff
@@ -0,0 +1,210 @@
...
+        if net_amount is None or vat_amount is None:
+            return
+
+        total_amount = net_amount + vat_amount
+
+        qs = order.refunds
+        if self.instance.pk:
+            qs = qs.exclude(pk=self.instance.pk)
+        currently_refunded = sum([
+            amount for amount in qs.values_list('total_amount', flat=True) if amount
```
reupen: Do we need [ and ] here?

datahub/omis/payment/admin.py
```diff
@@ -0,0 +1,210 @@
...
+        refund_remaining = order.total_cost - currently_refunded
+        if total_amount > refund_remaining:
+            self.add_error(
+                'net_amount',
+                ValidationError(
+                    _('Remaining amount that can be refunded: %(refund_remaining)s.'),
+                    params={
+                        'refund_remaining': refund_remaining
+                    },
+                    code='refund_insufficient'
```
reupen: Is the error that too much has been refunded? Then is `refund_insufficient` the right code?

datahub/omis/payment/admin.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def save_model(self, request, obj, form, change):
+        """
+        Populate total_amount from other fields and
+        created_by/modified_by from the logged in user.
+        """
+        if 'total_amount' in form.cleaned_data:
+            obj.total_amount = form.cleaned_data['total_amount']
+
+        if not change:
```
reupen: We need to solve this generally as well :cry:

datahub/omis/payment/admin.py
```diff
@@ -0,0 +1,210 @@
...
+from datahub.omis.order.constants import OrderStatus
+from .constants import RefundStatus
+from .models import Refund
+
+
+class RefundForm(forms.ModelForm):
+    """Form for adding/changing refund records via the django admin."""
+
+    APPROVED_MANDATORY_FIELDS = [
+        'level1_approved_on', 'level1_approved_by',
```
reupen: We've normally gone for one per line for long lists like these (to make diffs easier to follow). Not sure why we'd do differently for these ones.

PR #893, opened 2018-03-28
Make Elasticsearch paging consistent.

elcct:
Issue number: DH-1389

### Description of change

This adds `id` as additional sort field besides of default `_score` or chosen field. `id` is in every entity and is unique. Additional sorting by this field acts as a tiebreaker in case search result score (or primary sort field) is equal and makes search results deterministic and consistent between the pages 

### Checklist


elcct:
@reupen added some extra tests that test when field_order isn't None

PR #898, opened 2018-04-05
Add order to Headquarter Type.

elcct:
Issue number: DH-1646

### Description of change

This modifies Headquarter Type metadata so that it has additional `order` field. This enables Headquarter Types to be returned in particular order.

### Checklist


reupen:
We made HeadquarterType read-only in admin. So, I think we either need to update the order for the existing records in the migration, or make the order field editable in admin. (An option for the latter might be to add an editable_fields attribute to ReadOnlyAdmin and make it remove those from readonly_fields.)

elcct:
@reupen updated the admin now. I am not sure about the name, but couldn't come up with something better.

PR #899, opened 2018-04-05
Add contexts to services

reupen:
Issue number: DH-1643

### Description of change

Adds a multi-select contexts field to services. This will be used to control which services appear in different pages etc. and can be edited via admin.

Five contexts were added:

* event
* interaction
* investment_project_interaction
* service_delivery
* policy_feedback

The migration gives all existing services in the database the first four of those contexts.

A contexts__contains filter was also added to `/metadata/service/`.

Some values for the field had to be added to `fixtures/metadata/services.yaml` so I added some (effectively they will just be used by tests or new environments).

### Checklist


datahub/metadata/admin.py
```diff
@@ -80,6 +79,21 @@ class OrderedMetadataAdmin(admin.ModelAdmin):
...
+@admin.register(models.Service)
+class ServiceAdmin(DisableableMetadataAdmin):
+    """Admin for services."""
+
+    fields = ('id', 'name', 'contexts', 'disabled_on')
+    list_display = ('name', 'contexts_friendly', 'disabled_on')
+
+    def contexts_friendly(self, obj):
+        """Displays the contexts using their names (instead of their internal values)."""
+        return ', '.join(obj.CONTEXTS[context] for context in obj.contexts)
```
marcofucci: It would be nice to mirror what Django does with `ChoiceField`s and inject a `display_contexts_display` method on the model, what do you think?

datahub/metadata/test/test_views.py
```diff
@@ -138,6 +138,45 @@ def test_team_view(api_client):
...
+            'contexts': sorted(service.contexts),
+            'disabled_on': disabled_on,
+        }
+        assert len(services) == Service.objects.count()
+
+    @pytest.mark.parametrize('context', ('interaction', 'non-existent-context'))
+    def test_list_filter_by_contains(self, api_client, context):
+        """Test listing services, filtered by context."""
+        url = reverse(viewname='service')
+        response = api_client.get(url, data={'contexts__contains': context})
```
marcofucci: I'm not sure `contexts__contains` is a good name, I'd use `contexts=..`. or `contexts__in=`.
_contains_ might assume a substring of a context which would be an incorrect interpretation.

reupen: Not sure either of those are better options.

`contexts=` implies strict equality (comparing one array with another), which we aren't doing here.
`contexts__in=A` implies it's checking that contexts is in A (not that A is in contexts).

contains is the name of the operator used here, and is used in Python (`operator.contains()`) as well. If you know that contexts is a list, I don't think expecting contains to convert the list to a string and do a string-contains is a reasonable interpretation to make.

Open to any other suggestions for a name for it though.

datahub/metadata/admin.py
marcofucci: 👍 that looks a lot nicer, what's the problem with the admin?

datahub/metadata/test/test_views.py
marcofucci: So I went through it quickly and I might be missing something but it seems we are using `contexts` as many2many relationship. The implementation is an array but the logic is a m2m one.

The API django uses for that is `__in` (https://docs.djangoproject.com/en/2.0/topics/db/examples/many_to_many/) meaning you want the services with context in any of the items in the list or `=` if you are interested in only one item.

Using `__contains` doesn't seem to be consistent with anything else we are using.

reupen: Yes, but I'd argue that's confusing. It's doing the in for each item in publications there, it's effectively shorthand for `publications__id__in`. And as it's an inner join so you end up with rows repeated when they match more than one item in the operator argument. 

> or = if you are interested in only one item.

Similarly, that's evaluated on each publication with a rather clumsy syntax.

If we exposed either of those via the main API, it would be something like publications__id__in or publications__id to be consistent with the rest of the main API.

Anyway, how about `contexts__has` (which perhaps better encapsulates that it only takes one argument).

datahub/metadata/admin.py
reupen: Mainly that I had to tell it to use `get_contexts_display` explicitly – it's not very obvious (though this is still better than before).

datahub/metadata/test/test_views.py
marcofucci: Hm I partially see where you are coming from. To be honest your `__contains` or `__has` seems a better version of the django `=` which I agree could be confusing.
I think that the `__has` feels a lot better, still not convinced but go for it.

elcct: I'd vote for `__has` as well

marcofucci: What happens if we want to pass in a list of values, probably in OR? Would __has still make sense? I think it could, definitely better than __contains which django considers in AND in the ArrayField implementation.

reupen: I think if we wanted to pass more than one value, I would name it something like `__has_any`. We could go straight for that if we wanted, it might be useful to have.

marcofucci: Yeah, I think it could be a good idea to start with that already.

datahub/metadata/filters.py
```diff
@@ -7,12 +7,12 @@
 class ServiceFilterSet(FilterSet):
     """Filters for the service view."""
 
-    contexts__contains = CharFilter(name='contexts', method='filter_contexts__contains')
+    contexts__has_any = CharFilter(name='contexts', method='filter_contexts__contains')
```
marcofucci: should we change the method name as well?

datahub/metadata/test/test_views.py
```diff
@@ -164,17 +166,36 @@ def test_list(self, api_client):
...
         assert len(services) == Service.objects.count()
 
-    @pytest.mark.parametrize('context', ('interaction', 'non-existent-context'))
-    def test_list_filter_by_contains(self, api_client, context):
+    @pytest.mark.parametrize(
+        'contexts',
+        (
+            ['interaction'],
+            ['non-existent-context'],
+            ['interaction', 'service_delivery'],
```
marcofucci: I would use the constants instead of strings for the valid ones.

PR #901, opened 2018-04-06
Add the ability to make psycopg2 asynchronous using gevent

reupen:
Issue number: N/A

### Description of change

This add the ability to made psycopg2 asynchronous (when the server is run using Gunicorn) using psycogreen and the GUNICORN_ENABLE_ASYNC_PSYCOPG2 environment variable.

Although we were using gevent workers with Gunicorn, psycopg2 was not made asynchronous because it's unaffected by the patching that gevent does.

GUNICORN_ENABLE_ASYNC_PSYCOPG2 is disabled by default. We can enable it in a test environment first to make sure there are no undesired side effects.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the 
 | `ES_VERIFY_CERTS`  | No | |
 | `GUNICORN_ACCESSLOG`  | No | File to direct Gunicorn logs to (default=stdout). |
 | `GUNICORN_ACCESS_LOG_FORMAT`  | No |  |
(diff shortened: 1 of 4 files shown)


elcct:
This sounds good, just thinking how we could test that - maybe running front end tests against it?

reupen:
> This sounds good, just thinking how we could test that - maybe running front end tests against it?

I've run it against them locally, and nothing different happened compared to develop.

It could lead to a higher number of database connections in use at any one time, but that's limited by GUNICORN_WORKER_CONNECTIONS and the number of app instances (which we could reduce with this on),

One specific thing this should fix is the interaction search in admin blocking the entire API instance.

PR #905, opened 2018-04-11
Reorganise es funcs

marcofucci:
This reorganises the ES functions used to make a search.

At a very high level it:
- renames function names from `_get_...` to `_build_...`
- merges some functions
- renames variables and params to keep them consistent
- adds some more tests

### DIFF
--- a/datahub/search/query_builder.py
+++ b/datahub/search/query_builder.py
@@ -8,6 +8,9 @@ from elasticsearch_dsl.query import Bool, MatchPhrase, MultiMatch, Query, Term
 from datahub.search.apps import EXCLUDE_ALL, get_search_apps
 
 MAX_RESULTS = 10000
+FIELD_REMAPPING = {
+    'name': 'name_keyword',
+}
 
 
 class MatchNone(Query):
@@ -26,7 +29,7 @@ def get_basic_search_query(
         term,
         entities=None,
         permission_filters_by_entity=None,
-        field_order=None,
+        ordering=None,
         ignored_entities=(),
         offset=0,
         limit=100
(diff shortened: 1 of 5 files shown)


datahub/search/query_builder.py
```diff
@@ -49,10 +49,10 @@ def get_basic_search_query(
     # and the same query is always generated with the same inputs
     fields = sorted(fields)
 
-    query = _get_search_term_query(term, fields=fields)
+    query = _build_term_query(term, fields=fields)
     s = Search(index=settings.ES_INDEX).query(query)
```
elcct: Maybe `s` could be renamed to `search`?

datahub/search/query_builder.py
```diff
@@ -392,3 +338,52 @@ def _build_must_queries(filters, ranges, composite_field_mapping):
...
+        {field_name: sort_params},
+        'id'
+    )
+
+
+def _apply_aggs_to_query(query, aggregates):
+    """Applies aggregates to the query."""
+    for aggregate in aggregates:
+        # skip range and "query" filters as we can't aggregate them
+        if any(aggregate.endswith(x) for x in ('_before', '_after', '_trigram', '_exists')):
```
reupen: Can we put `('_before', '_after', '_trigram', '_exists')` into a named constant?

datahub/search/query_builder.py
```diff
@@ -392,3 +338,52 @@ def _build_must_queries(filters, ranges, composite_field_mapping):
...
+        'name': 'name_keyword',
+    }.get(field_name, field_name)
+
+    return query.sort(
+        {field_name: sort_params},
+        'id'
+    )
+
+
+def _apply_aggs_to_query(query, aggregates):
```
reupen: Maybe _add_aggs_to_query? Also not sure about the name of `aggregates` – maybe `fields` or `aggregation_fields`?

datahub/search/query_builder.py
```diff
@@ -392,3 +338,52 @@ def _build_must_queries(filters, ranges, composite_field_mapping):
...
+        # skip range and "query" filters as we can't aggregate them
+        if any(aggregate.endswith(x) for x in ('_before', '_after', '_trigram', '_exists')):
+            continue
+
+        query_aggs = query.aggs
+        if '.' in aggregate:
+            query_aggs = query_aggs.bucket(
+                aggregate,
+                'nested',
+                path=aggregate.split('.', 1)[0]
```
reupen: Should this be split or rsplit? Is it intentional that we flip between the two?

datahub/search/query_builder.py
```diff
@@ -294,103 +229,161 @@ def _get_exists_query(field, value):
...
             )
 
     if ranges:
-        must_filter.extend(_get_range_queries(ranges))
+        must_filter.extend(_build_range_queries(ranges))
 
     return must_filter
+
+
+def _apply_sorting_to_query(query, field_order):
```
reupen: Not sure `field_order` is a good name – it's not the order of the fields. Just ordering might be fine.

datahub/search/query_builder.py
```diff
@@ -294,103 +229,161 @@ def _get_exists_query(field, value):
...
+        'missing': '_first' if order == 'asc' else '_last'
+    }
+
+    # check if we sort by field in nested document (example: 'stage.name')
+    if '.' in field_name:
+        # extract and add path to nested document (example: 'stage')
+        sort_params['nested_path'] = field_name.split('.', 1)[0]
+
+    # remap field name if necessary
+    field_name = {
```
reupen: Can we make the dict literal a named constant? I think it's a bit buried here.

datahub/search/query_builder.py
```diff
@@ -294,103 +229,161 @@ def _get_exists_query(field, value):
...
+    """
+    Used by _build_field_query and always expecting value as a single value.
+    You should never need to use this, it's more likely you want _build_field_query instead.
+    """
+    # define field and value
+    real_field, real_value = field, value
+    if real_value is None:
+        real_value = False
+        if not real_field.endswith('_exists'):
+            real_field = f"{real_field.rsplit('.', maxsplit=1)[0]}_exists"
```
reupen: I was wondering what this logic was for, then I found it was being hit in one of the tests when filtering for headquarter_type=None. So effectively headquarter_type gets remapped to headquarter_type.id, and then back again... Doesn't seem ideal, but don't think we can do much about it here.

datahub/search/query_builder.py
```diff
@@ -294,103 +229,161 @@ def _get_exists_query(field, value):
...
+        *_build_field_queries(filters),
+        *_build_range_queries(ranges),
+    ]
 
 
-def _get_must_filter_query(filters, composite_filters, ranges):
-    """Gets "and" filter query."""
+def _build_must_queries(filters, ranges, composite_field_mapping):
+    """Builds a "must" filter query."""
     if filters is None:
```
reupen: filters is never None here. So we could get rid of this.

datahub/search/query_builder.py
marcofucci: This would probably need some tests as if I comment those lines out nothing fails :/

datahub/search/query_builder.py
elcct: Name was meant to hint the format of the string `[field]:[order]` -> `field_order`, but I guess that could be confusing. What about `ordering_template`?

marcofucci:
@reupen @elcct have another look.
I don't feel like I understand all the places where aggregations are used so I'd rather not touch split vs rsplit stuff. Probably one of you could take a better look at that.

PR #907, opened 2018-04-12
Feature flags feature.

elcct:
Issue number: DH-1604

### Description of change

This adds feature flags module that keeps feature flags. Module has one endpoint to be periodically consumed by the front end that lists all feature flags. This enables operator of Django admin to turn on or off future features if needed. 

### Checklist


marcofucci:
I was not part of the discussion but IMHO I would:
- remove pk from the API
- use a `code` field instead of `name`, make that unique
- have just `code` and `description` (probably created_on doesn't hurt but without returning it)
- have a flag for active / not active as boolean

datahub/featureflag/admin.py
```diff
@@ -0,0 +1,15 @@
...
+from datahub.core.admin import DisabledOnFilter
+from datahub.featureflag.models import FeatureFlag
+
+
+@admin.register(FeatureFlag)
+class FeatureFlagAdmin(admin.ModelAdmin):
+    """Feature flag admin."""
+
+    fields = ('id', 'name', 'description', 'disabled_on',)
+    list_display = ('name', 'description', 'created_by', 'created_on', 'disabled_on',)
```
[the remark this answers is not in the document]
elcct: I missed that :( I will update. Thanks

datahub/featureflag/models.py
```diff
@@ -0,0 +1,15 @@
+from django.conf import settings
+from django.db import models
+
+from datahub.core.models import BaseConstantModel, BaseModel
+
+MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class FeatureFlag(BaseConstantModel, BaseModel):
```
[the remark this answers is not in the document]
elcct: I think I would change to what @marcofucci has suggested. `active` boolean.

datahub/featureflag/serializers.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework import serializers
+
+from datahub.company.serializers import NestedAdviserField
+from datahub.core.serializers import ConstantModelSerializer
+
+
+class FeatureFlagSerializer(ConstantModelSerializer):
```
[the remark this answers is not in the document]
elcct: Very good point, I will update that.

elcct:
@marcofucci thank you for suggestions. 

Why having created_by/modified_by wouldn't be useful? I think that would be good to know who and when modified the flag.

elcct:
> Yes true but I don't think it's worth exposing them via the API which should probably return just list of (code, active), no?

I agree. I'll remove that from the API response.

> Is the general idea to delete a feature flag record when we are happy with the feature and it gets fully merged?

I think so. But when it is there it should be okay to track changes to it.

datahub/feature_flag/test/test_views.py
```diff
@@ -0,0 +1,49 @@
...
+            code=factory.Iterator(('def', 'abc',))
+        )
+        url = reverse('api-v3:feature-flag:collection')
+        response = self.api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+
+        response_data = response.json()
+        assert response_data == {
+            'count': 2,
```
reupen: I don't think it makes sense to use pagination in the response for these. Perhaps, even, the response should be in the form of a dict/object (like `{"code": {"description": "blah", "is_active": true}}`)?

datahub/feature_flag/views.py
```diff
@@ -0,0 +1,15 @@
...
+from datahub.oauth.scopes import Scope
+
+
+class FeatureFlagViewSet(CoreViewSetV3):
+    """Feature flag ViewSet v3."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
+    serializer_class = FeatureFlagSerializer
+    queryset = get_feature_flag_queryset()
```
marcofucci: Do we really need an external module for this queryset?

datahub/feature_flag/test/test_views.py
```diff
@@ -0,0 +1,49 @@
...
+from datahub.feature_flag.test.factories import FeatureFlagFactory
+
+# mark the whole module for db use
+pytestmark = pytest.mark.django_db
+
+
+class TestListFeatureFlags(APITestMixin):
+    """Tests for listing of feature flags."""
+
+    def test_can_list_feature_flags_having_no_permissions(self, api_client):
```
marcofucci: shouldn't it be `test_cannot...` ?

datahub/feature_flag/test/test_views.py
elcct: I can remove that. Then I think `description` also wouldn't really be needed in frontend, so that would leave us with simply `{"COFFEE_MACHINE": true, "CAT_FOOD_DISPENSER": false}` response?

datahub/feature_flag/test/test_views.py
```diff
@@ -0,0 +1,49 @@
...
+        """Should return 401"""
+        url = reverse('api-v3:feature-flag:collection')
+        response = api_client.get(url)
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    def test_can_list_feature_flags(self):
+        """List feature flags.."""
+        feature_flags = FeatureFlagFactory.create_batch(
+            2,
+            code=factory.Iterator(('def', 'abc',))
```
marcofucci: If you are defining these codes, it makes sense to assert those explicitly in the assert part. 
If you want to keep the assert bit as it is, you can remove this line I suppose instead.

datahub/feature_flag/test/test_views.py
marcofucci: yes I agree, we can get rid of the description as well.

elcct: Changed the response now, so it has a dictionary of flags and their active statuses.

datahub/feature_flag/views.py
elcct: Indeed that is an overkill. Removed.

datahub/feature_flag/test/test_views.py
elcct: Updated. Thanks!

datahub/feature_flag/test/test_views.py
elcct: I have changed this test now since the response has changed.

datahub/feature_flag/views.py
```diff
@@ -0,0 +1,16 @@
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from datahub.feature_flag.models import FeatureFlag
+
+
+@api_view()
+@permission_classes([IsAuthenticated])
+def get_feature_flags(request):
```
reupen: I think we should still try to use the generic views with a serialiser class, as otherwise the DRF API documentation feature won't be able to work out the format of the response.

elcct: I see. I have added generic view back. Thanks!

reupen:
We should set the verbose name of the app as it's showing up as feature_flag in admin.

datahub/feature_flag/models.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.conf import settings
+from django.db import models
+
+from datahub.core.models import BaseModel
+
+MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class FeatureFlag(BaseModel):
```
reupen: This needs a __str__ method as well.

PR #910, opened 2018-04-13
Correct various invalid ES model fields

reupen:
Issue number: DH-1657

### Description of change

This also adds some validation that the fields defined on ES models are correct. At the moment these are tests, though we could make them system checks if we prefer? https://docs.djangoproject.com/en/2.0/topics/checks/

### Checklist

### DIFF
--- a/datahub/search/investment/models.py
+++ b/datahub/search/investment/models.py
@@ -18,7 +18,6 @@ class InvestmentProject(DocType, MapDBModelToDict):
     approved_landed = Boolean()
     approved_non_fdi = Boolean()
     actual_land_date = Date()
-    actual_land_date_documents = dsl_utils.id_uri_mapping()
     business_activities = dsl_utils.id_name_mapping()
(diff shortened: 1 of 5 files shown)


datahub/search/investment/models.py
```diff
@@ -1,10 +1,35 @@
...
 from .. import dsl_utils
 from ..models import MapDBModelToDict
 
 
+def _referral_source_adviser_mapping():
+    """
+    Mapping for referral_source_adviser.
+
+    referral_source_adviser was incorrectly auto-created in the mapping, so this reflects how it
+    has been auto-created.
```
elcct: Does that mean the mapping is still incorrect? I am not sure if I understand this comment.

reupen: Yeah, it's not really incorrect, it's just different from contact_or_adviser_mapping. I've updated the docstring, it should make more sense now.

reupen:
@marcofucci @elcct Just fixed a small problem with the tests which identified country_lost_to as being a similar case to referral_source_adviser.

PR #912, opened 2018-04-16
Move interaction test views to submodules

marcofucci:
This splits and moves the interactions test views to different submodules to make them easy to maintain and refactor.

I haven't done much apart from grouping, renaming and moving the tests so you don't have to check the code of individual tests.

I'm not particularly happy with the `test_base.py` plus specific `test_...` modules but I think it's an improvement compared with what we had before. It's also a pity that with this we lose easy access to history of the test_views.py module.

What do you guys think?


elcct:
I don't like the `test_base.py` name, but can't think anything better.

PR #915, opened 2018-04-17
Only allow explcitly defined fields in Elasticsearch

reupen:
Issue number: DH-1658

### Description of change

This:

* explicitly defines all fields that were previously dynamically mapped by ES
* removes IGNORED_FIELDS from the ES models so that we have a whitelist of fields to include instead of a blacklist of fields to exclude
* sets the `dynamic` parameter to `strict` for all mapping types, so that it's impossible for fields to be added to the mapping implicitly

If we're concerned about forgetting to add new fields to the ES models we could add tests for each model to assert on {fields in DB} - {fields in ES}. But we have the checklist below as well, so that should be a reminder in itself to think about that.

I'll raise a separate PR to sort some of these attributes and dict keys alphabetically as they seem to have lost any meaningful order.

### Checklist

### DIFF
--- a/datahub/search/dsl_utils.py
+++ b/datahub/search/dsl_utils.py
@@ -13,6 +13,21 @@ EnglishText = partial(Text, analyzer='english_analyzer')
 SortableText = partial(Text, fielddata=True)
 
 
+class TextWithKeyword(Text):
+    """
+    Text field with keyword sub-field.
+
+    This definition is in line with the data type Elasticsearch uses for dynamically mapped text
+    fields.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialises the field, creating a keyword sub-field."""
+        super().__init__(*args, **kwargs, fields={
+            'keyword': Keyword(ignore_above=256)
+        })
+
+
 def contact_or_adviser_mapping(field, include_dit_team=False):
     """Mapping for Adviser/Contact fields."""
     props = {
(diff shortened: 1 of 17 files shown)


datahub/search/utils.py
```diff
@@ -4,3 +4,36 @@ class Echo:
...
+    return es_model._doc_type.mapping.properties._params['properties']
+
+
+def get_model_field_names(es_model):
+    """Gets the field names for an ES model."""
+    return get_model_fields(es_model).keys()
+
+
+def get_model_copied_to_field_names(es_model):
+    """Gets the names of fields that are copied to in an ES model."""
```
elcct: Maybe this comment could be reworded? Also wouldn't `get_model_copy_to_field_names` be better or something else as it doesn't sound clear to me.

datahub/search/dsl_utils.py
```diff
@@ -13,6 +13,21 @@
...
+    Text field with keyword sub-field.
+
+    This definition is in line with the data type Elasticsearch uses for dynamically mapped text
+    fields.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialises the field, creating a keyword sub-field."""
+        super().__init__(*args, **kwargs, fields={
+            'keyword': Keyword(ignore_above=256)
```
elcct: Could we tie this value to `CHAR_FIELD_MAX_LENGTH`? or why is it 256?

datahub/search/utils.py
reupen: I went with copied-to fields to better distinguish it from the fields that have copy_to set. I'll try to reword the docstring.

reupen:
@marcofucci Yes, I felt like there was a layer of abstraction missing. The main thing I don't like are all the copy-tos, though, we should be able to replace those with multi-fields. I think though that we pass too many functions around; dsl_utils (for fields) and dict_utils (for the mappings) feel like repetition.

PR #916, opened 2018-04-17
Refactor add interaction test views

marcofucci:
This:
- uses `pytest.mark.parametrize` to test different scenarios instead of having different tests
- adds a few validation cases not covered


datahub/interaction/test/views/test_interaction.py
```diff
@@ -11,7 +12,9 @@
 from datahub.event.test.factories import EventFactory
 from datahub.investment.test.factories import InvestmentProjectFactory
 from datahub.metadata.test.factories import TeamFactory
-from .test_common import NON_RESTRICTED_CHANGE_PERMISSIONS, NON_RESTRICTED_READ_PERMISSIONS
+from .test_common import (
+    NON_RESTRICTED_CHANGE_PERMISSIONS, NON_RESTRICTED_READ_PERMISSIONS, resolve_data
```
reupen: I think we should put these in a separate module rather than importing things from one test_ module into another.

datahub/interaction/serializers.py
```diff
@@ -128,18 +128,30 @@ class Meta:
                     ]),
                 ),
                 ValidationRule(
-                    'invalid_for_policy_feedback',
+                    'required',
```
reupen: Should we get rid of `AnyOfValidator('company', 'investment_project')` and fully replace it with rules?

datahub/interaction/serializers.py
```diff
@@ -128,18 +128,30 @@ class Meta:
...
+                    when=InRule('kind', [
+                        Interaction.KINDS.service_delivery,
+                        Interaction.KINDS.policy_feedback
+                    ]),
+                ),
+                ValidationRule(
+                    'invalid_for_non_interaction',
                     OperatorRule('investment_project', not_),
-                    when=EqualsRule('kind', Interaction.KINDS.policy_feedback),
+                    when=InRule('kind', [
```
reupen: This might benefit from being able to say 'kind is not interaction'. 'Not' could be a generic negation wrapper I guess. But this is fine as well.

datahub/interaction/test/views/test_common.py
```diff
@@ -38,6 +38,36 @@
...
+    """
+    Given a data dict with keys and values,
+    if a value is a callable, it resolves it
+    if a value is an object with a 'pk' attribute, it uses that instead.
+
+    It returns a new dict with resolved values.
+    """
+    resolved_data = {}
+    for key, value in data.items():
+        if callable(value):
```
reupen: I think the inner logic could be moved to a separate (internal) function. Then, you could use a dict comprehension here (and it'd be a bit more readable).

datahub/interaction/test/views/test_interaction.py
```diff
@@ -106,167 +120,110 @@ def test_add_company_interaction(self):
...
+                    'communication_channel': partial(random_obj_for_model, CommunicationChannel),
+                },
+                {
+                    'non_field_errors': [
+                        'One or more of company, investment_project must be provided.'
+                    ]
+                }
+            ),
+
+            # fields now allowed
```
reupen: Is this comment right?

datahub/interaction/test/views/test_interaction.py
marcofucci: 👍 what do you think about `utils`?

reupen: That'll do :+1:

datahub/interaction/serializers.py
reupen: Doesn't AnyIsNotBlankRule do pretty much the same thing as AnyOfValidator?

I just found it a bit confusing at the moment as the AnyOfValidator and the validation rules overlap (in relation to `company` and `investment_project`).

marcofucci: Managed to :)

PR #917, opened 2018-04-17
Fix companies house company search test.

elcct:
Issue number: N/A

### Description of change

This fixes test that occasionally fails due to non-deterministic pk. Adding sorting by `name` makes results consistent and makes test always pass.

### Checklist

### DIFF
--- a/datahub/search/companieshousecompany/tests/test_views.py
+++ b/datahub/search/companieshousecompany/tests/test_views.py
@@ -66,7 +66,8 @@ class TestSearchCompaniesHouseCompany(APITestMixin):
             (  # pagination
                 {
                     'limit': 1,
-                    'offset': 1
+                    'offset': 1,
+                    'sortby': 'name:asc',
                 },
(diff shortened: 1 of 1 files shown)


datahub/search/companieshousecompany/tests/test_views.py
```diff
@@ -66,7 +66,8 @@ def test_no_permissions(self):
             (  # pagination
                 {
                     'limit': 1,
-                    'offset': 1
+                    'offset': 1,
+                    'sortby': 'name:asc',
                 },
                 {'222'}
```
reupen: Why does 222 end up being the first one? Isn't 333 ('Cheetah') the first one alphabetically?

elcct: When content is sorted by name, the result is 333 ('Cheetah'), 222 ('Jaguarundi'), 111 ('Pallas') and in the test we request second page, which has 222.

PR #925, opened 2018-04-19
Improve coverage of restricted interaction test views

marcofucci:
This:

- adds a few more cases of non-restricted users accessing the interactions API
- fixes a bug in a test that did not use the right authentication when testing restricted interactions API

### DIFF
--- a/datahub/interaction/test/views/test_interaction.py
+++ b/datahub/interaction/test/views/test_interaction.py
@@ -1,5 +1,6 @@
 from datetime import date
 from functools import partial
+from itertools import chain
 
 import pytest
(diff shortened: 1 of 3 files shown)


datahub/interaction/test/views/test_interaction.py
```diff
@@ -24,10 +22,44 @@
 )
 
 
+NON_RESTRICTED_READ_PERMISSIONS = (
```
elcct: It looks like `test_common.py` wants to import that from `utils.py`.

marcofucci: 😬 thanks!

PR #926, opened 2018-04-20
Preparation for daily Elasticsearch sync

reupen:
Issue number: DH-1660

### Description of change

This makes a few changes in preparation for the ES sync Celery task:

1. Moves ES index creation and the updating of the mapping to a dedicated management command.

This stops the initialisation being done unnecessarily (e.g. for manage.py migrate), and also stops the Elasticsearch connection being shared amongst Celery workers (due to how elasticsearch-dsl manages connections, and how Celery forks). 

2. Changes sync_es functions to use a logger.

This is so that they can be used outside of the Django management command infrastructure.

3. Changes the Celery worker log format to include the logger name.

4. Sets the Elasticsearch logger log level to WARNING for Celery tasks (to cut down on noise).

### Checklist

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
-web: python manage.py collectstatic --noinput && python manage.py distributed_migrate --noinput && gunicorn config.wsgi --config config/gunicorn.py
+web: python manage.py collectstatic --noinput && python manage.py distributed_migrate --noinput && python manage.py init_es && gunicorn config.wsgi --config config/gunicorn.py
 init_rev: python manage.py createinitialrevisions
 celeryworker: celery worker -A config -l info
 celerybeat: celery beat -A config -l info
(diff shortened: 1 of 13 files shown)


Procfile
```diff
@@ -1,4 +1,4 @@
-web: python manage.py collectstatic --noinput && python manage.py distributed_migrate --noinput && gunicorn config.wsgi --config config/gunicorn.py
+web: python manage.py collectstatic --noinput && python manage.py distributed_migrate --noinput && python manage.py init_es && gunicorn config.wsgi --config config/gunicorn.py
```
marcofucci: do you think we could avoid this?

reupen: Could use a shell script if that's what you mean?

marcofucci: I meant, we run it manually when needed instead of running it every time, not sure if it's a good idea.

reupen: I think it's safe and it's equivalent to what we were doing. (There's a post_compile hook in the build pack, but I don't think that's a good place for it.)

PR #928, opened 2018-04-23
Command to update global hq.

elcct:
Issue number: DH-1608

### Description of change

This adds a command that updates given company's global hq if global hq doesn't already exist. 
It adds additional parameter `overwrite` which cause command to write global hq regardless if it exists. 

I also moved `simulate` param to the base command class as it was present in all commands.

### Checklist

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -21,6 +21,7 @@ from datahub.core.constants import HeadquarterType
 from datahub.core.serializers import (
     NestedRelatedField, PermittedFieldsModelSerializer, RelaxedURLField
 )
+from datahub.core.validate_utils import DataCombiner
 from datahub.core.validators import (
     AddressValidator,
     AllIsBlankRule,
@@ -236,7 +237,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         'uk_establishment_not_in_uk': ugettext_lazy(
             'A UK establishment (branch of non-UK company) must be in the UK.'
         ),
-        'global_headquarters_company_is_not_a_global_headquarters': ugettext_lazy(
+        'headquarter_type_is_not_global_headquarters': ugettext_lazy(
             'Company to be linked as global headquarters must be a global headquarters.'
         ),
         'invalid_global_headquarters': ugettext_lazy(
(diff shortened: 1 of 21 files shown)


datahub/dbmaintenance/management/commands/update_company_global_hq.py
```diff
@@ -0,0 +1,51 @@
...
+class Command(CSVBaseCommand):
+    """Command to update Company.global_headquarters."""
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--overwrite',
+            action='store_true',
+            dest='overwrite',
```
reupen: Shouldn't need to pass `dest` here.

datahub/dbmaintenance/management/commands/update_company_global_hq.py
```diff
@@ -0,0 +1,51 @@
...
+        if overwrite:
+            return True
+
+        # Assume companies with a current Global HQ are correct,
+        # as this data did not come from CDMS
+        return company.global_headquarters is None
+
+    def _process_row(self, row, simulate=False, overwrite=False, **options):
+        """Process one single row."""
+        company = Company.objects.get(pk=row['id'])
```
reupen: Could use `parse_uuid()` for `row['id']` here.

datahub/dbmaintenance/management/commands/update_company_global_hq.py
```diff
@@ -0,0 +1,51 @@
...
+        global_hq = None
+        global_hq_id = parse_uuid(row['global_hq_id'])
+        if global_hq_id is not None:
+            global_hq = Company.objects.get(pk=global_hq_id)
+
+        if self._should_update(company, overwrite=overwrite):
+            company.global_headquarters = global_hq
```
reupen: Any reason for not setting `global_headquarters_id` instead?

datahub/dbmaintenance/management/commands/update_company_global_hq.py
```diff
@@ -0,0 +1,51 @@
...
+            if simulate:
+                return
+
+            with reversion.create_revision():
+                company.save(
+                    update_fields=(
+                        'global_headquarters',
+                    )
+                )
+                reversion.set_comment('Global HQ data migration.')
```
reupen: Is it a data migration?

datahub/dbmaintenance/management/commands/update_company_global_hq.py
elcct: Good point - updated.

datahub/dbmaintenance/management/commands/update_company_global_hq.py
elcct: I forgot about that one, sure there is no reason for that. Fixed. Thanks!

datahub/dbmaintenance/management/commands/update_company_global_hq.py
elcct: It is a data correction. Fixed.

datahub/dbmaintenance/management/commands/update_company_global_hq.py
elcct: Updated, thanks!

marcofucci:
I think it's more complicated than this :(

records in datahub can be marked very incorrectly with means that:
- a subsidiary company could be marked as Global HQ
- there could be more companies marked as subsidiary than the one in the spreadsheet
- the selected company might be subsidiaries of different Global HQ companies

Therefore the script should:
- mark the right company as global HQ unmarking the old ones if necessary
- unmark all the subsidiaries and mark the ones in the spreadsheet

elcct:
@marcofucci assumption is that the subsidiaries data already in Data Hub is correct, as this feature has been enabled only recently (it is not old data). This command is only adding subsidiaries to the global HQ (chosen by us) and we already have a command that changes headquarter type.
We can use `overwrite` option which will "move" subsidiaries to the chosen global HQ if there is a need.

reupen:
Perhaps we could use CompanySerializer to do the updates for this one? Then, if there are any consistency errors, we'll at least know about them. (It might mean that we have to update modified_on for these updates, but I'm not sure that's necessarily wrong in this case.)

jim68000:
Agree with point 1 - it's an implicit part of the model that a subsidiary can only belong to a global hq, so it should be an error if this is not true

As far as override goes - my assumption is that any companies that re currently marked as subsidiaries of a global hq are correct - this is not data we migrated from CDMS, it's only possible to create this in Data Hub, so if users have created subsidiaries we should respect their knowledge. However there *may* be cases where we need to overwrite the user data, so override remains optional.

And as far as who creates the data - I'm guessing this will be the output of the cleanup tool?

marcofucci:
@jim68000 👍 

> And as far as who creates the data - I'm guessing this will be the output of the cleanup tool?

Ehm, unfortunately they said that they don't have enough time to help us sort out subsidiaries as well but they can just commit to One List specific matching tasks.

We might have to create a separate tool to send to individual global account managers and ask them to sort out their company's subsidiaries instead.

elcct:
@reupen @marcofucci now it uses `CompanySerializer` to validate the data. Also added extra validation for `Company` that was missing.

datahub/company/serializers.py
```diff
@@ -353,18 +353,24 @@ def validate_global_headquarters(self, global_headquarters):
...
-                )
-
             # checks if global_headquarters is not pointing to an instance of the model
             if self.instance == global_headquarters:
                 raise serializers.ValidationError(
                     self.error_messages['invalid_global_headquarters']
                 )
 
+            # checks if instance is global headquarters
+            if self.instance.headquarter_type_id == UUID(HeadquarterType.ghq.value.id):
```
reupen: The logic seems to be almost the same as `_validate_cannot_become_ghq_if_linked_to_another_ghq()`. 

I think we need to be careful getting things directly from the instance since there's nothing blocking a request updating that field as well. Since this rule involves more than one field (`headquarter_type` and `global_headquarters`) I think we should do it in `validate()` (or a validator on the class) using DataCombiner. We can still only make it perform that bit of validation if one of the relevant fields is being updated.

datahub/dbmaintenance/management/commands/update_company_global_hq.py
```diff
@@ -30,18 +31,26 @@ def _should_update(self, company, overwrite=False):
...
         if self._should_update(company, overwrite=overwrite):
-            company.global_headquarters_id = global_hq_id
-
+            global_hq_id = parse_uuid(row['global_hq_id'])
+            global_hq = {
+                'id': global_hq_id
+            } if global_hq_id is not None else None
+
+            data = {
+                'global_headquarters': global_hq,
```
reupen: I guess we need to clear `modified_by` if we're updating `modified_on`?

datahub/company/serializers.py
elcct: I changed the validation so now it is via `validate`.

datahub/dbmaintenance/management/commands/update_company_global_hq.py
elcct: I added clearing `modified_by`, so it is set to None.

datahub/company/serializers.py
```diff
@@ -310,62 +311,87 @@ class CompanySerializer(PermittedFieldsModelSerializer):
...
+        if total_subsidiaries > 0:
+            message = self.error_messages['global_headquarters_has_subsidiaries']
+            errors['headquarter_type'] = message
+
+        return errors
+
+    def _validate_cannot_become_ghq_if_linked_to_another_ghq(self, combiner):
+        errors = {}
+
+        headquarter_type_id = combiner.get_value_id('headquarter_type')
```
reupen: If `self.partial is True` I think we should only do this bit of validation if `headquarter_type` or `global_headquarters` is in `self.data`. So that the error doesn't trigger when saving a form without those fields.

datahub/company/serializers.py
```diff
@@ -310,62 +311,87 @@ class CompanySerializer(PermittedFieldsModelSerializer):
...
-        self,
-        headquarter_type_id
-    ):
-        """Raises an exception if new headquarter type is ghq and company is linked to ghq."""
+            message = self.error_messages['invalid_global_headquarters']
+            errors['global_headquarters'] = message
+
+        return errors
+
+    def _validate_ghq_has_ghq_headquarter_type(self, combiner):
```
reupen: This doesn't involve more than one field. So, it was fine in validate_global_headquarters() I think.

datahub/dbmaintenance/test/commands/test_update_company_global_hq.py
```diff
@@ -0,0 +1,365 @@
...
+    for company in companies:
+        company.refresh_from_db()
+
+    # Missing from data, fail
+    assert 'Company matching query does not exist' in caplog.text
+    assert len(caplog.records) == 1
+
+    assert company_needs_global_hq.global_headquarters == global_hq
+    assert company_should_get_new_global_hq.global_headquarters == global_hq
+    assert company_should_have_global_hq_removed.global_headquarters is None
```
marcofucci: It would be nice to check that `modified_by` is `None`.

datahub/company/serializers.py
```diff
@@ -310,62 +311,87 @@ class CompanySerializer(PermittedFieldsModelSerializer):
...
         ):
-            self._validate_cannot_remove_ghq_if_it_has_subsidiaries()
-            self._validate_cannot_become_ghq_if_linked_to_another_ghq(headquarter_type_id)
+            message = self.error_messages['subsidiary_cannot_be_a_global_headquarters']
+            errors['headquarter_type'] = message
 
-        return headquarter_type
+        return errors
+
+    def _validate_ghq_is_not_self(self, combiner):
```
marcofucci: What is the convention behind `_validate*` or `validate*` methods? 
I like similar methods to raise a validation exception and/or don't return anything but I see that some (core?) drf sometimes return some kind of data as well. Here we are returning errors which seems kind of weird.
Should we stick to one convention?

I'd say definitely raise exception if there's an error and I'm easy on returning the value or not (I slightly prefer not to return anything).

datahub/company/serializers.py
```diff
@@ -313,85 +313,60 @@ class CompanySerializer(PermittedFieldsModelSerializer):
...
-            self._validate_ghq_is_not_self,
-            self._validate_ghq_has_ghq_headquarter_type,
-        )
-
-        for validator in validators:
-            errors.update(validator(combiner))
-
-        if errors:
-            raise serializers.ValidationError(errors)
+        if any({'global_headquarters', 'headquarter_type'} & set(data)):
```
reupen: Should be able to simplify this to:
`if {'global_headquarters', 'headquarter_type'} & data.keys():`

datahub/company/serializers.py
```diff
@@ -236,7 +237,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         'uk_establishment_not_in_uk': ugettext_lazy(
             'A UK establishment (branch of non-UK company) must be in the UK.'
         ),
-        'global_headquarters_company_is_not_a_global_headquarters': ugettext_lazy(
+        'headquarter_type_is_not_global_headquarters': ugettext_lazy(
```
reupen: Not sure the new error code is better than the old one. What about combining them to make something like `global_headquarters_hq_type_is_not_global_headquarters`?

datahub/company/serializers.py
elcct: That's right. Done. Thanks!

datahub/company/serializers.py
elcct: This one is better. Changed. Thanks!

elcct:
@marcofucci I have two remaining commands to update name and trading_name (alias) that accept CSV. Is that also not needed? (those are to clean up company names from ghq etc)

PR #929, opened 2018-04-24
Schedule the full ES sync to run daily

reupen:
Issue number: DH-1660

### Description of change

This adds a couple of tasks for syncing all objects for a particular model, and schedules it to run overnight.

The new tasks are routed to a different Celery queue to separate long-running tasks from other tasks. Also a couple of extra Celery arguments were added to help avoid long-running tasks from hogging workers. We could also add additional commands to the Procfile so that we can start Celery worker processes with different arguments (e.g. for different queues), though I haven't done this for now as I think our needs remain fairly simple.

### Checklist

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -247,6 +247,11 @@ if REDIS_BASE_URL:
             'ssl_cert_reqs': ssl.CERT_NONE
         }
         CELERY_BROKER_USE_SSL = CELERY_REDIS_BACKEND_USE_SSL
+    CELERY_TASK_ROUTES = {
+        'datahub.search.tasks.*': {
+            'queue': 'long-running'
+        }
+    }
     CELERY_BEAT_SCHEDULE = {
         'refresh_pending_payment_gateway_sessions': {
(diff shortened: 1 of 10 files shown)


config/settings/common.py
```diff
@@ -247,6 +248,18 @@
...
+
+    # Increase timeout from one hour for long-running tasks
+    # (If the timeout is reached before a task, Celery will start it again. This
+    # would affect in particular any long-running tasks using acks_late=True.)
+    CELERY_BROKER_TRANSPORT_OPTIONS = {
+        'visibility_timeout': int(timedelta(hours=9).total_seconds())
+    }
+    CELERY_TASK_ROUTES = {
+        'datahub.search.tasks.*': {
+            'queue': 'long-running'
```
marcofucci: Do we want all future "search" tasks to run in the long-running queue? I imagine we will move single ES syncing to celery as well when an object changes and those are quick tasks.

reupen: No, I don't think we will. I suppose I can make it more specific now, in case we forget to change it later on :+1:

PR #931, opened 2018-04-25
Feature/update stages

jim68000:
Issue number: 

### Description of change

A command that allows us to change investment project stages via the management console. 

As this is my first attempt at committing something to Leeloo can you all take a _really hard look at what I've done please?_

### Checklist

### DIFF
--- a/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
+++ b/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
@@ -36,9 +36,7 @@ class Command(CSVBaseCommand):
         """Process one single row."""
         investment_project = InvestmentProject.objects.get(pk=row['id'])
 
-        old_stage = self.get_stage(row['old_stage'])
         new_stage = self.get_stage(row['new_stage'])
-
 
         investment_project.stage = new_stage
         if not simulate:
--- a/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
+++ b/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
@@ -53,9 +53,6 @@ class Command(CSVBaseCommand):
         investment_project.stage = self.get_stage(stage_id)
         investment_project.modified_by = self.get_adviser(adviser_id)
 
-        # if current_adviser:
-        #     investment_project.modified_by = current_adviser
-
         if not simulate:
             with reversion.create_revision():
                 investment_project.save(
--- a/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
+++ b/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from functools import lru_cache
 
 import reversion
@@ -23,17 +22,14 @@ class Command(CSVBaseCommand):
             help='If True it only simulates the command without saving the changes.',
         )
         parser.add_argument(
-            '--user',
-            default=False,
-            dest='user',
-            action='store',
+            '--adviser',
             help='UUID of a valid adviser to use as the modified_by value',
         )
 
     @lru_cache(maxsize=None)
     def get_stage(self, stage_id):
         """
-        :param company_id: uuid of the company
+        :param stage_id: uuid of the Investment project stage
         :return: instance of investment project stage with id == stage_id if it exists,
             None otherwise
         """
@@ -43,37 +39,34 @@ class Command(CSVBaseCommand):
 
     def get_adviser(self, adviser_id):
         """
-        :param adviser: uuid of the company
-        :return: instance of a data hub adviser,
+        :param adviser: uuid of an adviser
+        :return: instance of an adviser if they exist,
             None otherwise
         """
         if not adviser_id or adviser_id == 'null':
             return None
         return Advisor.objects.get(id=adviser_id)
 
-    def _process_row(self, row, simulate=False, user=False, **options):
+    def _process_row(self, row, simulate=False, adviser=False, **options):
         """Process one single row."""
-        id = parse_uuid(row['id'])
-        stage = parse_uuid(row['new_stage'])
-        adviser = parse_uuid(user)
+        investment_project_id = parse_uuid(row['investment_project_id'])
+        stage_id = parse_uuid(row['stage_id'])
+        adviser_id = parse_uuid(adviser)
 
-        investment_project = InvestmentProject.objects.get(pk=id)
+        investment_project = InvestmentProject.objects.get(pk=investment_project_id)
 
-        new_stage = self.get_stage(stage)
-        dh_user = self.get_adviser(adviser)
-        if not dh_user:
-            dh_user = investment_project.modified_by
-
+        new_stage = self.get_stage(stage_id)
+        current_adviser = self.get_adviser(adviser_id) or investment_project.modified_by
         investment_project.stage = new_stage
-        investment_project.modified_on = datetime.now()
-        investment_project.modified_by = dh_user
+
+        if current_adviser:
+            investment_project.modified_by = current_adviser
 
         if not simulate:
             with reversion.create_revision():
                 investment_project.save(
                     update_fields=(
                         'stage',
-                        'modified_on',
                         'modified_by'
                     )
                 )
--- a/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
+++ b/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
@@ -1,0 +1,67 @@
+from functools import lru_cache
+
+import reversion
+
+from datahub.investment.models import InvestmentProject
+from datahub.metadata.models import InvestmentProjectStage
+from ..base import CSVBaseCommand
+
+
+class Command(CSVBaseCommand):
+    """Command to update investment_project.stage."""
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            dest='simulate',
+            default=False,
+            help='If True it only simulates the command without saving the changes.',
+        )
+
+    @lru_cache(maxsize=None)
+    def get_stage(self, stage_id):
+        """
+        :param company_id: uuid of the company
+        :return: instance of Company with id == company_id if it exists,
+            None otherwise
+        """
+        if not stage_id or stage_id.lower().strip() == 'null':
+            return None
+        return InvestmentProjectStage.objects.get(id=stage_id)
+
+    def _should_update(self,
+                       investment_project,
+                       old_stage,
+                       ):
+        """
+        Checks if Investment project should be updated.
+
+        :param investment_project: instance of InvestmentProject
+        :param old_stage: instance of InvestmentProjectStage or none
+        :return: True if investment project needs to be updated
+        """
+        return investment_project.stage == old_stage
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        investment_project = InvestmentProject.objects.get(pk=row['id'])
+
+        old_stage = self.get_stage(row['old_stage'])
+        new_stage = self.get_stage(row['new_stage'])
+
+        if self._should_update(
+            investment_project,
+            old_stage,
+        ):
+            investment_project.stage = new_stage
+            if not simulate:
+                with reversion.create_revision():
+                    investment_project.save(
+                        update_fields=(
+                            'stage',
+                        )
+                    )
+                    reversion.set_comment('Stage migration.')
--- a/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
+++ b/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
@@ -1,7 +1,10 @@
+from datetime import datetime
 from functools import lru_cache
 
 import reversion
 
+from datahub.company.models import Advisor
+from datahub.dbmaintenance.utils import parse_uuid
 from datahub.investment.models import InvestmentProject
 from datahub.metadata.models import InvestmentProjectStage
 from ..base import CSVBaseCommand
@@ -16,34 +19,62 @@ class Command(CSVBaseCommand):
         parser.add_argument(
             '--simulate',
             action='store_true',
-            dest='simulate',
             default=False,
             help='If True it only simulates the command without saving the changes.',
+        )
+        parser.add_argument(
+            '--user',
+            default=False,
+            dest='user',
+            action='store',
+            help='UUID of a valid adviser to use as the modified_by value',
         )
 
     @lru_cache(maxsize=None)
     def get_stage(self, stage_id):
         """
         :param company_id: uuid of the company
-        :return: instance of Company with id == company_id if it exists,
+        :return: instance of investment project stage with id == stage_id if it exists,
             None otherwise
         """
-        if not stage_id or stage_id.lower().strip() == 'null':
+        if not stage_id or stage_id == 'null':
             return None
         return InvestmentProjectStage.objects.get(id=stage_id)
 
-    def _process_row(self, row, simulate=False, **options):
-        """Process one single row."""
-        investment_project = InvestmentProject.objects.get(pk=row['id'])
+    def get_adviser(self, adviser_id):
+        """
+        :param adviser: uuid of the company
+        :return: instance of a data hub adviser,
+            None otherwise
+        """
+        if not adviser_id or adviser_id == 'null':
+            return None
+        return Advisor.objects.get(id=adviser_id)
 
-        new_stage = self.get_stage(row['new_stage'])
+    def _process_row(self, row, simulate=False, user=False, **options):
+        """Process one single row."""
+        id = parse_uuid(row['id'])
+        stage = parse_uuid(row['new_stage'])
+        adviser = parse_uuid(user)
+
+        investment_project = InvestmentProject.objects.get(pk=id)
+
+        new_stage = self.get_stage(stage)
+        dh_user = self.get_adviser(adviser)
+        if not dh_user:
+            dh_user = investment_project.modified_by
 
         investment_project.stage = new_stage
+        investment_project.modified_on = datetime.now()
+        investment_project.modified_by = dh_user
+
         if not simulate:
             with reversion.create_revision():
                 investment_project.save(
                     update_fields=(
                         'stage',
+                        'modified_on',
+                        'modified_by'
                     )
                 )
-                reversion.set_comment('Stage migration.')
+                reversion.set_comment('Stage correction.')
--- a/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
+++ b/datahub/dbmaintenance/management/commands/update_investment_project_stage.py
@@ -16,13 +16,7 @@ class Command(CSVBaseCommand):
         """Define extra arguments."""
         super().add_arguments(parser)
         parser.add_argument(
-            '--simulate',
-            action='store_true',
-            default=False,
-            help='If True it only simulates the command without saving the changes.',
-        )
-        parser.add_argument(
-            '--adviser',
+            '--modified_by',
             help='UUID of a valid adviser to use as the modified_by value',
         )
 
@@ -33,7 +27,7 @@ class Command(CSVBaseCommand):
         :return: instance of investment project stage with id == stage_id if it exists,
             None otherwise
         """
-        if not stage_id or stage_id == 'null':
+        if not stage_id:
             return None
         return InvestmentProjectStage.objects.get(id=stage_id)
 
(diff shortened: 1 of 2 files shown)


datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,49 @@
...
+
+        investment_project.stage = new_stage
+        if not simulate:
+            with reversion.create_revision():
+                investment_project.save(
+                    update_fields=(
+                        'stage',
+                    )
+                )
+                reversion.set_comment('Stage migration.')
```
elcct: Is this a migration? Maybe correction would be more appropriate?

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,49 @@
...
+            action='store_true',
+            dest='simulate',
+            default=False,
+            help='If True it only simulates the command without saving the changes.',
+        )
+
+    @lru_cache(maxsize=None)
+    def get_stage(self, stage_id):
+        """
+        :param company_id: uuid of the company
```
elcct: I think comment should be updated as this is not looking for `company_id`.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,49 @@
...
+        :return: instance of Company with id == company_id if it exists,
+            None otherwise
+        """
+        if not stage_id or stage_id.lower().strip() == 'null':
+            return None
+        return InvestmentProjectStage.objects.get(id=stage_id)
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        investment_project = InvestmentProject.objects.get(pk=row['id'])
```
reupen: I'd use parse_uuid() from ` datahub.dbmaintenance.utils` for `row['id']` and `row['new_stage']` since it's a bit more robust.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,49 @@
...
+class Command(CSVBaseCommand):
+    """Command to update investment_project.stage."""
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            dest='simulate',
```
elcct: That should be derived from `--simulate`, so `dest` is not needed.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,49 @@
...
+        """Process one single row."""
+        investment_project = InvestmentProject.objects.get(pk=row['id'])
+
+        new_stage = self.get_stage(row['new_stage'])
+
+        investment_project.stage = new_stage
+        if not simulate:
+            with reversion.create_revision():
+                investment_project.save(
+                    update_fields=(
```
reupen: Do we still want modified_on (and modified_by) to remain unchanged for these kind of updates (when we aren't migrating data from a legacy system)?

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Done. I should probably set a slightly more readable colour scheme in PyCharm

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Done

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Good point. WIll correct

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Done

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Who should be the modified_by user - is there a system user?

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Done

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
reupen: `None` is an option, though passing it in somehow (as you said on Slack) is another one.

elcct: In the command I work on I set it to `None`. When running a command you could set any user you can think of, so I am not sure that would be much of help to have it there.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
marcofucci: Maybe if you push your changes, it might be easier to review the latest version.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            default=False,
+            help='If True it only simulates the command without saving the changes.',
+        )
+        parser.add_argument(
+            '--user',
+            default=False,
+            dest='user',
```
reupen: `dest='user'` and `action='store'` should be the defaults, so those can be omitted. Also None might make more sense for default (and that could also be omitted then as it's the default).

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        id = parse_uuid(row['id'])
+        stage = parse_uuid(row['new_stage'])
+        adviser = parse_uuid(user)
+
+        investment_project = InvestmentProject.objects.get(pk=id)
+
+        new_stage = self.get_stage(stage)
+        dh_user = self.get_adviser(adviser)
+        if not dh_user:
+            dh_user = investment_project.modified_by
```
reupen: It might be better to leave it as None in this case.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+
+        investment_project.stage = new_stage
+        investment_project.modified_on = datetime.now()
```
reupen: I'd expect modified_on to be set automatically if it's in update_fields. So this line may not be doing anything (check that though).

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
marcofucci: still showing up as `company_id`...

datahub/dbmaintenance/test/commands/test_update_investment_project_stage.py
```diff
@@ -0,0 +1,176 @@
...
+        {
+            'Body': BytesIO(bytes(csv_content, encoding='utf-8'))
+        },
+        expected_params={
+            'Bucket': bucket,
+            'Key': object_key
+        }
+    )
+    # set an explicit adviser id to make the changes
+    call_command('update_investment_project_stage', bucket, object_key, user=str(new_adviser.id))
```
reupen: Would also use freeze_time (from freezegun) so that you can assert on modified_on as well.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+            None otherwise
+        """
+        if not adviser_id or adviser_id == 'null':
+            return None
+        return Advisor.objects.get(id=adviser_id)
+
+    def _process_row(self, row, simulate=False, user=False, **options):
+        """Process one single row."""
+        id = parse_uuid(row['id'])
+        stage = parse_uuid(row['new_stage'])
```
marcofucci: `stage_id` might be a better name?

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        :return: instance of a data hub adviser,
+            None otherwise
+        """
+        if not adviser_id or adviser_id == 'null':
+            return None
+        return Advisor.objects.get(id=adviser_id)
+
+    def _process_row(self, row, simulate=False, user=False, **options):
+        """Process one single row."""
+        id = parse_uuid(row['id'])
```
marcofucci: `investment_project_id` might be a better name?

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        stage = parse_uuid(row['new_stage'])
+        adviser = parse_uuid(user)
```
marcofucci: `adviser_id` might be a better name?

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+    @lru_cache(maxsize=None)
+    def get_stage(self, stage_id):
+        """
+        :param company_id: uuid of the company
+        :return: instance of investment project stage with id == stage_id if it exists,
+            None otherwise
+        """
+        if not stage_id or stage_id == 'null':
```
marcofucci: I don't think this is ever going to be `null` as the `stage_id` param goes through `parse_uuid` which should already deal with that case if I'm not wrong.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+            return None
+        return InvestmentProjectStage.objects.get(id=stage_id)
+
+    def get_adviser(self, adviser_id):
+        """
+        :param adviser: uuid of the company
```
marcofucci: the comment should say `adviser_id` and _adviser_ not _company_

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        :return: instance of a data hub adviser,
```
marcofucci: `data hub` is probably unnecessary but not a big deal :)

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+            None otherwise
+        """
+        if not adviser_id or adviser_id == 'null':
```
marcofucci: Same as before, I don't think you'll ever get `adviser_id == 'null'`.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,80 @@
...
+        investment_project = InvestmentProject.objects.get(pk=id)
+
+        new_stage = self.get_stage(stage)
+        dh_user = self.get_adviser(adviser)
+        if not dh_user:
+            dh_user = investment_project.modified_by
+
+        investment_project.stage = new_stage
+        investment_project.modified_on = datetime.now()
+        investment_project.modified_by = dh_user
```
marcofucci: Shouldn't this be `None` if not specified via command param? In theory in incorrect to say that `investment_project.modified_by` made this change as we are changing `investment_project.modified_on` in any case.

datahub/dbmaintenance/test/commands/test_update_investment_project_stage.py
```diff
@@ -0,0 +1,176 @@
...
+
+    for investment_project in investment_projects:
+        investment_project.refresh_from_db()
+
+    # Adviser should change to reflect the command's user
+    assert investment_projects[0].modified_by == new_adviser
+    assert investment_projects[0].stage_id == (UUID(new_stages[0].value.id))
+    assert investment_projects[1].stage_id == (UUID(new_stages[1].value.id))
+    assert investment_projects[2].stage_id == (UUID(new_stages[2].value.id))
```
marcofucci: Probably minor but, it would be better to check `modified_by == new_adviser` on the other records as well.

marcofucci: `(UUID(new_stages[0].value.id))` you don't need the external round brackets.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: oops

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: removed

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: ok

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
elcct: and above in the comment should be `uuid of the adviser`, not company

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: Actually that's a good point - I'm not using old_stage at all. Removing that and renaming this.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: I've removed it so now have also removed the tests associated with it

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: done

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+from ..base import CSVBaseCommand
+
+
+class Command(CSVBaseCommand):
+    """Command to update investment_project.stage."""
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
```
reupen: If you update this branch against develop you can drop this as --simulate is added for all commands.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        :param adviser: uuid of an adviser
+        :return: instance of an adviser if they exist,
+            None otherwise
+        """
+        if not adviser_id or adviser_id == 'null':
+            return None
+        return Advisor.objects.get(id=adviser_id)
+
+    def _process_row(self, row, simulate=False, adviser=False, **options):
```
marcofucci: `adviser=None` would be better to keep the types consistent.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if current_adviser:
+            investment_project.modified_by = current_adviser
+
+        if not simulate:
+            with reversion.create_revision():
+                investment_project.save(
+                    update_fields=(
+                        'stage',
+                        'modified_by'
```
reupen: Still want `modified_on` in this list I think. (The previous comment was just saying that Django should set it automatically if it's in the fields being saved.)

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+            return None
+        return InvestmentProjectStage.objects.get(id=stage_id)
+
+    def get_adviser(self, adviser_id):
+        """
+        :param adviser: uuid of an adviser
+        :return: instance of an adviser if they exist,
+            None otherwise
+        """
+        if not adviser_id or adviser_id == 'null':
```
marcofucci: I still think that `adviser_id == 'null'` is not necessary as that case is covered by `parse_uuid`.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+    @lru_cache(maxsize=None)
+    def get_stage(self, stage_id):
+        """
+        :param stage_id: uuid of the Investment project stage
+        :return: instance of investment project stage with id == stage_id if it exists,
+            None otherwise
+        """
+        if not stage_id or stage_id == 'null':
```
marcofucci: I still think that `stage_id == 'null'` is not necessary as that case is covered by `parse_uuid`.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+            return None
+        return InvestmentProjectStage.objects.get(id=stage_id)
+
+    def get_adviser(self, adviser_id):
```
marcofucci: It would be good to `@lru_cache(maxsize=None)` this as well.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+    def _process_row(self, row, simulate=False, adviser=False, **options):
+        """Process one single row."""
+        investment_project_id = parse_uuid(row['investment_project_id'])
+        stage_id = parse_uuid(row['stage_id'])
+        adviser_id = parse_uuid(adviser)
+
+        investment_project = InvestmentProject.objects.get(pk=investment_project_id)
+
+        new_stage = self.get_stage(stage_id)
+        current_adviser = self.get_adviser(adviser_id) or investment_project.modified_by
```
reupen: I think we just want `investment_project.modified_by = self.get_adviser(adviser_id)` (to keep the record consistent).

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,73 @@
...
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            default=False,
+            help='If True it only simulates the command without saving the changes.',
+        )
+        parser.add_argument(
+            '--adviser',
```
marcofucci: why don't you call it `modified_by` instead of `adviser`?

elcct: do we really need to set `modified_by` at all? If that's the case wouldn't be good to add this to all commands where applicable, so that is consistent across the module? There is no validation to check if the adviser is the person who is running the command - that's why I am not sure about that.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: ok

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: ok

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: ok

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: ok

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: ok

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
jim68000: done

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,68 @@
...
+        investment_project_id = parse_uuid(row['investment_project_id'])
+        stage_id = parse_uuid(row['stage_id'])
+        adviser_id = parse_uuid(modified_by)
+
+        investment_project = InvestmentProject.objects.get(pk=investment_project_id)
+
+        investment_project.stage = self.get_stage(stage_id)
+        investment_project.modified_by = self.get_adviser(adviser_id)
+
+        # if current_adviser:
```
reupen: Can remove these commented out lines.

datahub/dbmaintenance/management/commands/update_investment_project_stage.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _process_row(self, row, simulate=False, modified_by=None, **options):
+        """Process one single row."""
+        investment_project_id = parse_uuid(row['investment_project_id'])
+        stage_id = parse_uuid(row['stage_id'])
+        adviser_id = parse_uuid(modified_by)
+
+        investment_project = InvestmentProject.objects.get(pk=investment_project_id)
+
+        investment_project.stage = self.get_stage(stage_id)
```
reupen: Just thought of something else – if the stage is going from Won to something else, the status will need to be knocked back from Won as well.

jim68000: That is a very good point.

PR #939, opened 2018-05-04
Add generic API client and Hawk request signing for requests to other APIs

reupen:
Issue number: N/A

### Description of change

This adds a generic API client and a Hawk authentication class compatible with requests.

These will be used for making requests to the company timeline API.

### Checklist

### DIFF
--- a/datahub/core/api_client.py
+++ b/datahub/core/api_client.py
@@ -1,0 +1,39 @@
+from logging import getLogger
+from urllib.parse import urljoin
+
+import requests
+
+
+logger = getLogger(__name__)
+
+
+class APIClient:
+    """Generic API client."""
+
+    # Prefer JSON to other content types
+    DEFAULT_ACCEPT = 'application/json;q=0.9,*/*;q=0.8'
+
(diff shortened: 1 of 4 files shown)


datahub/core/api_client.py
```diff
@@ -0,0 +1,80 @@
...
+        self._signing_algorithm = signing_algorithm
+        self._verify_response = verify_response
+
+    def __call__(self, request):
+        """Signs a request, and attaches a response verifier."""
+        sender = Sender({
+            'id': self._api_id,
+            'key': self._api_key,
+            'algorithm': self._signing_algorithm,
+        },
```
elcct: Is it possible to declare this dict outside of the Sender or improve formatting?

marcofucci:
> I've gone a bit further, perhaps, by making them all internal, but there should be no reason for anything else to use those attributes. For me, it's less noise as the public interface is smaller.

Yeah, that's what I meant, apart from `api_id` and `api_key` the other ones could be probably changed dynamically without crazy/weird problems.

PR #940, opened 2018-05-06
Added policy feedback to services metadata

ztolley:
The policy feedback service is required by the frontend to be able to allow users to add policy feedback interactions


reupen:
After some discussion, we:

1. Need the context filtering in the front end first
2. Will probably resurrect Ravi's PR for this https://github.com/uktrade/data-hub-leeloo/pull/890

PR #942, opened 2018-05-09
Add policy feedback service via migration

reupen:
Issue number: N/A

### Description of change

Reopening this (previously https://github.com/uktrade/data-hub-leeloo/pull/890) as context filtering has been implemented in the front end.

A migration seems like the easiest way to get the service in all environments as the front end will depend on a policy feedback service existing.

### Checklist


reupen:
> +1 so we could maybe get rid of the services from the admin?

No, that gets a fair bit of use. I'm not a huge fan of adding this via a migration but it seems like the best option available.

marcofucci:
Yeah I actually have to do something similar so I see what you mean.

PR #943, opened 2018-05-09
Investment Project Proposition API

elcct:
Issue number: DHI-37 DHI-38 DHI-45 DHI-46

### Description of change

This adds Proposition API for Investment Project. Proposition can be created, completed or abandoned. It shouldn't be possible to edit proposition once it has been created.
When completing or abandoning, user has to provide reason.
Once completed or abandoned its status shouldn't change.
API allows to get single proposition, list, create, complete and abandon.

### Checklist


datahub/investment/proposition/exceptions.py
```diff
@@ -0,0 +1,11 @@
+from django.utils.translation import ugettext_lazy as _
+from rest_framework import status
+from rest_framework.exceptions import APIException
+
+
+class Conflict(APIException):
```
reupen: Can we move the OMIS one to the core app instead of having two Conflict exception classes?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
...
+    completed_on = models.DateTimeField(blank=True, null=True)
+    completed_by = models.ForeignKey(
+        'company.Advisor', on_delete=models.CASCADE, related_name='+', null=True
+    )
+
+    def __str__(self):
+        """Human readable representation of the object."""
+        return self.name
+
+    @transaction.atomic
```
reupen: Is `@transaction.atomic` serving a purpose here? I can only see one database operation in the function body.

datahub/investment/proposition/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+
+from datahub.company.serializers import NestedAdviserField
+from datahub.core.serializers import NestedRelatedField
+from datahub.investment.proposition.models import Proposition
+
+
+class CreatePropositionSerializer(serializers.ModelSerializer):
+    """Proposition serialiser for create endpoint."""
+
+    investment_project = NestedRelatedField(
```
reupen: I think we should create a `NestedInvestmentProjectField` class (instead of specifying `extra_fields=('name', 'project_code')` each time there's a nested investment project).

datahub/investment/proposition/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+        'investment.InvestmentProject',
+        extra_fields=('name', 'project_code')
+    )
+    adviser = NestedAdviserField()
+
+    deadline = serializers.DateField()
```
reupen: Are these explicitly defined in order to make them required? I think we should use extra_kwargs on Meta for that:

1. It will preserve other arguments from the model field (e.g. max_length)
2. It's a bit more obvious which attributes are being overridden.

Same comment for the other serialisers.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+                'last_name': adviser.last_name,
+                'name': adviser.name,
+                'id': str(adviser.pk),
+            },
+            'deadline': '2018-02-10',
+            'status': 'ongoing',
+            'name': 'My proposition.',
+            'scope': 'Very broad scope.',
+            'reason_abandoned': '',
+            'created_on': instance.created_on.isoformat().replace('+00:00', 'Z'),
```
reupen: I think it makes more sense to use freeze_time for this. Also you can use format_date_or_datetime() from `datahub.core.test_utils`.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+            'created_by': {
+                'first_name': instance.created_by.first_name,
```
reupen: Similarly, I think it makes more sense to assert that the created_by user is the user that made the request.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+            'abandoned_by': None,
+            'completed_details': '',
+            'completed_on': None,
+            'completed_by': None
+        }
+
+    def test_cannot_created_with_fields_missing(self):
+        """Test that proposition cannot be created without required fields."""
+        url = reverse('api-v3:investment:proposition:collection')
+        response = self.api_client.post(
```
reupen: Should be able to fit this on one line.

datahub/investment/proposition/views.py
```diff
@@ -0,0 +1,89 @@
...
+from datahub.oauth.scopes import Scope
+
+MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class PropositionViewSet(CoreViewSet):
+    """ViewSet for public facing proposition endpoint."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
```
reupen: It's removing DjangoCrudPermission – shouldn't we enforce permissions on this? How do the association permissions for LEPs and DAs fit in to this?

datahub/investment/proposition/urls.py
```diff
@@ -0,0 +1,29 @@
...
+
+proposition_complete = PropositionViewSet.as_view({
+    'post': 'complete',
+})
+
+proposition_abandon = PropositionViewSet.as_view({
+    'post': 'abandon',
+})
+
+public_urls = [
```
reupen: Why `public_urls`?

datahub/investment/proposition/views.py
```diff
@@ -0,0 +1,89 @@
...
+        'investment_project',
+        'adviser',
+        'abandoned_by',
+        'completed_by',
+    )
+    filter_backends = (
+        DjangoFilterBackend,
+        OrderingFilter,
+    )
+    filter_fields = ('adviser_id', 'investment_project_id', 'status',)
```
reupen: Shouldn't we put propositions below `investment/<uuid:pk>/` instead of using a query param to filter by project?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
+"""Investment project proposition models."""
+
+import uuid
+
+from django.conf import settings
+from django.db import models, transaction
+from django.utils.timezone import now
+
+from datahub.core.models import (
```
reupen: This can fit on one line.

datahub/investment/proposition/serializers.py
```diff
@@ -0,0 +1,106 @@
...
+    investment_project = NestedRelatedField(
+        'investment.InvestmentProject',
+        extra_fields=('name', 'project_code')
+    )
+    adviser = NestedAdviserField()
+    deadline = serializers.DateField()
+    name = serializers.CharField()
+    created_by = NestedAdviserField()
+    abandoned_by = NestedAdviserField()
+    completed_by = NestedAdviserField()
```
marcofucci: do we need all these fields?

datahub/investment/proposition/test/factories.py
```diff
@@ -0,0 +1,31 @@
...
+
+class PropositionFactory(factory.django.DjangoModelFactory):
+    """Investment project proposition factory."""
+
+    id = factory.LazyFunction(uuid.uuid4)
+
+    investment_project = factory.SubFactory(InvestmentProjectFactory)
+    adviser = factory.SubFactory(AdviserFactory)
+
+    deadline = date(2018, 5, 20)
```
marcofucci: can we use `factory.Faker('future_date')` maybe?

datahub/investment/proposition/test/test_models.py
```diff
@@ -0,0 +1,12 @@
...
+import pytest
+
+from datahub.investment.proposition.test.factories import PropositionFactory
+
+pytestmark = pytest.mark.django_db
+
+
+def test_proposition():
+    """Tests that proposition can be created."""
+    PropositionFactory()
```
marcofucci: not sure we really need this but it's not a big deal to keep it I guess?

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+                'name': 'My proposition.',
+                'scope': 'Very broad scope.',
+                'adviser': adviser.pk,
+                'deadline': '2018-02-10',
+            },
+            format='json',
+        )
+        assert response.status_code == status.HTTP_201_CREATED
+        response_data = response.json()
+        instance = Proposition.objects.get(pk=UUID(response_data['id']))
```
marcofucci: I don't think you need the `UUID()` bit but it's find if you want.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+        response_data = response.json()
+        assert response_data == {
+            'investment_project': ['This field is required.'],
+            'adviser': ['This field is required.'],
+            'deadline': ['This field is required.'],
+            'name': ['This field is required.'],
+            'scope': ['This field is required.'],
+        }
+
+    def test_deadline_validation(self):
```
marcofucci: Why are you testing just the deadline and not the other fields? To be honest I don't even think we need this test as we rely on the `DateField` to validate it for us.

I would personally use a parametrised test for the fields required test above and maybe if you really want a few extra e.g. invalid investment_project, invalid adviser, invalid deadline

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+
+from datahub.company.test.factories import AdviserFactory
+from datahub.core.test_utils import APITestMixin
+from datahub.investment.proposition.models import Proposition
+from datahub.investment.test.factories import InvestmentProjectFactory
+from .factories import PropositionFactory
+
+
+class TestCreateProposition(APITestMixin):
+    """Base tests for the create proposition view."""
```
marcofucci: Why `Base`?

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        response_data = response.json()
+        assert response_data['deadline'] == [
+            'Date has wrong format. Use one of these formats instead: YYYY[-MM[-DD]].'
+        ]
+
+
+class TestUpdateProposition(APITestMixin):
+    """Base tests for the update proposition view."""
```
marcofucci: Why `Base`?

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+    """Tests for the list propositions view."""
+
+    def test_filtered_by_investment_project(self):
+        """List of propositions filtered by investment project."""
+        investment_project = InvestmentProjectFactory()
+
+        PropositionFactory.create_batch(3)
+        propositions = PropositionFactory.create_batch(
+            3, investment_project=investment_project
+        )
```
marcofucci: I would create a new more not linked to this `investment_project` to test the filtering bit.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+        assert actual_ids == expected_ids
+
+    def test_filtered_by_adviser(self):
+        """List of propositions filtered by adviser."""
+        adviser = AdviserFactory()
+
+        PropositionFactory.create_batch(3)
+        propositions = PropositionFactory.create_batch(
+            3, adviser=adviser
+        )
```
marcofucci: I would create a new more not linked to this `adviser` to test the filtering bit.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+        })
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['count'] == 1
+        assert response_data['results'][0]['status'] == proposition_status
+
+
+class TestGetProposition(APITestMixin):
+    """Base tests for get proposition view."""
```
marcofucci: Why `Base`?

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,422 @@
...
+            'abandoned_on': None,
+            'abandoned_by': None,
+            'completed_details': '',
+            'completed_on': None,
+            'completed_by': None,
+        }
+
+
+class TestCompleteProposition(APITestMixin):
+    """Base tests for the complete proposition view."""
```
marcofucci: Why `Base`?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+    name = models.CharField(max_length=MAX_LENGTH)
+    scope = models.TextField(blank=True)
+
+    reason_abandoned = models.TextField(blank=True)
+    abandoned_on = models.DateTimeField(blank=True, null=True)
+    abandoned_by = models.ForeignKey(
+        'company.Advisor', on_delete=models.CASCADE, related_name='+', null=True
+    )
```
marcofucci: I think you want `blank=True` as well.

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+    completed_details = models.TextField(blank=True)
+    completed_on = models.DateTimeField(blank=True, null=True)
+    completed_by = models.ForeignKey(
+        'company.Advisor', on_delete=models.CASCADE, related_name='+', null=True
+    )
```
marcofucci: I think you want `blank=True` as well.

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
...
+    """Proposition model."""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    investment_project = models.ForeignKey(
+        'investment.InvestmentProject', on_delete=models.CASCADE, related_name='proposition'
+    )
+    adviser = models.ForeignKey(
+        'company.Advisor', on_delete=models.CASCADE, related_name='+'
+    )
+    deadline = models.DateField(blank=True, null=True)
```
marcofucci: Isn't `deadline` always required?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
...
+    status = models.CharField(
+        max_length=MAX_LENGTH, choices=PropositionStatus, default=PropositionStatus.ongoing
+    )
+
+    name = models.CharField(max_length=MAX_LENGTH)
+    scope = models.TextField(blank=True)
```
marcofucci: Isn't `scope` always required?

datahub/investment/proposition/test/test_views.py
elcct: Left from copy and paste :blush: Will fix. Thanks!

datahub/investment/proposition/test/test_models.py
elcct: I forgot to delete that...

datahub/investment/proposition/test/test_views.py
elcct: That's true. I have removed it.

datahub/investment/proposition/serializers.py
elcct: I think you're right - since we can only modify this record once, `modified_by` would do. I'll fix. Thanks!

datahub/investment/proposition/urls.py
elcct: Yes, that makes no sense. Will correct. Thanks!

datahub/investment/proposition/models.py
elcct: That is true, it doesn't help. I'll remove it. Thanks!

datahub/investment/proposition/test/test_views.py
elcct: Isn't this achieved by previous line?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class Proposition(BaseModel):
+    """Proposition model."""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    investment_project = models.ForeignKey(
+        'investment.InvestmentProject', on_delete=models.CASCADE, related_name='proposition'
+    )
+    adviser = models.ForeignKey(
```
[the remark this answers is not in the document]
elcct: We just use `adviser` throughout propositions. That would be a person responsible for creating a proposition.

datahub/investment/proposition/models.py
elcct: I removed this field now.

datahub/investment/proposition/models.py
elcct: I removed this field now.

datahub/investment/proposition/test/test_views.py
marcofucci: oh missed that 👍

datahub/investment/proposition/models.py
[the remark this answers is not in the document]
marcofucci: So this is the same as `created_by`?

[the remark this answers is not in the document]
elcct: No, because that could be added by someone else.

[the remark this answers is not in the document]
elcct: I meant responsible for creating physical proposition, not on Data Hub.

datahub/investment/proposition/views.py
elcct: Good catch - I wasn't sure about that. Now confirmed that it should have the same permission structure as Interactions. I have copied over permission functionality from Interaction, but I think I need to add tests specifically for that.

datahub/investment/proposition/test/test_views.py
marcofucci: same for this, looks good.

datahub/investment/proposition/models.py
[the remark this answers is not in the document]
marcofucci: The thing that we are talking about it for so long suggest that there could be a better name for this field, I'm still not sure what this represents. Do you mean `assigned_to`?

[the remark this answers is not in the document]
elcct: That sounds good.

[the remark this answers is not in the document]
elcct: I changed field name to `assigned_to`. Thanks!

datahub/core/permissions.py
```diff
@@ -163,6 +166,9 @@ def get_model_action_for_view_action(http_method, view_action, many_to_many=Fals
     if http_method == 'OPTIONS':
         return 'read'
 
+    if view_action is None:
```
reupen: When was this happening?

datahub/core/permissions.py
```diff
@@ -15,6 +16,8 @@
     'archive': 'change',
     'unarchive': 'change',
     'metadata': 'read',
+    'complete': 'change',
+    'abandon': 'change',
```
marcofucci: These don't seem generic enough to be added to core. It would be better if these were defaults and the Permission class could override them  to add view specific ones.

datahub/investment/proposition/test/test_views.py
```diff
@@ -0,0 +1,488 @@
...
+
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        response_data = response.json()
+        assert response_data == {
+            'assigned_to': ['This field is required.'],
+            'deadline': ['This field is required.'],
+            'name': ['This field is required.'],
+            'scope': ['This field is required.'],
+        }
+
```
marcofucci: It would be nice to have some tests on missing permissions.

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+    The following codenames mean that the user can only read, change and add propositions for
+    investment projects that they are associated with:
+
+    read_associated_investmentproject_proposition
+    change_associated_investmentproject_proposition
+    add_associated_investmentproject_proposition
+
+    They cannot read, change or add propositions that do not relate to an investment project.
```
reupen: Don't all propositions relate to a project?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _change_status(self, status, by, details):
+        """Change status of a proposition."""
+        if self.status != PropositionStatus.ongoing:
+            raise APIConflictException(
+                f'The action cannot be performed in the current status {self.status}.'
+            )
+        self.status = status
+        self.modified_by = by
+        self.modified_on = now()
```
reupen: Doesn't modified_on update automatically?

datahub/investment/proposition/models.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    name = models.CharField(max_length=MAX_LENGTH)
+    scope = models.TextField()
+
+    details = models.TextField()
+
+    def __str__(self):
+        """Human readable representation of the object."""
+        return self.name
```
reupen: Should it also include the project's name for disambiguation?

datahub/investment/proposition/permissions.py
```diff
@@ -0,0 +1,122 @@
...
+        checker = InvestmentProjectPropositionAssociationChecker()
+        request = self.serializer.context['request']
+        view = self.serializer.context['view']
+
+        if not checker.should_apply_restrictions(request, view.action):
+            return
+
+        investment_project = attrs.get('investment_project')
+
+        if investment_project is None:
```
reupen: This shouldn't be required here as investment_project isn't optional.

datahub/investment/proposition/test/factories.py
```diff
@@ -0,0 +1,31 @@
...
+
+    id = factory.LazyFunction(uuid.uuid4)
+
+    investment_project = factory.SubFactory(InvestmentProjectFactory)
+    assigned_to = factory.SubFactory(AdviserFactory)
+
+    deadline = factory.Faker('future_date')
+    status = PropositionStatus.ongoing
+
+    name = factory.Sequence(lambda n: f'name {n}')
```
reupen: Would be nice to use factory.Faker() here.

datahub/investment/proposition/test/test_views.py
elcct: I have added more tests which uncovered some issues with the permissions on propositions that I have also fixed.

datahub/core/permissions.py
elcct: I have modified `get_model_action_for_view_action` so that it can be called with extra mappings by `InvestmentProjectAssociationCheckerBase`'s `should_apply_restrictions`. It is then overriden by `InvestmentProjectPropositionAssociationChecker` with `extra_view_to_action_mapping` related to Propositions.

datahub/investment/proposition/models.py
elcct: Removed. Thanks!

datahub/investment/proposition/models.py
elcct: That is true. Removed. Thanks!

datahub/investment/proposition/models.py
elcct: Added investment project's name. Thanks!

datahub/investment/proposition/permissions.py
elcct: That is true. I should have removed that. Thanks!

datahub/investment/proposition/test/factories.py
elcct: Updated with `factory.Faker('text')`. Thanks!

datahub/core/permissions.py
```diff
@@ -161,14 +159,27 @@ def _check_actual_object_permission(self, request, view, obj):
         return True
 
 
-def get_model_action_for_view_action(http_method, view_action, many_to_many=False):
+def get_model_action_for_view_action(
+    http_method,
+    view_action,
+    many_to_many=False,
+    extra_view_to_action_mapping=None,
+    extra_many_to_many_action_mapping=None
```
reupen: Not sure we need extra_view_to_action_mapping and extra_many_to_many_action_mapping, what about one argument for both to reduce complexity?

datahub/core/permissions.py
```diff
@@ -161,14 +159,27 @@ def _check_actual_object_permission(self, request, view, obj):
...
     if http_method == 'OPTIONS':
         return 'read'
 
     if view_action is None:
         raise APIMethodNotAllowedException()
 
-    mapping = _MANY_TO_MANY_VIEW_TO_ACTION_MAPPING if many_to_many else _VIEW_TO_ACTION_MAPPING
+    if many_to_many:
+        mapping = _MANY_TO_MANY_VIEW_TO_ACTION_MAPPING
+        if isinstance(extra_many_to_many_action_mapping, dict):
```
reupen: Makes more sense to check if it's not None here. Mainly because if something with the wrong type is passed, it shouldn't be silently ignored.

datahub/investment/proposition/test/test_views.py
reupen: This still seems relevant.

reupen:
I think some additional tests are needed for when the project ID in the URL is invalid.

elcct:
Thanks @reupen. That uncovered that I forgot to add a check for non existent project. That is fixed now and I added additional tests.

datahub/investment/proposition/models.py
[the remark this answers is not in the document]
yozzo: This change from a semantical point of view makes sense, however I'll need to push back on renaming `adviser` to `assigned_to` in this isolated case because on the front-end it creates technical debt: 
* preselecting the registered users in the names dropdown not working anymore 
* fetching the names will still be referring to Advisers
* other sub apps (interactions) use the adviser key
Unless we roll out this change to all apps and sub-apps, I would keep it the same across all of them. Yes, it's a really good name for this type of property, and it serves the purpose, but in our case it just adds an extra layer of complexity.

marcofucci:
@yozzo I'm not following, does it mean that we can't have fields in different resources with different names?
Can you add a couple of links to the different frontend places to highlight the problem?

datahub/core/permissions.py
```diff
@@ -158,11 +159,23 @@ def _check_actual_object_permission(self, request, view, obj):
...
+):
     """Gets the model action corresponding to a view action."""
     if http_method == 'OPTIONS':
         return 'read'
 
-    mapping = _MANY_TO_MANY_VIEW_TO_ACTION_MAPPING if many_to_many else _VIEW_TO_ACTION_MAPPING
+    if view_action is None:
+        raise APIMethodNotAllowedException()
+
+    mapping = _MANY_TO_MANY_VIEW_TO_ACTION_MAPPING \
```
reupen: This should fit on one line (as it was before).

datahub/core/permissions.py
```diff
@@ -158,11 +159,23 @@ def _check_actual_object_permission(self, request, view, obj):
...
+        if many_to_many else _VIEW_TO_ACTION_MAPPING
+
+    if extra_view_to_action_mapping is not None:
+        mapping.update(extra_view_to_action_mapping)
```
reupen: Won't this mutate _VIEW_TO_ACTION_MAPPING or _MANY_TO_MANY_VIEW_TO_ACTION_MAPPING?

yozzo:
@marcofucci I'm not saying we can't have fields with different names. What I'm saying is that Pawel and I started working on this feature at the same time and shared together a mock data structure that was using `adviser` for that property.  I did try to refactor with `assigned_to` on my end, but it's not a quick change, and will set us back timewise.

Same naming conventions are used for investment projects interactions 

https://www.datahub.staging.uktrade.io/investment-projects/0002d942-6199-e211-a939-e4115bead28a/interactions/create/interaction

and for investment team members 

https://www.datahub.staging.uktrade.io/investment-projects/eab95483-965d-e411-985c-e4115bead28a/edit-team-members

https://www.datahub.staging.uktrade.io/investment-projects/3d998321-ef74-e611-af73-e4115bead28a/team

datahub/core/permissions.py
elcct: It will. I should have use the copy of those dicts. That is now fixed and uncovered a problem where this extra mapping was not passed in `ViewBasedModelPermissions`. Fixed that as well. Thanks!

datahub/core/permissions.py
elcct: Now that I added copy it won't fit :(

datahub/core/permissions.py
reupen: OK, but PEP 8 style is to prefer wrapping in ( and ) for line continuation.

elcct: I am sorry, you're right. I have updated it. Thanks

marcofucci:
@yozzo oh I meant links to the bits of the actual code that makes refactoring harder, not the screens. Nothing wrong with the world `adviser` in other contexts where its meaning is obvious (e.g. teams as you posted).

It would be nice to keep the backend clean but I'll leave it to @elcct, I've hit the approve button.

PR #944, opened 2018-05-10
Add company timeline view

reupen:
Issue number: N/A

### Description of change

This adds a view to retrieve the timeline for a company using the reporting service.

A new route has been added at `/v3/company/<id>/timeline`. A new permission `company.read_company_timeline` has been also added.

### Checklist


datahub/company/timeline/views.py
```diff
@@ -0,0 +1,40 @@
...
+            f'company.{CompanyPermission.read_company}',
+            f'company.{CompanyPermission.read_company_timeline}',
+        ),
+    )
+    queryset = Company.objects.all()
+    serializer_class = TimelineEventSerializer
+
+    def list(self, request, *args, **kwargs):
+        """Lists timeline events (paginated)."""
+        company = self.get_object()
```
marcofucci: This (`self.get_object`, `queryset` etc.) works but it's a bit weird as it refers to a company model whilst this ViewSet is for timeline events instead. Not a big deal I suppose.

datahub/company/timeline/client.py
```diff
@@ -0,0 +1,80 @@
...
+        Initialises the client.
+
+        The API URL and key are taken from settings if not explicitly provided.
+        """
+        if api_url is None:
+            api_url = settings.REPORTING_SERVICE_API_URL
+        if api_id is None:
+            api_id = settings.REPORTING_SERVICE_API_ID
+        if api_key is None:
+            api_key = settings.REPORTING_SERVICE_API_KEY
```
marcofucci: I'm not sure `api_url=None, api_id=None, api_key=None` adds anything here, I suspect reading from settings should be enough. In tests, we can override settings to test stuff.

datahub/company/test/timeline/test_client.py
```diff
@@ -0,0 +1,128 @@
...
+                'data_source': 'companies_house.companies',
+                'datetime': 'Mon, 31 Dec 2017 00:00:00 GMT',
+                'description': 'Accounts filed',
+            }]
+        }
+    },
+    '/api/v1/company/events/?companies_house_id=125695': {
+        'status_code': status.HTTP_404_NOT_FOUND
+    },
+    '/api/v1/company/events/?companies_house_id=125696': {
```
marcofucci: A small thing but using a company number very different from `125695` would make scanning the code easier as it would make it obvious that the numbers are different.

datahub/company/test/timeline/test_views.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    def test_permission_is_denied(self, permission_codenames):
+        """Test that a 403 is returned for users without the correct permissions."""
+        user = create_test_user(permission_codenames=permission_codenames)
+        api_client = self.create_api_client(user=user)
+        company = CompanyFactory()
+        url = reverse('api-v3:company:timeline-collection', kwargs={'pk': company.pk})
+
+        response = api_client.get(url)
+        assert response.status_code == status.HTTP_403_FORBIDDEN
```
marcofucci: I would add another couple of tests:

1. when the response from the API is different from what we expect and the serializer.is_valid fails
2. when the response form the API returns a 500 error.

elcct: I agree with above.

reupen:
> I've added a couple of comments but I'm also worried that the name reporting service misleads a bit. Why don't we call it timeline service or something similar?

I started off calling it company timeline API, but I changed it because:

- the service is actually called something like data platform reporting service
- it does more than just company timeline

See https://github.com/uktrade/dt07-reporting/blob/develop/api.md.

So calling it company timeline API is wrong and won't make sense if we start using anything else from the service. Any other suggestions for names?

datahub/company/timeline/views.py
reupen: Yeah, it could be a method on CompanyViewSet in theory, but I went for keeping all the timeline stuff separate so that it's mostly self-contained. I agree that it's a little bit odd, but it does save us from duplicating that logic.

datahub/company/timeline/client.py
reupen: Have removed these now.

datahub/company/test/timeline/test_client.py
reupen: I've updated these.

datahub/company/test/timeline/test_views.py
reupen: I've added a couple more tests for those cases.

reupen:
I've renamed the settings and classes so that they refer to the 'data science company API' instead.

PR #946, opened 2018-05-14
Additional interaction service types.

elcct:
Issue number: N/A

### Description of change

This has Interaction Service types required for SPI.


marcofucci:
I'm not sure they look like services 🤔 
E.g. _Investment Enquiry – Transferred to LEP (IST use only)_, what does it mean?

elcct:
That means project has been given to someone else - LEP in this case. Each new project will eventually have an interaction with one of those eight outcomes. Two of those count towards SPI.

elcct:
Given limited time the goal was to reuse existing features to enable measurement of the SPI. I understand this is not ideal.

PR #947, opened 2018-05-15
Add update_one_list_fields command to update One List fields

marcofucci:
This adds a new django command to update `classification` and `one_list_account_owner` with the values from the CSV.

If `reset_unmatched` is set, all the records not in the CSV will have those fields set to None.

### DIFF
--- a/datahub/dbmaintenance/management/commands/update_one_list_fields.py
+++ b/datahub/dbmaintenance/management/commands/update_one_list_fields.py
@@ -1,0 +1,165 @@
+from functools import lru_cache
+from logging import getLogger
+
+import reversion
+from django.db.models import Q
+
+from datahub.company.models import Advisor, Company
+from datahub.dbmaintenance.utils import parse_uuid
+from datahub.metadata.models import CompanyClassification
+from ..base import CSVBaseCommand
+
+
+logger = getLogger(__name__)
+
+
+class Command(CSVBaseCommand):
+    """
+    Command to update:
+    - Company.classification
+    - Company.one_list_account_owner
+
+    If --reset-unmatched=true, all the records not in the CSV will have
+    the fields above set to None.
+    """
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--reset-unmatched',
+            action='store_true',
+            default=False,
+            help=(
+                'If true, Data Hub records not present '
+                'in the CSV will have classification and one_list_account_owner set to None.'
+            )
+        )
+
+    def _handle(self, *args, simulate=False, reset_unmatched=False, **options):
+        """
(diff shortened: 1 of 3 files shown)


datahub/dbmaintenance/management/commands/update_one_list_fields.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        company = Company.objects.get(pk=row['id'])
+        classification_id = parse_uuid(row['classification_id'])
+        one_list_account_owner_id = parse_uuid(row['one_list_account_owner_id'])
+
+        # remove company.pk from the list of companies to reset
+        self.companies_to_reset.pop(company.pk, None)
```
reupen: If there's an error earlier in the function, how does company get removed from `self.companies_to_reset`?

datahub/dbmaintenance/test/commands/test_update_one_list_fields.py
```diff
@@ -0,0 +1,277 @@
...
+    values against new ones.
+    """
+    for field in fields:
+        setattr(company, f'_prev_{field}', getattr(company, field))
+
+
+def assert_did_not_change(company, *fields):
+    """
+    Assert that the company fields did not change.
+    It assumes you called `save_prev_fields` before the potential change.
```
elcct: Maybe comment should be impersonal, like "It assumes `save_prev_fields` has been called before the potential change."

datahub/dbmaintenance/management/commands/update_one_list_fields.py
marcofucci: What kind of error? I could move this to the line immediately after getting the company maybe?

reupen: Any error, doesn't CSVBaseCommand.process_row() swallow up exceptions?

I'm not sure the ignore errors behaviour makes sense with --reset-unmatched.

marcofucci: I'm not sure to understand:

- `handle` calls `process_row` which wraps `_process_row` in a try/except
- we want to avoid resetting the company, even if there are exceptions so we still want to exclude that company from the `self.companies_to_reset` dict before any logic.

I wanted to wrap everything into an atomic transaction but keeping the same logic that we have always based on is nice.

I actually moved that line a bit further up which looks a lot better.

reupen: It's better with `self.companies_to_reset.pop(company.pk, None)` at the top but it's still weird to have that in an exception handler. I'll click approve though since there's less risk having it at the top.

marcofucci:
@elcct any more comments before I merge?

marcofucci:
@elcct any more comments before I merge?

PR #949, opened 2018-05-15
New Tier B - Capital Investment 

marcofucci:
This adds a new `Tier B - Capital Investment` and gets rid of the `CompanyClassification` constants from the tests.

The record has already been created in production and I will load the fixture in other environments manually.


fixtures/metadata/companies.yaml
```diff
@@ -2,21 +2,24 @@
...
+  fields: {name: "Tier A1 - Tomorrow's Champions"}
+- model: metadata.companyclassification
+  pk: 7e0c261a-d447-e411-985c-e4115bead28a
+  fields: {name: 'Tier A2 - Global Partners'}
 - model: metadata.companyclassification
   pk: bb1bf800-8d53-e311-aef3-441ea13961e2
   fields: {name: 'Tier B - Global Accounts'}
+- model: metadata.companyclassification
+  pk: 23ef2218-37f7-4abf-aacb-7c49f65ee1e3
+  fields: {name: 'Tier B - Global Accounts (Capital Investment)'}
```
reupen: Seems a bit confusing to have two tiers named tier B – no chance of a tier B1 and B2?

marcofucci: Good point, I've asked.

marcofucci: It looks like they want to keep it this way _Adding numbers to the tiers starts to suggest more than is intended_ (meaning a specific case of B).

PR #950, opened 2018-05-15
Add project manager first assigned on field to Investment Project.

elcct:
Issue number: N/A

### Description of change

This adds `project_manager_first_assigned_on` DateTime field to Investment Project. This enables measurement of SPI 2. 
This field is set with current datetime when project manager is first time assigned.
This field doesn't need to be exposed in the views and is intended to be write once.
Field will be used by upcoming reporting module.


datahub/investment/models.py
```diff
@@ -386,6 +389,10 @@ def __init__(self, *args, **kwargs):
     def save(self, *args, **kwargs):
         """Updates the stage log after saving."""
         adding = self._state.adding
+
+        if self.project_manager and self.project_manager_first_assigned_on is None:
```
reupen: Won't this set this date incorrectly for projects already in the system (with a project manager)? I think you'll need to track if the field has changed as well (and the old value was None).

elcct: This is true. I have added such check and additional test. Thanks!

datahub/investment/models.py
```diff
@@ -382,10 +385,19 @@ def __init__(self, *args, **kwargs):
...
     def save(self, *args, **kwargs):
         """Updates the stage log after saving."""
         adding = self._state.adding
+
+        if (
+            self.__project_manager_id is None
+            and self.project_manager
+            and self.project_manager_first_assigned_on is None
+        ):
+            self.project_manager_first_assigned_on = self.modified_on
```
marcofucci: Isn't `modified_on` calculated when saving the record to the db? At this point I think it's too early.

datahub/investment/test/test_views.py
```diff
@@ -1485,13 +1485,56 @@ def test_patch_team_success(self):
...
+        assert project.project_manager_first_assigned_on is None
+
+        url = reverse('api-v3:investment:investment-item',
+                      kwargs={'pk': project.pk})
+        request_data = {
+            'project_manager': {
+                'id': str(adviser_1.id)
+            }
+        }
+        response = self.api_client.patch(url, data=request_data, format='json')
```
marcofucci: It would be better to freeze the time here again to a different value, otherwise the same value is used in all cases (creation, change etc.).

datahub/investment/models.py
elcct: It is. I can only say I wrote that late :blush: this worked because it saved previously saved modified_on date. Fixed. Thanks!

datahub/investment/test/test_views.py
elcct: That is a good idea. Added.

datahub/investment/models.py
```diff
@@ -382,12 +386,22 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
         self.__stage_id = self.stage_id
+        self.__project_manager_id = self.project_manager_id
 
     def save(self, *args, **kwargs):
         """Updates the stage log after saving."""
         adding = self._state.adding
+        if (
+            self.__project_manager_id is None
+            and self.project_manager
```
reupen: Shouldn't we be checking self.project_manager_id here? As if self.project_manager_id is updated directly, self.project_manager won't update automatically (while changes to self.project_manager do update self.project_manager_id).

datahub/investment/models.py
```diff
@@ -382,12 +386,22 @@ def __init__(self, *args, **kwargs):
...
+            and self.project_manager_first_assigned_on is None
+        ):
+            self.project_manager_first_assigned_on = now()
+
         super().save(*args, **kwargs)
+
         self._update_stage_log(adding)
+        self.__project_manager = self.project_manager
```
reupen: Should that be `self.__project_manager_id = self.project_manager_id`?

elcct: Of course! :blush: Thank you. Fixed.

datahub/investment/models.py
elcct: That is true. I have update it. Thanks

PR #953, opened 2018-05-17
Only Project Manager or Project Assurance Adviser can move project to Verify Win

elcct:
Issue number: DHI-49

### Description of change

This restricts movement of the investment project stage from active to verify win only to project manager or project assurance adviser. 
If anyone else tries to move the project it will return a validation error.
Moving project back from verify win to active or from won to verify win is not restricted.
This is a part of SPI 4 work.

### DIFF
--- a/datahub/investment/serializers.py
+++ b/datahub/investment/serializers.py
@@ -165,6 +165,12 @@ class NestedInvestmentProjectStageLogSerializer(serializers.ModelSerializer):
 class IProjectSerializer(PermittedFieldsModelSerializer):
     """Serialiser for investment project endpoints."""
 
+    default_error_messages = {
+        'only_pm_or_paa_can_move_to_verify_win': ugettext_lazy(
+            'Only Project Manager or Project Assurance Adviser can move project to Verify Win.'
+        )
+    }
+
     incomplete_fields = serializers.SerializerMethodField()
     project_code = serializers.CharField(read_only=True)
     investment_type = NestedRelatedField(meta_models.InvestmentType)
@@ -250,6 +256,8 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
             data['allow_blank_estimated_land_date'] = False
             data['allow_blank_possible_uk_regions'] = False
 
+        self._check_if_investment_project_can_be_moved_to_verify_win(data)
+
         fields = None
         if self.partial and 'stage' not in data:
             fields = data.keys()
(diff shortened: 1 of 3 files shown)


datahub/investment/serializers.py
```diff
@@ -262,6 +270,21 @@ def validate(self, data):
 
         return data
 
+    def _check_if_investment_project_can_be_moved_to_verify_win(self, data):
+        # only project manager or project assurance adviser can move a project to verify win
+        if self.instance and 'stage' in data:
+            if (
+                str(data['stage'].id) == InvestmentProjectStage.verify_win.value.id
```
reupen: Looks a bit complex here, can you use some variables to simplify the if statement? Like allowed_user_ids, current_user_id etc.

Also I think you'll want to check if the stage was moved from any earlier stage to verify_win. Elsewhere that's done by e.g. checking if the old stage's order is less than `InvestmentProjectStage.verify_win.value.order`.

elcct: I agree this is a bit complex. I will update. Thanks!

elcct:
@reupen I would expect that front end knows the id of currently logged in adviser and has access to project manager and project assurance adviser fields when retrieving investment project details. If those fields don't match then Move to verify win button shouldn't be shown. This PR is only to prevent such action in case that is not implemented in the front end. When unauthorised user tries to move the project, frontend should show an error returned by the API.

datahub/investment/test/test_views.py
```diff
@@ -1187,16 +1201,59 @@ def test_change_stage_verify_win_success(self):
...
+            r_and_d_budget=True,
+            non_fdi_r_and_d_budget=True,
+            associated_non_fdi_r_and_d_project=InvestmentProjectFactory(),
+            new_tech_to_uk=True,
+            export_revenue=True,
+            address_1='12 London Road',
+            address_town='London',
+            address_postcode='SW1A 2AA',
+            delivery_partners=[random_obj_for_model(InvestmentDeliveryPartner)],
+            average_salary_id=constants.SalaryRange.below_25000.value.id,
```
marcofucci: Do you need all of these fields set here? They only add distract you away from the main purpose of the test.

reupen: I suppose you could take a VerifyWinInvestmentProjectFactory and override the stage to be active, @elcct.

reupen:
> @reupen I would expect that front end knows the id of currently logged in adviser and has access to project manager and project assurance adviser fields when retrieving investment project details. If those fields don't match then Move to verify win button shouldn't be shown. This PR is only to prevent such action in case that is not implemented in the front end. When unauthorised user tries to move the project, front end should show an error returned by the API.

That's fine but it means that the logic is repeated and controlled in two places. It's not great for maintainability. I think a flag in the response and some information about who can move the project to the next stage would be better, but up to you.

PR #959, opened 2018-05-23
Add adviser export to the admin site

reupen:
Issue number: N/A

### Description of change

Adds a simple reporting framework for admin and the first report (an export of all advisers).

I haven't added a link to the reports page from the index yet as it seems that that's not simple (looks like it'll be easier in Django 2.1, though).

### Checklist

### DIFF
--- a/datahub/admin_report/report.py
+++ b/datahub/admin_report/report.py
@@ -110,6 +110,11 @@ def get_reports_by_model(user):
     return reports_by_model
 
 
+def report_exists(report_id):
+    """Checks if a report exists."""
+    return report_id in _registry
+
+
 def get_report_by_id(report_id, user):
     """
     Gets a report instance for using its ID.
--- a/datahub/admin_report/report.py
+++ b/datahub/admin_report/report.py
@@ -1,0 +1,122 @@
+from typing import Sequence
+
+from django.core.exceptions import PermissionDenied
+from django.db.models import Model
+from django.utils.timezone import now
+
+from datahub.core.exceptions import DataHubException
+
+_registry = {}
+
+
+class _ReportMeta(type):
+    """
+    Metaclass for report definitions.
+
+    Used to auto-register report definitions.
+    """
+
+    def __new__(mcs, name, bases, namespace, **kwargs):  # noqa: N804
+        """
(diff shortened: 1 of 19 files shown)


datahub/company/test/test_admin_report.py
```diff
@@ -0,0 +1,64 @@
...
+            permission_codenames=('read_advisor',),
+            is_staff=True,
+            password=self.PASSWORD,
+        )
+
+        client = self.create_client(user=user)
+        response = client.get(url)
+        assert response.status_code == status.HTTP_200_OK
+        # 7 = header + test user + the 5 test advisers
+        assert len(response.getvalue().decode('utf-8').splitlines()) == 7
```
elcct: Perhaps it would be good to check if those lines indeed contain the data we need.

datahub/admin_report/report.py
```diff
@@ -0,0 +1,127 @@
+from typing import Sequence
+
+from django.core.exceptions import PermissionDenied
+from django.db.models import Model
+from django.utils.timezone import now
+
+from datahub.core.exceptions import DataHubException
+
+_registry = {}
```
elcct: Maybe it would be good to give a comment what this `_registry` keeps.

datahub/admin_report/report.py
```diff
@@ -0,0 +1,127 @@
...
+        _registry[cls.id] = cls()
+
+    def check_permission(self, user):
+        """Checks whether the user has permission for this report."""
+        return user.has_perms(self.permissions_required)
+
+    def get_filename(self):
+        """Gets the filename (excluding extension) to use for the report."""
+        today = now().date().isoformat()
+        return self.filename_template.format(name=self.name, today=today)
```
marcofucci: Could we add hours and seconds as well and maybe format it with year.month.day-hours.minutes in case we are going to store it somewhere and we want it ordered?

datahub/company/test/test_admin_report.py
reupen: Normally I would, but the CSV generation code is in a different app. It's also a bit tricky as the test user ends up in there. The generation of the rows is tested separately in the following test.

PR #970, opened 2018-06-03
Allow client to specify null as the headquarter_type in a company search

ztolley:
The frontend needs to be able to search for a list of companies, filtered on the headquarter type and specify that it wants to include companies with a null headquarters.

Previously when a null was passed it would result in an error from the serializer. This change allows nulls to be passed.


reupen:
The test at https://github.com/uktrade/data-hub-leeloo/blob/dd85c31cb48d1f22204697174bc7212dfc24d728/datahub/search/company/test/test_views.py#L115-L165 is passing because it's submitting the data as a form instead of JSON.

If you change the test to pass format='json' it will fix the test. Then we should be able to get rid of NullStringUUIDField (since it isn't used anywhere else) and use `StringUUIDField(allow_null=True)` instead.

reupen:
@ztolley It failed Flake8 – just a bit of white space to remove.

marcofucci:
@reupen yeah if that means getting rid of that `format=json` in tests

PR #975, opened 2018-06-07
Filter interaction by event and include contact job title in result

ztolley:
Adds a filter to the interactions API to allow interactions to be fetched that are associated with a given event, to find people who have attended an event. The view also allows that result to be sorted by contact last name and first name.

The Screen that lists the attendees needs to show the associated contact job title so the interaction serializer adds that.

### DIFF
--- a/datahub/interaction/serializers.py
+++ b/datahub/interaction/serializers.py
@@ -40,7 +40,9 @@ class InteractionSerializer(serializers.ModelSerializer):
     }
 
     company = NestedRelatedField(Company)
-    contact = NestedRelatedField(Contact)
+    contact = NestedRelatedField(Contact, extra_fields=(
+        'name', 'first_name', 'last_name', 'job_title'
+    ))
     dit_adviser = NestedAdviserField()
     created_by = NestedAdviserField(read_only=True)
     dit_team = NestedRelatedField(Team)
--- a/datahub/interaction/test/views/test_common.py
+++ b/datahub/interaction/test/views/test_common.py
@@ -8,6 +8,7 @@ from datahub.company.test.factories import AdviserFactory, CompanyFactory, Conta
 from datahub.core.constants import Service, Team
 from datahub.core.reversion import EXCLUDED_BASE_MODEL_FIELDS
 from datahub.core.test_utils import APITestMixin, create_test_user, random_obj_for_model
+from datahub.event.test.factories import EventFactory
 from datahub.investment.test.factories import InvestmentProjectFactory
 from datahub.metadata.test.factories import TeamFactory
 from ..factories import CompanyInteractionFactory, EventServiceDeliveryFactory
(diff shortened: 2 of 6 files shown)


datahub/interaction/test/views/test_common.py
```diff
@@ -117,6 +118,25 @@ def test_filtered_by_investment_project(self):
         expected_ids = {str(i.id) for i in project_interactions}
         assert actual_ids == expected_ids
 
+    def test_filtered_by_event(self):
+        """List of interactions filtered by event"""
+        contact = ContactFactory()
+        event = EventFactory()
+
+        CompanyInteractionFactory.create_batch(3, contact=contact)
+        service_deliveries = EventServiceDeliveryFactory.create_batch(3, event=event)
```
reupen: Would be nice to have a few service deliveries with a different event as well (i.e. add a `EventServiceDeliveryFactory.create_batch(3)` as well).

datahub/interaction/views.py
```diff
@@ -33,6 +33,6 @@ class InteractionViewSet(CoreViewSet):
         OrderingFilter,
         PolicyFeedbackPermissionFilter,
     )
-    filter_fields = ['company_id', 'contact_id', 'investment_project_id']
-    ordering_fields = ('date', 'created_on')
+    filter_fields = ['company_id', 'contact_id', 'investment_project_id', 'event_id']
```
reupen: Seems like both of these lists are in a bit of a random order, can we made them both alphabetical?

reupen:
Can we add a test (or two) for the new sorting options?

datahub/interaction/serializers.py
```diff
@@ -18,6 +19,12 @@
 from .permissions import HasAssociatedInvestmentProjectValidator, KindPermissionValidator
 
 
+NestedContactField = partial(
```
reupen: Not sure it's worthwhile defining a NestedContactField as it's only used once. Could just do `contact = NestedRelatedField(Contact, extra_fields=('name', 'job_title'))`

datahub/interaction/test/views/test_common.py
```diff
@@ -117,6 +121,87 @@ def test_filtered_by_investment_project(self):
...
+        })
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['count'] == len(interactions)
+        expected_names = sorted(interaction.contact.last_name for interaction in interactions)
+        actual_names = [result['contact']['last_name'] for result in response_data['results']]
+        assert expected_names == actual_names
+
+    def test_sort_by_contact_firstname(self):
```
marcofucci: You should be able to merge the 2 sort by contact test and use purest.parametrize for testing the two cases.

ztolley: oo.. never heard of them but will give them a try

datahub/interaction/test/views/test_common.py
```diff
@@ -117,6 +122,74 @@ def test_filtered_by_investment_project(self):
...
+        service_deliveries = EventServiceDeliveryFactory.create_batch(3, event=event)
+
+        url = reverse('api-v3:interaction:collection')
+        response = self.api_client.get(url, {'event_id': event.id})
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['count'] == 3
+
+        actual_ids = {i['id'] for i in response_data['results']}
```
reupen: `i` is a bit misleading as it's not an index or number. I'd just use `result` for this one and `service_delivery` for the next one.

stuaxo: Can confirm it gets confusing - ```i``` is used in other export-* code like this and took a few moments of staring at to work out what was going on.

PR #981, opened 2018-06-12
Include public bodies in business types metadata

marcofucci:
This changes the metadata _Govement Department_ to _Government Department or other public body_.

As a consequence, the database row will be changed by the post_migrate signal when migrate is called.


datahub/company/constants.py
```diff
@@ -16,7 +16,10 @@ class BusinessTypeConstant(Enum):
 
     charity = Constant('Charity', '9dd14e94-5d95-e211-a939-e4115bead28a')
     company = Constant('Company', '98d14e94-5d95-e211-a939-e4115bead28a')
-    government_dept = Constant('Government Dept', '9cd14e94-5d95-e211-a939-e4115bead28a')
+    government_dept_or_other_public_body = Constant(
+        'Government Department or other public body',
```
reupen: Lower case d?

PR #988, opened 2018-06-18
(feat) Activity stream endpoint for pen testing

michalc:
Issue number: Activity stream:24

https://trello.com/c/mcxkHcAl/24-stub-feed-from-datahub-with-fake-data-for-pen-testing

### Description of change

    A trivial "Hello World" style endpoint, that exposes no data, but will be used
    for pen testing the authentication on the activity stream endpoint.

    It is required that we have 2 separate factors used for authentication. We
    have:

      - Client IP checking using X-Forwarded-For header
      - Hawk authentication

    In PaaS, and similar cloud providers, the client IP is in the X-Forwarded-For
    header, which (apparently) has measures to prevent spoofing, and from
    discussions with their support team, it is acceptable to use this in
    conjunction with Hawk for authentication.

    We deliberately do not use django-ipware or similar to determine the client IP.
    The django-ipware library does use the X-Forwarded-For header, and I suspect it
    would work, but it _also_

      - Has configured (by default) a chain of HTTP
        headers/Django request properties to fallback on if X-Forwarded-For header.
      - Has code relating to non-publically routable IP addresses
      - Has code for the X-Forwarded-For header giving it's IP addresses in a
        non-standard order.

    This is all unnecessary in our case, and so has the primary effect of
    increasing the attack surface area, when all that it would replace is:

       remote_address = request.META['HTTP_X_FORWARDED_FOR'].split(',')[0].strip()

    Ideally we would use network-level protection on the client IP. However, this
    is not possible at this time in PaaS.

### Checklist

* [x] ~Have any relevant search models been updated?~ N/A
* [x] ~Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?~ N/A
* [x] ~Have any relevant select-/prefetch-related field lists in the views and search apps been updated?~ N/A
* [x] ~Has the admin site been updated (for new models, fields etc.)?~ N/A


config/settings/test.py
```diff
@@ -33,10 +33,4 @@
 
 GOVUK_PAY_URL='https://payments.example.com/'
 
-CACHES = {
-    'default': {
-        'BACKEND': 'django.core.cache.backends.dummy.DummyCache'
-    }
-}
-
```
marcofucci: Better to keep this and use `local_memory_cache` in your tests if you have to.

datahub/activitystream/views.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+            raise AuthenticationFailed(INCORRECT)
+
+        if 'HTTP_AUTHORIZATION' not in request.META:
+            raise AuthenticationFailed(NOT_PROVIDED)
+
+        try:
+            _raise_exception_if_not_authentic(request)
+        except HawkFail as e:
+            log.warning('Failed authentication {e}'.format(e=e))
```
marcofucci: We try to use f strings instead of format.

datahub/activitystream/test/test_activitystream_views.py
```diff
@@ -0,0 +1,273 @@
...
+        HTTP_X_FORWARDED_FOR='1.2.3.4',
+    )
+
+    assert response.status_code == 401
+    error = b'{"detail":"Incorrect authentication credentials."}'
+    assert response.content == error
+
+
+@pytest.mark.django_db
+def test_if_incorrect_domain_then_401_returned(client):
```
reupen: I think at least some of these tests could be parametrised. Could you have a look at that as it should tidy things up (and improve readability)?

datahub/activitystream/views.py
```diff
@@ -0,0 +1,115 @@
...
+from django.core.cache import cache
+from mohawk import Receiver
+from mohawk.exc import HawkFail
+from rest_framework.authentication import BaseAuthentication
+from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.viewsets import ViewSet
+
+log = logging.getLogger(__name__)
```
marcofucci: Maybe rename it to `logger` to keep it consistent with the rest of the codebase.

datahub/activitystream/views.py
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+
+def _seen_nonce(access_key_id, nonce, _):
+    cache_key = 'activity_stream:{access_key_id}:{nonce}'.format(
+        access_key_id=access_key_id, nonce=nonce)
+    seen_cache_key = cache.get(cache_key, False)
+
+    # cache.add only adds key if it isn't present
+    cache.add(cache_key, True, timeout=60)
```
reupen: Can we put the timeout in a named constant (possibly in settings, as we tend to put them in there)?

datahub/activitystream/views.py
```diff
@@ -0,0 +1,115 @@
...
+
+    return {
+        'id': settings.ACTIVITY_STREAM_ACCESS_KEY_ID,
+        'key': settings.ACTIVITY_STREAM_SECRET_ACCESS_KEY,
+        'algorithm': 'sha256',
+    }
+
+
+def _seen_nonce(access_key_id, nonce, _):
+    cache_key = 'activity_stream:{access_key_id}:{nonce}'.format(
```
reupen: We generally use f-strings (unless the format string is a template that's passed around).

datahub/activitystream/views.py
reupen: We normally use logger (rather the log) for logger variable names.

michalc: Done!

datahub/activitystream/views.py
michalc: Done!

config/settings/test.py
michalc: Done!

datahub/activitystream/views.py
```diff
@@ -0,0 +1,110 @@
...
+        if 'HTTP_AUTHORIZATION' not in request.META:
+            raise AuthenticationFailed(NOT_PROVIDED)
+
+        try:
+            _raise_exception_if_not_authentic(request)
+        except HawkFail as e:
+            logger.warning(f'Failed authentication {e}')
+            raise AuthenticationFailed(INCORRECT)
+
+        return (_ActivityStreamUser(), None)
```
reupen: Not sure about this, as django-oauth-toolkit doesn't override the user if the token isn't associated with one. Since we're working with both, it might be beneficial to be consistent.

datahub/activitystream/views.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def authenticate_header(self, request):
+        """This is returned as the WWW-Authenticate header when
+        AuthenticationFailed is raised. DRF also requires this
+        to send a 401 (as opposed to 403)
+        """
+        return 'Hawk'
+
+    def authenticate(self, request):
+        if 'HTTP_X_FORWARDED_FOR' not in request.META:
```
reupen: Doesn't this stop us from running this outside of PaaS (e.g. locally)?

datahub/activitystream/views.py
```diff
@@ -0,0 +1,110 @@
...
+            logger.warning(
+                'Failed authentication: no X-Forwarded-For header passed'
+            )
+            raise AuthenticationFailed(INCORRECT)
+
+        remote_address = \
+            request.META['HTTP_X_FORWARDED_FOR'].split(',')[0].strip()
```
reupen: Could pass `maxsplit=1` to `str.split()` (if only to indicate that the other IP addresses are intentionally discarded). Also would avoid `\` – would probably use an intermediate variable here.

docker-compose.env
```diff
@@ -18,3 +18,6 @@ RESOURCE_SERVER_AUTH_TOKEN=sso-token
 DATA_SCIENCE_COMPANY_API_URL=http://dt07-reporting/
 DATA_SCIENCE_COMPANY_API_ID=dt07-reporting-id
 DATA_SCIENCE_COMPANY_API_KEY=dt07-reporting-id
+ACTIVITY_STREAM_IP_WHITELIST=1.2.3.4
```
reupen: These also need to go in `config/settings/sample.env`.

datahub/activitystream/views.py
```diff
@@ -0,0 +1,110 @@
...
+from mohawk.exc import HawkFail
+from rest_framework.authentication import BaseAuthentication
+from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.viewsets import ViewSet
+
+logger = logging.getLogger(__name__)
+
+NOT_PROVIDED = 'Authentication credentials were not provided.'
```
reupen: I didn't find the names of these descriptive enough when reading the code further down – would suggest something like `NO_CREDENTIALS_MESSAGE` and `INCORRECT_CREDENTIALS_MESSAGE`.

datahub/activitystream/views.py
```diff
@@ -0,0 +1,110 @@
...
+
+class ActivityStreamViewSet(ViewSet):
+    """List-only view set for the activity stream"""
+
+    authentication_classes = (_ActivityStreamAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def list(self, request):
+        """A single page of activities"""
+        return Response({'secret': 'content-for-pen-test'})
```
reupen: Shouldn't we be using `Receiver.respond()` to generate the `Server-Authorization` response header?

datahub/activitystream/views.py
```diff
@@ -0,0 +1,110 @@
...
+from rest_framework.viewsets import ViewSet
+
+logger = logging.getLogger(__name__)
+
+NOT_PROVIDED = 'Authentication credentials were not provided.'
+INCORRECT = 'Incorrect authentication credentials.'
+
+
+def _lookup_credentials(access_key_id):
+    if access_key_id != settings.ACTIVITY_STREAM_ACCESS_KEY_ID:
```
marcofucci: Isn't this vulnerable to timing attacks? If would be safer to use `constant_time_compare` instead.

datahub/activitystream/views.py
michalc: You can just add the header in your favourite client, e.g. Postman. (I would lean to doing this rather than making the production code more complex).

datahub/activitystream/views.py
michalc: Can you clarify, perhaps with an example of what would be consistent?

datahub/activitystream/views.py
michalc: Done!

datahub/activitystream/views.py
michalc: Done!

config/settings/test.py
```diff
@@ -33,10 +33,4 @@
 
 GOVUK_PAY_URL='https://payments.example.com/'
 
-CACHES = {
-    'default': {
-        'BACKEND': 'django.core.cache.backends.dummy.DummyCache'
```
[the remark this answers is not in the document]
michalc: Done!

README.md
```diff
@@ -208,6 +208,9 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the
 
 | Variable name | Required | Description |
 | ------------- | ------------- | ------------- |
+| `ACTIVITY_STREAM_ACCESS_KEY_ID` | Yes | A non-secret access key ID, corresponding to `ACTIVITY_STREAM_ACCESS_KEY_ID`. The holder of the secret key can access the activity stream endpoint by Hawk authentication. |
```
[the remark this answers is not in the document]
michalc: Done!

datahub/activity_stream/views.py
```diff
@@ -0,0 +1,116 @@
...
+INCORRECT_CREDENTIALS_MESSAGE = 'Incorrect authentication credentials.'
+
+
+def _lookup_credentials(access_key_id):
+    if not constant_time_compare(access_key_id,
+                                 settings.ACTIVITY_STREAM_ACCESS_KEY_ID):
+        raise HawkFail(f'No Hawk ID of {access_key_id}')
+
+    logger.warning(settings.ACTIVITY_STREAM_ACCESS_KEY_ID)
+    logger.warning(settings.ACTIVITY_STREAM_SECRET_ACCESS_KEY)
```
marcofucci: Any reason why you are logging the keys?

michalc: Was in there while I was debugging

michalc: Removed

marcofucci:
@michalc just a quick note, it's easier to review code that is not constantly squashed so we tend to push commits generated from review comments to the branch and then when the PR is accepted the author can squash before merging as he wishes ;)

michalc:
@marcofucci 

> @michalc just a quick note, it's easier to review code that is not constantly squashed so we tend to push commits generated from review comments to the branch and then when the PR is accepted the author can squash before merging as he wishes ;)

Understood!

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -0,0 +1,228 @@
...
+    assert response.status_code == status.HTTP_401_UNAUTHORIZED
+    assert response.json() == expected_json
+
+
+@pytest.mark.django_db
+def test_if_authentication_passed_but_61_seconds_in_past_401_returned(api_client):
+    """If the Authorization header is generated 61 seconds in the past, then a
+    401 is returned
+    """
+    past = datetime.datetime.now() + datetime.timedelta(seconds=-61)
```
reupen: I'd say `datetime.datetime.now() - datetime.timedelta(seconds=61)` would be clearer (rather than + ... -).

datahub/activitystream/views.py
reupen: I'm not really sure why we can't fall back to `REMOTE_ADDR` if `HTTP_X_FORWARDED_FOR` is not set. But up to you.

datahub/activity_stream/views.py
```diff
@@ -0,0 +1,114 @@
...
+    cache.add(cache_key, True,
+              timeout=settings.ACTIVITY_STREAM_NONCE_EXPIRY_SECONDS)
+
+    if seen_cache_key:
+        logger.warning(f'Already seen nonce {nonce}')
+
+    return seen_cache_key
+
+
+def _raise_exception_if_not_authentic(request):
```
marcofucci: Not sure this is a good name, wouldn't `authorise` be better?

datahub/activity_stream/views.py
```diff
@@ -0,0 +1,114 @@
...
+        request.build_absolute_uri(),
+        request.method,
+        content=request.body,
+        content_type=request.content_type,
+        seen_nonce=_seen_nonce,
+    )
+
+
+class _ActivityStreamUser(AnonymousUser):
+    username = 'activity_stream_user'
```
marcofucci: It would be good to store the non-secret access key on the user as there could be more than one in the future. I understand this might complicate the implementation so see how tricky it is and I'm okay to leave it out if it's too unnecessarily complex for now.

michalc: Although I think it'll be really easy, since there is just the one, if the only benefit would be seen in a future where there are more than one, then I would be more pro not doing it until that future happens.

datahub/activitystream/views.py
reupen: With client credentials (not linked to a user), django-oauth-toolkit ends up returning `None, access_token`. So `request.user` ends up being `None` and `request.auth` is an instance of the AccessToken.

It works because it inspects `request.auth` and `request.successful_authenticator` as required:
https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/contrib/rest_framework/permissions.py

Maybe _ActivityStreamUser is useful, but:

* it returns `True` for both is_authenticated and is_anonymous (which seems unconventional)
* the permission class doesn't seem to be actually having an effect, because the authentication class raises an error on authentication failure

.circleci/config.yml
```diff
@@ -11,6 +11,9 @@ jobs:
           POSTGRES_DB=datahub
 
     environment:
+      ACTIVITY_STREAM_IP_WHITELIST: 1.2.3.4
```
reupen: It seems like these specific values are required for the tests. So they should either (also) be set in `config/settings/test.py`, or overridden in the tests.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -0,0 +1,228 @@
...
+def _auth_header(key_id, secret_key, url, method, content, content_type):
+    return mohawk.Sender({
+        'id': key_id,
+        'key': secret_key,
+        'algorithm': 'sha256',
+    },
+        url,
+        method,
+        content=content,
+        content_type=content_type,
```
marcofucci: This is not so easy to read. I'd store the first dict in a var or align that to the other params.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -0,0 +1,228 @@
...
+@pytest.mark.parametrize(
+    'get_kwargs,expected_json',
+    (
+        (
+            # If no X-Forwarded-For header
+            dict(
+                content_type='',
+                HTTP_AUTHORIZATION=_auth_header(
+                    'some-id', 'some-secret', _url(), 'GET', '', '',
+                )
```
marcofucci: These params specified here don't seem to help scanning the tests.
I'd use defaults for the params in `_auth_header` and only specify the ones that make these specific cases special when needed.
E.g. 
```
_auth_header()
``` 
in this case and 

```
_auth_header(path='incorrect')
``` 
for incorrect path

etc.

datahub/activitystream/views.py
michalc: Will opt to leave as-is: there isn't a production use-case for `REMOTE_ADDR`.

michalc:
@reupen To check before merging, are you ok with the state of the code? (All good if you have further comments/changes you would think be good)

datahub/activity_stream/views.py
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+
+def _seen_nonce(access_key_id, nonce, _):
+    """Returns if the passed access_key_id/nonce combination has been
+    used within settings.ACTIVITY_STREAM_NONCE_EXPIRY_SECONDS
+    """
+    cache_key = f'activity_stream:{access_key_id}:{nonce}'
+    seen_cache_key = cache.get(cache_key, False)
+
```
reupen: This seems like a race condition (separate get and add). cache.add() is supposed to return whether the value was actually stored, so can we use that instead of calling cache.get() first?

PR #998, opened 2018-06-25
Add Elasticsearch mapping migration management command

reupen:
Issue number: n/a

### Description of change

This adds a command to migrate Elasticsearch mappings. This also switches us from multiple-mapping-types per index to one-index-per-mapping-type.

How the migration works is described in `docs/Elasticsearch migrations.md` (would be interested in any suggestions on how the process could be improved...)

Note that there is still some follow-up work to be done:

- we should be using `_doc` for mapping type names, as that's the direction ES is going in
- we might need to consider how many shards we have per index
- we're a bit inconsistent about whether things are defined on the search app or model (I had to define stuff on the model here as the app wasn't always available)
- we could probably simplify some of the permissions logic (omit indexes from the query instead of filtering based on mapping type)
- tidying up some of the mappings, for example using `index=False` a bit more liberally and replacing nested fields with object fields where appropriate

### Checklist


datahub/search/migrate.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def migrate_app(search_app):
+    """Migrates a search app to a new index (if its mapping is out of date)."""
+    app_name = search_app.name
+    es_model = search_app.es_model
+
+    target_mapping_hash = es_model.get_target_mapping_hash()
+    needs_migration = es_model.get_current_mapping_hash() != target_mapping_hash
+    if needs_migration:
```
marcofucci: I think it might be nicer to add this check to the es_model instead to have it centralised.
Something like `es_model.needs_migration()`

datahub/search/models.py
```diff
@@ -13,17 +24,115 @@ class BaseESModel(DocType):
...
+            )
+        return read_indices, next(iter(write_indices))
+
+    @classmethod
+    def get_index_prefix(cls):
+        """Gets the prefix used for indices and aliases."""
+        return f'{settings.ES_INDEX_PREFIX}-{cls._doc_type.name}-'
+
+    @classmethod
+    def get_target_mapping_hash(cls):
```
elcct: What target means here?

datahub/search/migrate.py
elcct: I agree, since es_model has methods related to future index it would makes sense for the model to also be able to tell if it needs to be migrated.

datahub/search/models.py
```diff
@@ -13,17 +24,115 @@ class BaseESModel(DocType):
...
+    def get_read_and_write_indices(cls):
+        """Gets the indices currently referenced by the read and write aliases."""
+        read_indices, write_indices = get_indices_for_aliases(
+            cls.get_read_alias(), cls.get_write_alias()
+        )
+        if len(write_indices) != 1:
+            raise DataHubException(
+                'Unexpected alias state; write alias references multiple indices',
+            )
+        return read_indices, next(iter(write_indices))
```
marcofucci: Why wouldn't you use `get_write_index` instead of replicating the same logic?

datahub/search/models.py
```diff
@@ -13,17 +24,115 @@ class BaseESModel(DocType):
...
+        mapping_data = serialise_mapping(cls._doc_type.mapping.to_dict())
+        return blake2b(mapping_data, digest_size=16).hexdigest()
+
+    @classmethod
+    def get_current_mapping_hash(cls):
+        """Extracts and returns the mapping hash from the current index name."""
+        current_write_index = cls.get_write_index()
+        prefix = cls.get_index_prefix()
+        if not current_write_index.startswith(prefix):
+            return ''
```
marcofucci: I know that this is for the first time when the index doesn't have any prefix but I'd still log this case as a warning in case it happens again in the future instead of _failing_ silently.

datahub/search/models.py
```diff
@@ -13,17 +24,115 @@ class BaseESModel(DocType):
...
+
+    @classmethod
+    def get_target_index_name(cls):
+        """Generates a unique name for the index based on its mapping."""
+        mapping_hash = cls.get_target_mapping_hash()
+        prefix = cls.get_index_prefix()
+        return f'{prefix}{mapping_hash}'
+
+    @classmethod
+    def initialise_index(cls, force_update_mapping=False):
```
elcct: This seems a bit big, could this be split into separate read and write parts?

datahub/search/elasticsearch.py
```diff
@@ -123,38 +121,109 @@ def get_client():
...
+    """Helper class for making multiple alias updates atomically."""
 
-        if index_settings:
-            index.settings(**index_settings)
-        index.create()
+    def __init__(self):
+        """Initialises the instance with an empty list of pending operations."""
+        self.actions = []
 
+    def add_indices_to_alias(self, alias, indices):
```
marcofucci: Not sure `add` is a good verb. Would _associate_alias_with_indices_ be better?

datahub/search/models.py
reupen: Target as in target (or desired) state. Could go with desired instead?

datahub/search/models.py
reupen: The idea was to retrieve them in one operation so that it reflects a single state (while if two operations were used, there's a small risk of something updating the aliases between the operations).

So I'll make get_write_index() use this function (to get rid of the duplication).

datahub/search/models.py
reupen: I'll add a warning (though we should be able to make it an error once all environments have been migrated).

datahub/search/models.py
reupen: Not sure it's worth splitting it like that – could add a simpler function to add indexes to an alias to get rid of a few lines. (Also five lines will go ones the legacy stuff is removed.)

datahub/search/elasticsearch.py
reupen: Not too fussed. It's add and remove in the underlying operation though, and maybe they emphasise more that it's a set? Happy to call them anything that makes sense though.

datahub/search/bulk_sync.py
```diff
@@ -1,24 +1,14 @@
...
-    """
-    search_apps = get_search_apps()
-
-    # if models empty, assume all models
-    return [search_app for search_app in search_apps if not models or search_app.name in models]
+BULK_INDEX_TIMEOUT_SECS = 300
+BULK_DELETION_TIMEOUT_SECS = 300
 
 
 def sync_app(item, batch_size=None):
```
marcofucci: Can we rename `item` to `search_app` as we are refactoring this?

datahub/search/models.py
```diff
@@ -46,7 +155,7 @@ def db_object_to_dict(cls, dbmodel):
         return result
 
     @classmethod
-    def db_objects_to_es_documents(cls, dbmodels):
+    def db_objects_to_es_documents(cls, dbmodels, index=None):
```
marcofucci: I'm not sure we need the extra `index` param here and in `es_document` as we are always using the write_index so the logic in `es_document` should be okay without `index`. I might missing something though.

datahub/search/migrate.py
```diff
@@ -0,0 +1,67 @@
...
+        _perform_migration(search_app)
+        return
+
+    if len(es_model.get_read_and_write_indices()[0]) != 1:
+        logger.info(f'Possibly incomplete {app_name} search app migration detected')
+        _schedule_resync(search_app)
```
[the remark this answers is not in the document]
reupen: I could check if the `migrate_model()` lock as active for that search app, but it won't be foolproof (as the task could still be queued etc.)

I think doing a similar check in the migrate_model task would avoid most problems (i.e. don't do anything if there's only one read index).

datahub/search/bulk_sync.py
```diff
@@ -55,3 +49,33 @@ def sync_app(item, batch_size=None):
                         f'{rows_processed*100//total_rows}%')
 
     logger.info(f'{model_name} rows processed: {rows_processed}/{total_rows} 100%.')
+
+
+def _delete_documents(index, complete_index_actions):
```
marcofucci: Could we rename `complete_index_actions` to `es_docs` or similar? I know that it's not 100% accurate but `complete_index_actions` is slightly confusing if you don't reference where this is called from. What do you think?

datahub/search/migrate_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    delete_index,
+    get_aliases_for_index,
+    start_alias_transaction,
+)
+
+
+logger = getLogger(__name__)
+
+
+def resync_after_migrate(search_app):
```
marcofucci: Would it be possible to add this logic to the standard `sync_app` maybe with a param if we want to be safe?

We could also add the `advisory_lock` to the `sync_app` to instead of in the `migrate_model`.

I find the the logic around `migrate_model`, `resync_after_migrate` and `_schedule_resync` potentially not necessary. It takes a lot of mental effort to understand what's going on.

datahub/search/elasticsearch.py
marcofucci: I think the ES docs tend to use _associate_ if I'm not wrong https://www.elastic.co/guide/en/elasticsearch/reference/6.2/indices-aliases.html

reupen: I'll change it :slightly_smiling_face:

datahub/search/models.py
elcct: What do you think about word "nascent"?

datahub/search/models.py
elcct: I see. It looks like after removing legacy stuff it should be much better, so I am fine with this.

datahub/search/migrate_utils.py
reupen: > Would it be possible to add this logic to the standard sync_app maybe with a param if we want to be safe?

We could do, if you think it would help (I was trying to keep the migration logic separate).

> I find the the logic around migrate_model, resync_after_migrate and _schedule_resync potentially not necessary. It takes a lot of mental effort to understand what's going on.

Yes, it's not simple, though I'm not sure how it can be simplified (any suggestions? :slightly_smiling_face: )

The hope was to automate it at some point (the same way `./manage.py migrate` is) so it's been written with that in mind (and the zero downtime...)

datahub/search/models.py
```diff
@@ -13,17 +28,104 @@ class BaseESModel(DocType):
...
+
+            create_alias(cls.get_write_alias(), index_name)
+        elif force_update_mapping:
+            cls.init(cls.get_write_alias())
+
+        if not read_alias_exists:
+            create_alias(cls.get_read_alias(), cls.get_write_index())
+
+    @classmethod
+    def create_index(cls, index_name):
```
marcofucci: I don't find it clear when to use `initialise_index` and when to use `create_index`. 
It seems like `create_index` creates and initialises the mapping as well so why would I need the `initialise_index` method? If `initialise_index` initialises the index, why `create_index` is not calling it? Maybe it needs a better name.

datahub/search/elasticsearch.py
```diff
@@ -123,38 +121,122 @@ def get_client():
...
+
+@contextmanager
+def start_alias_transaction():
+    """Returns a context manager that can be used to update indices atomically."""
+    alias_updater = AliasUpdater()
+    yield alias_updater
+    alias_updater.commit()
+
+
+def create_alias(alias, index):
```
marcofucci: Why do we keep this if we have `AliasUpdater`? Having 2 versions adds confusion. I've also noticed that we have a `create_alias` django command, do we need that?

marcofucci: and the same for `delete_alias`.

datahub/search/tasks.py
```diff
@@ -30,3 +36,18 @@ def sync_model(search_app_name):
     """
     search_app = get_search_app(search_app_name)
     sync_app(search_app)
+
+
+@shared_task(bind=True, acks_late=True, priority=7, max_retries=5, default_retry_delay=60)
+def migrate_model(self, search_app_name, new_mapping_hash):
```
marcofucci: I'm not sure this is a good name. It's not migrating the model after all, it's just triggering a resync followed by migration cleanup.

datahub/search/elasticsearch.py
reupen: `delete_alias()` is a bit different as it removes the alias completely (rather than dissociating it from a single index). It's only used for test set-up though (to clean up after any test runs that didn't terminate cleanly).

`create_alias()` is there as it's a bit cumbersome to use `start_alias_transaction()` for a single association operation (same reason both exist in the Elasticsearc API, presumably).

I could make `create_alias()` use `start_alias_transaction()` rather than the equivalent operation in the ES API if we prefer. (Also will rename it I guess to `associate_alias_with_index` to be consistent with the new names for `AliasUpdater`.)

datahub/search/apps.py
```diff
@@ -35,12 +35,12 @@ def __init__(self, mod):
         self.mod = mod
         self.mod_signals = f'{self.mod}.signals'
 
-    def init_es(self):
+    def init_es(self, force_update_mapping=False):
         """
         Makes sure mappings exist in Elasticsearch.
         This call is idempotent.
```
elcct: Is this comment still relevant in the context of `force_update_mapping` parameter?

datahub/search/apps.py
```diff
@@ -70,14 +70,34 @@ def get_permission_filters(self, request):
...
+
+def get_search_apps_by_name(app_names=None):
+    """
+    Returns the apps for a particular set of app names.
+
+    :param app_names: list of search app names to return app instances for, None for all
+    """
+    search_apps = get_search_apps()
+
+    # if app_names empty, assume all apps
```
elcct: Here it says that about app_names being empty, but function comment says about being None. Maybe would be good to make comments consistent.

datahub/search/models.py
reupen: `sync_app()` takes a snapshot of what the aliases point to at the start of the operation and uses those for all bulk insertions and deletions. The reason was so that it doesn't leave any particular index half-populated or with its contents partially deleted (in case the aliases are updated in the middle of the operation for some reason).

datahub/search/models.py
reupen: It's a nice word, but I don't think it works here (the index might not exist yet, or it might've existed for a long time...)

elcct: I see. In that case for a lack of a better word "target" would do...

datahub/search/models.py
```diff
@@ -13,17 +28,119 @@ class BaseESModel(DocType):
...
+        """
+        read_alias_exists = alias_exists(cls.get_read_alias())
+        write_alias_exists = alias_exists(cls.get_write_alias())
+        if not write_alias_exists:
+            # Handle migration from the legacy single-index set-up
+            # TODO: Remove once all environments have been migrated to the new structure
+            if settings.ES_LEGACY_INDEX and index_exists(settings.ES_LEGACY_INDEX):
+                index_name = settings.ES_LEGACY_INDEX
+            else:
+                index_name = cls.get_target_index_name()
```
marcofucci: I'm not sure when would this else logic be executed, is it still in case of the legacy index?

marcofucci: Oh I see, in case of new environments.

marcofucci: It would be good if we could centralise the creation and the setup of an index instead of being here for the first time and in `migrate._perform_migration` afterwards.
Also, it's not clear that `set_up_index_and_aliases` really sets up the index and the aliases the first time only ignoring potential needed migrations.

Could we use the same migration logic for migrating from zero to the initial setup somehow?

datahub/search/models.py
```diff
@@ -13,17 +28,119 @@ class BaseESModel(DocType):
...
+        return f'{prefix}{mapping_hash}'
+
+    @classmethod
+    def is_migration_needed(cls):
+        """Returns whether the active mapping is out of date and a migration is needed."""
+        target_mapping_hash = cls.get_target_mapping_hash()
+        return cls.get_current_mapping_hash() != target_mapping_hash
+
+    @classmethod
+    def was_migration_started(cls):
```
marcofucci: It would be better if you could move this to the `migrate` module as it's an implementation detail (2 indices means ongoing migration) that this model doesn't know anything about.

datahub/search/management/commands/init_es.py
```diff
@@ -12,6 +13,26 @@ class Command(BaseCommand):
...
+        parser.add_argument(
+            '--force-update-mapping',
+            action='store_true',
+            help='Attempts to update the mapping if the index and alias already exist.',
+        )
+
     def handle(self, *args, **options):
         """Executes the command."""
-        init_es()
+        with advisory_lock('leeloo_init_es'):
```
marcofucci: Can we get rid of the `init_es` command and logic and always use the `migrate_es` one? We could add an `update_in_place` (or similar) option there to cover the case with of _force-update-mapping_.

datahub/search/migrate.py
```diff
@@ -0,0 +1,65 @@
...
+    app_name = search_app.name
+    es_model = search_app.es_model
+
+    if es_model.is_migration_needed():
+        _perform_migration(search_app)
+        return
+
+    if es_model.was_migration_started():
+        logger.info(f'Possibly incomplete {app_name} search app migration detected')
+        _schedule_resync(search_app)
```
marcofucci: Minor but if we could avoid this it would be great. If you call `migrate_es` twice, the second call would re-schedule a resync (so 2 resync tasks in total). I can't see anything bad happening in the code apart from an unnecessary second re-sync. 

I guess you added this in case the first call fails leaving the system in an inconsistent state so that the second call can re-sync and complete the work?

datahub/search/signals.py
```diff
@@ -51,10 +52,17 @@ def disconnect(self):
...
     instance = db_model.objects.get(pk=pk)
-    doc = search_model.es_document(instance)
+    doc = search_model.es_document(instance, index=write_index)
     elasticsearch.bulk(actions=(doc, ), chunk_size=1)
 
+    # If a migration is in progress, remove old versions of the document from indices that are
+    # being migrated from
+    remove_indices = read_indices - {write_index}
+    delete_document(search_model, pk, indices=remove_indices)
+
```
marcofucci: Is this to avoid 2 objects to be returned when searching and when a migration is in progress? I'm not sure it's worth this here as it's buried away and not in the migration module.

datahub/search/management/commands/init_es.py
reupen: That was basically my initial idea – get rid of `init_es` and use `migrate_es` for the same purpose. I haven't done it yet though, mainly because this is quite a large change and I'd feel more comfortable once it's been tested in staging and it's in production. (By large change, I mean both the splitting up of the index into multiple indexes and the migration logic). I could still make `migrate_es` do the initialisation as well so that we're halfway there (and mark `init_es` as deprecated).

(I also would prefer to get rid of `--force-update-mapping` I think – that was similarly just a fallback to maintain the current workflow.)

datahub/search/migrate.py
reupen: Yes, it's so that there's a way to retry if something goes wrong (pretty badly wrong really).

The second resync shouldn't happen, because `resync_after_migrate()` has a check at the start of it. (Or have you seen something else happen locally?) The second task could end up waiting for the lock for a while, I could change the behaviour there so that it just exits if it can't get the lock.

marcofucci: oh yeah I see the check in `resync_after_migrate` now. I think returning immediately if there's already a lock would be nice actually.

datahub/search/signals.py
marcofucci: I'm saying that maybe we don't delete anything at this point. `sync_app` after the migration should delete these documents from the old index anyway, right?

reupen: Yes, they'll get deleted eventually, but that could be some time later. In the meantime both an outdated and current version of the document would be returned by searches.

marcofucci: I understand and that would not be ideal but this migration logic is in a random place and it's not triggered by the migration logic. If we wanted to be consistent and we had this here we should have changed the `sync_app` to check for migration data automatically as well and just have the migration command to set up and trigger a re-sync that would pick things up.
In this form, this adds too many moving parts.

reupen: No moving parts means suspending writes to ES completely while the migration is going on. Which might work if it was a quick process, but as it stands it's potentially a very long process depending on which models are being migrated.

marcofucci: hm okay so how about we:

1. move the `post_batch_callback` call to `elasticsearch.bulk`
2. we change `sync_app` to just call `bulk(..., post_batch_callback=post_batch_callback)`
3. we pass in `_sync_app_post_batch_callback` to `bulk` in the `datahub.search.signals._sync_es` so that at least we use the same logic?

The names would need to be refactored a bit of course.

reupen: I had a look at that and it's a bit more complicated:

1. sync_app() takes a snapshot of the indexes that the alias points at, which are passed to post_batch_callback. So you would need to pass a post_batch_callback_context as well (or do something different).

2. It gets a bit messy as the return value from the real bulk() call will be lost if post_batch_callback() raises an exception. I don't think it matters in the current code paths, but it doesn't seem like good design.

(And, also, the current post_batch_callback would end up calling bulk() inside a bulk() which is a bit confusing.)

I'll move the common logic in sync_app() and sync_model() to another function, which should solve all problems...

datahub/search/sync_async.py
```diff
@@ -0,0 +1,29 @@
...
+    sync_objects(
+        es_model,
+        [instance],
+        read_indices,
+        write_index,
+        post_batch_callback=delete_from_secondary_indices_callback,
+    )
+
+
+def sync_object_async(search_model, db_model, pk):
```
marcofucci: We can get rid of `search_model` and get it dynamically using `search.apps.get_search_app_by_model`.
It can be in another PR but we just need to remember it, do you want me to create an issue?

datahub/search/management/commands/init_es.py
```diff
@@ -12,6 +13,26 @@ class Command(BaseCommand):
...
+            logger.info('Creating Elasticsearch indices and initialising mappings...')
+
+            for app in get_search_apps_by_name(options['model']):
+                app.init_es()
```
marcofucci: `force_update_mapping` is not used nor passed in.

datahub/search/models.py
marcofucci: 👍 worth updating the docs saying that if index is not passed in the default write index is used, just to make it clear.

datahub/search/bulk_sync.py
```diff
@@ -55,3 +43,21 @@ def sync_app(item, batch_size=None):
                         f'{rows_processed*100//total_rows}%')
 
     logger.info(f'{model_name} rows processed: {rows_processed}/{total_rows} 100%.')
+
+
+def sync_objects(es_model, model_objects, read_indices, write_index, post_batch_callback=None):
+    """Syncs an iterable of model instances to Elasticsearch."""
+    actions = list(
+        es_model.db_objects_to_es_documents(model_objects, index=write_index)
+    )
```
marcofucci: I find it a bit odd that `actions` is converted to a list when in `db_objects_to_es_documents` it uses yield for I guess memory related issues. This is also the only place where we call `db_objects_to_es_documents`.

datahub/search/elasticsearch.py
```diff
@@ -123,38 +121,121 @@ def get_client():
...
+def index_exists(index_name):
+    """Checks if an index exists."""
+    client = get_client()
+    return client.indices.exists(index_name)
 
-    :returns: True if index_name exists
-    """
-    return get_client().indices.exists(index=name)
 
+def create_index(index_name, es_model):
```
marcofucci: I don't think this should be tight to `es_model` as this module is less about specific models and more about indices and aliases. I know there's a one to one match now but still, it would be cleaner without it.

datahub/search/elasticsearch.py
```diff
@@ -123,38 +121,121 @@ def get_client():
...
+    """Deletes an index."""
+    logger.info(f'Deleting the {index_name} index...')
+    client = get_client()
+    client.indices.delete(index_name)
+
+
+def get_indices_for_alias(alias):
+    """Gets the indices referenced by an alias."""
+    client = get_client()
+    return client.indices.get_alias(name=alias).keys()
```
marcofucci: Can't we get rid of `get_indices_for_alias` and always use `get_indices_for_aliases`? We could even check if aliases is len(1) return that instead of a list...

datahub/search/management/commands/migrate_es.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Command(BaseCommand):
+    """Command for migrating an Elasticsearch index."""
+
+    help = "Creates the Elasticsearch index (if necessary) and updates the index's mapping."
+
+    def add_arguments(self, parser):
+        """Handle arguments."""
+        parser.add_argument(
+            '--model',
```
marcofucci: I know we started calling this arg `model` in `sync_app` but I wonder if we should refactor and use `app_name` instead (or use app_name in new commands). I think that `model` is confusing and it could mean different things (also django related) and in theory that is the search app name not the model name.

marcofucci: another option is to introduce a `model-label` argument instead and use `get_search_app_by_model`.

datahub/search/apps.py
```diff
@@ -23,12 +25,14 @@ def __init__(self, mod):
...
 
-    def init_es(self):
+    def init_es(self, force_update_mapping=False):
         """
-        Makes sure mappings exist in Elasticsearch.
-        This call is idempotent.
+        Creates the index and aliases for this app if they don't already exist.
+
+        If force_update_mapping is True and the write alias already exists, an attempt
+        is made to update to update the existing mapping in place.
```
elcct: `to update to update`

datahub/search/bulk_sync.py
reupen: Yes, I think it was like that to calculate num_actions. We can use the batch size for chunk_size though, and get num_actions from the return value of bulk(). I'll change it.

datahub/search/sync_async.py
reupen: Yes, I thought having both was redundant when I started working on this PR, though my first thought was to rejig things so that it receives a search app instead. You way is a much smaller change, though.

Yes, we should add a separate task (also for some of the points in my opening comment).

datahub/search/elasticsearch.py
reupen: I'll change it to take a mapping instance instead to decouple them (also should get rid of an API call...)

datahub/search/elasticsearch.py
reupen: Is your concern just that we are using ES in two different ways? It's only used once anyway, so I can get rid of it.

datahub/search/management/commands/migrate_es.py
reupen: Yes, I agree that we should change it. I think changing the name of the argument to something like `--app` or `--search-app` is a good option. We could also use positional arguments (a bit more similar to `./manage.py migrate`).

datahub/search/bulk_sync.py
reupen: Ah, it's a bit more complicated as actions is also passed to post_batch_callback...

datahub/search/tasks.py
```diff
@@ -30,3 +36,27 @@ def sync_model(search_app_name):
...
+def complete_model_migration(self, search_app_name, new_mapping_hash):
+    """
+    Completes a migration by performing a full resync, updating aliases and removing old indices.
+    """
+    search_app = get_search_app(search_app_name)
+    if search_app.es_model.get_target_mapping_hash() != new_mapping_hash:
+        logger.warning(
+            f'Unexpected target mapping hash, an old app instance may have received this task. '
+            f'Rescheduling {search_app_name} search app migration...'
+        )
```
marcofucci: This could also happen when/if a new app instance receives this task and the mapping has changed...

conftest.py
```diff
@@ -175,15 +175,29 @@ def _es_client(worker_id):
 @pytest.fixture(scope='session')
 def _setup_es_indexes(_es_client):
     """Sets up ES and makes the client available."""
-    _create_test_index(_es_client, settings.ES_INDEX)
-
     # Create models in the test index
     for search_app in get_search_apps():
+        # Clean up in case of any aborted test runs
+        index = search_app.es_model.get_target_index_name()
```
marcofucci: `index_name` would be a better name

datahub/search/migrate_utils.py
```diff
@@ -0,0 +1,67 @@
...
+        with start_alias_transaction() as alias_transaction:
+            alias_transaction.dissociate_indices_from_alias(read_alias, indices_to_remove)
+    else:
+        logger.warning(f'No indices to remove for the {read_alias} alias')
+
+    for index in indices_to_remove:
+        _delete_old_index(index)
+
+
+def _delete_old_index(index):
```
marcofucci: Why wouldn't you have this code where it's called instead of using a different function?

datahub/search/bulk_sync.py
marcofucci: okay so maybe we can get rid of the yield and return a list directly in `db_objects_to_es_documents`

datahub/search/bulk_sync.py
```diff
@@ -55,3 +43,21 @@ def sync_app(item, batch_size=None):
...
+    num_actions = len(actions)
+    bulk(
+        actions=actions,
+        chunk_size=num_actions,
```
marcofucci: didn't we decide to avoid using len() for `chunk_size` and keep a fixed constant instead?

datahub/search/elasticsearch.py
```diff
@@ -123,38 +121,114 @@ def get_client():
...
+        client = get_client()
+        client.indices.update_aliases(body={
+            'actions': self.actions
+        })
+        self.actions = []
+
+
+@contextmanager
+def start_alias_transaction():
+    """Returns a context manager that can be used to update indices atomically."""
```
marcofucci: As `_AliasUpdater` is private, a bit of doc would be great including how you can use `alias_updater` (e.g. its methods).

datahub/search/management/commands/migrate_es.py
marcofucci: Added an issue https://github.com/uktrade/data-hub-leeloo/issues/1034 just to keep track of it.

conftest.py
```diff
@@ -175,15 +175,29 @@ def _es_client(worker_id):
...
+            delete_alias(write_alias)
+
+        # Create indices and aliases
         search_app.init_es()
 
     yield _es_client
 
-    _es_client.indices.delete(settings.ES_INDEX)
+    for search_app in get_search_apps():
+        delete_index(search_app.es_model.get_target_index_name())
```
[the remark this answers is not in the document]
reupen: What do you mean? They're two separate for-loops.

datahub/search/query_builder.py
```diff
@@ -19,10 +18,15 @@ class MatchNone(Query):
     name = 'match_none'
 
 
-def delete_document(model, document_id):
+def delete_document(model, document_id, indices=None, ignore_404_responses=True):
```
marcofucci: This should go in `search.deletion.py`

datahub/search/tasks.py
reupen: You mean we should change the wording in the warning?

conftest.py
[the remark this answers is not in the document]
marcofucci: Oh true sorry, reviewing on github is not always ideal :/

datahub/search/bulk_sync.py
reupen: Yes, but this was like that before. There's a lot in this PR already so I think we can review that separately.

datahub/search/tasks.py
marcofucci: yeah I meant that

PR #999, opened 2018-06-25
Add delete_orphans command

marcofucci:
Issue number: NA

### Description of change

This:
- adds a new django app to hold all cleanup related logic
- adds a new `delete_orphans` django command

`./manage.py delete_orphans <model> [--simulate=True]` will delete all orphaned `model`s.
An orphaned object is an unreferenced 6-month-old record.

The tests also test that the tests fully cover all the cases just so that we don't forget anything when adding/changing fields.

I'm not convinced that _cleanup_ is a good enough name for this app.
Also, it's not as readable as I wanted so any suggestion is much appreciated.

I've made the following changes from the original ticket:
- archived records deleted as well
- the query uses `modified_on` instead of `created_on` to determine what records are considered orphaned
- the field used to determine what records are considered orphaned for events is `end_date`

This might change when Anthony is back.

Kudos to @reupen for implementing the main logic.

### Checklist

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -71,6 +71,7 @@ LOCAL_APPS = [
     'datahub.search.apps.SearchConfig',
     'datahub.user',
     'datahub.dbmaintenance',
+    'datahub.cleanup',
     'datahub.omis.core',
     'datahub.omis.order',
     'datahub.omis.market',
--- a/datahub/cleanup/management/commands/delete_orphans.py
+++ b/datahub/cleanup/management/commands/delete_orphans.py
@@ -1,0 +1,118 @@
+from datetime import timedelta
+from logging import getLogger
+from secrets import token_urlsafe
+
+from django.apps import apps
+from django.core.management.base import BaseCommand
+from django.db.models import Exists, OuterRef
+from django.template.defaultfilters import capfirst
+from django.utils.timezone import now
+
+
+logger = getLogger(__name__)
+
+
+class OrphanDeletionConfig:
+    """
+    Defines the config values with related defaults used to determine
+    when a record is considered orphan.
+    """
+
+    def __init__(
+        self,
+        days_before_orphaning=30 * 6,  # 6 months
+        date_field='modified_on'
+    ):
+        """Initialises the objects with overridable defaults."""
+        self.days_before_orphaning = days_before_orphaning
+        self.date_field = date_field
+
+
+# CONFIGS FOR ALL MODELS ALLOWED TO BE CLEANED UP
+ORPHANING_CONFIGS = {
+    'company.Contact': OrphanDeletionConfig(),
+    'company.Company': OrphanDeletionConfig(),
+    'event.Event': OrphanDeletionConfig(date_field='end_date')
+}
+
+
+def get_related_fields(model):
+    """
+    Returns all the fields of `model` that hold the link between referencing objects
+    and the referenced object (`model`).
+
+    :param model: orphaned model class
+    :returns: list of fields of `model` that hold references via dependent objects
+    """
+    return [
+        f for f in model._meta.get_fields(include_hidden=True)
+        if (f.one_to_many or f.one_to_one or f.many_to_many or f.many_to_one)
+        and f.auto_created
+        and not f.concrete
+        and not f.field.model._meta.auto_created
+    ]
+
+
+def get_unreferenced_objects_query(model):
+    """
+    :param model: orphaned model class
+    :returns: queryset for unreferenced objects
+    """
+    fields = get_related_fields(model)
+
+    identifiers = [f'ann_{token_urlsafe(6)}' for _ in range(len(fields))]
+
+    qs = model.objects.all()
+    for identifier, field in zip(identifiers, fields):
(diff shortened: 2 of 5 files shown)


config/settings/common.py
```diff
@@ -71,6 +71,7 @@
     'datahub.search.apps.SearchConfig',
     'datahub.user',
     'datahub.dbmaintenance',
+    'datahub.cleanup',
```
elcct: Could this live in dbmaintenance app?

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            default=False,
+            help='If True it only simulates the command and prints the SQL query.',
+        )
+
+    def handle(self, *args, **options):
+        """Main login for the actual command."""
```
elcct: Logic?

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -0,0 +1,116 @@
...
+    :returns: queryset for unreferenced objects
+    """
+    fields = get_related_fields(model)
+
+    identifiers = [f'ann_{token_urlsafe(6)}' for _ in range(len(fields))]
+
+    qs = model.objects.all()
+    for identifier, field in zip(identifiers, fields):
+        related_field = field.field
+        subquery = related_field.model.objects.filter(
```
reupen: I think we can add on an `.only('pk')` here to generate shorter queries.

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -0,0 +1,203 @@
...
+    },
+}
+
+
+def test_mappings():
+    """
+    Test that `MAPPINGS` includes all the data necessary for covering all the cases.
+    This is to avoid missing tests when new fields and models are added or changed.
+    """
+    for model_name in delete_orphans.ORPHANING_CONFIGS:
```
reupen: Can we parametrise or use pytest_generate_tests()?

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -0,0 +1,203 @@
...
+            pytest.fail(f'Please add test cases for deleting orphaned {model}')
+
+        related_fields = delete_orphans.get_related_fields(model)
+        expected_related_deps = {(field.field.model, field.field.name) for field in related_fields}
+        related_deps_in_mapping = {
+            (dep_factory._meta.model, dep_field_name)
+            for dep_factory, dep_field_name in mapping['dependent_models']
+        }
+
+        missing_dep_mappings = (expected_related_deps - related_deps_in_mapping)
```
reupen: Not sure these brackets are helping – I think they can go.

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -0,0 +1,203 @@
...
+        if missing_dep_mappings:
```
reupen: Do we want to check for the opposite case as well (something in `related_deps_in_mapping` but not in `expected_related_deps`)?

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -0,0 +1,203 @@
...
+        - a record with other objects referencing it doesn't get deleted
+        - a record without any objects referencing it but not old enough
+            doesn't get deleted
+        - a record without any objets referencing it and old gets deleted
+    """
+    orphaning_config = delete_orphans.ORPHANING_CONFIGS[model_name]
+    mapping = MAPPINGS[model_name]
+    model_factory = mapping['factory']
+
+    tot_model_records = 0
```
reupen: Is tot short for total? Total isn't much longer, so may as well go for that.

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -0,0 +1,116 @@
...
+            **{f'{config.date_field}__lt': now() - timedelta(config.days_before_orphaning)}
+        ).order_by('-modified_on')
+
+        logger.info(f'{capfirst(model._meta.verbose_name_plural)} to delete: {qs.count()}')
+
+        if options['simulate']:
+            logger.info(f'SQL:\n{qs.query}')
+        else:
+            qs.delete()
+            logger.info('Done')
```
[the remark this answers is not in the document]
marcofucci: I think both is better so that we keep one in case of `simulate=True`.

datahub/cleanup/test/commands/test_delete_orphans.py
marcofucci: I had that originally but I think that this case should be covered by the `test_run`. Any obsolete mapping will result in more objects being deleted in the `test_run`.

config/settings/common.py
marcofucci: yeah, I thought about it but I thought we decided that dbmaintenance would be temporary whilst we will keep  this indefinitely.

PR #1002, opened 2018-06-27
Block editing of archived contacts

reupen:
Issue number: n/a

### Description of change

This blocks direct editing of archived contacts (via the API).

(This doesn't consider consider current contacts that belong to an archived company, as we need to discuss a bit more how that will work.)

### Checklist

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -27,6 +27,7 @@ from datahub.core.validators import (
     AllIsBlankRule,
     AnyIsNotBlankRule,
     EqualsRule,
+    NotArchivedValidator,
     OperatorRule,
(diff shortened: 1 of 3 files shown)


marcofucci:
@reupen 
> (This doesn't consider consider current contacts that belong to an archived company, as we need to discuss a bit more how that will work.)

I guess we should archive all contacts when a company gets archived so that the logic in this PR will be sufficient.

marcofucci:
Yes very true. My initial opinion (that could change) would be to have a confirmation screen that tells you all the things it will archive and ask you to confirm, django style. That would allow us to keep it non reversible (if you made a mistake it's your fault, we warned you...).

reupen:
That could be sufficient, yep (and it's a simpler design...)

PR #1011, opened 2018-07-02
Delete ES docs on django delete

marcofucci:
Issue number: N/A

### Description of change

This:
- adds a context manager that collects and deletes the objects from ES when they are deleted from the db
- creates a search app that can be used to test generic logic

### Checklist


datahub/cleanup/conftest.py
```diff
@@ -0,0 +1 @@
+from datahub.search.conftest import _es_client, _setup_es_indexes, setup_es  # noqa: F401
```
marcofucci: This is ugly but I can't find a better way, do you guys have any suggestion?

marcofucci: I think this breaks the tests in circleCI 😞

config/settings/common.py
```diff
@@ -214,6 +214,16 @@
 
 MPTT_ADMIN_LEVEL_INDENT = 30
 
+SEARCH_APPS = [
+    'datahub.search.companieshousecompany.CompaniesHouseCompanySearchApp',
+    'datahub.search.company.CompanySearchApp',
+    'datahub.search.contact.ContactSearchApp',
+    'datahub.search.event.EventSearchApp',
+    'datahub.search.investment.InvestmentSearchApp',
```
reupen: Sort alphabetically? (Interaction before investment.)

conftest.py
```diff
@@ -83,3 +89,95 @@ def _synchronous_submit_to_thread_pool(fn, *args, **kwargs):
...
+    for search_app in get_search_apps():
+        search_app.init_es()
+        search_app.connect_signals()
+
+    yield _es_client
+
+    _es_client.indices.delete(settings.ES_INDEX)
+
+    for search_app in get_search_apps():
+        search_app.disconnect_signals()
```
reupen: Hmm, guess it was like this before but looks like the signals only get disconnected at the end of the pytest session. That means they'll be active in non-search tests that run after search ones, no? Can the signal bits be moved to the setup_es fixture instead so the signal handlers are only active for the search tests?

datahub/search/test/test_deletion.py
```diff
@@ -0,0 +1,121 @@
...
+    obj.delete()
+
+    collector.disconnect()
+
+    assert collector.deletions == {
+        SimpleModel: [es_doc]
+    }
+
+    assert SimpleModel.objects.count() == 0
+    setup_es.count(settings.ES_INDEX, doc_type=search_app.name)['count'] == 1
```
reupen: Is this (and line 102) missing an assert?

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -100,19 +101,21 @@ def add_arguments(self, parser):
...
-        model_verbose_name = capfirst(model._meta.verbose_name_plural)
-        logger.info(f'{model_verbose_name} to delete: {qs.count()}')
-
-        if options['simulate']:
-            logger.info(f'SQL:\n{qs.query}')
-        else:
-            deleted = qs.delete()
-            logger.info(f'{model_verbose_name} deleted: {deleted}')
+
+        with update_es_after_deletions():
```
reupen: Should be able to use `update_es_after_deletions()` as a decorator to get rid of the indentation (it should already work as a decorator as it uses `contextlib.contextmanager`).

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -177,15 +181,23 @@ def test_run(orphaning_mapping):
     # 3 + 1 in case of self-references
     total_model_records = 3 + (1 if dep_factory == model_factory else 0)
 
+    setup_es.indices.refresh()
+
     model = apps.get_model(model_name)
+    search_app = get_search_app_by_model(model)
+
     assert model.objects.count() == total_model_records
+    setup_es.count(settings.ES_INDEX, doc_type=search_app.name)['count'] == total_model_records
```
reupen: Same question about asserts as further down.

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -200,14 +212,24 @@ def test_simulate(model_name, caplog):
...
+
     management.call_command(delete_orphans.Command(), model_name, simulate=True)
+
     assert model.objects.count() == 3
+    setup_es.count(settings.ES_INDEX, doc_type=search_app.name)['count'] == 3
 
     # check that 3 records would have been deleted
-    assert len(caplog.records) == 2
-    assert 'to delete: 3' in caplog.records[0].message
+    assert any(
```
reupen: Might be simpler to do `assert 'to delete: 3' in caplog.text`.

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -100,19 +101,21 @@ def add_arguments(self, parser):
     def handle(self, *args, **options):
```
reupen: Not necessarily for this PR, but it would be nice to set `help` on the command as well.

datahub/search/apps.py
```diff
@@ -78,10 +66,23 @@ def get_search_app(app_name):
...
 
+@lru_cache(maxsize=None)
+def get_search_app_by_model(model):
+    """
+    :returns: a single search app (by django model)
+    :param model: django model for the search app
+    :raises LookupError: if it can't find the search app
+    """
+    for search_app in get_search_apps():
+        if search_app.queryset.model == model:
```
reupen: It might make more sense to use `is` since we're checking if two objects are the same object. Doesn't matter too much though, since it ends up doing the same thing.

datahub/search/deletion.py
```diff
@@ -0,0 +1,122 @@
...
+    :raises DataHubException: in case of non 404 errors
+    """
+    delete_actions = (
+        _create_delete_action(index, es_doc['_type'], es_doc['_id'])
+        for es_doc in es_docs
+    )
+
+    _, errors = bulk(
+        actions=delete_actions,
+        chunk_size=len(es_docs),
```
reupen: I think we might still want to set a fixed chunk size as there's still an upper limit on request size (although we'll probably never hit it). Not sure how much load deletions puts on ES either...

We should also really be setting max_chunk_bytes too (making it a configurable setting – it defaults to 100 MB when we have a max 10 MB request size in most cases).

conftest.py
marcofucci: Thanks for spotting that!

reupen:
Something else to consider is that the interaction search app does have a post_delete signal receiver, so we might need to disconnect some existing signal receiver while a Collector is active as well...

marcofucci:
> Something else to consider is that the interaction search app does have a post_delete signal receiver, so we might need to disconnect some existing signal receiver while a Collector is active as well...

True :/

datahub/search/deletion.py
```diff
@@ -0,0 +1,141 @@
...
+        """Starts listening to post_delete signals on all search models."""
+        for receiver in self.signal_receivers_to_add:
+            receiver.disconnect()
+
+        for receiver in self.signal_receivers_to_disable:
+            receiver.connect()
+
+    def _collect(self, sender, instance, **kwargs):
+        """Logic that gets run on post_delete."""
+        model = instance.__class__
```
reupen: Shouldn't we still be using on_commit (like the normal signals) so that it only gets collected if the transaction is committed?

datahub/search/deletion.py
```diff
@@ -0,0 +1,141 @@
...
+
+            # set up the receivers to add when collecting the deleted objects
+            self.signal_receivers_to_add.append(
+                SignalReceiver(post_delete, model, self._collect)
+            )
+
+            # get the existing post/pre_delete receivers that need to be
+            # disabled in the meantime
+            for receiver in search_app.get_signals_receivers():
+                if receiver.signal in (post_delete, pre_delete):
```
reupen: There's also a post_delete signal for InvestmentProjectTeamMember which it makes less sense to disable. Although it probably won't do much harm disabling it, we should probably only disable the delete signals targeted at the same model as the search app.

datahub/search/deletion.py
```diff
@@ -0,0 +1,141 @@
...
+
+    WARNING: this temporarily disable existint post_delete and pre_delete signal
+    receivers defined in all the SearchApps to avoid side effects.
+    """
+    collector = Collector()
+    collector.connect()
+
+    yield collector
+
+    collector.disconnect()
```
reupen: We'll need to disconnect the signals even if an exception is raised during the yield. Looks like that's missing at the moment.

datahub/search/deletion.py
marcofucci: It's not as easy as it sounds as django deletes objects in cascade and the related delete signals are triggered as well so it's safer to delete all of them.

datahub/search/deletion.py
```diff
@@ -0,0 +1,141 @@
...
+    all the django objects deleted in the block.
+
+    This works by listening to the `post_delete` django signal, collecting all the deleted
+    django objects and deleting all of them from ES in bulk when exiting from the context manager.
+
+    WARNING: this temporarily disable existint post_delete and pre_delete signal
+    receivers defined in all the SearchApps to avoid side effects.
+    """
+    collector = Collector()
+    collector.connect()
```
elcct: I wonder if something like `.collect()` and `.release()` would be more appropriate?

datahub/search/test/test_deletion.py
```diff
@@ -0,0 +1,156 @@
...
+        assert receiver.connect.called
+
+    assert collector.deletions == {
+        SimpleModel: [es_doc]
+    }
+
+    assert SimpleModel.objects.count() == 0
+    assert setup_es.count(settings.ES_INDEX, doc_type=search_app.name)['count'] == 1
+
+    # delete
```
elcct: Is this comment necessary?

datahub/search/test/search_support/simplemodel/signals.py
```diff
@@ -0,0 +1,17 @@
+from django.db.models.signals import post_delete, pre_delete
+
+from ..models import SimpleModel as DBSimpleModel
+from ....signals import SignalReceiver
+
+
+def on_delete_callback(sender, instance, **kwargs):
```
elcct: Maybe name should also indicate that this is dummy function?

datahub/search/deletion.py
```diff
@@ -0,0 +1,141 @@
...
+        collector = Collector()
+        collector.connect()  # starts listening to post_delete signals
+
+        # delete django objects
+
+        collector.disconnect()  # stops listening to post_delete signals
+
+        collector.delete_from_es()  # deletes the collected objects from ES
+
+    WARNING: this temporarily disable existint post_delete and pre_delete signal
```
elcct: `existint` - looks like a typo

datahub/search/deletion.py
marcofucci: Better to collect at this point and delete only on transaction.commit I guess

datahub/search/deletion.py
marcofucci: The problem is that `collect` doesn't really collect, it only connects the right signal receivers so it would be wrong. The collection bit happens at the time the objects are deleted.

datahub/search/deletion.py
marcofucci: 👍 `max_chunk_bytes=10MB` and `chunk_size=10000` should be fine I suppose?

datahub/search/deletion.py
reupen: Ah, yes that's a good point. I suppose that problem exists generally though :thinking:

datahub/search/deletion.py
```diff
@@ -0,0 +1,148 @@
...
+        '_type': _type,
+        '_id': _id,
+    }
+
+
+class Collector:
+    """
+    Collects all the deleted django objects so that they can be deleted from ES.
+
+    Most of the time you will want to use the context manager
```
reupen: Is there any reason to keep the option of directly using Collector open? Why don't we just say 'Don't use this, use the context manager `update_es_after_deletions` instead' and make Collector internal?

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -106,9 +107,10 @@ def handle(self, *args, **options):
         model = apps.get_model(model_name)
         config = ORPHANING_CONFIGS[model_name]
 
+        orphaning_datetetime = today(tzinfo=utc) - timedelta(config.days_before_orphaning)
```
reupen: Typo in datetetime.

datahub/search/deletion.py
```diff
@@ -11,6 +12,7 @@
 
 
 BULK_DELETION_TIMEOUT_SECS = 300
+BULK_CHUCK_SIZE = 10000
```
reupen: Typo in chunk :)

datahub/search/deletion.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def delete_from_es(self):
+        """Deletes all the deleted django models from ES."""
+        transaction.on_commit(self._delete_from_es)
+
+
+@contextmanager
+def update_es_after_deletions():
+    """
+    Returns a context manager that can be used to automatically delete from ES
```
reupen: Might be worth mentioning that it can be used as a decorator as well.

datahub/search/deletion.py
reupen: Hmm, it's a bit complicated I think. If you start a transaction while a collector is active it could still misbehave. I'm not sure we can stop an an atomic block existing while a collector is active, so maybe just documenting it would be fine.

datahub/search/deletion.py
marcofucci: Because I don't like internal/private stuff, I thought it was obvious 😆

datahub/search/deletion.py
marcofucci: What do you mean?

datahub/search/deletion.py
reupen: It's covered by PEP 8 if that's what you mean :slightly_smiling_face: 

https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces (also if you like the Hitchiker's Guide: http://docs.python-guide.org/en/latest/writing/style/#we-are-all-responsible-users)

datahub/search/deletion.py
reupen: Just that if we e.g. delete an investment project outside of the collector, and it cascades the delete to the team member, it's going to try and update the project in ES as well... Not important for this PR, anyway.

datahub/search/deletion.py
marcofucci: 👍 good point

datahub/search/deletion.py
marcofucci: Yeah I know, that's true for OMIS as well. We could add a check before getting the object from the db before syncing. But yeah, not in this PR.

datahub/search/deletion.py
marcofucci: Yeah, I'm okay if others want to keep stuff private but I like to share all my code, developers can do what they want with this Collector if they find it useful for some reasons.

PR #1019, opened 2018-07-10
Add proposition document upload

elcct:
Issue number: DHI-111

### Description of change

This updates existing Document Upload functionality, so that it is re-usable more easily. 
Common views and base model has been moved to Documents app.
S3 dependent code in virus scanning has been removed.

Because we have separate AWS credentials for different type of documents I have added `bucket` field to the shared Document model, so that it could be possible to find where file has actually been uploaded.

The flow looks like this:
- Client calls POST `/v3/investment/<ip-uuid>/proposition/<p-uuid>/document` with filename, gets back proposition document details with signed upload url, where file needs to be uploaded to
- Once uploaded, client POST to `/v3/investment/<ip-uuid>/proposition/<p-uuid>/document/<d-uuid>/upload-callback` with empty body. This initiates virus scanning.
- Once scanning has completed, the download endpoint will return signed S3 URL to the document.

Existing document upload endpoints could be removed in separate PR.

### Checklist


config/settings/common.py
```diff
@@ -359,3 +359,9 @@
 REPORT_AWS_SECRET_ACCESS_KEY = env('REPORT_AWS_SECRET_ACCESS_KEY', default='')
 REPORT_BUCKET = env('REPORT_BUCKET', default='')
 REPORT_AWS_REGION= env('REPORT_AWS_REGION', default='')
+
+# Investment Project documents (propositions)
+INVESTMENT_DOCUMENT_AWS_ACCESS_KEY_ID = env('INVESTMENT_DOCUMENT_AWS_ACCESS_KEY_ID', default='')
```
reupen: Do we need a separate bucket for just investment project documents? What happens to DOCUMENTS_BUCKET and the credentials for it?

elcct: I learned we are not supposed to upload to the DOCUMENTS_BUCKET as it stores legacy documents. I think it should be removed from Leeloo (I could do that in another PR)

INVESTMENT_DOCUMENT_BUCKET will be shared between propositions and evidence that is coming next.

datahub/documents/views.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return response
+
+    def upload_complete_callback(self, request, *args, **kwargs):
+        """File upload done callback."""
+        doc = self.get_object()
+        serializer = UploadStatusSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        submit_to_thread_pool(
```
reupen: We should really move the virus scanning to Celery so that we can more easily add retry logic if it fails etc.

datahub/documents/av_scan.py
```diff
@@ -1,46 +1,23 @@
-from contextlib import closing
+from io import BytesIO
 from logging import getLogger
 
 import requests
 from django.conf import settings
 from django.utils.timezone import now
+from django.utils.translation import ugettext_lazy as _
```
reupen: We should probably use `gettext_lazy` as we don't support Python 2. The Django docs say:

> The u prefix on the functions below comes from a difference in Python 2 between unicode and bytestrings. If your code doesn’t support Python 2, use the functions without the u.

datahub/documents/models.py
```diff
@@ -18,9 +18,12 @@ class Document(BaseModel, ArchivableModel):
     """General model for keeping track of user uploaded documents."""
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    bucket = models.CharField(
```
reupen: I'm not sure about putting the actual bucket in the model as the bucket is also specified in the configuration. It could be some identifier that maps to a bucket in the configuration?

datahub/documents/serializers.py
```diff
@@ -13,3 +13,9 @@ class DocumentSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Document
         fields = '__all__'
+
+
+class UploadStatusSerializer(serializers.Serializer):
+    """Serializer for upload status endpoints."""
+
+    status = serializers.ChoiceField(choices=('success',))
```
reupen: Elsewhere we use `model_utils.Choices`, I'm not sure there's a reason to be different here.

datahub/documents/models.py
```diff
@@ -18,9 +18,12 @@ class Document(BaseModel, ArchivableModel):
...
+        max_length=settings.CHAR_FIELD_MAX_LENGTH,
+        blank=True,
```
reupen: Doesn't seem like it should be allowed for it to be blank?

datahub/documents/models.py
```diff
@@ -43,30 +51,58 @@ def name(self):
...
+
+
+class BaseDocumentModel(BaseModel):
+    """Base document model."""
+
+    PATH_TEMPLATE = 'investment-documents/{filename}'
+    BUCKET = settings.DOCUMENTS_BUCKET
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    filename = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH)
```
reupen: Would it make more sense to store the (original) file name in the Document model?

datahub/documents/models.py
```diff
@@ -43,30 +51,58 @@ def name(self):
         """Model name."""
         return self.path
 
-    def generate_signed_url(self):
+    def __str__(self):
+        """String repr."""
+        return f'Document(filename="{self.filename}", av_clean={self.av_clean})'
+
+
+class BaseDocumentModel(BaseModel):
```
reupen: It seems a bit confusing to have a BaseDocumentModel model and a Document model (when the latter doesn't inherit from the former). Maybe we rename this one – not sure what to, but could be something like AbstractEntityDocument.

datahub/investment/proposition/migrations/0002_proposition_document.py
```diff
@@ -0,0 +1,39 @@
...
+                ('proposition', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='documents', to='proposition.Proposition')),
+            ],
+            options={
+                'verbose_name': 'investment project proposition document',
+                'permissions': (('read_propositiondocument', 'Can read proposition document'),),
+            },
+        ),
+        migrations.AlterUniqueTogether(
+            name='propositiondocument',
+            unique_together={('proposition', 'filename')},
```
reupen: Is there a logical reason to not allow the same file name twice or is it just because of the S3 key format?

datahub/documents/models.py
```diff
@@ -43,30 +51,58 @@ def name(self):
...
+    """Base document model."""
+
+    PATH_TEMPLATE = 'investment-documents/{filename}'
```
reupen: Not sure about encouraging the overriding of the key format – seems like it will lead to disorganisation. Are we able to go for a generic format – something like `<model-label>/<object-id>/<file>`? We might want a unique identifier added to the file name as well.

datahub/documents/models.py
marcofucci: Yes I agree, I had to check to code to fully understand the difference. 
I would also move `get_signed_url` and `get_signed_upload_url` to the `Document` model.

datahub/documents/models.py
marcofucci: I agree, `filename` in both models is very confusing.

datahub/investment/proposition/views.py
```diff
@@ -122,3 +127,52 @@ def get_additional_data(self, create):
     def _check_project_exists(self):
         if not InvestmentProject.objects.filter(pk=self.kwargs['project_pk']).exists():
             raise Http404(self.non_existent_project_error_message)
+
+
+class PropositionDocumentViewSet(DocumentViewSet):
+    """Proposition Document ViewSet."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (
```
reupen: OK if you want to add the permissions in a separate PR, but like this we also run the risk of it ending up in production without proper permissions. Maybe we could put it behind a feature flag? Might even be able to use a permission class for that...

datahub/documents/models.py
```diff
@@ -43,30 +51,58 @@ def name(self):
...
+            investment_doc = cls(
+                document=doc,
+                **kwargs,
+            )
+            investment_doc.save()
+
+        return investment_doc
 
 
 @receiver(post_delete, sender=Document)
```
marcofucci: To be honest I would not use a signal here, it just doesn't make sense as signals are made for decoupling things whilst this doc is already an S3 one.
This would also fix the problem with the expected `doc.client` sometimes set when you call `.delete` on `BaseDocumentModel` but not set when calling `delete` on `Document`.

datahub/documents/serializers.py
marcofucci: Not even sure why the endpoint accepts a body with the only option of status=success. Are we going to allow other options as well like 'cancel' etc.? If we do we can add that later on, I think we should just get rid of it and accept a POST with an empty body.

datahub/documents/views.py
```diff
@@ -0,0 +1,68 @@
...
+from rest_framework import status
+from rest_framework.response import Response
+
+from datahub.core.thread_pool import submit_to_thread_pool
+from datahub.core.viewsets import CoreViewSet
+from datahub.documents.av_scan import virus_scan_document
+from datahub.documents.serializers import UploadStatusSerializer
+
+
+class DocumentViewSet(CoreViewSet):
```
marcofucci: This is not really `DocumentViewSet` but `DocumentModelViewSet` or whatever we rename that to, isn't it?

marcofucci: It would also be good to prefix it with `Base` so that it's clear that it's not a complete Viewset.

datahub/documents/models.py
```diff
@@ -43,30 +51,58 @@ def name(self):
...
         return self.path
 
-    def generate_signed_url(self):
+    def __str__(self):
+        """String repr."""
+        return f'Document(filename="{self.filename}", av_clean={self.av_clean})'
+
+
+class BaseDocumentModel(BaseModel):
+    """Base document model."""
```
marcofucci: Some docs on how to use this + what views/serializers to use would be great.

datahub/investment/proposition/utils.py
```diff
@@ -0,0 +1,17 @@
+import boto3
+from django.conf import settings
+
+
+def get_proposition_s3_client():
+    """Get S3 client singleton."""
+    s3 = getattr(get_proposition_s3_client, 's3_instance', None)
+    if not s3:
```
marcofucci: could use `functools.lru_cache` instead of this?

datahub/documents/views.py
```diff
@@ -0,0 +1,68 @@
...
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
+
+    def create(self, request, *args, **kwargs):
+        """Create and one-time upload URL generation."""
+        response = super().create(request, *args, **kwargs)
+        document = self.get_queryset().get(pk=response.data['id'])
+        client = kwargs.get('client')
```
marcofucci: As `client` is the same throughout this class it would make more sense to have an abstract `get_client` method in this class that gets overridden in the subclasses and here just use `.get_client` instead?

reupen:
If we're using multiple buckets for documents (with differing credentials), it looks like there's a bit more work to do to organise things:

1. Organise the settings e.g.:

    ```python
    DOCUMENT_BUCKETS = {
        'investment': {  # I will call 'investment' the bucket ID
            'bucket': 'xxx',
            'aws_access_key_id': 'xxx',
            ...
        }
    }
            
    ```

2. An enum of bucket IDs (probably in config so that the settings can use it)

3. Single `get_s3_client()` implementation that takes a bucket ID and gets the relevant parameters from the settings

datahub/documents/av_scan.py
```diff
@@ -64,24 +41,23 @@ def _process_document(document_pk: str):
...
-    response = s3_client.get_object(Bucket=bucket, Key=key)
-    with closing(response['Body']):
+def _scan_object(original_filename: str, download_url: str):
+    """Virus scans a file stored on remote server."""
+    with requests.get(download_url, stream=True) as response:
+        if response.status_code != 200:
+            raise VirusScanException('Source file not found.')
         return _scan_raw_file(
-            original_filename, S3StreamingBodyWrapper(response), response['ContentType']
+            original_filename, response.content, response.headers['content-type']
```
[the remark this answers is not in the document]
elcct: This is indeed what we want. Thank you

datahub/documents/models.py
elcct: I am going to change it to identifier that maps to the bucket configuration.

datahub/documents/serializers.py
elcct: I understand this endpoint is not great and I have not spent much time on this :( I am going to improve it soon. Thanks!

datahub/investment/proposition/serializers.py
```diff
@@ -48,10 +48,45 @@ def abandon(self):
...
+        model = PropositionDocument
+        fields = ('id', 'filename', 'created_by', 'created_on', 'url',)
+        read_only_fields = fields
+
+
 class PropositionSerializer(serializers.ModelSerializer):
     """Proposition serialiser for view only endpoints."""
 
     investment_project = NestedInvestmentProjectField()
+    documents = NestedPropositionDocumentSerializer(many=True, read_only=True)
```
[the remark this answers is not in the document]
elcct: What would be the reason for not including it here? It is for front-end convenience and typically it should contain no more than 1-2 documents.

datahub/documents/models.py
elcct: Since document is tied to the model, I think generic path like `<model-name>/<YYYY-mm-dd>/<pk>` would do. Since we are not exposing that to the user, it doesn't really need human readable form. What do you think?

reupen: Which date and pk are those in the example?

elcct: I would update PR today. That would be something like this:

```
    @classmethod
    def create_document_path(cls, document_id):
        """Create document path."""
        today = now().strftime('%Y-%m-%d')
        return f'{cls._meta.model_name}/{today}/{document_id}'

    @classmethod
    def create_from_declaration_request(cls, **kwargs):
        """Create proposition document along with the corresponding Document."""
        with transaction.atomic():
            document_id = uuid.uuid4()
            doc = Document(
                id=document_id,
                bucket_id=cls.BUCKET,
                path=cls.create_document_path(document_id),
            )
            doc.save()
            investment_doc = cls(
                document=doc,
                **kwargs,
            )
            investment_doc.save()

        return investment_doc
```

elcct: This is actually good point as I would also like to redirect user to S3, so that original file name should stay in the path. I would also like to avoid of the file being overwritten. What do you think about:
`f'{cls._meta.model_name}/{today}/{document_id}/{original_filename}'`

`{today}` grouping is only so that when someone tries to download S3 bucket, he/she wouldn't end up in million files/directories in one directory. 

If we stayed with {document_id} as a file name, then we would have to proxy S3 download in order to restore original filename?

datahub/documents/views.py
elcct: I have removed that as I went with @reupen suggestion to organise the bucket settings.

datahub/investment/proposition/utils.py
elcct: I have removed this function, but now used `functools.lru_cache` in its equivalent in `documents/utils.py`.

datahub/documents/models.py
elcct: It has been replaced with `bucket_id` that contains key to bucket settings.

datahub/documents/models.py
elcct: I have replaced `filename` with `original_filename` instead.

datahub/documents/models.py
elcct: I have renamed this model to `AbstractEntityDocument`, and also moved S3 related methods to `Document` model.

datahub/investment/proposition/migrations/0002_proposition_document.py
elcct: I have removed that.  Thanks!

datahub/documents/views.py
elcct: I have renamed this Viewset to `BaseDocumentModelViewSet`. Thanks!

datahub/documents/models.py
reupen: > If we stayed with {document_id} as a file name, then we would have to proxy S3 download in order to restore original filename?

Not necessarily, ResponseContentDisposition (mentioned in the last comment) should be able to take care of that. How does the download work between the front end and the back end?

elcct: Document will be accessible via link to frontend, frontend then calls backend and if it gets redirect to the S3, it will return that redirect to the user. User then downloads file directly from S3.

reupen: Doesn't a redirect seem like an odd way to pass a URL from the back end to the front end? A redirect would imply to me that the front end is following the redirect, but from what you're saying it's extracting the URL from the initial 302 response and then redirecting the user to that. Is that right?

elcct: You are correct. I could replace it with a JSON containing the signed download URL instead.

datahub/investment/proposition/test/test_views.py
```diff
@@ -1097,3 +1112,240 @@ def test_cannot_abandon_proposition_without_details(self):
...
+                'doc_pk': doc.pk
+            }
+        )
+
+        response = self.api_client.delete(url)
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert mock_submit.called is False
+
+    @pytest.mark.usefixtures('synchronous_thread_pool', 'synchronous_on_commit')
+    @pytest.mark.parametrize('s3_stubber', ['investment'], indirect=True)
```
reupen: I wonder if there's a more direct way we could handle this? One option might be to turn s3_stubber into s3_stubber_factory and use request.addfinalizer() in it (to avoid a `with` in the test).

datahub/documents/av_scan.py
```diff
@@ -64,33 +28,37 @@ def _process_document(document_pk: str):
...
 
-def _scan_s3_object(original_filename, bucket, key):
-    """Virus scans a file stored in S3."""
-    s3_client = get_s3_client()
-    response = s3_client.get_object(Bucket=bucket, Key=key)
-    with closing(response['Body']):
+def _scan_object(name: str, download_url: str):
+    """Virus scans a file stored on remote server."""
+    with requests.get(download_url, stream=True) as response:
+        if response.status_code != 200:
```
reupen: You should be able to use `response.raise_for_status()` and reraise a `VirusScanException` if it raises an `HTTPError` (from requests.exceptions).

datahub/investment/proposition/test/test_views.py
elcct: I tried to use factory, but that needed more changes in the test, so I abandoned it and did this way instead. I could redo it once I fix other things.

datahub/documents/models.py
elcct: Now the endpoint returns JSON with `{"document_url": "https://..."}`. I also included `original_filename` in the document path so that the file would download under its name when browser is redirected to the destination S3 URL.

datahub/investment/proposition/views.py
elcct: I have added permissions in the similar flavour as for Proposition themselves. I think I need to write more tests for those permissions though.

datahub/documents/models.py
elcct: I am going to write docs this week.

datahub/documents/models.py
```diff
@@ -33,56 +37,82 @@ class Document(BaseModel, ArchivableModel):
...
+    original_filename = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH)
+    document = models.OneToOneField(Document, on_delete=models.PROTECT)
+
+    class Meta:
+        abstract = True
+
+    def delete(self, using=None, keep_parents=False):
+        """Ensure document is removed when entity document is being deleted."""
+        result = super().delete(using, keep_parents)
+        self.document.delete(using, keep_parents)
```
reupen: Doesn't this contradict the `on_delete=models.PROTECT` above? Or have I missed something? Maybe it's to do with bulk delete (on a QuerySet)?

elcct: I left it as in original implementation, but when I think about it that is indeed not right. I'll look into it. Thanks!

datahub/documents/tasks.py
```diff
@@ -0,0 +1,28 @@
+from celery.task import task
+from django_pglocks import advisory_lock
+
+from .av_scan import process_document
+from .models import Document
+from .utils import delete_s3_obj
+
+
+@task(ignore_result=True)
```
reupen: I believe this decorator is deprecated. I though shared_task was the one to use: http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html#using-the-shared-task-decorator

datahub/documents/views.py
```diff
@@ -0,0 +1,87 @@
...
+    def download(self, request, *args, **kwargs):
+        """Provides download information."""
+        model_document = self.get_object()
+
+        if not model_document.document.scanned_on:
+            raise TemporarilyUnavailableException()
+
+        url = model_document.document.get_signed_url()
+        if not url:
+            return HttpResponseForbidden()
```
reupen: Should we raise PermissionDenied (or another API exception) rather than return a response? Seems a bit inconsistent to use raise for one and return for the other.

conftest.py
```diff
@@ -49,10 +49,16 @@ def api_client():
...
+@pytest.fixture
+def s3_stubber(request):
+    """
+    S3 stubber using the botocore Stubber class.
+
+    bucket_id can be specified by a parameter, it defaults to 'default'.
+    """
+    bucket_id = request.param if hasattr(request, 'param') else 'default'
+
+    s3_client = get_s3_client_for_bucket(bucket_id)
```
marcofucci: I don't think you are ever using a param different from 'default' so you can get rid of request and request.param.

datahub/documents/views.py
```diff
@@ -0,0 +1,87 @@
...
+                    model_document.document.get_signed_url(unsafe=True),
+                )
+            )
+            status = upload_statuses.virus_scanning_scheduled
+        else:
+            status = (
+                upload_statuses.completed
+                if model_document.document.scanned_on
+                else upload_statuses.virus_scanning_in_progress
+            )
```
marcofucci: I've noticed that `Document.delete` triggers the task to delete the object from S3. 
Wouldn't make more sense to move this logic to the model as well for consistency?

I'm thinking about moving `status` to the model and having a method `schedule_av_scan` which returns that status and schedules the task if needed.
Also having another method `get_upload_status` which returns `complete` or `virus_scanning_in_progress` would be an extra nice touch as well.

datahub/documents/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+def delete_s3_document(document_id):
+    """Handle document delete."""
+    document = Document.objects.get(id=document_id)
```
marcofucci: By the time this is run, the document object might not be in the database any more.

datahub/documents/av_scan.py
```diff
@@ -1,58 +1,23 @@
...
-    anti-virus service.
-
-    Any errors are logged and sent to Sentry.
-    """
-    with advisory_lock(f'av-scan-{document_pk}'):
-        _process_document(document_pk)
-
-
-def _process_document(document_pk: str):
+def process_document(document_pk: str, download_url: str):
```
marcofucci: This should probably return without doing anything if the document has already been scanned.

marcofucci: Also by the time this is run, the document might have been deleted and might not exist.

datahub/documents/tasks.py
elcct: :D very good point!

conftest.py
elcct: One or more tests for proposition documents relies on that.

datahub/documents/views.py
marcofucci: Isn't clear enough that 403 is returned because the file has a virus anyway? Not sure it is.

datahub/documents/views.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if not model_document.document.scanned_on:
+            raise TemporarilyUnavailableException()
+
+        url = model_document.document.get_signed_url()
+        if not url:
+            return HttpResponseForbidden()
+
+        serializer = DownloadInfoSerializer(data={'document_url': url})
+        serializer.is_valid(raise_exception=True)
```
marcofucci: Not sure this is validating anything. Either get rid of this or move all the checks to the serializer instead.

datahub/investment/proposition/serializers.py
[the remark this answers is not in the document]
marcofucci: I'd say for consistency, you also already have an endpoint for listing all documents, isn't it?

datahub/investment/test/test_views.py
```diff
@@ -3227,36 +3230,39 @@ def test_documents_list_is_filtered_by_project(self):
...
+        assert response_data['original_filename'] == 'test.txt'
         assert response_data['project']['id'] == str(project.pk)
 
         doc = IProjectDocument.objects.get(pk=response_data['id'])
-        assert doc.filename == 'test.txt'
+        assert doc.original_filename == 'test.txt'
         assert doc.doc_type == 'total_investment'
         assert doc.project.pk == project.pk
-        assert 'signed_upload_url' in response.data
+        assert response_data['signed_upload_url'] == 'http://document-about-ocelots'
```
marcofucci: It would be nice to assert the whole response data instead of just fragments of it.

datahub/investment/serializers.py
```diff
@@ -480,12 +481,10 @@ def create(self, validated_data):
...
         )
 
-
-class UploadStatusSerializer(serializers.Serializer):
-    """Serializer for upload status endpoints."""
-
-    status = serializers.ChoiceField(choices=('success',))
+    def get_signed_url(self, obj):
+        """Gets signed URL."""
+        return obj.document.get_signed_url()
```
marcofucci: Isn't this inconsistent with the proposition download going through an extra endpoint?

datahub/documents/views.py
elcct: DRF doesn't let me access `serializer.data` without calling `is_valid`. I know this look pointless :( Do you have better idea? Or should I just serialize a dictionary to JSON?

datahub/investment/serializers.py
marcofucci: I think another PR is better, this is already pretty big.

datahub/investment/proposition/test/test_views.py
elcct: Ok I have removed that.

conftest.py
elcct: I am sorry I was wrong. I removed the code depending on this and forgot about it :( Removed the request bit.

datahub/documents/views.py
marcofucci: Oh yeah I forgot, nah that's fine.

conftest.py
```diff
@@ -49,10 +49,10 @@ def api_client():
     return APIClient()
 
 
-@pytest.fixture()
-def s3_stubber():
-    """S3 stubber using the botocore Stubber class"""
-    s3_client = get_s3_client()
+@pytest.fixture
+def s3_stubber(request):
```
reupen: Is it using `request` now?

datahub/documents/av_scan.py
```diff
@@ -1,96 +1,71 @@
...
```
reupen: Maybe we can come up with a better name for this function now that it's public.

datahub/documents/av_scan.py
```diff
@@ -1,96 +1,71 @@
...
 
-def _scan_raw_file(filename, file_object, content_type):
+def _scan_raw_file(name, content, content_type):
     """Virus scans a file-like object."""
     multipart_fields = {
         'file': (
-            filename,
-            file_object,
+            name,
+            content,
```
reupen: There is a v2 version of the API which we probably want to switch to (see the 'man'). It also gives a reason for not passing the scan which we may want to store or log (it can be because files are encrypted).

datahub/documents/av_scan.py
```diff
@@ -1,96 +1,71 @@
...
     doc.av_clean = is_file_clean
     doc.save()
 
 
-def _scan_s3_object(original_filename, bucket, key):
-    """Virus scans a file stored in S3."""
-    s3_client = get_s3_client()
-    response = s3_client.get_object(Bucket=bucket, Key=key)
-    with closing(response['Body']):
+def _scan_object(name: str, download_url: str):
```
reupen: I'm not sure `_scan_object` is a good name for this now that it's not tied to S3.

datahub/documents/av_scan.py
elcct: I have not used v2 API because we are not storing the reason for failure of the scan. I could think about what we could do with those responses and create another PR.

datahub/documents/tasks.py
elcct: I updated it so it doesn't take document_id.

datahub/investment/test/test_views.py
elcct: I have changed asserts so now it looks at whole response.

datahub/documents/models.py
```diff
@@ -1,27 +1,32 @@
...
 class Document(BaseModel, ArchivableModel):
     """General model for keeping track of user uploaded documents."""
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4)
-    path = models.CharField(
-        max_length=settings.CHAR_FIELD_MAX_LENGTH,
-        unique=True,
-    )
+    bucket_id = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH, default='default')
+    path = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH, unique=True)
```
reupen: Shouldn't `unique=True` be removed from path now it is bucket_id + path? (And a unique constraint added for the pair?)

datahub/documents/models.py
```diff
@@ -33,56 +38,111 @@ class Document(BaseModel, ArchivableModel):
...
+    def __str__(self):
+        """String repr."""
+        return f'Document(pk="{self.pk}", av_clean={self.av_clean})'
+
+    def delete(self, using=None, keep_parents=False):
+        """Ensure document is deleted from S3."""
+        result = super().delete(using, keep_parents)
+
+        if self.uploaded_on:
+            from .tasks import delete_s3_document
```
reupen: If we're going to use a local import, are we able to do it in the task instead? Since it seems like we could encounter circular imports more commonly there.

datahub/documents/models.py
```diff
@@ -33,56 +38,111 @@ class Document(BaseModel, ArchivableModel):
...
+            delete_s3_document.apply_async(args=(self.bucket_id, self.path,))
+
+        return result
+
+    def schedule_av_scan(self):
+        """Schedule AV scan."""
+        if not self.scan_initiated_on:
+            from .tasks import virus_scan_document
```
reupen: Same question as above.

datahub/documents/models.py
```diff
@@ -33,56 +38,111 @@ class Document(BaseModel, ArchivableModel):
...
+        """Get status of the upload."""
+        if self.scan_initiated_on is None:
+            return upload_statuses.not_virus_scanned
+        return (
+            upload_statuses.completed
+            if self.scanned_on
+            else upload_statuses.virus_scanning_in_progress
+        )
+
+    def get_signed_url(self, unsafe=False):
```
reupen: I think we could use a better name than `unsafe` e.g. `allow_unsafe` or `allow_infected`.

conftest.py
elcct: It doesn't. Removed. Thanks!

datahub/documents/av_scan.py
elcct: I have renamed it to `perform_virus_scan` - it says more what it does.

datahub/documents/models.py
```diff
@@ -33,56 +38,111 @@ class Document(BaseModel, ArchivableModel):
...
+    @classmethod
+    def create_document_path(cls, document_id, original_filename):
+        """Create document path."""
+        today = now().strftime('%Y-%m-%d')
+        return f'{cls._meta.model_name}/{today}/{document_id}/{original_filename}'
+
+    @classmethod
+    def create_from_declaration_request(cls, original_filename, **kwargs):
+        """Create entity document along with the corresponding Document."""
+        with transaction.atomic():
```
reupen: Can we use `transaction.atomic` as a decorator here?

datahub/documents/av_scan.py
elcct: I have renamed it to `_download_and_scan_file`.

datahub/documents/models.py
elcct: I am going to change to `allow_unsafe` as file may not necessarily be infected.

datahub/documents/models.py
```diff
@@ -33,56 +38,111 @@ class Document(BaseModel, ArchivableModel):
...
+        """Ensure document is removed when entity document is being deleted."""
+        result = super().delete(using, keep_parents)
+        self.document.delete(using, keep_parents)
+        return result
 
-    transaction.on_commit(
-        lambda: submit_to_thread_pool(delete_document)
-    )
+    @property
+    def upload_status(self):
```
reupen: Is there a reason that this is a property while the get_upload_status it calls on Document is a method? Seems like it would be better to be consistent about it if there isn't a reason for it.

datahub/documents/models.py
```diff
@@ -1,27 +1,32 @@
...
 
 from datahub.core.models import ArchivableModel, BaseModel
-from datahub.core.thread_pool import submit_to_thread_pool
-from datahub.core.utils import delete_s3_obj, sign_s3_url
+from .utils import sign_s3_url
 
 logger = getLogger(__name__)
 
 
+upload_statuses = Choices(
```
reupen: Sometimes we use UPPER_SNAKE_CASE for Choices instances, other times UpperCamelCase, but I haven't seen lower_snake_case used before. I would vote for caps over UpperCamelCase as it's not a type.

datahub/documents/tasks.py
```diff
@@ -0,0 +1,23 @@
+from celery.task import task
+from django_pglocks import advisory_lock
+
+from .av_scan import process_document
+from .utils import delete_s3_obj
+
+
+@task(ignore_result=True)
```
reupen: Is there a reason for using ignore_result? Also I left a comment earlier about using shared_task instead of task.

elcct: I am sorry, I haven't get to it yet. I'll update it soon. Thanks!

datahub/documents/models.py
elcct: Not sure why I didn't do it. Seems obvious! Thanks for spotting this.

datahub/documents/views.py
```diff
@@ -0,0 +1,72 @@
...
+from datahub.documents.av_scan import TemporarilyUnavailableException
+from .serializers import DownloadInfoSerializer, UploadStatusSerializer
+
+
+class BaseDocumentModelViewSet(CoreViewSet):
+    """Documents ViewSet."""
+
+    def list(self, request, *args, **kwargs):
+        """Custom pre-filtered list."""
+        queryset = self.filter_queryset(self.get_queryset())
```
reupen: This looks the same as the default implementation – is there a reason to override it?

datahub/documents/models.py
```diff
@@ -33,56 +38,111 @@ class Document(BaseModel, ArchivableModel):
...
+
+    def get_signed_url(self, unsafe=False):
         """Generate pre-signed download URL, but only if doc is AV clean."""
-        if self.av_clean:
-            return sign_s3_url(settings.DOCUMENTS_BUCKET, self.path)
+        if self.av_clean or unsafe:
+            return sign_s3_url(
+                self.bucket_id,
+                self.path,
+            )
```
reupen: Because there is an explicit `return <expression>`, it should use an explicit `return None` rather than an implicit one (as per PEP 8). But, tbh, I think raising an exception in that case would be better (if, one day, `sign_s3_url()` returned `None` for some reason, we wouldn't know the difference).

datahub/investment/proposition/views.py
```diff
@@ -122,3 +127,34 @@ def get_additional_data(self, create):
...
+        IsAssociatedToInvestmentProjectPropositionDocumentFilter,
+    )
+
+    def get_queryset(self):
+        """Returns proposition documents queryset."""
+        return PropositionDocument.objects.select_related(
+            'proposition__investment_project'
+        ).filter(
+            proposition_id=self.kwargs['pk'],
+            proposition__investment_project_id=self.kwargs['project_pk'],
```
reupen: I might have missed something, but where does `project_pk` come from? Why do we need to filter by it and `proposition_id`?

datahub/documents/views.py
```diff
@@ -0,0 +1,72 @@
...
+        serializer = DownloadInfoSerializer(data={'document_url': url})
+        serializer.is_valid(raise_exception=True)
+        return Response(serializer.data)
+
+    def perform_destroy(self, instance):
+        """Perform destroy in transaction/savepoint mode."""
+        with transaction.atomic():
+            return super().perform_destroy(instance)
+
+    def get_object(self):
```
reupen: Does this need to be overridden? Does it do the same thing as setting `lookup_url_kwarg` on the class would?

datahub/investment/proposition/urls.py
```diff
@@ -21,9 +21,47 @@
...
     path('proposition/<uuid:pk>', proposition_item, name='item'),
     path('proposition/<uuid:pk>/complete', proposition_complete, name='complete'),
     path('proposition/<uuid:pk>/abandon', proposition_abandon, name='abandon'),
+    path(
+        'proposition/<uuid:pk>/document',
+        proposition_document_collection,
+        name='document-collection'
+    ),
+    path(
+        'proposition/<uuid:pk>/document/<uuid:doc_pk>',
```
reupen: Not sure `pk` is a good name for the proposition ID. I would call it `proposition_pk`.

datahub/documents/models.py
elcct: I have removed circular imports by using `apps.get_model` in the av_scan.

datahub/documents/models.py
elcct: Same answer as above.

datahub/documents/models.py
elcct: Done.

datahub/documents/models.py
elcct: Maybe "convenience", but changed to use `get_upload_status`.

datahub/documents/models.py
elcct: Changed to UPPER_SNAKE_CASE :)

datahub/documents/tasks.py
elcct: I have replaced it with @shared_task

datahub/investment/proposition/views.py
elcct: It is included in `investment/<uuid:project_pk>/` path. So that full url would be `/v3/investment/<project_pk>/proposition/<proposition_pk>/document...`

elcct: Filtering by investment project is in case someone manipulates URL and try to use different investment project and/or proposition to access document. It would look "weird" if URL composed that way could return a document.

reupen: > It is included in investment/<uuid:project_pk>/ path.

Ah, now it makes sense :+1:

datahub/documents/views.py
elcct: I removed custom bits from this code and have not checked if that became same as default implementation :( Thanks!

datahub/documents/views.py
elcct: It doesn't need to. Changed to `lookup_url_kwarg`. Thanks!

datahub/investment/proposition/urls.py
elcct: Yes, it doesn't look great. I have replaced it with `proposition_pk`.

datahub/documents/models.py
elcct: That is true, I shouldn't have left it like this. I added explicit None for now.

datahub/investment/test/test_views.py
elcct: I am not going to change those tests here, because that will be removed (legacy document stuff).

datahub/documents/views.py
elcct: This is indeed inconsistent. Added message to PermissionDenied that the file contains a virus.

datahub/investment/proposition/serializers.py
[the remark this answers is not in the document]
elcct: I gave it a second thought and I think you're right. Mixing that in the propositions response is messy. I have removed that now.

datahub/documents/views.py
```diff
@@ -0,0 +1,52 @@
...
+    def download(self, request, *args, **kwargs):
+        """Provides download information."""
+        model_document = self.get_object()
+
+        if not model_document.document.scanned_on:
+            raise TemporarilyUnavailableException()
+
+        url = model_document.document.get_signed_url()
+        if not url:
+            raise PermissionDenied('File contains a virus.')
```
marcofucci: Can you test if `document.av_clean` instead? It's more obvious that the file contains a virus in that case.

datahub/documents/views.py
```diff
@@ -0,0 +1,52 @@
...
+
+        serializer = DownloadInfoSerializer(data={'document_url': url})
+        serializer.is_valid(raise_exception=True)
+        return Response(serializer.data)
```
marcofucci: I'd still move all the checks to the serializer and have `is_valid` raise all the exceptions if needed. It seems slightly cleaner.

datahub/documents/models.py
marcofucci: A bit of inconsistency here in the way you use `transaction.atomic` on create (on the method) and delete (on the view).

datahub/investment/proposition/views.py
```diff
@@ -122,3 +130,34 @@ def get_additional_data(self, create):
...
+    """Proposition Document ViewSet."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (
+        IsAuthenticatedOrTokenHasScope,
+        PropositionDocumentModelPermissions,
+        IsAssociatedToInvestmentProjectPropositionDocumentPermission,
+    )
+    serializer_class = PropositionDocumentSerializer
+    queryset = PropositionDocument.objects
```
marcofucci: It would be good to get rid of this line to make it less confusing, you have `get_queryset` already.

datahub/documents/serializers.py
```diff
@@ -11,5 +11,17 @@ class DocumentSerializer(serializers.ModelSerializer):
     """
 
     class Meta:
-        model = models.Document
+        model = Document
```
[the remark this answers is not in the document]
marcofucci: Also, is `DocumentSerializer` actually used anywhere?

datahub/documents/models.py
elcct: @reupen @marcofucci so I changed `on_delete` to `models.CASCADE` and the view would now delete corresponding document (which would then delete entity model as well). I also added back a signal to delete s3 file. It seems like the s3 stuff (file upload) is happening outside of the app (client uploads to the given S3 url), so it would make sense for the delete to happen separately.

datahub/documents/serializers.py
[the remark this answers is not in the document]
elcct: Serializer removed.

datahub/documents/models.py
```diff
@@ -33,56 +39,120 @@ class Document(BaseModel, ArchivableModel):
...
+
+    @classmethod
+    def create_document_path(cls, document_id, original_filename):
+        """Create document path."""
+        today = now().strftime('%Y-%m-%d')
+        return f'{cls._meta.model_name}/{today}/{document_id}/{original_filename}'
+
+    @classmethod
+    @transaction.atomic
+    def create_from_declaration_request(cls, original_filename, **kwargs):
```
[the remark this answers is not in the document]
elcct: Good idea. I have created manager instead.

datahub/documents/models.py
```diff
@@ -33,56 +39,120 @@ class Document(BaseModel, ArchivableModel):
...
+                    self.get_signed_url(allow_unsafe=True),
+                )
+            )
+            return UPLOAD_STATUSES.virus_scanning_scheduled
+
+        return self.get_upload_status()
+
+    def get_upload_status(self):
+        """Get status of the upload."""
+        if self.scan_initiated_on is None:
```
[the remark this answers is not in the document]
elcct: I have changed those checks to be consistent. Thanks!

datahub/documents/av_scan.py
```diff
@@ -1,97 +1,71 @@
...
+        return
+
+    if doc.scanned_on is not None or doc.scan_initiated_on is not None:
         return
 
     doc.scan_initiated_on = now()
     doc.save()
 
-    is_file_clean = _scan_s3_object(doc.filename, doc.s3_bucket, doc.s3_key)
+    is_file_clean = _download_and_scan_file(str(doc.pk), download_url)
```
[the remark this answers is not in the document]
elcct: I am looking into it. Thanks!

datahub/documents/av_scan.py
```diff
@@ -1,97 +1,71 @@
...
-    anti-virus service.
-
-    Any errors are logged and sent to Sentry.
-    """
-    with advisory_lock(f'av-scan-{document_pk}'):
-        _process_document(document_pk)
-
-
-def _process_document(document_pk: str):
+def perform_virus_scan(document_pk: str, download_url: str):
```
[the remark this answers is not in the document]
elcct: I have added a comment. Thanks

config/settings/common.py
```diff
@@ -354,8 +355,23 @@
...
-REPORT_AWS_REGION= env('REPORT_AWS_REGION', default='')
+DOCUMENT_BUCKETS = {
+    'default': {
+        'bucket': env('DOCUMENTS_BUCKET'),
+        'aws_access_key_id': env('AWS_ACCESS_KEY_ID', default=''),
+        'aws_secret_access_key': env('AWS_SECRET_ACCESS_KEY', default=''),
+        'aws_region': env('AWS_DEFAULT_REGION', default=''),
+    },
+    'investment': {
+        'bucket': env('INVESTMENT_DOCUMENT_BUCKET', default=''),
```
reupen: Given that we already have `DOCUMENTS_BUCKET` as an environment variable, I think this should be `INVESTMENT_DOCUMENTS_BUCKET` (plural documents).

datahub/documents/av_scan.py
```diff
@@ -1,97 +1,80 @@
...
-        return _scan_raw_file(
-            original_filename, S3StreamingBodyWrapper(response), response['ContentType']
+def _download_and_scan_file(name: str, download_url: str):
+    """Virus scans a file stored on remote server."""
+    with requests.get(download_url, stream=True) as response:
+        try:
+            response.raise_for_status()
+        except HTTPError as e:
+            raise VirusScanException(
+                f'Problem with the source file (status_code={e.response.status_code}).'
```
reupen: I'm not sure what 'Problem with the source file' means. It makes it sound like there's something wrong with the file itself when that might not be the case. I would go for something like 'Problem downloading file to be scanned'.

reupen:
Some more logging around the scanning would be good (might want to avoid logging the full pre-signed URLs, although probably difficult to avoid completely).

config/settings/common.py
elcct: I am going to remove `DOCUMENTS_BUCKET` and unused old investment document upload in the next PR, so that wouldn't be an issue. 
If I change document to plural, I would have to change report as well.

datahub/documents/av_scan.py
elcct: Maybe `Unable to download the source file for scanning`?

reupen: I'm not sure what 'source' is referring to, is there a destination file?

datahub/documents/signals.py
```diff
@@ -0,0 +1,12 @@
...
+
+from .models import Document
+from .tasks import delete_s3_document
+
+
+@receiver(post_delete, sender=Document, dispatch_uid='document_post_delete')
+def delete_s3_object_on_document_delete(sender, instance, **kwargs):
+    """Deletes S3 object after document is deleted."""
+    if instance.uploaded_on:
+        delete_s3_document.apply_async(args=(instance.bucket_id, instance.path,))
```
reupen: I think handling deletions in this way could lead to things ending up in an inconsistent state. It could happen in at least a couple of ways:

- the S3 object deletion fails (but the upload record is already deleted from the database)
- the transaction is rolled back (which could easily happen as we use ATOMIC_REQUESTS)

It might be a a more reliable model if we use the statuses for deletions instead of actually deleting from the database (e.g. add `deletion_pending` and `deleted` statuses). Anything in those statuses could be excluded from the API response.

datahub/documents/av_scan.py
elcct: Good point. I'd leave "source" then.

datahub/documents/views.py
elcct: I have removed these and replaced with just view's serializer.

datahub/documents/views.py
elcct: Done.

datahub/documents/av_scan.py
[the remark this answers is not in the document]
elcct: I have added additional status `virus_scanning_failed`, there is also additional field `av_reason` that stores exception converted to string. This field would be used when v2 virus scan API will be implemented.

datahub/documents/models.py
elcct: I have added docs in the `docs/Document upload.md`.

datahub/documents/signals.py
elcct: Good idea. I have changed it so that when deletion is requested, a document gets `deletion_pending` status and a task to delete the document and the file is being scheduled. I have also disabled deletion in the admin for now, so that I could add proper admin delete functionality later as this PR is getting bigger and bigger.

datahub/documents/av_scan.py
```diff
@@ -29,21 +28,25 @@ def perform_virus_scan(document_pk: str, download_url: str):
...
 
     document.mark_scan_initiated()
 
-    is_file_clean = _download_and_scan_file(str(document.pk), download_url)
+    try:
+        is_file_clean = _download_and_scan_file(str(document.pk), download_url)
+    except Exception as e:
+        document.mark_scan_failed(str(e))
+        logger.exception(f'Virus scanning of document with ID {document_pk} failed.')
+        raise e
```
reupen: Better to use `raise` on its own for re-raising.

datahub/documents/av_scan.py
```diff
@@ -29,21 +28,25 @@ def perform_virus_scan(document_pk: str, download_url: str):
...
+
+    # TODO: add reason from v2 API response
+    document.mark_as_scanned(is_file_clean, '')
 
-    document.mark_as_scanned(is_file_clean, 'what')
+    logger.info(f'Virus scanning of Document with ID {document_pk} completed.')
```
reupen: Might be nice to include if the document was clean or not?

datahub/documents/tasks.py
```diff
@@ -1,14 +1,24 @@
...
-from .utils import delete_s3_obj
+from .utils import delete_s3_obj, get_document_by_pk
 
 
 @shared_task
-def delete_s3_document(bucket_id, path):
+def delete_document(document_pk):
     """Handle document delete."""
-    delete_s3_obj(bucket_id, path)
+    with transaction.atomic():
```
reupen: Should be able to put `transaction.atomic()` as a decorator (after @shared_task).

datahub/documents/tasks.py
```diff
@@ -1,14 +1,24 @@
...
+        document = get_document_by_pk(document_pk)
+        if document and document.status == 'deletion_pending':
```
reupen: Is this a string literal instead of using UPLOAD_STATUSES to avoid a circular import? I think we might have to bite the bullet and use local imports with tasks...

datahub/documents/tasks.py
elcct: Updated. Thanks!

datahub/documents/av_scan.py
elcct: Yes, that will be helpful. Added.

datahub/documents/av_scan.py
elcct: True. Fixed. Thanks :)

datahub/documents/tasks.py
elcct: Yes, this is to avoid circular import. I am going to add local import then.

elcct: Updated.

datahub/documents/utils.py
```diff
@@ -1,10 +1,31 @@
...
+    This is to avoid circular imports from av_scan and tasks.
+    """
+    try:
+        document_model = apps.get_model('documents', 'Document')
+        document = document_model.objects.get(pk=document_pk)
+    except ObjectDoesNotExist:
+        logger.warning(f'Document with ID {document_pk} does not exist.')
+        return None
+
+    return document
```
reupen: Looks like this return could go straight in the `try:` block?

datahub/documents/av_scan.py
```diff
@@ -105,9 +89,9 @@ def _scan_raw_file(filename, file_object, content_type):
     )
     response.raise_for_status()
     if response.text not in ('OK', 'NOTOK'):
-        raise VirusScanException(f'Unexpected response from AV service: {response.content}')
+        content = response.content.decode('utf-8')
```
reupen: Not sure we should explicitly decode the response as UTF-8 if it's unexpected as we don't know what's in it. For bytes would use response.content as it is or for a string would just use response.text.

datahub/documents/tasks.py
```diff
@@ -0,0 +1,34 @@
...
+
+from .av_scan import perform_virus_scan
+from .utils import delete_s3_obj, get_document_by_pk
+
+
+@shared_task
+@transaction.atomic
+def delete_document(document_pk):
+    """Handle document delete."""
+    from .models import UPLOAD_STATUSES
```
reupen: Probably would be good to keep a blank line after local imports (when used)...

datahub/documents/av_scan.py
elcct: That is true. I think I meant to use .text :) Fixed, thanks!

datahub/documents/tasks.py
elcct: Sure, missed that :( fixed.

datahub/documents/views.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def download(self, request, *args, **kwargs):
+        """Provides download information."""
+        entity_document = self.get_object()
+
+        if not entity_document.document.scanned_on:
+            raise TemporarilyUnavailableException()
+
+        if not entity_document.document.av_clean:
+            raise PermissionDenied('File contains a virus.')
```
reupen: Could be that it was password-protected as well. I would say something less specific (e.g. 'file did not pass virus scanning').

elcct: This is a good point - updated the text. Thanks!

datahub/documents/av_scan.py
```diff
@@ -1,97 +1,81 @@
...
-    with closing(response['Body']):
-        return _scan_raw_file(
-            original_filename, S3StreamingBodyWrapper(response), response['ContentType']
+    document.mark_scan_initiated()
+
+    try:
+        is_file_clean = _download_and_scan_file(str(document.pk), download_url)
+    except Exception as e:
+        document.mark_scan_failed(str(e))
+        logger.exception(f'Virus scanning of document with ID {document_pk} failed.')
```
reupen: Won't the exception get logged anyway by Celery? Maybe just a logger.error() will be fine.

elcct: That's what I though, just I can see we always put something as a message when using logger.exception. I can change it to logger.error. Thanks!

datahub/documents/av_scan.py
```diff
@@ -1,97 +1,81 @@
...
+def _download_and_scan_file(document_pk: str, download_url: str):
+    """Virus scans a file stored on remote server."""
+    with requests.get(download_url, stream=True) as response:
+        try:
+            response.raise_for_status()
+        except HTTPError as e:
+            raise VirusScanException(
+                f'Unable to download the document with ID {document_pk} '
+                f'for scanning (status_code={e.response.status_code}).'
+            )
```
reupen: I would add `from e` at the end (`raise ... from e`).

Also we need to standardise on `e` or `exc` when using `except ... as`...

datahub/documents/models.py
```diff
@@ -31,58 +39,142 @@ class Document(BaseModel, ArchivableModel):
     scanned_on = models.DateTimeField(
         null=True, blank=True,
     )
+
     av_clean = models.NullBooleanField(db_index=True)
+    av_reason = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH, blank=True)
```
reupen: I think a TextField (without max_length) would be better for this, who knows how long the reasons could be...

datahub/documents/models.py
```diff
@@ -31,58 +39,142 @@ class Document(BaseModel, ArchivableModel):
...
-    def generate_signed_upload_url(self):
+    def schedule_av_scan(self):
+        """Schedule AV scan."""
+        if not self.scan_initiated_on:
+            self.mark_scan_scheduled()
+
+            virus_scan_document.apply_async(
+                args=(
+                    str(self.pk),
+                    self.get_signed_url(allow_unsafe=True),
```
reupen: Can the task get the pre-signed URL? It won't necessarily be executed straight away.

datahub/documents/models.py
```diff
@@ -31,58 +39,142 @@ class Document(BaseModel, ArchivableModel):
...
+        """Marks document that scan has been initiated."""
+        self.scan_initiated_on = now()
+        return self._update_status(UPLOAD_STATUSES.virus_scanning_in_progress)
+
+    def mark_scan_failed(self, reason):
+        """Marks document that scan has failed."""
+        self.av_reason = reason
+        return self._update_status(UPLOAD_STATUSES.virus_scanning_failed)
+
+    def _update_status(self, status, save=True):
```
reupen: I can't see where save=False is passed, do we need that option?

datahub/documents/av_scan.py
elcct: I missed that :( I prefer `e`, but I see that we use `exc` more often, so I changed to `exc`. Thanks!

datahub/documents/models.py
elcct: This is a good idea as the char field may be indeed too short. Changed.

datahub/documents/models.py
elcct: Yes, sending the URL like this is also a bit redundant. I have removed that. Thanks!

datahub/documents/models.py
```diff
@@ -31,58 +39,142 @@ class Document(BaseModel, ArchivableModel):
...
+        self.status = status
+        if save:
+            self.save()
+        return self.status
+
+    def get_signed_url(self, allow_unsafe=False):
+        """Generate pre-signed download URL, but only if doc is AV clean."""
```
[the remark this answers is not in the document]
elcct: That is true. I have updated the docstring. Thanks!

datahub/documents/models.py
elcct: I thought I'd need it, but it is not being used. Removed. Thanks!

datahub/documents/models.py
```diff
@@ -31,58 +39,142 @@ class Document(BaseModel, ArchivableModel):
...
+    def mark_as_scanned(self, av_clean, av_reason):
+        """Marks document as scanned."""
+        self.scanned_on = now()
+        self.av_clean = av_clean
+        self.av_reason = av_reason
+        return self._update_status(UPLOAD_STATUSES.virus_scanned)
+
+    def mark_scan_scheduled(self):
+        """Marks document scan has been scheduled."""
+        self.uploaded_on = now()
```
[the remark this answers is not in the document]
elcct: I am not sure if I like this. That would bring some "magic" to the model which could make it more difficult to understand what is happening? I have not used it before though, so maybe my worries are unfounded.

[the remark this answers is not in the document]
marcofucci: yeah I agree, too magic.

datahub/documents/av_scan.py
```diff
@@ -1,97 +1,81 @@
...
-
-
-def virus_scan_document(document_pk: str):
-    """Virus scans an uploaded document.
-
-    This is intended to be run in the thread pool executor. The file is streamed from S3 to the
-    anti-virus service.
-
-    Any errors are logged and sent to Sentry.
+def perform_virus_scan(document_pk: str, download_url: str):
```
marcofucci: Maybe a stupid question but: why do you need the download_url to be passed in as argument?
You could have just `document_pk` and inside the function get `download_url = document.get_signed_url(allow_unsafe=True)`.

datahub/documents/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+def virus_scan_document(document_pk: str):
+    """Virus scans an uploaded document.
+
+    The file is streamed from S3 to the anti-virus service.
+
+    Any errors are logged and sent to Sentry.
+    """
+    with advisory_lock(f'av-scan-{document_pk}'):
+        document = get_document_by_pk(document_pk)
+        download_url = document.get_signed_url(allow_unsafe=True)
```
marcofucci: `document` can be None

datahub/documents/av_scan.py
marcofucci: Also, you actually need `download_url` in `_download_and_scan_file` only.

elcct: The idea was to have this independent from the model, but since this is already tied to the `Document` and I pass `document_pk`  (used as a name really) I can move it down to the `_download_and_scan_file` and this could be separated when there will be a need for that.

datahub/documents/tasks.py
elcct: That's true - added a check. Thanks

datahub/documents/models.py
```diff
@@ -31,58 +39,141 @@ class Document(BaseModel, ArchivableModel):
...
+        if self.av_clean or allow_unsafe:
+            return sign_s3_url(
+                self.bucket_id,
+                self.path,
+            )
+        return None
+
+    def get_signed_upload_url(self):
         """Generate pre-signed upload URL."""
         return sign_s3_url(
```
marcofucci: Might be worth adding an assert that `self.scan_initiated_on is None` as you should never call this method in case it's not.

datahub/documents/models.py
```diff
@@ -31,58 +39,141 @@ class Document(BaseModel, ArchivableModel):
...
+        or allow_unsafe is set.
+        """
+        if self.av_clean or allow_unsafe:
+            return sign_s3_url(
+                self.bucket_id,
+                self.path,
+            )
+        return None
+
+    def get_signed_upload_url(self):
```
marcofucci: Might be worth adding an assert that self.scan_initiated_on is None as you should never call this method in case it's not.

elcct: I agree it is worth adding. Done.

datahub/documents/views.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+
+
+class BaseEntityDocumentModelViewSet(CoreViewSet):
+    """Documents ViewSet."""
+
+    lookup_url_kwarg = 'entity_document_pk'
+    filter_backends = (
+        ExcludeDocumentStatusDeletionPendingFilter,
+    )
```
marcofucci: I'm convinced that it's a good idea to use `filter_backends` for ignoring objects in deletion pending. 
Maybe I would change the default manager to never return objects in delete pending and define another special method on the queryset or another manager for accessing those objects as well that the task can use. This would avoid accidental access to objects in pending deletion but would maybe make things less clear as well.

That being said, if you and @reupen are okay with the current implementation I'm okay with that too.

elcct: No, you're right. I think use of `filter_backend` for this is wrong - as you pointed out it is easy to forget and also cumbersome. I think I forgot that I could use manager for that :blush:, so thanks for pointing that out! I have removed this `filter_backend` and did it through manager.

datahub/documents/models.py
```diff
@@ -31,58 +39,156 @@ class Document(BaseModel, ArchivableModel):
...
+    """Base document model."""
+
+    BUCKET = 'default'
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    original_filename = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH)
+    document = models.OneToOneField(Document, on_delete=models.CASCADE)
+
+    objects = EntityDocumentManager()
+    all_objects = BaseEntityDocumentManager()
```
reupen: Not sure 'objects' and 'all_objects' are clearly differentiated enough (also EntityDocumentManager vs BaseEntityDocumentManager) – maybe adding another method on EntityDocumentManager instead (like `including_pending_deletion()`) would be clearer?

(Otherwise, I would try to name all_objects, EntityDocumentManager and BaseEntityDocumentManager more descriptively.)

datahub/documents/models.py
```diff
@@ -31,58 +39,156 @@ class Document(BaseModel, ArchivableModel):
...
-    transaction.on_commit(
-        lambda: submit_to_thread_pool(delete_document)
-    )
+    def get_queryset(self):
+        """Exclude objects with document having deletion pending as status."""
+        return super().get_queryset().exclude(document__status=UPLOAD_STATUSES.deletion_pending)
+
+
+class AbstractEntityDocumentModel(BaseModel):
+    """Base document model."""
```
reupen: It would be nice to add a small note about the behaviour regarding pending_deletion.

datahub/documents/test/test_av_scan.py
```diff
@@ -19,76 +20,118 @@
...
+            'Content-Length': '1000',
+        }
+    )
     requests_mock.post(
         'http://av-service/',
         text='OK',
         status_code=status.HTTP_400_BAD_REQUEST,
     )
 
     with pytest.raises(HTTPError) as exc:
```
reupen: I see it was like this before, but `exc` should be called `excinfo` really as it's not the exception by itself.

datahub/documents/test/test_tasks.py
```diff
@@ -0,0 +1,63 @@
...
+        original_filename='test.txt',
+        my_field='lions',
+    )
+    document = entity_document.document
+    document.uploaded_on = now()
+    document.mark_deletion_pending()
+
+    bucket_name = get_bucket_name(document.bucket_id)
+
+    s3_stubber.add_response('delete_object', {
```
reupen: This should be using add_client_error(). Errors will result in an exception being raised.

datahub/investment/proposition/models.py
```diff
@@ -54,6 +56,50 @@ class PropositionPermission(StrEnum):
...
+    project that was created by an adviser in the same team, or an adviser in the same team has
+    been linked to the project).
+
+    Note that permissions on other models are independent of permissions on proposition documents.
+    Also note that if both *_all_* and *_associated_investmentproject_* permissions are
+    assigned to the same user,  the *_all_* permission will be the effective one.
+    """
+
+    read_all = 'read_all_propositiondocument'
+    read_associated_investmentproject = 'read_associated_investmentproject_propositiondocument'
```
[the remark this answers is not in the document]
elcct: I am happy to do that. It seems like we have a convention with including investmentproject to these.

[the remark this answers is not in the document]
reupen: It could stay but I was just thinking about making them shorter. I did it originally for interactions because there are non-investment-project interactions as well, but you're right it's in other places as well now.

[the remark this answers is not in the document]
elcct: I have removed it. I will have similar set of permissions for Evidence, so it will be good to keep it shorter.

datahub/documents/test/test_tasks.py
elcct: That's right. I have changed this test and added another one to test response other than 204, that is not a client error. Also changed the document delete function going with @marcofucci suggestion, so that the document is no longer checked in the task, but in `perform_delete_document` function.

datahub/documents/utils.py
```diff
@@ -67,14 +68,47 @@ def sign_s3_url(bucket_id, key, method='get_object', expires=3600):
...
+            )
+            status_code = response['ResponseMetadata']['HTTPStatusCode']
+            if status_code != 204:
+                raise DocumentDeleteException(
+                    f'Invalid response for Document with ID {document_pk} from S3'
+                    f' (status_code={status_code}).'
+                )
+
+        document.delete()
+    else:
```
reupen: I would deal with the error case first, so:

```
if document.status != UPLOAD_STATUSES.deletion_pending:
   raise ...
```

And then reduce the indentation of the rest of the code a level.

datahub/documents/utils.py
```diff
@@ -67,14 +68,47 @@ def sign_s3_url(bucket_id, key, method='get_object', expires=3600):
...
+
+            client = get_s3_client_for_bucket(bucket_id)
+            bucket_name = get_bucket_name(bucket_id)
+
+            response = client.delete_object(
+                Bucket=bucket_name,
+                Key=document.path,
+            )
+            status_code = response['ResponseMetadata']['HTTPStatusCode']
+            if status_code != 204:
```
reupen: I'm not really sure why we need to check the status code, the client library does whatever checks are needed for us. But it's not a big deal.

datahub/documents/models.py
```diff
@@ -31,58 +39,165 @@ class Document(BaseModel, ArchivableModel):
...
-    @property
-    def s3_bucket(self):
-        """The S3 bucket where this document is stored."""
-        return settings.DOCUMENTS_BUCKET
 
-    @property
-    def s3_key(self):
-        """The S3 key corresponding to this bucket."""
-        return self.path
+class EntityDocumentManager(models.Manager):
```
marcofucci: I don't know if we should have a manager that excludes records in deletion pending for the Document model as well to keep things consistent? What do you guys think?

datahub/documents/utils.py
elcct: Good point. I should have done that. Thanks!

datahub/documents/utils.py
elcct: That is true. I have removed it. Thanks!

datahub/documents/models.py
elcct: I think that accessing `Document` would be sort of "low level" thing and it shouldn't be exposed to the clients? For example I believe admin should see all documents, also those with deletion_pending status. So if I added this kind of manager to Documents it wouldn't be used.

PR #1023, opened 2018-07-12
Add clean-up command for deleting very old interactions

reupen:
Issue number: n/a

### Description of change

This adds a `delete_old_records` command for deleting very old interactions.

It also adds a common base class for the `delete_old_records` and `delete_orphans` commands, which has some changes from the original command:

* `--simulate` now uses a transaction with rollback to fully simulate the deletion (including any errors and which exact models were deleted)
* the printing of the return value from QuerySet.delete() was improved
* `--only-print-query` was added (only prints the SQL query)

The tests are shared between commands at the moment, but we can change that later on if that doesn't prove practical.

### Checklist

  _Nothing about these commands in the README at the moment_


datahub/cleanup/management/commands/delete_old_records.py
```diff
@@ -0,0 +1,20 @@
+from logging import getLogger
+
+from dateutil.relativedelta import relativedelta
+
+from datahub.cleanup.cleanup_config import ModelCleanupConfig
+from datahub.cleanup.management.commands._base_command import BaseCleanupCommand
+
+logger = getLogger(__name__)
```
marcofucci: Is this needed?

reupen: Oh no, thanks!

conftest.py
```diff
@@ -49,6 +49,33 @@ def api_client():
     return APIClient()
 
 
+class _ReturnValueTracker:
+    def __init__(self, cls, method_name):
+        self.return_values = []
+        self.cls = cls
+        self.method_name = method_name
```
marcofucci: I'm not sure we need to store `cls` and `method_name` on `self`.

conftest.py
```diff
@@ -49,6 +49,33 @@ def api_client():
...
+            return_value = self.original_method(obj, *args, **kwargs)
+            self.return_values.append(return_value)
+            return return_value
+
+        return _spy
+
+
+@pytest.fixture
+def track_return_values(monkeypatch):
+    """Fixture that tracks the return values of a method."""
```
marcofucci: I'd add a few lines to document its usage with an example.

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+            (ContactFactory, 'company'),
+            (ShallowInvestmentProjectFactory, 'intermediate_company'),
+            (ShallowInvestmentProjectFactory, 'investor_company'),
+            (ShallowInvestmentProjectFactory, 'uk_company'),
+            (OrderFactory, 'company'),
+            (CompanyFactory, 'global_headquarters'),
+            (CompanyFactory, 'parent'),
+            (BusinessLeadFactory, 'company'),
+        ),
+        'implicit_related_models': (),
```
marcofucci: It would be good to change the factory and add `export_to_countries` and `future_interest_countries` here for completeness.

(I also realised today that the SQL does not print these tables which is something we might want to add probably in another PR).

Ideally it would be good to have `implicit_related_models` build dynamically from the model but it can be in a different PR.

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+        ),
+        'implicit_related_models': (
+            'event.Event_teams',
+            'event.Event_related_programmes',
+        ),
+    },
+    'interaction.Interaction': {
+        'factory': CompanyInteractionFactory,
+        'dependent_models': (),
+        'implicit_related_models': (),
```
marcofucci: Same here for `policy_areas`.

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+
+    for receiver in disconnected_signal_receivers:
+        receiver.connect()
+
+
+@pytest.fixture(params=(
+    (command_cls, model_name, config)
+    for command_cls in COMMAND_CLASSES
+    for model_name, config in command_cls.CONFIGS.items()
+), ids=_format_tuple)
```
marcofucci: This is not really easy to read, I'd do
```python
@pytest.fixture(
    params=(
        (command_cls, model_name, config)
        for command_cls in COMMAND_CLASSES
        for model_name, config in command_cls.CONFIGS.items()
    ), 
    ids=_format_tuple
)
```

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+        config,
+        MAPPINGS[model_name],
+        dep_factory,
+        dep_field_name,
+    )
+    for command_cls in COMMAND_CLASSES
+    for model_name, config in command_cls.CONFIGS.items()
+    for dep_factory, dep_field_name in MAPPINGS[model_name]['dependent_models']
+), ids=_format_cleanup_mapping
+)
```
marcofucci: Same here re. formatting

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+
+
+def _format_tuple(value):
+    return ', '.join(str(item) for item in value)
+
+
+@pytest.fixture(params=COMMAND_CLASSES, ids=str)
+def cleanup_command(request):
+    """Fixture that parametrises tests for each clean-up command."""
+    yield request.param()
```
marcofucci: I think this is unnecessary and it's used only once in this module.
`@pytest.mark.parametrize('cleanup_command', COMMAND_CLASSES)` is easier to read IMHO.

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+    },
+    'interaction.Interaction': {
+        'factory': CompanyInteractionFactory,
+        'dependent_models': (),
+        'implicit_related_models': (),
+    },
+}
+
+
+def _format_tuple(value):
```
marcofucci: I'd use a better name e.g. `iterable_as_string` or something like that

datahub/cleanup/management/commands/_base_command.py
```diff
@@ -0,0 +1,96 @@
...
+                for deletion_model, model_deletion_count in deletions_by_model.items():
+                    logger.info(f'{deletion_model}: {model_deletion_count}')
+
+                if is_simulation:
+                    logger.info(f'Rolling back deletions...')
+                    raise SimulationRollback()
+        except SimulationRollback:
+            logger.info(f'Deletions rolled back')
+
+    def _get_query(self, model):
```
marcofucci: I'd add this to the command class so that it can be extended if needed (without the underscore 😆 )

marcofucci:
The rollback in simulation looks nice but I wonder how much value it adds as we'll always run it in staging first. I guess it doesn't hurt to have it anyway.

datahub/cleanup/management/commands/_base_command.py
marcofucci: oh sorry didn't see it was already on the class, I'm not sure it needs to be private.

datahub/cleanup/management/commands/delete_old_records.py
```diff
@@ -0,0 +1,16 @@
...
+class Command(BaseCleanupCommand):
+    """Command for deleting very old records (as per the data retention policy)."""
+
+    help = ('Irrevocably deletes very old records for a model, using the criteria defined in the '
+            'DIT Data Hub retention policy. A simulation can be performed using the --simulate '
+            'argument.')
+
+    CONFIGS = {
+        'interaction.Interaction': ModelCleanupConfig(relativedelta(years=10), 'date'),
+    }
```
marcofucci: I don't think this should use `get_unreferenced_objects_query`. 
Delete old records should not check for references, it works at the moment because interactions don't have any dependent models but if you add another model the logic doesn't hold.

I'd use a generic queryset in the common bit and use the `get_unreferenced_objects_query` bit only in the `delete_orphans` command.

reupen: I thought we were going to use a variant of `get_unreferenced_objects_query()` for the other old record commands that filters related records? That's why I went for that.

reupen:
> The rollback in simulation looks nice but I wonder how much value it adds as we'll always run it in staging first. I guess it doesn't hurt to have it anyway.

Even in staging it's useful if you want to test the command without actually deleting anything. The old simulation was not really a simulation (and did not tell you what would actually be deleted)...

datahub/cleanup/management/commands/_base_command.py
reupen: It's easy to change it to public later if there's a reason to. There's no reason right now, I only moved it out of handle() because that function was getting quite big.

datahub/cleanup/management/commands/delete_old_records.py
marcofucci: Yeah, I've just realised that this piece of work was the retention stuff. 
The problem is that if you add another model at the moment it acts like the orphaned command, it does not look for related objects not referenced for a long time. You don't see any problems at the moment as the interactions don't have any dependant objects.

datahub/cleanup/cleanup_config.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    Clean-up configuration for a model.
+
+    Defines the criteria for determining which records should be cleaned up.
+    """
+
+    # Records older than this will be cleaned up
+    age_threshold: relativedelta
+    # The field to use for determining the age of a record
+    date_field: str = 'modified_on'
```
marcofucci: Isn't is obvious that it's a string? We don't need the `str`

reupen: Yep

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+            config,
+            MAPPINGS[model_name],
+            dep_factory,
+            dep_field_name,
+        )
+        for command_cls in COMMAND_CLASSES
+        for model_name, config in command_cls.CONFIGS.items()
+        for dep_factory, dep_field_name in MAPPINGS[model_name]['dependent_models']
+    ),
+    ids=_format_cleanup_mapping,
```
marcofucci: Wondering if `ids=lambda p: _format_iterable((*p[:2], *p[-1:]))` could work here to avoid an extra function `_format_cleanup_mapping` used just for this. Not sure if I like it or not though.

datahub/cleanup/test/commands/test_common.py
```diff
@@ -0,0 +1,373 @@
...
+
+@pytest.fixture(
+    params=(
+        (command_cls, model_name, config)
+        for command_cls in COMMAND_CLASSES
+        for model_name, config in command_cls.CONFIGS.items()
+    ),
+    ids=_format_iterable,
+)
+def cleanup_args(request):
```
marcofucci: Not sure about `cleanup_args`, would `cleanup_configs` be clearer maybe?

datahub/cleanup/management/commands/_base_command.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        simulation_group = parser.add_mutually_exclusive_group()
+        simulation_group.add_argument(
+            '--simulate',
+            action='store_true',
+            help='Simulates the command by performing the deletions and rolling them back. Also'
+                 'prints the SQL query.',
+        )
+        simulation_group.add_argument(
+            '--only-print-query',
```
marcofucci: Can we change it to `only-print-queries`, I'll add other queries later on and it would be silly to add backward incompatibility if we know that already.

datahub/cleanup/test/commands/test_common.py
reupen: It might work if we used a named tuple or similar, otherwise readability suffers.

datahub/cleanup/test/commands/test_common.py
reupen: Yeah, I was struggling with names. cleanup_configs didn't work, because a clean-up config is one of the things cleanup_args contains :cry:

PR #1031, opened 2018-07-20
Delete orphaned versions

marcofucci:
### Description of change

This adds a new command `delete_orphaned_versions` which deletes all the `reversion.Version` records related to objects no longer in the database.
`reversion.Revision` records without versions are deleted as well.

`--model-label` can be passed in to specify the model(s) to consider. 
All models are used if model_label is not passed in.

### Checklist

~* [ ] Have any relevant search models been updated?~
~* [ ] Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?~
~* [ ] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?~
~* [ ] Has the admin site been updated (for new models, fields etc.)?~
~* [ ] Has the README been updated (if needed)?~

### DIFF
--- a/datahub/cleanup/management/commands/delete_orphaned_versions.py
+++ b/datahub/cleanup/management/commands/delete_orphaned_versions.py
@@ -1,0 +1,53 @@
+from collections import Counter
+from logging import getLogger
+
+import reversion
+from django.apps import apps
+from django.core.management import BaseCommand
+from django.db import transaction
+from reversion.models import Revision, Version
+
+logger = getLogger(__name__)
+
+
+def _get_all_model_labels():
+    return [model._meta.label for model in reversion.get_registered_models()]
+
+
+class Command(BaseCommand):
+    """Deletes all django versions for models that no longer exist in the database."""
+
+    def __repr__(self):
+        """Python representation (used for parametrised tests)."""
+        module_name = self.__class__.__module__.rsplit('.', maxsplit=1)[1]
+        return f'{module_name}.{self.__class__.__name__}()'
+
+    def add_arguments(self, parser):
+        """Handle arguments."""
+        parser.add_argument(
+            '--model-label',
+            action='append',
+            choices=_get_all_model_labels(),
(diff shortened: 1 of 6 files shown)


marcofucci:
Not sure about the command name, another option could be `delete_unreferenced_versions`.

datahub/cleanup/management/commands/delete_orphaned_versions.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __repr__(self):
+        """Python representation (used for parametrised tests)."""
+        module_name = self.__class__.__module__.rsplit('.', maxsplit=1)[1]
+        return f'{module_name}.{self.__class__.__name__}()'
+
+    def add_arguments(self, parser):
+        """Handle arguments."""
+        parser.add_argument(
+            '--model_label',
```
reupen: How about `--model-label`?

datahub/cleanup/management/commands/delete_orphaned_versions.py
```diff
@@ -0,0 +1,74 @@
...
+    return [model._meta.label for model in reversion.get_registered_models()]
+
+
+class DeletionCounter:
+    """Counter counting all items deleted also grouping them by model."""
+
+    def __init__(self):
+        """Initialises the object."""
+        self.total_deleted = 0
+        self.deletions_by_model = defaultdict(int)
```
reupen: Are we able to use collections.Counter? It accepts dicts. It could replace `self.deletions_by_model` (or we could get rid of DeletionCounter entirely).

datahub/cleanup/management/commands/delete_orphaned_versions.py
marcofucci: 👍 Is it for consistency or there's another reason as well?

datahub/cleanup/management/commands/delete_orphaned_versions.py
marcofucci: yeah, we can ignore the `total_deleted` as well I guess.

PR #1036, opened 2018-07-26
Add company core team endpoint

marcofucci:
### Description of change

This adds a new endpoint `/v3/company/<company-pk>/core-team` which returns a list with only the global account manager in it at the moment.

The next step will be to create a new table for team members and add those to the returned list.

This also refactor `NestedRelatedField` to allow explicit field mapping as well.

### Checklist

* [ ] ~Have any relevant search models been updated?~
* [ ] ~Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?~
* [ ] ~Has the admin site been updated (for new models, fields etc.)?~
* [ ] ~Has the README been updated (if needed)?~

### DIFF
--- a/datahub/company/views.py
+++ b/datahub/company/views.py
@@ -4,6 +4,7 @@ from django_filters import FilterSet
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import mixins, viewsets
 from rest_framework.filters import OrderingFilter
+from rest_framework.response import Response
 
 from datahub.core.audit import AuditViewSet
 from datahub.core.mixins import ArchivableViewSetMixin
@@ -20,6 +21,7 @@ from .queryset import get_contact_queryset
 from .serializers import (
     AdviserSerializer,
     CompaniesHouseCompanySerializer,
+    CompanyCoreTeamMemberSerializer,
     CompanySerializer,
     ContactSerializer,
 )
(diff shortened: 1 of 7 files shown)


datahub/core/serializers.py
```diff
@@ -75,7 +90,10 @@ def __init__(self, model, extra_fields=('name',), **kwargs):
                        model)
 
         self.pk_field = UUIDField()
-        self._fields = extra_fields
+        self._fields = [
+            field if type(field) == tuple else (field, ReadOnlyUUIDAwareField())
```
reupen: `isinstance(field, tuple)` instead of `type(field) == tuple`?

datahub/core/serializers.py
```diff
@@ -4,7 +4,19 @@
...
 from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
-from rest_framework.fields import UUIDField
+from rest_framework.fields import ReadOnlyField, UUIDField
+
+
+class ReadOnlyUUIDAwareField(ReadOnlyField):
+    """Same as `ReadOnlyField` but it also converts UUIDs to strings."""
+
+    def to_representation(self, value):
```
reupen: Is this actually needed (as opposed to just using `ReadOnlyField`)? I know it's what `_value_for_json()` was doing before, but maybe it was wrong (or became unnecessary). If I comment out this method the only test that fails is `TestNestedRelatedField.test_to_representation_extra_fields`.

marcofucci: Yeah, I don't like this and I started with `ReadOnlyField` but then I noticed that test failing so I put back the conversion.

I think DRF converts things for us during the API response but when calling to_representation manually UUID is returned instead.

Not sure what's best.

datahub/company/views.py
```diff
@@ -58,6 +60,38 @@ class CompanyViewSet(ArchivableViewSetMixin, CoreViewSet):
...
+
+    def list(self, request, *args, **kwargs):
+        """Lists core team members."""
+        company = self.get_object()
+        objs = []
+        if company.one_list_account_owner:
+            objs.append(
+                {
+                    'adviser': company.one_list_account_owner,
+                    'is_global_account_manager': True
```
reupen: If there isn't some detail I'm missing, I think we should be consistent with the field names in the API (the adviser came from `one_list_account_owner`, but the name of this field includes  `global_account_manager`).

marcofucci: yeah, the fields `account_manager` is going as it's legacy. 

The field `one_list_account_owner` is probably going to go as well and moved to a CoreTeamMember table. If not, it will be renamed as `one_list_account_owner` is not the correct name.

reupen: Also not sure what will come with the team members when they are added but I guess there will be some kind of role? Is a boolean consistent with that?

datahub/company/views.py
```diff
@@ -58,6 +60,38 @@ class CompanyViewSet(ArchivableViewSetMixin, CoreViewSet):
     )
 
 
+class CoreTeamViewSet(CoreViewSet):
+    """
+    Views for the core team of a company.
```
reupen: What does core mean here? Maybe that could be elaborated on.

Also it looks tied to Company so maybe Company should be in the name?

datahub/company/views.py
marcofucci: We haven't done enough user research on roles yet as it seems that the global account manager is the only thing really needed and the other roles are very loose and change depending on the context.

We're therefore starting with a loose implementation (only boolean field for global account manager) and move to a more strict role based one if needed.

datahub/company/views.py
```diff
@@ -60,11 +60,28 @@ class CompanyViewSet(ArchivableViewSetMixin, CoreViewSet):
     )
 
 
-class CoreTeamViewSet(CoreViewSet):
+class CompanyCoreTeamViewSet(CoreViewSet):
     """
     Views for the core team of a company.
 
-    At the moment, this only includes company.one_list_account_owner.
+    When a company is account managered, a core team is established.
```
elcct: `managered` - should be `managed`?

PR #1041, opened 2018-07-31
Add CompanyCoreTeamMember model and update endpoint

marcofucci:
Issue number: N/A

### Description of change

This:
- adds a new CompanyCoreTeamMember model
- changes the `/company/<pk>/core-team` endpoint to return that list along with the global account manager
- enables adding/changing the core team from the django admin

### Checklist

* ~[ ] Have any relevant search models been updated?~

Updated company admin:

screen shot 2018-07-31 at 10 51 58


datahub/company/views.py
```diff
@@ -96,15 +75,36 @@ class CompanyCoreTeamViewSet(CoreViewSet):
...
-                    'adviser': company.one_list_account_owner,
+                    'adviser': global_account_manager,
                     'is_global_account_manager': True
                 }
             )
 
+        # add all other core members excluding the global account manager
+        # who might have already been added
+        team_members = company.core_team_members.exclude(
+            adviser_id=getattr(global_account_manager, 'id', None)
```
reupen: Not sure about the use of `getattr()` here. It makes it seem like it's checking if `id` exists on `global_account_manager` when it's checking that `global_account_manager` is not `None`.

datahub/company/views.py
```diff
@@ -61,28 +61,7 @@ class CompanyViewSet(ArchivableViewSetMixin, CoreViewSet):
...
-    This team is called "core team" because it's official and does not change
-    often. Usually, a wider team around a company is established as well.
-    This team includes specialists and other advisers needed for short-term
-    and more reactive support.
-
-
-    At the moment, this endpoint only includes company.one_list_account_owner
-    representing the global account manager.
-    """
+    """Views for the core team of a company."""
```
reupen: I guess this uses Company permissions and CompanyCoreTeamMember ones aren't checked? Which is probably OK, but it would be good to mention it in the docstring.

datahub/company/views.py
```diff
@@ -96,15 +75,36 @@ class CompanyCoreTeamViewSet(CoreViewSet):
...
+        team_members = company.core_team_members.exclude(
+            adviser=global_account_manager
+        ).select_related(
+            'adviser',
+            'adviser__dit_team',
+            'adviser__dit_team__uk_region',
+            'adviser__dit_team__country',
+        )
+        for team_member in team_members:
+            objs.append(
```
reupen: Would normally prefer `list.extend()` with an iterable to `list.append()` in a loop.

marcofucci: yeah slightly better, I'd agree

datahub/company/views.py
```diff
@@ -96,15 +82,37 @@ class CompanyCoreTeamViewSet(CoreViewSet):
...
+                    'adviser': global_account_manager,
                     'is_global_account_manager': True
                 }
             )
 
+        # add all other core members excluding the global account manager
+        # who might have already been added
+        team_members = company.core_team_members.exclude(
+            adviser=global_account_manager
+        ).select_related(
```
reupen: These are in an undefined order at the moment, right? Would be nice to sort them (if only to make it deterministic).

datahub/company/models/company.py
```diff
@@ -188,6 +190,49 @@ def has_valid_trading_address(self):
...
+
+    When a company is account managed, a core team is established.
+    This usually includes:
+    - one and only one global account manager
+    - a local account manager from the country where the company is based
+    - one or more local account managers from the country where the company
+        is exporting to or investing in
+
+    However, this layout is not always as strict.
+    Other roles might exist and an single person can also have multiple roles.
```
elcct: `an single person` should be `a single person`?

datahub/company/views.py
```diff
@@ -64,24 +64,10 @@ class CompanyCoreTeamViewSet(CoreViewSet):
     """
     Views for the core team of a company.
 
-    When a company is account managed, a core team is established.
-    This usually includes:
-    - one and only one global account manager
-    - a local account manager from the country where the company is based
-    - one or more local account managers from the country where the company
-        is exporting to or investing in
+    The permissions to access this resource are inherited from the company one only.
```
elcct: Not sure what company one is.

PR #1047, opened 2018-08-01
Tidy up company admin and standardise created/modified fields

marcofucci:
Issue number: N/A

### Description of change

This:
- groups fields in fieldsets on the company admin change page
- defines a new `BaseModelAdminMixin` for all `ModelAdmin` with model extending `core.BaseModel`. This updates created_by/modified_by automatically and can show those fields as readonly descriptions in the change form (see screenshot).

### Checklist

* ~[ ] Have any relevant search models been updated?~
* ~[ ] Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?~
* ~[ ] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?~

localhost_8000_admin_company_company_cf5a637f-6056-41ae-8694-dd6790a63794_change

### DIFF
--- a/datahub/company/admin.py
+++ b/datahub/company/admin.py
@@ -4,7 +4,7 @@ from django.contrib.auth.admin import UserAdmin
 from django.db import models
 from reversion.admin import VersionAdmin
 
-from datahub.core.admin import ReadOnlyAdmin
+from datahub.core.admin import BaseModelAdminMixin, ReadOnlyAdmin
 from datahub.metadata.admin import DisableableMetadataAdmin
 from .models import (
     Advisor,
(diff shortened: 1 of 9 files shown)


datahub/company/admin.py
```diff
@@ -34,9 +34,95 @@ class CompanyCoreTeamMemberInline(admin.TabularInline):
...
+                    'uk_region',
+                    'employee_range',
+                    'turnover_range',
+                    'classification',
+                    'one_list_account_owner',
+                )
+            }
+        ),
+        (
+            'HIERARCHY',
```
reupen: If we're using caps can we change the adviser one as well?

marcofucci: I don't particularly like the uppercase bit but Django converts to uppercase in css in other places and it looks weird to have both.

Unfortunately I'm not able to use any of the existing selectors so I'd have to define a new css just for this. 

Unless you have a better idea I'm going to change the adviser fieldsets to uppercase, not ideal but ¯\_(ツ)_/¯

reupen: Yeah, that's fine. Was just commenting on the inconsistency :)

datahub/omis/payment/admin.py
```diff
@@ -193,34 +194,23 @@ class RefundAdmin(admin.ModelAdmin):
...
     list_select_related = (
         'order',
     )
 
     def save_model(self, request, obj, form, change):
-        """
-        Populate total_amount from other fields and
-        created_by/modified_by from the logged in user.
-        """
+        """Populate total_amount from other fields and."""
```
reupen: Trailing 'and' here.

datahub/core/admin.py
```diff
@@ -65,6 +69,55 @@ def get_readonly_fields(self, request, obj=None):
...
+    It updates created_by and modified_by automatically from the logged in user.
+
+    It also adds support for descriptive versions of created_on/by and modified_on/by,
+    using only two admin "fields": 'created' and 'modified'.
+    To use them just add 'created' and 'modified' to `readonly_fields` and `fields`
+    instead of created_on/by and modified_on/by.
+    """
+
+    def _get_description_for_timed_event(self, event_on, event_by):
+        text = []
```
reupen: I would use something like `text_parts` as I found it a bit confusing that `text` was a list of strings.

PR #1054, opened 2018-08-03
Don't use nested 

reupen:
### Description of change

The main use case for nested queries is when:

- you have an array of objects in a field
- you need to filter on two sub-fields of that field with an AND operator (for example, company.contacts.first_name = a AND company.contacts.surname = b)

Since we have no such cases of this, and because the nested fields used in filters use `include_in_parent=True`, we don't need to use nested queries at all.

This replaces them with normal queries, and is one of the steps on the way to replacing most or all of our nested fields with normal objects.

(There is a slight change in logic in `_build_single_field_query()`, but this should not be of any practical difference.)

This also removes a few nested array fields from the sort fields for companies and investment projects as having them as sort fields doesn't seem to make much sense.

Lastly, nested_id_uri_field is removed as it was unused.

### Checklist


datahub/search/query_builder.py
```diff
@@ -231,29 +231,22 @@ def _build_single_field_query(field, value):
...
-        query = Q('match', **{real_field: field_query})
-
-    # return nested or plain query
-    if '.' in real_field:
-        return Q('nested', path=real_field.rsplit('.', maxsplit=1)[0], query=query)
-    return query
+    if field.endswith('_exists'):
+        return _build_exists_query(field, value)
+
+    # Implicit exists query
```
elcct: Capital i

PR #1057, opened 2018-08-07
Update interactions export from search

reupen:
### Description of change

This changes how exports work from search (starting with interactions – the other exports have been removed for now).

Unfortunately, it's a bit complicated as the search is performed from the filter pages (which uses Elasticsearch) but the data needs to come from the RDBMS (for multiple reasons, including that the data needed in the exports comes from a lot of models so getting it from Elasticsearch isn't practical).

So the process is to get a list of IDs that match the query from Elasticsearch and then those IDs with the Django ORM. (Another approach would be to put the IDs in a temporary table, but it starts getting more complicated so I haven't gone down that route.)

The number of results is limited to 5000, so it's not huge and performance should be reasonable. (Note: the front end won't show the export button at all if there are more than 5000 results.)

A placeholder set of fields is included in the interactions export for now – will update them in a follow-up PR.

### Checklist


reupen:
Just noticed that the normal entity search view and the export view use get_search_by_entity_query a bit differently – will push another commit to fix that.

datahub/core/csv.py
```diff
@@ -0,0 +1,32 @@
...
+        yield writer.writerow(row)
+
+
+def create_csv_response(rows, field_titles, filename):
+    """Creates a CSV HTTP response."""
+    # TODO: Use additional FileResponse.__init__() arguments when Django 2.1 is released
+    # See https://code.djangoproject.com/ticket/16470
+    response = FileResponse(
+        csv_iterator(rows, field_titles),
+        content_type=CSV_CONTENT_TYPE
```
stuaxo: You might be able to use `as_attachment=True, filename=filename` here to avoid setting them later.

reupen: There's a TODO above about that :) It needs Django 2.1, we're on Django 2.0 at the moment.

stuaxo: Fair enough :)  I haven't used it before..

Funny streaming FileField == 0 latency, vs the raw speed of sendfile, a shame there's not a world with the possibility to have both.

marcofucci:
@reupen it looks good but it would be good to have live document on who needs this and how they will be using the data. Let's wait for Anthony before merging this.

marcofucci:
@reupen I've added a blocked label so that we don't forget about this ;)

datahub/core/csv.py
stuaxo: Now this is blocked we might get to Django 2.1 in the meantime :grimacing:

datahub/core/csv.py
```diff
@@ -0,0 +1,44 @@
...
+    """Returns an iterator producing CSV formatted data from provided input."""
+    try:
+        yield BOM_UTF8
+        writer = DictWriter(EchoUTF8(), fieldnames=field_titles.keys())
+
+        yield writer.writerow(field_titles)
+        for row in rows:
+            yield writer.writerow(row)
+    except Exception:
+        # Because CSV responses are normally streamed, a 200 response will already have been
```
elcct: will -> would?

marcofucci:
Discussed and agreed to:

1. limit the number of columns and rows in the export
2. use data analytics to monitor its usage
3. ideally add some copy saying that this subject to change without notice

PR #1060, opened 2018-08-09
Add management command for updating email marketing preferences

reupen:
### Description of change

Adds a management command for updating Contact.accepts_dit_email_marketing using a CSV file.

This command also disables search signal receivers for contacts due to the large number of tasks that would otherwise be generated. (Syncing can manually be performed afterwards, if required.)

### Checklist


datahub/search/signals.py
```diff
@@ -44,3 +49,27 @@ def disconnect(self):
...
+    signal_receivers = [
+        receiver for search_app in get_search_apps()
+        for receiver in search_app.get_signal_receivers()
+        if receiver.sender == model and receiver.is_connected
+    ]
+
+    for receiver in signal_receivers:
+        receiver.disconnect()
+
+    yield
```
marcofucci: what if there's an exception?

datahub/search/test/test_signals.py
```diff
@@ -0,0 +1,75 @@
+from unittest.mock import Mock
+
+import pytest
+
+from datahub.company.models import Contact
+from datahub.company.test.factories import CompanyFactory, ContactFactory
+from datahub.interaction.test.factories import CompanyInteractionFactory
```
marcofucci: It would be better to use `datahub.search.test.test_support` instead and update it with a related model. This is just to minimise dependencies (e.g. changing interactions will not require a rewrite of this).

PR #1062, opened 2018-08-10
Remove iprojectdocument.

elcct:
### Description of change

This removes unused `IProjectDocument` model and its related views etc. Because we use the default S3 bucket for migration commands I renamed `DOCUMENTS_BUCKET` env variable to `DEFAULT_BUCKET`. 

### Checklist


.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       AWS_DEFAULT_REGION: eu-west-2
       AWS_ACCESS_KEY_ID: foo
       AWS_SECRET_ACCESS_KEY: bar
-      DOCUMENTS_BUCKET: baz
+      DEFAULT_BUCKET: baz
```
marcofucci: I think you forgot to update `docker-compose.env` and `sample.env`

elcct: Oh :( Fixed now. Thank you!

elcct:
@marcofucci I have updated all environments. I will delete old key after the release.

PR #1063, opened 2018-08-10
Remove company.parent field from django

marcofucci:
### Description of change

The field `parent` was not used so this gets rid of it in django.

The database field is kept to avoid errors whilst deploying and a future PR is needed to get rid of it completely.

### Checklist


datahub/company/migrations/0029_remove-parent.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0028_core-team'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
reupen: We might get some errors while the apps are being switched over this way, but up to you.

marcofucci: :( true

PR #1064, opened 2018-08-10
Evidence models.

elcct:
### Description of change

This adds evidence sub app and models. `EvidenceDocument` - that keeps investment projects' evidence document and `EvidenceTag` metadata used to tag evidence documents.


datahub/investment/evidence/models.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+
+    name = models.CharField(max_length=MAX_LENGTH)
+
+    def __str__(self):
+        """Human readable representation of the object."""
+        return f'{self.investment_project.name} - {self.name}'
+
+
+class EvidenceDocument(AbstractEntityDocumentModel):
```
reupen: What about a __str__() for this (or AbstractEntityDocumentModel)?

datahub/investment/evidence/models.py
```diff
@@ -0,0 +1,42 @@
...
+from django.conf import settings
+from django.db import models
+
+from datahub.core.models import BaseModel
+from datahub.documents.models import AbstractEntityDocumentModel
+
+MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class Evidence(BaseModel):
```
reupen: I'm not sure Evidence is a good name since it's a mass noun. What exactly is an 'evidence'? Can this be more specific?

datahub/investment/evidence/models.py
```diff
@@ -0,0 +1,42 @@
...
+    """Evidence model."""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    investment_project = models.ForeignKey(
+        'investment.InvestmentProject', on_delete=models.CASCADE, related_name='evidence'
+    )
+
+    name = models.CharField(max_length=MAX_LENGTH)
```
reupen: What would the 'name' of an evidence be? Can this be more descriptive?

datahub/investment/evidence/models.py
elcct: Good idea. I have added `__str__()` on `EvidenceDocument`.

datahub/investment/evidence/models.py
elcct: I changed it to `EvidenceGroup`, it is short and fixes the problem with the word. Thanks!

datahub/investment/evidence/models.py
elcct: I think it now does make sense when it belongs to `EvidenceGroup` - as the name of the group?

reupen: Yes, I think it's OK now.

elcct:
@reupen @marcofucci Updated the model with the `EvidenceDocument` as the main object that can be tagged with `EvidenceTag` registered in metadata.

PR #1071, opened 2018-08-14
Allow sorting interactions by adviser and subject in endpoint

marcofucci:
### Description of change

This adds the ability to sort interactions by `dit_adviser` `first_name` or `last_name` and by `subject` in the collection endpoint.

The related test has been changed to make it always deterministic.

### DIFF
--- a/datahub/interaction/test/views/test_common.py
+++ b/datahub/interaction/test/views/test_common.py
@@ -1,4 +1,5 @@
 from datetime import date, datetime
+from functools import reduce
 from operator import attrgetter
 from random import sample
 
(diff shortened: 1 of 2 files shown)


datahub/interaction/test/views/test_common.py
```diff
@@ -146,36 +147,82 @@ def test_filtered_by_event(self):
...
+                'subject': 'dolor',
+            },
+        ]
+
+        interactions = []
+        for data in data_list:
+            creation_time = data.pop('created_on')
+            with freeze_time(creation_time):
+                interactions.append(
+                    EventServiceDeliveryFactory(**data)
```
reupen: This doesn't seem to be working as expected for `contact__first_name` and `contact__last_name` – probably related to the `factory.LazyAttribute` used in `InteractionFactoryBase`.

marcofucci: That's very odd but it was indeed happening. The new version is slightly less nice to read but works :/

PR #1072, opened 2018-08-14
Update to Django 2.1

reupen:
### Description of change

The majority of changes here are to rename our read permissions so that they start with view_ instead of read_.

The `/whoami/` endpoint has also been amended to return the view permissions both as `app.view_*` and `app.read_*` codenames. This is for backwards compatibility – we will need to update the front end to check for `app.view_*` permissions.

(We can also separate ReadOnlyAdmin into a pure-read-only version and the current partially read-only version, but that can be done separately.)

At some point we can also migrate from `NullBooleanField` to `BooleanField(null=True)`, but that will need to wait for DRF support.

### Checklist


datahub/investment/models.py
```diff
@@ -501,7 +498,7 @@ class InvestmentProjectTeamMember(models.Model):
 
     No default permissions are defined on this model as permissions from the InvestmentProject
     model are used and enforced instead. This is to avoid unnecessary complexity in the
-    permissions model, where permissions on both models would need to be checked. (A custom read
+    permissions model, where permissions on both models would need to be checked. (A read
```
marcofucci: I guess you can get rid of the whole sentence _A read permission is also not defined..._

datahub/company/test/test_company_views.py
```diff
@@ -127,12 +127,12 @@ def test_sort_by_created_on(self):
         assert expected_timestamps == actual_timestamps
 
     def test_list_companies_without_read_document_permission(self):
```
marcofucci: `test_without_view_company_document ` would be better

datahub/company/test/test_company_views.py
```diff
@@ -147,13 +147,13 @@ def test_list_companies_without_read_document_permission(self):
         )
 
     def test_list_companies_with_read_document_permission(self):
```
marcofucci: `test_with_view_company_document ` would be better

datahub/company/test/test_company_views.py
```diff
@@ -172,13 +172,13 @@ class TestGetCompany(APITestMixin):
     """Tests for getting a company."""
 
     def test_get_company_without_read_document_permission(self):
```
marcofucci: `test_without_view_company_document ` would be better

datahub/company/test/test_contact_views.py
```diff
@@ -576,13 +576,13 @@ def test_view(self):
         }
 
     def test_get_contact_without_read_document_permission(self):
```
marcofucci: `test_without_view_company_document ` would be better

datahub/company/test/test_contact_views.py
```diff
@@ -616,12 +616,12 @@ def test_all(self):
         assert response.data['count'] == 5
 
     def test_all_without_read_document_permission(self):
```
marcofucci: `test_without_view_contact_document ` would be better

datahub/company/test/test_contact_views.py
```diff
@@ -637,13 +637,13 @@ def test_all_without_read_document_permission(self):
         )
 
     def test_all_with_read_document_permission(self):
```
marcofucci: `test_with_view_contact_document` would be better

PR #1073, opened 2018-08-15
Fix report key creation.

elcct:
### Description of change

`settings.REPORT_BUCKET` is no longer available and `get_bucket_name` function should be used to get bucket name. This fixes the problem with faulty `_get_report_key` function, preventing report bucket key from being created.


elcct:
I have removed bucket name from the report key as it is not necessary.

PR #1076, opened 2018-08-17
Add investment project export

reupen:
### Description of change

Adds exporting of investment project search results. This includes a few utility functions that can be used to help build queries for reports.

I think there's some scope to share a few testing utilities between tests for the different data exports. However, I'll leave that for the next one as there's quite a bit here already.

### Checklist


datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+from datahub.core.test.support.factories import BookFactory, PersonFactory
+from datahub.core.test.support.models import Book, Person
+
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.mark.parametrize('num_authors', range(3))
+def test_get_string_agg_subquery(num_authors):
+    """Test get_string_agg_subquery()."""
```
marcofucci: It would be good to have more descriptive and detailed docs so that the reader doesn't have to spend much time trying to figure out that the code is doing and what the final check expects.

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+        author_names=get_string_agg_subquery(Book, 'authors__first_name')
+    )
+    author_names_str = queryset.first().author_names
+    actual_author_names = author_names_str.split(', ') if author_names_str else []
+    expected_author_names = [author.first_name for author in authors]
+    assert actual_author_names == expected_author_names
+
+
+def test_get_aggregate_subquery():
+    """Test get_aggregate_subquery()."""
```
marcofucci: Same as above re. docs

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    proofreader = PersonFactory()
+    books = BookFactory.create_batch(5, proofreader=proofreader)
```
marcofucci: It would be good to test when books = [] as well

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    queryset = Person.objects.annotate(
+        max_published=get_aggregate_subquery(Person, Max('proofread_books__published_on'))
```
marcofucci: if the model of the subquery is different from the model of the main query, the SQL is still working fine but of course the value is empty because the types (and the pk values) are different. 
It's not obvious so maybe you could add some text if the stringdocs of the function? You are doing a pk=pk inside the subquery after all.

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+    annotated_book = queryset.first()
+    assert annotated_book.genre_name == book.get_genre_display()
+
+
+class TestGetFullNameExpression:
+    """Tests for get_full_name_expression()."""
+
+    def test_full_name_annotation(self):
+        """Tests the generation of full names."""
```
marcofucci: Same comment re. docs.

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    Test that get_aggregate_subquery() raises an error when passed a non-aggregate
+    expression.
+    """
+    with pytest.raises(ValueError):
+        get_aggregate_subquery(Person, Left('proofread_books__name', 5))
+
+
+@pytest.mark.parametrize('genre', ('horror', 'non_fiction', 'invalid-option'))
+def test_get_choices_as_case_expression(genre):
+    """Test get_choices_as_case_expression()."""
```
marcofucci: Same comment re. docs.

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,85 @@
...
+def test_get_aggregate_subquery_raises_error_on_on_aggregate_expression():
+    """
+    Test that get_aggregate_subquery() raises an error when passed a non-aggregate
+    expression.
+    """
+    with pytest.raises(ValueError):
+        get_aggregate_subquery(Person, Left('proofread_books__name', 5))
+
+
+@pytest.mark.parametrize('genre', ('horror', 'non_fiction', 'invalid-option'))
```
marcofucci: Add a case for None?

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    assert actual_author_names == expected_author_names
+
+
+class TestGetAggregateSubquery:
+    """Tests for get_aggregate_subquery()."""
+
+    @pytest.mark.parametrize('num_books', (0, 5))
+    def test_with_max_aggregate_expression(self, num_books):
+        """
+        Test that Max() can be used to calucalte the maximum published-on date for the books a
```
marcofucci: 'calucalte ' typo

datahub/core/test/test_query_utils.py
```diff
@@ -0,0 +1,96 @@
...
+from datahub.core.test.support.factories import BookFactory, PersonFactory
+from datahub.core.test.support.models import Book, Person
+
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.mark.parametrize('num_authors', range(3))
+def test_get_string_agg_subquery(num_authors):
+    """Test that get_string_agg_subquery() builds full names for Book.authors."""
```
marcofucci: Not sure it builds _full names_, it joins all the first_name(s)

PR #1079, opened 2018-08-20
Fix streaming to AV service and use of V2 API.

elcct:
### Description of change

This fixes streaming of the file to AV service. `StreamingIterator` was incorrectly used as it is apparently doesn't update its `len` property. A custom wrapper is being used instead with an added test to check if the file is being encoded correctly.
This update also contains a change to use V2 version of the AV API.

### Checklist


datahub/documents/av_scan.py
```diff
@@ -4,14 +4,28 @@
...
 from requests_toolbelt.multipart.encoder import MultipartEncoder
-from requests_toolbelt.streaming_iterator import StreamingIterator
 
 from .exceptions import VirusScanException
 from .utils import get_document_by_pk
 
 logger = getLogger(__name__)
 
 
+class StreamWrapper:
```
reupen: We had a similar wrapper before:

https://github.com/uktrade/data-hub-leeloo/blob/v5.1.0/datahub/documents/av_scan.py#L18

Things I prefer about it:

- it uses self._remaining_bytes so it's a bit clearer what's going on
- it implements __len__ instead of setting self.len, and has a docstring to explain why the behaviour is a bit weird

datahub/documents/av_scan.py
```diff
@@ -4,14 +4,28 @@
...
+    """Wrap stream to be compatible with MultipartEncoder."""
+
+    def __init__(self, data, length):
+        """Initialise StreamWrapper."""
+        self.len = int(length) if length else None
+        self._data = data
+
+    def read(self, chunk_size):
+        """Read data in chunks."""
+        chunk = self._data.read(chunk_size) or b''
```
reupen: Is the `or b''` needed?

datahub/documents/av_scan.py
```diff
@@ -4,14 +4,28 @@
...
 
 logger = getLogger(__name__)
 
 
+class StreamWrapper:
+    """Wrap stream to be compatible with MultipartEncoder."""
+
+    def __init__(self, data, length):
+        """Initialise StreamWrapper."""
+        self.len = int(length) if length else None
```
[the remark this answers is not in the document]
elcct: It shouldn't be in this case. Removed. Thanks.

datahub/documents/av_scan.py
elcct: It is not. Removed. Thanks.

datahub/documents/av_scan.py
elcct: Good points. I replaced it with our old slightly modified code. It is indeed better.

PR #1080, opened 2018-08-21
Update interactions search export columns

reupen:
### Description of change

Updates the columns output in the interactions search results export. Also tidies up a few other things around the exports.

### Checklist

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -320,6 +320,11 @@ if REDIS_BASE_URL:
 
 # FRONTEND
 DATAHUB_FRONTEND_BASE_URL = env('DATAHUB_FRONTEND_BASE_URL', default='http://localhost:3000')
+DATAHUB_FRONTEND_URL_PREFIXES = {
+    'company': f'{DATAHUB_FRONTEND_BASE_URL}/companies',
+    'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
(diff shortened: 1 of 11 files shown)


datahub/core/query_utils.py
```diff
@@ -83,3 +84,13 @@ def get_full_name_expression(field_name):
         ),
         default=None,
     )
+
+
+def get_front_end_url_expression(model_name, pk_expression):
```
elcct: We use `FRONTEND`, `front-end` and `front_end` - seems a bit inconsistent.

reupen: Yeah, I know. It's two words so I normally separate it, but named DATAHUB_FRONTEND_URL_PREFIXES to match DATAHUB_FRONTEND_BASE_URL.

Front end vs front-end is attributive use (as an adjective) vs on its own (front-end developer vs the front end).

config/settings/common.py
```diff
@@ -321,6 +321,10 @@
 
 # FRONTEND
 DATAHUB_FRONTEND_BASE_URL = env('DATAHUB_FRONTEND_BASE_URL', default='http://localhost:3000')
+DATAHUB_FRONTEND_URL_PREFIXES = {
+    'company': f'{DATAHUB_FRONTEND_BASE_URL}/companies',
+    'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
```
marcofucci: Can you add the omis one as well?

PR #1085, opened 2018-08-21
Add evidence document upload endpoints

elcct:
### Description of change

This adds evidence document upload endpoints and an admin for evidence tags.

### Checklist

* [X] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?


datahub/investment/evidence/models.py
```diff
@@ -1,15 +1,61 @@
...
+    """
+
+    view_all = 'view_all_evidencedocument'
+    view_associated_investmentproject = 'view_associated_evidencedocument'
+    change_all = 'change_all_evidencedocument'
+    change_associated_investmentproject = 'change_associated_evidencedocument'
+    add_all = 'add_all_evidencedocument'
+    add_associated_investmentproject = 'add_associated_evidencedocument'
+    delete = 'delete_evidencedocument'
+    delete_associated_investmentproject = \
```
reupen: Won't this fit on one line?

datahub/investment/evidence/models.py
```diff
@@ -29,7 +75,42 @@ class EvidenceDocument(AbstractEntityDocumentModel):
...
+            ),
+            (
+                EvidenceDocumentPermission.delete_associated_investmentproject.value,
+                'Can delete evidence document for associated investment projects'
+            ),
+        )
+        default_permissions = (
+            'add_all',
+            'change_all',
+            'delete',
```
reupen: Should have `'view_all'` at the end as well.

datahub/investment/evidence/serializers.py
```diff
@@ -0,0 +1,57 @@
...
+class EvidenceDocumentSerializer(serializers.ModelSerializer):
+    """Serializer for Investment Project Evidence Documents."""
+
+    comment = serializers.CharField(required=False)
+    tags = NestedRelatedField(EvidenceTag, many=True)
+    av_clean = serializers.BooleanField(source='document.av_clean', read_only=True)
+    created_by = NestedAdviserField(read_only=True)
+    status = serializers.CharField(source='document.status', read_only=True)
+    uploaded_on = serializers.DateTimeField(source='document.uploaded_on', read_only=True)
+    created_by = NestedAdviserField(read_only=True)
```
reupen: `created_by` is defined twice here.

changelog/investment/evidence-document-upload.feature
```diff
@@ -0,0 +1 @@
+It is now possible to upload the evidence for given investment project.
```
reupen: Small thing – would add an 'a' before given (or just replace 'given' with 'an').

elcct: Updated. Thanks!

datahub/investment/evidence/serializers.py
```diff
@@ -0,0 +1,57 @@
...
+from datahub.investment.evidence.permissions import (
+    EvidenceDocumentHasAssociatedInvestmentProjectValidator
+)
+from datahub.investment.serializers import NestedInvestmentProjectField, NestedRelatedField
+
+
+class EvidenceDocumentSerializer(serializers.ModelSerializer):
+    """Serializer for Investment Project Evidence Documents."""
+
+    comment = serializers.CharField(required=False)
```
reupen: Are we able to use extra_kwargs and/or read_only_fields for fields no using NestedRelatedField? As it preserves other attributes like max_length (when they exist).

datahub/investment/evidence/serializers.py
elcct: Oh! Thanks! Removed.

datahub/investment/evidence/models.py
elcct: Good point. Added.

datahub/investment/evidence/models.py
elcct: I will fit. Changed.

datahub/investment/evidence/test/utils.py
```diff
@@ -0,0 +1,18 @@
+from datahub.investment.test.factories import InvestmentProjectFactory
+from .factories import EvidenceTagFactory
+from ..models import (
```
reupen: This should fit on one line.

datahub/investment/evidence/serializers.py
elcct: I moved `comment` to `extra_kwargs`. Thanks!

datahub/investment/evidence/test/utils.py
elcct: Good point. Updated.

datahub/investment/evidence/test/test_views.py
```diff
@@ -0,0 +1,446 @@
...
+import pytest
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.core.test_utils import APITestMixin, create_test_user, format_date_or_datetime
+from datahub.documents.models import Document, UPLOAD_STATUSES
+from datahub.investment.test.factories import InvestmentProjectFactory
+from .factories import EvidenceTagFactory
+from .utils import create_evidence_document
+from ..models import (
```
reupen: Will this fit on one line?

datahub/investment/evidence/test/test_views.py
```diff
@@ -0,0 +1,446 @@
...
+        url = reverse('api-v3:investment:evidence-document:document-collection', kwargs={
+            'project_pk': investment_project.pk,
+        })
+
+        user = create_test_user(permission_codenames=permissions)
+        api_client = self.create_api_client(user=user)
+
+        response = api_client.post(url, format='json', data={
+            'original_filename': 'test.txt',
+            'tags': [et.pk for et in evidence_tags],
```
reupen: Minor point but I think tag would make a better shortened name for the variable than 'et'.

datahub/investment/evidence/test/test_views.py
```diff
@@ -0,0 +1,446 @@
...
+        })
+
+        assert response.status_code == status.HTTP_201_CREATED
+        response_data = response.data
```
reupen: Normally we use response.json(), any reason this is different?

datahub/investment/evidence/test/test_views.py
```diff
@@ -0,0 +1,446 @@
...
+            (True, status.HTTP_200_OK),
+            (False, status.HTTP_403_FORBIDDEN,)
+        )
+    )
+    @patch('datahub.documents.models.sign_s3_url')
+    def test_document_download(self, sign_s3_url, av_clean, expected_status):
+        """Tests download of individual document."""
+        sign_s3_url.return_value = 'http://what'
+
+        user = create_test_user(permission_codenames=NON_RESTRICTED_VIEW_PERMISSIONS[0])
```
reupen: I think passing `(EvidenceDocumentPermission.view_all,)` would be clearer than passing `NON_RESTRICTED_VIEW_PERMISSIONS[0]`.

datahub/investment/evidence/test/test_views.py
elcct: It will. Fixed.

datahub/investment/evidence/test/test_views.py
elcct: Yes, you're right, `tag` would be less taxing on the brain :) Changed.

datahub/investment/evidence/test/test_views.py
elcct: There is no reason. I have changed it to `.json()`. Thanks!

datahub/investment/evidence/serializers.py
```diff
@@ -0,0 +1,58 @@
...
+from datahub.investment.evidence.permissions import (
+    EvidenceDocumentHasAssociatedInvestmentProjectValidator
+)
+from datahub.investment.serializers import NestedInvestmentProjectField, NestedRelatedField
+
+
+class EvidenceDocumentSerializer(serializers.ModelSerializer):
+    """Serializer for Investment Project Evidence Documents."""
+
+    tags = NestedRelatedField(EvidenceTag, many=True)
```
reupen: I think allow_empty defaults to True, is no tags allowed?

datahub/investment/evidence/admin.py
```diff
@@ -0,0 +1,9 @@
+from django.contrib import admin
+
+from datahub.metadata.admin import DisableableMetadataAdmin
+from .models import EvidenceTag
+
+
+@admin.register(EvidenceTag)
+class EvidenceTagAdmin(DisableableMetadataAdmin):
```
reupen: Do we have some initial EvidenceTag data to load?

datahub/investment/permissions.py
```diff
@@ -90,6 +92,8 @@ def should_apply_restrictions(self, request, view_action):
             'model_name': self.model._meta.model_name,
             'action': action
         }
+        if request.user.has_perm(self.standard_permission_template.format(**format_kwargs)):
```
reupen: Not sure about this, where we've have an *_associated_* permission before the normal one has had _all_ in the middle. Shouldn't we rename `delete_evidencedocument` to `delete_all_evidencedocument` instead?

elcct: I changed it so that it uses `delete_all_evidencedocument`. Thanks!

datahub/investment/evidence/admin.py
elcct: I have added initial EvidenceTag data.

datahub/investment/permissions.py
```diff
@@ -47,7 +48,8 @@ class InvestmentProjectModelPermissions(ViewBasedModelPermissions):
             _PermissionTemplate.stage_to_won,
         ),
         'delete': (
-            _PermissionTemplate.standard,
+            _PermissionTemplate.all,
```
reupen: Is this change needed? Since EvidenceDocumentModelPermissions is also defined.

elcct: I mean to revert it to its original form :( Fixed. Thanks!

datahub/investment/permissions.py
```diff
@@ -163,6 +164,55 @@ def _get_filter_field_prefix(self):
         return f'{self.model_attribute}__' if self.model_attribute else ''
 
 
+class HasAssociatedInvestmentProjectValidator:
+    """
+    Validator which enforces association permissions when adding or updating associated object.
+    """
+
+    required_message = 'This field is required.'
```
reupen: Is required_message message used?

datahub/investment/permissions.py
```diff
@@ -163,6 +164,55 @@ def _get_filter_field_prefix(self):
         return f'{self.model_attribute}__' if self.model_attribute else ''
 
 
+class HasAssociatedInvestmentProjectValidator:
+    """
+    Validator which enforces association permissions when adding or updating associated object.
```
reupen: It's only for adding, isn't it (not updating)?

I think 'when adding a related object' would be a bit clearer, but in any case it would be good to add an example to make it clearer what it's for.

Also would be good to mention that it needs to be subclassed with `checker` and `non_associated_investment_project_message` overridden.

datahub/investment/evidence/test/test_views.py
```diff
@@ -0,0 +1,594 @@
...
+                'project_pk': investment_project.pk,
+            }
+        )
+        api_client = self.create_api_client(user=user)
+        response = api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+
+        if not allowed:
```
reupen: Shouldn't it return a 403? Why doesn't it :thinking:

elcct: I think that's a behaviour when a parent entity wasn't an investment project, so you would have an access to the list, but restricted objects would be filtered out. But since the parent is an investment project it would make sense to disallow the entire endpoint.

reupen: Yeah, I guess it comes from interactions. I assume propositions is the same so I won't say to change it, but it would be cleaner if it just blocked creations and listing with a 403.

reupen:
Main other thing I would add is to add the new permissions to the groups in `metadata/fixtures/teams.yaml` (won't affect production, but at least new environments and the acceptance tests will get them).

changelog/investment/evidence-endpoint.api
```diff
@@ -0,0 +1,16 @@
...
+``POST /v3/investment/<investment project pk>/evidence`` creates new evidence document upload.
+``GET /v3/investment/<investment project pk>/evidence/<evidence document pk>`` gets details of evidence document
+``DELETE /v3/investment/<investment project pk>/evidence/<evidence document pk>`` deletes given evidence document.
+``POST /v3/investment/<investment project pk>/evidence/<evidence document pk>/upload_callback`` notifies that file upload has been completed and initiates virus scanning.
+``GET /v3/investment/<investment project pk>/evidence/<evidence document pk>/download`` returns a signed URL to the document file object.
+Following permissions are required to use the endpoints:
+``evidence.add_all_evidencedocument``
+``evidence.view_all_evidencedocument``
+``evidence.change_all_evidencedocument``
+``evidence.delete_evidencedocument``
```
reupen: This should be updated to `evidence.delete_all_evidencedocument`.

changelog/investment/evidence-document-upload.feature
```diff
@@ -0,0 +1 @@
+It is now possible to upload the evidence for a given investment project.
```
reupen: I would just elaborate a little bit to make it clearer that it's about documents. Could just change 'the evidence' to 'evidence documents'.

changelog/investment/evidence-endpoint.api
elcct: Good catch! Fixed. Thanks

changelog/investment/evidence-document-upload.feature
elcct: Changed. Thanks

datahub/investment/evidence/test/test_views.py
elcct: It is fixed now. Thanks!

datahub/investment/permissions.py
elcct: Updated the comment.

datahub/investment/permissions.py
elcct: It is not. Removed.

elcct:
Added appropriate permissions to the `teams.yaml`.

datahub/investment/evidence/test/test_views.py
```diff
@@ -0,0 +1,600 @@
...
+        })
+
+        api_client = self.create_api_client(user=user)
+        response = api_client.post(url, format='json', data={
+            'original_filename': 'test.txt',
+            'tags': [tag.pk for tag in evidence_tags],
+        })
+        response_data = response.json()
+
+        if not allowed:
```
reupen: Since we have changed listing to return a 403, shouldn't we do this one as well?

elcct: Oh no. Yes! Not sure why I thought it is okay to do just for the list. Thanks

datahub/investment/evidence/serializers.py
```diff
@@ -0,0 +1,58 @@
...
+            'investment_project',
+            'original_filename',
+            'url',
+            'status',
+        )
+        read_only_fields = (
+            'investment_project', 'av_clean', 'url', 'created_on', 'status',
+        )
+        validators = (
+            EvidenceDocumentHasAssociatedInvestmentProjectValidator(),
```
reupen: Is this still needed?

elcct: It's not, actually a couple of more stuff wasn't needed as it turns out, so I removed it. Thanks! Looks much simpler now.

PR #1088, opened 2018-08-23
Add the ability to export company search results

reupen:
### Description of change

Adds the ability to export company search results in the same way as the other exports.

Also removes some unused test code from the original export code.

### Checklist


datahub/company/models/company.py
```diff
@@ -154,6 +155,7 @@ class Meta:
         permissions = (
             (CompanyPermission.view_company_document.value, 'Can view company document'),
             (CompanyPermission.view_company_timeline.value, 'Can view company timeline'),
+            (CompanyPermission.export_company.value, 'Can export company'),
```
elcct: Maybe can `Can export company search results` would be better?

reupen: Just keeping it in line with the other export permissions (and the default Django permissions). I'm not sure 'search results' is helpful either, as that sounds more like global search than the collection pages?

PR #1097, opened 2018-08-29
Updated proposition completed endpoint.

elcct:
### Description of change

This makes `details` field no longer required to complete a proposition. `details` are still required to abandon the proposition. Proposition also needs to have at least one document uploaded to be completed.

### Checklist


datahub/investment/proposition/models.py
```diff
@@ -142,7 +143,13 @@ def complete(self, by, details):
 
         :param by: the adviser who marked the proposition as complete
         :param details: details of completion
+        :raises ValidationError: when trying to complete proposition without uploaded documents
+        :raises APIConflictException: when proposition status is not ongoing
         """
+        if self.documents.filter(document__status=UPLOAD_STATUSES.virus_scanned).count() == 0:
+            raise ValidationError({
```
reupen: It seems a bit odd to raise a DRF exception from the model, but I see we're already doing that so OK.

PR #1098, opened 2018-08-29
Improved testing of tags in evidence documents endpoints.

elcct:
### Description of change

This ensures the order of tags in the endpoints' response so that their validity can be asserted reliably.

### Checklist


changelog/investment/evidence-tests.bugfix
```diff
@@ -0,0 +1 @@
+Tests of ``tags`` in the evidence document endpoints responses are now improved.
```
reupen: I think I'd be inclined to put this in the internal changes section since it's not that interesting to anyone using the API.

elcct: That's a good idea. Moved.

PR #1100, opened 2018-08-30
Puts new proposition complete behaviour behind a feature flag.

elcct:
### Description of change

This puts the new behaviour of proposition complete functionality behind the feature flag. If `proposition-documents` feature flag is active, new functionality becomes enabled.
This adds a suite of tests for old behaviour in `TestLegacyCompleteProposition` class.
It also adds a utility function in the feature_flag app that gets the "state" of the feature flag.

### Checklist

### DIFF
--- a/datahub/investment/proposition/test/test_views.py
+++ b/datahub/investment/proposition/test/test_views.py
@@ -74,7 +74,7 @@ NON_RESTRICTED_DELETE_PERMISSIONS = (
 
 
 @pytest.fixture
-def get_proposition_document_feature_flag():
+def proposition_document_feature_flag():
     """Gets proposition document feature flag."""
     yield FeatureFlagFactory(code=FEATURE_FLAG_PROPOSITION_DOCUMENT)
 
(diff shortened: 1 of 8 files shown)


datahub/investment/proposition/test/test_views.py
```diff
@@ -647,7 +656,11 @@ class TestCompleteProposition(APITestMixin):
     """Tests for the complete proposition view."""
 
     @pytest.mark.parametrize('permissions', NON_RESTRICTED_CHANGE_PERMISSIONS)
-    def test_non_restricted_user_can_complete_proposition(self, permissions):
+    def test_non_restricted_user_can_complete_proposition(
+        self,
+        permissions,
+        get_proposition_document_feature_flag
```
reupen: It would probably be easier to apply the fixture to the class using `@pytest.mark.usefixtures('get_proposition_document_feature_flag')` (instead of each of these tests).

Also, I'm not sure prefixing the feature name with `get_` is a good idea since it's creating something. I think `proposition_document_feature_flag` is fine.

elcct: Very good points. This way is much better. Thanks!

PR #1103, opened 2018-08-30
Remove noise from news

elcct:
### Description of change

This removes unnecessary news.


marcofucci:
It would also be nice to add an empty line between each item in `evidence-endpoint.api` to have a better resulting rst. You can run `towncrier --version dev --draft` to check locally.

elcct:
Oh your comment appeared after I clicked merge :(

PR #1104, opened 2018-08-31
Rename proposition document permissions

elcct:
### Description of change

This fixes typo in delete associated proposition document and renames delete proposition document permission to be consistent with other entities.

### Checklist


datahub/investment/proposition/models.py
```diff
@@ -99,9 +99,9 @@ class PropositionDocumentPermission(StrEnum):
     change_associated_investmentproject = 'change_associated_propositiondocument'
     add_all = 'add_all_propositiondocument'
     add_associated_investmentproject = 'add_associated_propositiondocument'
-    delete = 'delete_propositiondocument'
+    delete_all = 'delete_all_propositiondocument'
     delete_associated_investmentproject = \
```
reupen: Will this fit on one line? Also, shouldn't `_investmentproject` be removed from the name (it's not in the value)?

datahub/investment/proposition/models.py
```diff
@@ -185,6 +181,7 @@ class Meta:
             ),
         )
         default_permissions = (
+            'view_all',
```
reupen: The view permission is the fourth element in default_permissions.

datahub/investment/proposition/models.py
elcct: It will. I keep forgetting that. `_investmentproject` indeed is pointless there, so I am going to remove it from all permissions there whilst at it.

reupen:
So does that mean we have permissions with codenames `view_propositiondocument` and `view_proposition` to delete?

elcct:
I think we don't have `view_proposition` and `view_propositiondocument`.

datahub/investment/proposition/models.py
elcct: Updated, thanks!

reupen:
> I think we don't have view_proposition and view_propositiondocument.

Oh, yes, you're right.

datahub/investment/proposition/migrations/0006_rename_permissions.py
```diff
@@ -0,0 +1,52 @@
...
+
+def rename_permissions(apps, schema_editor):
+    """
+    This fixes the typo in delete_associated_propositiondocument and renames delete_propositiondocument to
+    delete_all_propositiondocument so it is consistent with corresponding permission of evidencedocument.
+    It also removed _investmentproject_ from proposition permissions to make it consistent.
+    """
+    permission_model = apps.get_model('auth', 'Permission')
+
+    permission_model.objects.filter(
```
reupen: Not too important, but I think I'd have created a list of tuples with the old and new names and then iterated on that to do the actual renaming.

datahub/investment/proposition/migrations/0006_rename_permissions.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    permission_model.objects.filter(
+        codename='delete_propositiondocument'
+    ).update(
+        codename='delete_all_propositiondocument'
+    )
+
+    permission_model.objects.filter(
+        codename='view_associated_investmentproject_proposition'
```
reupen: Just bear in mind the front-end dependency some of the the proposition permission renamings (e.g. https://github.com/uktrade/data-hub-frontend/blob/e233002f5c8dfcc5102e97b2407089c64f795f50/src/apps/investment-projects/constants.js#L34) – need to think about how to manage that.

elcct: That's a good point. I think front-end could add `view_associated_proposition` to the list and then remove the old one once this update is on all environments?

elcct: It also appears that front-end has a typo, so that would be a good opportunity to address that.

PR #1106, opened 2018-08-31
Add the ability to export OMIS order search results

reupen:
### Description of change

Adds the ability to export OMIS orders search results in the same way as other exports.

### Checklist


datahub/omis/order/test/test_admin.py
```diff
@@ -38,7 +38,11 @@ def test_403_if_no_permissions(self):
         user = create_test_user(
             is_staff=True,
             password=self.PASSWORD,
-            permission_codenames=('add_order', 'delete_order', 'view_order')
+            permission_codenames=(
+                OrderPermission.add,
+                OrderPermission.delete,
+                OrderPermission.view,
+            )
```
marcofucci: This is okay but we are using hardcoded permissions in tests everywhere else. It would be good if we agree on one or the other option and we apply it everywhere.

datahub/search/omis/views.py
```diff
@@ -57,3 +69,60 @@ class SearchOrderParams:
...
+            Refund.objects.annotate(
+                total_refund=Cast(
+                    # We use Func rather than Sum to stop Django grouping on Refund.pk
+                    Func('net_amount', function='sum'),
+                    DecimalField(max_digits=19, decimal_places=2),
+                ) / 100
+            ).filter(
+                order=OuterRef('pk'),
+                status=RefundStatus.approved
+            ).order_by(
```
marcofucci: Is this `order_by` needed?

reupen: Yes, there's a default order by that needs to be cleared (it's not needed, but more importantly it breaks the query).

reupen: Hmm, I might be able to use Sum() with the approach in that link..

changelog/omis/omis-export.api
```diff
@@ -0,0 +1,3 @@
+``POST /v3/search/order/export`` was added for exporting OMIS order search
+results as a CSV file with up to 5000 rows. The ``order.export_order``
+permission was also added and is required to use this endpoint.
```
marcofucci: One line to avoid problems with github release notes?

PR #1109, opened 2018-09-06
Add support for CORS

ztolley:
This change uses the django-cors-headers module to add support for CORS.
This enables the use of Axios in the frontend express app as well as direct calls to the API from the browser.


reupen:
Closing for now as the API is intentionally not accessible via XHR requests.

PR #1110, opened 2018-09-11
Add user events for document deletion

elcct:
### Description of change

This adds two user event logs for evidence and proposition document deletion.

### DIFF
--- a/datahub/investment/evidence/test/test_views.py
+++ b/datahub/investment/evidence/test/test_views.py
@@ -1,7 +1,10 @@
+import datetime
 from operator import attrgetter, itemgetter
 from unittest.mock import patch
 
 import pytest
+from django.utils.timezone import utc
+from freezegun import freeze_time
 from rest_framework import status
 from rest_framework.reverse import reverse
 
@@ -9,6 +12,7 @@ from datahub.company.test.factories import TeamFactory
 from datahub.core.test_utils import APITestMixin, create_test_user, format_date_or_datetime
 from datahub.documents.models import Document, UPLOAD_STATUSES
 from datahub.investment.test.factories import InvestmentProjectFactory
+from datahub.user_event_log.models import USER_EVENT_TYPES, UserEvent
 from .factories import EvidenceTagFactory
 from .utils import create_evidence_document
 from ..models import EvidenceDocument, EvidenceDocumentPermission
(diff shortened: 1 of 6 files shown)


datahub/investment/evidence/test/test_views.py
```diff
@@ -566,6 +601,17 @@ def test_document_delete(self, delete_document, permissions, associated, allowed
...
+        assert UserEvent.objects.count() == 1
+
+        user_event = UserEvent.objects.first()
+
+        assert user_event.adviser == user
+        assert user_event.type == USER_EVENT_TYPES.evidence_document_delete
+        assert user_event.timestamp == frozen_time
+        assert user_event.api_url_path == url
+        user_event.data['tags'] = sorted(user_event.data['tags'], key=itemgetter('name'))
+        assert user_event.data == expected_user_event_data
```
marcofucci: It would be better to create its own test for this instead of extending the original one with an extra check.

datahub/investment/proposition/test/test_views.py
```diff
@@ -1716,13 +1720,37 @@ def test_document_delete(self, delete_document, permissions):
...
         delete_document.assert_called_once_with(args=(document_pk, ))
 
+        assert UserEvent.objects.count() == 1
+
+        user_event = UserEvent.objects.first()
+        assert user_event.adviser == user
+        assert user_event.type == USER_EVENT_TYPES.proposition_document_delete
+        assert user_event.timestamp == frozen_time
+        assert user_event.api_url_path == url
+        assert user_event.data == expected_user_event_data
```
marcofucci: It would be better to create its own test for this instead of extending the original one with an extra check.

marcofucci: It would be good to have an extra one that tests that if an exception happens during a delete, the user log is not recorded.

datahub/investment/evidence/test/test_views.py
marcofucci: It would be good to have an extra one that tests that if an exception happens during a delete, the user log is not recorded.

elcct: Good point. Moved to its own test.

datahub/investment/proposition/test/test_views.py
elcct: Good point. Moved to its own test.

elcct: I have added an additional test to check that exception doesn't cause user event to be created.

datahub/investment/evidence/test/test_views.py
elcct: I have added an additional test to check that exception doesn't cause user event to be created.

datahub/investment/proposition/test/test_views.py
```diff
@@ -1745,6 +1749,93 @@ def test_document_delete_without_permission(self, delete_document):
...
+        url = reverse(
+            'api-v3:investment:proposition:document-item',
+            kwargs={
+                'proposition_pk': proposition.pk,
+                'project_pk': proposition.investment_project.pk,
+                'entity_document_pk': entity_document.pk
+            }
+        )
+        user = create_test_user(
+            permission_codenames=(PropositionDocumentPermission.delete_all,),
```
marcofucci: Should we use a pytest param as the test above for permissions?

datahub/investment/evidence/test/test_views.py
```diff
@@ -586,6 +590,117 @@ def test_document_delete_without_permission(self, delete_document):
...
+
+    @patch.object(Document, 'mark_deletion_pending')
+    def test_document_delete_failure_wont_create_user_event_log(
+        self,
+        mark_deletion_pending,
+    ):
+        """Tests document deletion failure won't create user event log."""
+        mark_deletion_pending.side_effect = Exception('No way!')
+        user = create_test_user(
+            permission_codenames=(EvidenceDocumentPermission.delete_all,),
```
marcofucci: Should we use a pytest param as the test above for permissions?

datahub/investment/proposition/test/test_views.py
elcct: I thought that would be enough as other test checks if the delete endpoint can be reached with valid permissions, so this one only tests behaviour when there is a failure.

marcofucci: Fair enough 👍

PR #1113, opened 2018-09-14
Fix proposition document list endpoint

elcct:
### Description of change

~This fixes wrong look-up field for associated object filter of proposition document list view so that restricted users can now access the list endpoint.
Ideally, the endpoint should return 403 when a user is not associated with the investment project. That means the filter wouldn't be needed at all. I could tackle that in another PR.~

This checks if a restricted user is associated with the investment project when accessing proposition document endpoints. It returns 403 if a restricted user isn't associated.

This also fixes delete endpoint, where a restricted user could delete non-associated proposition document.

### Checklist

### DIFF
--- a/datahub/investment/proposition/permissions.py
+++ b/datahub/investment/proposition/permissions.py
@@ -1,10 +1,8 @@
-from rest_framework.exceptions import ValidationError
-
 from datahub.core.permissions import IsAssociatedToObjectPermission, ViewBasedModelPermissions
 from datahub.core.utils import StrEnum
-from datahub.investment.models import InvestmentProject
 from datahub.investment.permissions import (
-    InvestmentProjectAssociationCheckerBase, IsAssociatedToInvestmentProjectFilter
+    InvestmentProjectAssociationCheckerBase,
+    IsAssociatedToInvestmentProjectPermissionMixin,
 )
 from datahub.investment.proposition.models import Proposition, PropositionDocument
 
@@ -64,7 +62,10 @@ class InvestmentProjectPropositionAssociationChecker(
     associated_permission_template = _PermissionTemplate.associated
 
 
-class IsAssociatedToInvestmentProjectPropositionPermission(IsAssociatedToObjectPermission):
+class IsAssociatedToInvestmentProjectPropositionPermission(
+    IsAssociatedToInvestmentProjectPermissionMixin,
+    IsAssociatedToObjectPermission,
+):
     """Permission class based on InvestmentProjectPropositionAssociationChecker."""
 
     checker_class = InvestmentProjectPropositionAssociationChecker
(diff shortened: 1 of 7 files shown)


datahub/investment/proposition/test/test_views.py
```diff
@@ -1693,6 +1945,107 @@ def test_document_upload_schedule_virus_scan(
...
+        }
+        virus_scan_document_apply_async.assert_called_once_with(
+            args=(str(entity_document.document.pk), )
+        )
+
+    def test_restricted_user_cannot_schedule_virus_scan_for_non_associated_document(
+        self
+    ):
+        """
+        Test that restricted user cannot schedule a virus scan for non associated document
```
reupen: The first line of the docstring should end with a full stop (also should be a hyphen in non-associated).

datahub/investment/proposition/test/test_views.py
```diff
@@ -1693,6 +1945,107 @@ def test_document_upload_schedule_virus_scan(
...
+            original_filename='test.txt',
+            created_by=project_creator,
+        )
+
+        url = reverse(
+            'api-v3:investment:proposition:document-item-callback',
+            kwargs={
+                'proposition_pk': proposition.pk,
+                'project_pk': proposition.investment_project.pk,
+                'entity_document_pk': entity_document.pk
```
reupen: We should try and be consistent about when we use trailing commas. Would suggest that we follow the examples here: https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas (which means adding a trailing comma on this line).

datahub/investment/proposition/permissions.py
```diff
@@ -73,71 +71,22 @@ def get_actual_object(self, obj):
...
-    When subclassing, checker and non_associated_investment_project_message have to be overridden.
-    """
-
-    non_associated_investment_project_message = None
-    checker = None
-
-    def __init__(self):
-        """
-        Initialises the validator.
+    def has_permission(self, request, view):
```
reupen: This has_permission() looks the same as the one in IsAssociatedToInvestmentProjectPropositionPermission – can we move it to a common base class or a standalone function?

datahub/investment/proposition/views.py
```diff
@@ -171,3 +174,7 @@ def destroy(self, request, *args, **kwargs):
         data['proposition_id'] = entity_document.proposition_id
         record_user_event(request, USER_EVENT_TYPES.proposition_document_delete, data=data)
         return super().destroy(request, *args, **kwargs)
+
+    def _check_proposition_exists(self):
+        if not Proposition.objects.filter(pk=self.kwargs['proposition_pk']).exists():
+            raise Http404(self.non_existent_proposition_error_message)
```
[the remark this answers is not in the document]
elcct: Good catch! Added a test to cover it. Thanks

PR #1115, opened 2018-09-18
Revert make links in CSV exports clickable in Excel

reupen:
### Description of change

This makes URLs in CSV exports unclickable again (in Excel). This is because hyperlinks to Data Hub pages don't behave correctly when clicked on in Excel. Excel tries to open the page itself and gets redirected to SSO, and then opens SSO in the browser. The user then ends up getting redirected to the Data Hub home page instead of the originally requested URL, as the front end stores the original URL in the user's session.

See https://support.microsoft.com/kb/899927 for further information.

### Checklist

### DIFF
--- a/changelog/revert-hyperlinks-in-exports.feature
+++ b/changelog/revert-hyperlinks-in-exports.feature
@@ -1,0 +1,1 @@
+URLs in CSV exports and reports are no longer clickable when the CSV file is opened in Excel. This is because the links do not behave correctly when clicked on in Excel (see https://support.microsoft.com/kb/899927 for further information on why).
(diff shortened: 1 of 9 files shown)


stuaxo:
Is there a missing feature in SSO or datahub to get it to go to the right page after SSO ?

reupen:
Yes, it involves both the FE and SSO I think. Right now, the FE stores the originally requested URL in the session (and the redirect URL given to SSO points to `/oauth/callback`).

To make this work, the originally requested URL would need to be passed to SSO as part of the query string rather than relying on the session (and the value would need to be validated by at least one of the apps to make sure it hasn't been manipulated).

PR #1119, opened 2018-09-24
Add the ability to load email addresses to opt out of marketing via the admin site

reupen:
### Description of change

Adds a button on the contact change list page in the admin site that takes you to a form to load a CSV file of email addresses to opt out of email marketing.

Also splits up datahub.company.admin into a subpackage containing multiple module.

Would suggest looking at the two commits separately.

### Checklist

### DIFF
--- a/datahub/company/admin/adviser.py
+++ b/datahub/company/admin/adviser.py
@@ -1,0 +1,59 @@
+from django.contrib import admin
+from django.contrib.auth.admin import UserAdmin
+from reversion.admin import VersionAdmin
+
+from datahub.company.models import Advisor
+
+
+@admin.register(Advisor)
+class AdviserAdmin(VersionAdmin, UserAdmin):
+    """Adviser admin."""
+
+    fieldsets = (
+        (None, {
+            'fields': (
+                'email',
+                'password'
+            )
+        }),
+        ('PERSONAL INFO', {
+            'fields': (
+                'first_name',
+                'last_name',
+                'contact_email',
+                'telephone_number',
+                'dit_team'
+            )
+        }),
+        ('PERMISSIONS', {
+            'fields': (
+                'is_active',
+                'is_staff',
+                'is_superuser',
+                'groups',
+                'user_permissions'
+            )
+        }),
+        ('IMPORTANT DATES', {
+            'fields': (
+                'last_login',
+                'date_joined'
+            )
+        }),
+    )
+    add_fieldsets = (
+        (None, {
+            'classes': ('wide',),
+            'fields': ('email', 'password1', 'password2'),
+        }),
+    )
+    list_display = ('email', 'first_name', 'last_name', 'dit_team', 'is_active', 'is_staff',)
+    search_fields = (
+        '=pk',
+        'first_name',
+        'last_name',
+        'email',
+        '=dit_team__pk',
+        'dit_team__name',
+    )
+    ordering = ('email',)
--- a/datahub/company/admin/contact.py
+++ b/datahub/company/admin/contact.py
@@ -1,8 +1,142 @@
-from django.contrib import admin
+import csv
+import io
+from logging import getLogger
+from typing import NamedTuple
+
+import reversion
+from chardet import UniversalDetector
+from django import forms
+from django.contrib import admin, messages as django_messages
+from django.contrib.admin.templatetags.admin_urls import admin_urlname
+from django.core.exceptions import PermissionDenied, ValidationError
+from django.core.validators import FileExtensionValidator
+from django.http import HttpResponseRedirect
+from django.template.response import TemplateResponse
+from django.urls import path, reverse
+from django.utils.translation import gettext
 from reversion.admin import VersionAdmin
 
 from datahub.company.models import Contact
 from datahub.core.admin import BaseModelAdminMixin
+from datahub.search.signals import disable_search_signal_receivers
+
+
+logger = getLogger(__name__)
+
+
+class LoadEmailMarketingOptOutsForm(forms.Form):
+    """Form used for loading a CSV file to opt out contacts from email marketing."""
+
+    HEADER_DECODE_ERROR_MESSAGE = gettext('There was an error decoding the file contents.')
+    BODY_DECODE_ERROR_MESSAGE = gettext(
+        'There was an error decoding the text in the file provided. No records have been '
+        'modified.',
+    )
+    NO_EMAIL_COLUMN_MESSAGE = gettext('This file does not contain an email column.')
+
+    email_list = forms.FileField(
+        label='Email list (CSV file)',
+        validators=[FileExtensionValidator(allowed_extensions=('csv',))],
+    )
+
+    def clean_email_list(self):
+        """Validates the uploaded CSV file and creates a CSV DictReader from it."""
+        # This could be an instance of InMemoryUploadedFile or TemporaryUploadedFile
+        # (depending on the file size)
+        file_field = self.cleaned_data['email_list']
+
+        # Guess the file encoding (primarily to check for a UTF-8 BOM)
+        encoding_detector = UniversalDetector()
+        for chunk in file_field.chunks():
+            encoding_detector.feed(chunk)
+            if encoding_detector.done:
+                break
+
+        detection_result = encoding_detector.close()
+        encoding = detection_result['encoding']
+
+        file_field.seek(0)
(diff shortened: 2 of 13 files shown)


datahub/company/admin/contact.py
```diff
@@ -36,3 +96,108 @@ class ContactAdmin(BaseModelAdminMixin, VersionAdmin):
...
     )
+
+    def get_urls(self):
+        """Gets the URLs for this model."""
+        return [
+            *super().get_urls(),
+            path(
+                'load-email-marketing-opt-outs',
+                self.admin_site.admin_view(self.opt_out_form),
+                name='company_contact_load-email-marketing-opt-outs',
```
marcofucci: It would be better to get app_label and model_name from the model e.g. https://github.com/django/django/blob/master/django/contrib/admin/options.py#L612

datahub/company/admin/contact.py
```diff
@@ -36,3 +96,108 @@ class ContactAdmin(BaseModelAdminMixin, VersionAdmin):
...
+        self.message_user(request, success_msg, messages.SUCCESS)
+
+        if opt_out_res.num_non_matching_email_addresses:
+            warning_msg = (
+                f'{opt_out_res.num_non_matching_email_addresses} email addresses did not match a '
+                f'contact'
+            )
+            self.message_user(request, warning_msg, messages.WARNING)
+
+        return HttpResponseRedirect(reverse('admin:company_contact_changelist'))
```
marcofucci: Same here re. app_label and model_name.

datahub/company/admin/contact.py
```diff
@@ -36,3 +96,108 @@ class ContactAdmin(BaseModelAdminMixin, VersionAdmin):
...
+        num_contacts_updated = 0
+        num_non_matching_email_addresses = 0
+
+        for row in data:
+            email = row['email']
+
+            if not email:
+                continue
+
+            contacts = Contact.objects.filter(email=email)
```
marcofucci: what about `email_alternative`?

marcofucci: shouldn't it be case insensitive? And probably trim it?

marcofucci: Just read the copy in the template _alternative email addresses are ignored_

datahub/company/admin/contact.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class LoadEmailMarketingOptOutsForm(forms.Form):
+    """Form used for loading a CSV file to opt out contacts from email marketing."""
+
+    DECODE_ERROR_MESSAGE = gettext('There was an error decoding the file contents.')
+    NO_EMAIL_COLUMN_MESSAGE = gettext('This file does not contain an email column.')
+
+    email_list = forms.FileField(
+        label='Email list (CSV file)',
```
elcct: This says `CSV file`, but allows `.txt`.

datahub/company/test/admin/test_contact.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+
+        client = self.create_client(user=user)
+        response = client.get(change_list_url)
+        assert response.status_code == status.HTTP_200_OK
+
+        load_opt_outs_url = reverse(
+            'admin:company_contact_load-email-marketing-opt-outs',
+        )
+        assert load_opt_outs_url not in response.rendered_content
```
marcofucci: It would also be good to assert something true to test that you are on the page you expect the user to be.

datahub/company/admin/contact.py
```diff
@@ -0,0 +1,203 @@
...
+                num_contacts_matched += 1
+
+                if contact.accepts_dit_email_marketing:
+                    num_contacts_updated += 1
+                    contact.accepts_dit_email_marketing = False
+                    contact.modified_by = request.user
+                    contact.save()
+
+            if not contacts:
+                num_non_matching_email_addresses += 1
```
elcct: Not sure if that is in the scope, but I would like to know which emails didn't match, so I can adjust / improve the list.

datahub/company/test/admin/test_contact.py
```diff
@@ -0,0 +1,337 @@
...
+    def test_does_not_allow_file_without_email_column(self):
+        """Test that the form rejects a CSV file that doesn't contain an email column."""
+        file = io.BytesIO(b'test\r\nrow')
+        file.name = 'test.csv'
+
+        url = reverse(
+            'admin:company_contact_load-email-marketing-opt-outs',
+        )
+        response = self.client.post(url, data={
+            'email_list': file
```
elcct: This field name may be confusing - as it says `email_list` but it takes a file. Since the endpoint offers only one purpose, maybe just the `file` or similar as a field name would be sufficient?

datahub/company/admin/contact.py
reupen: Not 100% convinced – the Django code is generic while this is not, and we already have loads of instances where admin route names are referenced like this in reverse() calls. But if we were going to do this, it would be better to add a function that generates the names (given a model and the last part of the name) instead of formatting them everywhere.

datahub/company/admin/contact.py
reupen: I would have liked to have done a preview even, but unfortunately that increases the complexity greatly as the data has to be kept around somewhere for the actual load.

We could display the first N warnings on a page (to avoid pagination and storage problems), though I don't really want to add any more complexity to this PR. So, perhaps we could log warnings for now and see how frequently they happen.

datahub/company/admin/contact.py
marcofucci: Using a generic logic here would make refactoring easy in case for example we want to move the model somewhere else. It doesn't add much complexity so I don't know why we wouldn't want it.

datahub/company/admin/contact.py
elcct: I see, in that case, indeed it would be great to add logging there so that we can see how common is to have non-matching emails.

datahub/company/admin/contact.py
```diff
@@ -36,3 +96,108 @@ class ContactAdmin(BaseModelAdminMixin, VersionAdmin):
...
+
+        if request.method != 'POST':
+            return self._opt_out_form_response(request)
+
+        form = LoadEmailMarketingOptOutsForm(request.POST, request.FILES)
+        if not form.is_valid():
+            return self._opt_out_form_response(request, form)
+
+        try:
+            opt_out_res = self._process_opt_outs(request, form.cleaned_data['email_list'])
```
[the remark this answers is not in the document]
reupen: Yes, it would probably sit better on the form. form.save() is for `ModelForm`s though, no?

datahub/company/admin/contact.py
reupen: Yes, I could get rid of .txt to simplify things.

datahub/company/test/admin/test_contact.py
reupen: I'm not sure file is too helpful either as it's non-specific and it gets converted to a csv.DictReader instance in the form. Open to other suggestions, though.

datahub/company/admin/contact.py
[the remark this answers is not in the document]
marcofucci: yeah a `Form` doesn't really have any `save` method but it's still good to create a `save` or equivalent method for this logic. It doesn't have to be called `save`.

datahub/company/admin/contact.py
```diff
@@ -36,24 +44,87 @@ def clean_email_list(self):
...
+
     @classmethod
     def _validate_columns(cls, csv_reader):
         try:
             fieldnames = csv_reader.fieldnames
         except UnicodeError as exc:
-            raise ValidationError(cls.DECODE_ERROR_MESSAGE, code='unicode-error') from exc
+            raise ValidationError(
+                cls.HEADER_DECODE_ERROR_MESSAGE,
+                code='header-unicode-error'
```
elcct: Is comma missing here?

datahub/company/admin/adviser.py
```diff
@@ -0,0 +1,59 @@
...
+
+@admin.register(Advisor)
+class AdviserAdmin(VersionAdmin, UserAdmin):
+    """Adviser admin."""
+
+    fieldsets = (
+        (None, {
+            'fields': (
+                'email',
+                'password'
```
marcofucci: trailing comma

PR #1120, opened 2018-09-24
Improve OAuth security

elcct:
### Description of change

This modifies the behaviour of `OAuthLibCore` so that it only parses the request body if the content type is `application/x-www-form-urlencoded`.
That blocks parsing of JSON body before the authentication is performed.

### DIFF
--- a/datahub/oauth/backend.py
+++ b/datahub/oauth/backend.py
@@ -1,0 +1,19 @@
+from django.http.multipartparser import parse_header
+from oauth2_provider.oauth2_backends import OAuthLibCore
+from rest_framework import HTTP_HEADER_ENCODING
+
+
+class ContentTypeAwareOAuthLibCore(OAuthLibCore):
+    """Extends the default OAuthLibCore to limit the use of request body."""
+
+    def extract_body(self, request):
+        """
+        Returns POST contents if content type is application/x-www-form-urlencoded.
+
+        Refer to https://tools.ietf.org/html/rfc6750#section-2.2 for more information.
+        """
+        content_type = request.content_type
+        base_media_type, _ = parse_header(content_type.encode(HTTP_HEADER_ENCODING))
+        if base_media_type == 'application/x-www-form-urlencoded':
+            return request.POST.items()
+        return ()
--- a/datahub/oauth/test/test_backend.py
+++ b/datahub/oauth/test/test_backend.py
@@ -1,0 +1,107 @@
+from unittest.mock import patch
+from urllib.parse import urlencode
+
+import pytest
+from django.conf import settings
+from oauth2_provider.contrib.rest_framework.permissions import IsAuthenticatedOrTokenHasScope
+from rest_framework import status
+from rest_framework.response import Response
+
+from datahub.core.test_utils import APITestMixin
+from datahub.core.viewsets import CoreViewSet
+from datahub.oauth.scopes import Scope
+
+
+@pytest.fixture
+def oauth2_backend_class(monkeypatch):
+    """Ensure ContentTypeAwareOAuthLibCore is being set."""
+    monkeypatch.setitem(
+        settings.OAUTH2_PROVIDER,
+        'OAUTH2_BACKEND_CLASS',
+        'datahub.oauth.backend.ContentTypeAwareOAuthLibCore',
+    )
+
+
+class RestrictedAccessViewSet(CoreViewSet):
+    """DRF ViewSet to test authentication."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
+
+
+class TestContentTypeAwareOAuthLibCore(APITestMixin):
+    """Tests for ContentTypeAwareOAuthLibCore."""
+
+    @pytest.mark.parametrize(
+        'token',
+        (
+            'Bearer token',
+            '',
+            'Cool cats',
+        )
+    )
+    @patch.object(RestrictedAccessViewSet, 'create')
+    def test_request_is_not_parsed_if_oauth2_token_is_invalid_and_content_type_is_application_json(
+        self,
+        create,
+        token,
+        oauth2_backend_class,
+        api_request_factory,
+    ):
+        """
+        Tests that if the oauth2 token is invalid and contenty type is application.json,
+        the request is not being parsed.
+        """
+        request = api_request_factory.post(
+            '/',
+            # sending invalid JSON
+            data=b'{"what": "cat}',
+            content_type='application/json',
+            Authorization=token,
+        )
+        my_view = RestrictedAccessViewSet.as_view(
+            actions={'post': 'create'}
+        )
+        response = my_view(request)
+
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+        create.assert_not_called()
+
+    @pytest.mark.parametrize(
+        'content_type,expected_authorized',
+        (
+            ('application/x-www-form-urlencoded; charset=utf-8', True),
+            ('application/x-www-form-urlencoded;', True),
+            ('application/cats-on-mars', False),
+        )
+    )
+    @patch.object(RestrictedAccessViewSet, 'create')
+    def test_request_is_parsed_if_oauth2_token_is_in_the_form(
+        self,
+        create,
+        content_type,
+        expected_authorized,
+        oauth2_backend_class,
+        api_request_factory,
+    ):
+        """Tests that the request is being parsed if the auth token is in the form."""
+        if expected_authorized:
+            create.return_value = Response(data={'result': True})
+        data = {
+            'access_token': self.get_token(Scope.internal_front_end),
+        }
+        request = api_request_factory.post(
+            '/',
+            data=urlencode(data),
+            content_type=content_type,
+        )
+        my_view = RestrictedAccessViewSet.as_view(
+            actions={'post': 'create'}
+        )
+        response = my_view(request)
+        if expected_authorized:
+            assert response.status_code == status.HTTP_200_OK
+            create.assert_called_once()
+        else:
+            assert response.status_code == status.HTTP_401_UNAUTHORIZED
+            create.assert_not_called()
--- a/datahub/oauth/test/test_scopes.py
+++ b/datahub/oauth/test/test_scopes.py
@@ -1,5 +1,6 @@
 from base64 import b64encode
 from unittest import mock
+from urllib.parse import urlencode
 
 import pytest
 from django.test.utils import override_settings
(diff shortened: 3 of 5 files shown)


datahub/core/__init__.py
```diff
@@ -1 +0,0 @@
-default_app_config = 'datahub.core.apps.CoreConfig'
```
reupen: Is this change meant to be here – why is it being removed?

elcct: Oh thanks, I had some stuff in there and deleted too much :blush:.

datahub/core/test/test_auth.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class RestrictedAccessViewSet(CoreViewSet):
+    """DRF ViewSet to test authentication."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
+
+
+class TestStrictOAuth2Authentication(APITestMixin):
```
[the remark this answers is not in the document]
elcct: Good idea. I moved it to `oauth` app and also renamed it. Thanks!

datahub/oauth/test/test_scopes.py
```diff
@@ -31,10 +31,13 @@ def test_creating_a_token_default_scope(self):
...
             HTTP_AUTHORIZATION=_create_auth_header(app.client_id, app.client_secret)
         )
+        data = {'grant_type': 'client_credentials'}
         url = reverse('token')
-        response = client.post(url, data={
-            'grant_type': 'client_credentials',
-        })
+        response = client.post(
+            url,
+            urlencode(data),
```
reupen: It would be better to keep data= here as data is a keyword argument.

elcct: I missed that. Thanks!

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,51 @@
...
+    @pytest.mark.parametrize(
+        'token',
+        (
+            'Bearer token',
+            '',
+            'Cool cats',
+        )
+    )
+    @patch.object(RestrictedAccessViewSet, 'create')
+    def test_request_is_not_parsed_if_oauth2_token_is_invalid(self, create, token):
```
marcofucci: I know that this is indirectly testing `ContentTypeAwareOAuthLibCore` but `settings.OAUTH2_BACKEND_CLASS` could change and this would break without obvious reasons.
Maybe set `settings.OAUTH2_BACKEND_CLASS` to `ContentTypeAwareOAuthLibCore` before the tests.

I know that if the settings change we wouldn't know about it because there isn't anything testing the setting but I think leaving it like this is obscure.

Also, I'd add an extra test for a successful call with content_type='application/x-www-form-urlencoded'.

datahub/oauth/backend.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def extract_body(self, request):
+        """
+        Returns POST contents if content type is application/x-www-form-urlencoded.
+
+        Refer to https://tools.ietf.org/html/rfc6750#section-2.2 for more information.
+        """
+        if request.content_type == 'application/x-www-form-urlencoded':
+            return request.POST.items()
+        return ''
```
reupen: Looking at OAuthLibCore._extract_params(), it calls urlencode() on the return value from here, which wants an iterable of pairs. Although an empty string works, I'm not sure it quite makes sense. I would stick with an empty list or tuple (or similar).

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,51 @@
...
+        """Tests that if the oauth2 token is invalid, the request is not being parsed."""
+        create.return_value = Response(data={'result': True})
```
reupen: Do we need to set a return value if we're expecting it not to be called?

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,51 @@
...
+from oauth2_provider.contrib.rest_framework.permissions import IsAuthenticatedOrTokenHasScope
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.test import APIRequestFactory
+
+from datahub.core.test_utils import APITestMixin
+from datahub.core.viewsets import CoreViewSet
+from datahub.oauth.scopes import Scope
+
+factory = APIRequestFactory()
```
reupen: Why do we instantiate this here?

datahub/oauth/test/test_scopes.py
```diff
@@ -120,9 +133,11 @@ def test_scope_not_allowed(self, test_urls, grant_type):
     def test_expired_token(self, test_urls, grant_type):
         """Tests a test view with an expired token and the required scope."""
         application = self.get_application(grant_type=grant_type)
-        access_token = AccessTokenFactory(application=application,
-                                          scope=TestScope.test_scope_1,
-                                          expires=Faker('past_datetime', tzinfo=utc))
+        access_token = AccessTokenFactory(
+            application=application,
+            scope=TestScope.test_scope_1,
+            expires=Faker('past_datetime', tzinfo=utc)
```
reupen: OK, the formatting was wrong before, but we should add a trailing comma on this line if we're changing it.

reupen:
Also was going to mention that I'd make the commit message a bit more descriptive – the text in the PR could go in the commit message body.

datahub/oauth/test/test_backend.py
elcct: I have added a fixture that sets `OAUTH2_BACKEND_CLASS` for each test. I also added a test that checks succesful call with content_type='application/x-www-form-urlencoded'. Thanks!

datahub/oauth/backend.py
elcct: Good catch! In the examples, I checked they have used an empty string. I should have checked if that is correct. Changed to return an empty tuple. Thanks

datahub/oauth/test/test_backend.py
elcct: We don't need it, I forgot to delete that. Removed.

datahub/oauth/test/test_backend.py
elcct: There is no need to. I made it use `api_request_factory` fixture instead. Thanks!

datahub/oauth/test/test_scopes.py
elcct: Indeed, missed that. Fixed, thanks!

elcct:
When I squash the commits I will include PR description in the commit message.

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,97 @@
...
+
+from datahub.core.test_utils import APITestMixin
+from datahub.core.viewsets import CoreViewSet
+from datahub.oauth.scopes import Scope
+
+
+@pytest.fixture
+def oauth2_backend_class():
+    """Ensure ContentTypeAwareOAuthLibCore is being set."""
+    _oauth2_backend_class = None
```
reupen: You could just use `monkeypatch` and its `setitem()` method I would have thought.

elcct: Oh, that way is indeed much simpler. Thanks!

datahub/oauth/backend.py
```diff
@@ -0,0 +1,15 @@
...
+class ContentTypeAwareOAuthLibCore(OAuthLibCore):
+    """Extends the default OAuthLibCore to limit the use of request body."""
+
+    def extract_body(self, request):
+        """
+        Returns POST contents if content type is application/x-www-form-urlencoded.
+
+        Refer to https://tools.ietf.org/html/rfc6750#section-2.2 for more information.
+        """
+        if request.content_type == 'application/x-www-form-urlencoded':
```
reupen: Although it seems like passing parameters with `application/x-www-form-urlencoded` isn't really a thing, I think I would still use either `cgi.parse_header()` or the DRF way to get the content type from the header:

https://github.com/encode/django-rest-framework/blob/master/rest_framework/request.py#L32

Just because it's more consistent and it's more in line with what was happening before.

elcct: That is true - it would have failed if content type also contained encoding information. I have improved the check and tests. Thanks!

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,104 @@
...
+from datahub.oauth.scopes import Scope
+
+
+@pytest.fixture
+def oauth2_backend_class(monkeypatch):
+    """Ensure ContentTypeAwareOAuthLibCore is being set."""
+    monkeypatch.setitem(
+        settings.OAUTH2_PROVIDER,
+        'OAUTH2_BACKEND_CLASS',
+        'datahub.oauth.backend.ContentTypeAwareOAuthLibCore'
```
marcofucci: trailing comma missing

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        response = my_view(request)
+
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+        create.assert_not_called()
+
+    @pytest.mark.parametrize(
+        'content_type,authorized',
+        (
+            ('application/x-www-form-urlencoded; charset=utf-8', True,),
```
marcofucci: you don't need the trailing comma after `True`, the same for the other lines

elcct:
I haven't but will look into that.

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,104 @@
...
+    @pytest.mark.parametrize(
+        'token',
+        (
+            'Bearer token',
+            '',
+            'Cool cats',
+        )
+    )
+    @patch.object(RestrictedAccessViewSet, 'create')
+    def test_request_is_not_parsed_if_oauth2_token_is_invalid(
```
marcofucci: This is not 100% accurate, isn't it? The request body is not parsed if oauth2 token is invalid *and* the content_type is _application/json_. Same for the docstring.

datahub/oauth/test/test_backend.py
```diff
@@ -0,0 +1,104 @@
...
+        my_view = RestrictedAccessViewSet.as_view(
+            actions={'post': 'create'}
+        )
+        response = my_view(request)
+
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+        create.assert_not_called()
+
+    @pytest.mark.parametrize(
+        'content_type,authorized',
```
marcofucci: I think it would be slightly clearer to add expected in the _authorized_ variable name somewhere otherwise it seems a param of the test instead of an expectation to check.

PR #1145, opened 2018-10-01
Add trailing commas to all OMIS apps

reupen:
### Description of change

This adds trailing commas to all of the OMIS-related apps where the trailing commas were missing. This includes some related code reformatting and the removal of an incorrect trailing comma.

### Checklist


datahub/omis/core/utils.py
```diff
@@ -47,7 +47,7 @@ def gen():
         # but it's unlikely and if the transaction is atomic, it would not put
         # the db in an inconsistent state.
         start_count = len(
-            model.objects.select_for_update().filter(created_on__date=current_date.date())
+            model.objects.select_for_update().filter(created_on__date=current_date.date()),
```
marcofucci: This one looks a bit weird, should we do an exception and don't have it here?

reupen: What do you mean? TBH I think it could just use an advisory lock instead of select_for_update().

If we want to avoid the comma though, I'd assign the query set to a variable first and call len() with that.

reupen:
> The len and list with extra trailing commas look a bit weird but I'm easy.

Yeah there aren't really any exceptions (with flake8-commas), and it's probably easier not to have any anyway.

PR #1172, opened 2018-10-10
Add company selection part of duplicate company merging tool

reupen:
### Description of change

This is the first part of a feature being added to the admin site to deal with duplicate companies.

This part allows the selection of companies that are duplicates and viewing the companies that have been selected. The selection of companies is done via the company change page so that the existing company search on the change list can be reused for this purpose.

As the feature is incomplete, it is currently behind a feature flag. The terminology used at the moment refers to merging companies, though this may change.

The logic for the views has been placed in a new class named CompanyMergeViews, to help keep things organised, and also to make it easier to reuse in case we decide to use a similar pattern elsewhere.

Subsequent parts will include the selection of which company should be kept as the active record, and the transfer of contacts and interactions from one company to another.

The actual tool will be limited to two companies, though the logic for that will follow in another PR.

Since it's an incomplete feature in the admin site, I haven't added a news fragment in this PR.

### Checklist

### DIFF
--- a/datahub/company/admin/company.py
+++ b/datahub/company/admin/company.py
@@ -3,6 +3,7 @@ from django.contrib import admin
 from django.db import models
 from reversion.admin import VersionAdmin
 
+from datahub.company.admin.company_merge_list import CompanyMergeViews
 from datahub.company.models import Company, CompanyCoreTeamMember
 from datahub.core.admin import BaseModelAdminMixin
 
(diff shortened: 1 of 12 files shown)


marcofucci:
This seems quite complex compared to an admin form with 2 `ModelChoiceField`s which would also be easier to understand for the staff user.

Any reason why you didn't implement it that way?

reupen:
You mean a pair of giant selects (drop-down lists)? I had a look at the autocomplete widget for admin as well but it's tied to model fields so there was no easy way to reuse that.

marcofucci:
I meant using the `ForeignKeyRawIdAdminField` widget like Django does everywhere else.

You might have to set the widget on the field programmatically in the init of the form as it expects an admin site instance I think.

screen shot 2018-10-11 at 10 57 11

reupen:
I tried using that at first, but similarly it's tied to a model field and I didn't have much luck using outside of ModelAdmin or a ModelForm. If you have an example of that I would happily switch to it...

marcofucci:
yeah you are right, after looking at the code it seems like it's tied to a field.

Would it be good to write a widget that borrows some logic from that? It would be useful in other cleansing related admin tool I guess we will build in the future?

That widget doesn't seem very complicated.

reupen:
Replaced by https://github.com/uktrade/data-hub-leeloo/pull/1178

PR #1174, opened 2018-10-11
Fix investment estimated land date validation

marcuspp:
### Description of change
When creating a new investment project estimated land date is required. This change means that the check is carried out even when other required fields are missing. Previously this was only done when all other required fields had been populated.

### How to test
- On master
- Navigate to a Company -> Investment -> Add new investment project
- Select a type of investment and a source of foreign equity investment
- Press continue
- Don't enter any project data and hit save, estimated land date won't be highlighted as required.
- On this branch
- Repeat Hit save, estimated land date will now be highlighted
- Find an existing Investment that does not have an estimated land date, make sure that it can be saved without an estimated land date

### Checklist

### DIFF
--- a/datahub/investment/serializers.py
+++ b/datahub/investment/serializers.py
@@ -21,7 +21,7 @@ from datahub.investment.models import (
     Involvement,
     SpecificProgramme,
 )
-from datahub.investment.validate import validate
+from datahub.investment.validate import REQUIRED_MESSAGE, validate
 
 CORE_FIELDS = (
     'id',
@@ -245,6 +245,12 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     proposal_deadline = serializers.DateField(required=False, allow_null=True)
     stage_log = NestedInvestmentProjectStageLogSerializer(many=True, read_only=True)
 
+    def validate_estimated_land_date(self, value):
+        """Validate estimated land date."""
+        if value or (self.instance and self.instance.allow_blank_estimated_land_date):
+            return value
+        raise serializers.ValidationError(REQUIRED_MESSAGE)
+
     def validate(self, data):
         """Validates the object after individual fields have been validated.
 
@@ -256,22 +262,12 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
         """
         if not self.instance:
             # Required for validation as DRF does not allow defaults for read-only fields
-            data['allow_blank_estimated_land_date'] = False
             data['allow_blank_possible_uk_regions'] = False
 
         self._check_if_investment_project_can_be_moved_to_verify_win(data)
         self._check_if_investment_project_can_be_moved_to_won(data)
-
-        fields = None
-        if self.partial and 'stage' not in data:
-            fields = data.keys()
-        errors = validate(self.instance, data, fields=fields)
-
-        if errors:
-            raise serializers.ValidationError(errors)
-
+        self._validate_for_stage(data)
         self._update_status(data)
-
         return data
 
     def _check_if_investment_project_can_be_moved_to_verify_win(self, data):
@@ -306,6 +302,14 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
                     'stage': self.default_error_messages['only_ivt_can_move_to_won'],
                 }
                 raise serializers.ValidationError(errors)
+
+    def _validate_for_stage(self, data):
+        fields = None
+        if self.partial and 'stage' not in data:
+            fields = data.keys()
+        errors = validate(self.instance, data, fields=fields)
+        if errors:
+            raise serializers.ValidationError(errors)
 
     def get_incomplete_fields(self, instance):
         """Returns the names of the fields that still need to be completed in order to
--- a/datahub/investment/test/test_serializers.py
+++ b/datahub/investment/test/test_serializers.py
@@ -2,6 +2,7 @@ import pytest
 
 from datahub.company.test.factories import AdviserFactory
 from datahub.investment.serializers import (
+    IProjectSerializer,
     IProjectTeamMemberListSerializer,
     IProjectTeamMemberSerializer,
 )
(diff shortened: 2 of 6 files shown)


datahub/investment/test/test_serializers.py
```diff
@@ -127,3 +128,20 @@ def test_team_member_list_update_add_only():
     assert updated_team_members[1].adviser == new_team_member_data[1]['adviser']
     assert updated_team_members[1].role == new_team_member_data[1]['role']
     assert project.team_members.count() == 2
+
+
+def test_estimated_land_date_is_required_for_new_project():
+    """Test creating new project requires an estimated land date."""
+    project_data = dict(estimated_land_date=None)
```
reupen: We use `{` and `}` for dict literals – we only use dict() for converting a sequence (list of tuples etc.) to a dict.

reupen:
We use the imperative mood and present tense for all commit messages, so, for example,

>  Added change log for estimated land date fix

should be

>  Add change log for estimated land date fix

(although in general we wouldn't create a separate commit for that).

The commit messages will always start with a verb.

marcuspp:
> We use the imperative mood and present tense for all commit messages, so, for example,
> 
> > Added change log for estimated land date fix
> 
> should be
> 
> > Add change log for estimated land date fix
> 
> (although in general we wouldn't create a separate commit for that).
> 
> Commit messages will always start with a verb.

Yeah sorry I was trying to make sure I followed that convention and forgot the change log hence a separate commit.

datahub/investment/serializers.py
```diff
@@ -245,6 +245,14 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     proposal_deadline = serializers.DateField(required=False, allow_null=True)
     stage_log = NestedInvestmentProjectStageLogSerializer(many=True, read_only=True)
 
+    def validate_estimated_land_date(self, value):
+        """Additional validation for estimated land date."""
+        if value or self.instance and self.instance.allow_blank_estimated_land_date:
```
reupen: Not sure about the logic here – shouldn't it be:

```
if value or (self.instance and self.instance.allow_blank_estimated_land_date):
    return value

raise serializers.ValidationError(REQUIRED_MESSAGE)
```
?

datahub/investment/test/test_serializers.py
```diff
@@ -127,3 +128,28 @@ def test_team_member_list_update_add_only():
...
+    assert not serializer.is_valid()
+    assert 'estimated_land_date' in serializer.errors.keys()
+
+
+def test_estimated_land_date_can_be_blank_if_allowed():
+    """Tests updating estimated land date to a blank when if allowed."""
+    project = InvestmentProjectFactory(allow_blank_estimated_land_date=True)
+    project_data = {'estimated_land_date': None}
+    serializer = IProjectSerializer(project, data=project_data, partial=True)
+    assert serializer.is_valid() is True
```
reupen: Just `assert serializer.is_valid()` would be fine – only need `is True` or `is False` when we care about the type (False vs None etc.).

datahub/investment/test/test_serializers.py
```diff
@@ -127,3 +128,28 @@ def test_team_member_list_update_add_only():
...
+    assert 'estimated_land_date' in serializer.errors.keys()
+
+
+def test_estimated_land_date_cannot_be_erased_if_value_already_present():
+    """Tests updating estimated land date cannot be set to a blank if not allowed."""
+    project = InvestmentProjectFactory()
+    project_data = {'estimated_land_date': None}
+    serializer = IProjectSerializer(project, data=project_data, partial=True)
+    assert not serializer.is_valid()
+    assert 'estimated_land_date' in serializer.errors.keys()
```
reupen: Would be nice to check that the error is the expected one as well.

datahub/investment/test/test_serializers.py
```diff
@@ -127,3 +128,28 @@ def test_team_member_list_update_add_only():
...
     assert updated_team_members[1].role == new_team_member_data[1]['role']
     assert project.team_members.count() == 2
+
+
+def test_estimated_land_date_is_required_for_new_project():
+    """Tests estimated land date is required for new projects."""
+    project_data = {'estimated_land_date': None}
+    serializer = IProjectSerializer(data=project_data)
+    assert not serializer.is_valid()
+    assert 'estimated_land_date' in serializer.errors.keys()
```
reupen: Would be nice to check that the error is the expected one as well.

datahub/investment/serializers.py
```diff
@@ -307,6 +301,17 @@ def _check_if_investment_project_can_be_moved_to_won(self, data):
                 }
                 raise serializers.ValidationError(errors)
 
+    def _additional_validation(self, data):
+        if not self.instance:
```
reupen: This seems like an odd place to mutate `data`. I think this and the next two lines would be better at the start of validate().

datahub/investment/serializers.py
```diff
@@ -245,6 +245,14 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     proposal_deadline = serializers.DateField(required=False, allow_null=True)
     stage_log = NestedInvestmentProjectStageLogSerializer(many=True, read_only=True)
 
+    def validate_estimated_land_date(self, value):
+        """Additional validation for estimated land date."""
```
reupen: Generally our docstrings for functions start with a verb – so either 

> Validate estimated land date.

or 

> Validates estimated land date.

(We have both forms in the code base at present, though the former is what PEP 257 says.)

datahub/investment/test/test_serializers.py
```diff
@@ -127,3 +128,28 @@ def test_team_member_list_update_add_only():
     assert updated_team_members[1].adviser == new_team_member_data[1]['adviser']
     assert updated_team_members[1].role == new_team_member_data[1]['role']
     assert project.team_members.count() == 2
+
+
+def test_estimated_land_date_is_required_for_new_project():
```
reupen: I would be nice to group these tests using a class. 

The existing tests weren't, but they were all testing IProjectTeamMemberListSerializer, so I would put those in a class as well.

datahub/investment/test/test_views.py
```diff
@@ -542,22 +542,23 @@ def test_create_project_conditional_failure(self):
             'intermediate_company': {
                 'id': str(intermediate_company.id),
             },
-            'referral_source_activity': {
```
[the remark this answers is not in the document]
marcuspp: I changed the test to reflect what actually happens on the site and what gets posted but I think that I should of kept both the original test and added this one as now there isn't a test that tests the conditional checks in validate.py.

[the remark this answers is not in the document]
reupen: It looks like test_create_project_fail and test_create_project_fail_none were covering the first pass of validation – not sure why estimated_land_date isn't coming up as an error there with this changes, though.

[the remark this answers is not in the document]
marcuspp: If the field is missing from the post `validate_<field>` function unfortunately isn't called hence why its not picked up with an empty post test. I'm aware this isn't ideal but this should be an improvement.

datahub/investment/serializers.py
```diff
@@ -245,6 +245,14 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     proposal_deadline = serializers.DateField(required=False, allow_null=True)
     stage_log = NestedInvestmentProjectStageLogSerializer(many=True, read_only=True)
 
+    def validate_estimated_land_date(self, value):
+        """Validate estimated land date."""
+        if value or (self.instance and self.instance.allow_blank_estimated_land_date):
+            return value
+        if not self.instance or self.instance.estimated_land_date:
```
marcofucci: Is this _if_ needed? It should always raise a ValidationError at this point I think.

datahub/investment/test/test_serializers.py
```diff
@@ -14,116 +15,140 @@
...
+
+class TestIProjectSerializer:
+    """Tests for IProjectSerializer."""
+
+    def test_estimated_land_date_is_required_for_new_project(self):
+        """Tests estimated land date is required for new projects."""
+        project_data = {'estimated_land_date': None}
+        serializer = IProjectSerializer(data=project_data)
+        assert not serializer.is_valid()
+        assert 'estimated_land_date' in serializer.errors.keys()
```
marcofucci: It would be good to assert the actual error here.

datahub/investment/test/test_serializers.py
```diff
@@ -14,116 +15,140 @@
...
+
+    def test_estimated_land_date_cannot_be_erased_if_value_already_present(self):
```
marcofucci: This is not 100% true, isn't it? If `estimated_land_date` is set and `allow_blank_estimated_land_date` is True, you can still erase `estimated_land_date` and the system is happy.

If this is okay, maybe change the name of this test and add an extra test to cover this case as well?

datahub/investment/test/test_serializers.py
```diff
@@ -14,116 +15,142 @@
...
+
+class TestIProjectSerializer:
+    """Tests for IProjectSerializer."""
+
+    def test_estimated_land_date_is_required_for_new_project(self):
+        """Tests estimated land date is required for new projects."""
+        project_data = {'estimated_land_date': None}
+        serializer = IProjectSerializer(data=project_data)
+        assert not serializer.is_valid()
+        assert 'estimated_land_date' in serializer.errors.keys()
```
reupen: `.keys()` is not needed here and in the next test.

datahub/investment/serializers.py
```diff
@@ -307,6 +303,14 @@ def _check_if_investment_project_can_be_moved_to_won(self, data):
                 }
                 raise serializers.ValidationError(errors)
 
+    def _additional_validation(self, data):
```
reupen: Just one small last comment, I don't think `_additional_validation` is a good name for this function. It could sound more like a function, and I think 'additional' underplays what it does a bit.

The docstring for validate() is `Validates an investment project for the current stage.` so maybe this could be something like `_validate_for_stage`.

PR #1189, opened 2018-10-23
Enhance SPI 2 with information who first assigned a project manager

elcct:
### Description of change

This adds new column to the investment project model - `project_manager_first_time_assigned_by` which stores who assigned the project manager for the first time. 
It moves the logic that determines if project manager has been assigned for the first time from the model to serializer and admin. This is because at the model level we don't know who performed the assignment. 
That columns is now included in the SPI report.

### Checklist


datahub/investment/serializers.py
```diff
@@ -245,6 +246,14 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     proposal_deadline = serializers.DateField(required=False, allow_null=True)
     stage_log = NestedInvestmentProjectStageLogSerializer(many=True, read_only=True)
 
+    def save(self, **kwargs):
+        """Saves when and who assigned a project manager for the first time."""
+        if 'project_manager' in self.validated_data and self.instance.project_manager is None:
```
reupen: Won't `self.instance` be `None` on project creation? Although I know `project_manager` wouldn't be set on creation with the current flow in the front end, would still be better to check `(self.instance is None or self.instance.project_manager is None)` instead of just `self.instance.project_manager is None` I think.

datahub/investment/admin.py
```diff
@@ -62,6 +63,25 @@ class InvestmentProjectAdmin(BaseModelAdminMixin, VersionAdmin):
         'modified_by',
     )
 
+    def save_model(self, request, obj, form, change):
+        """
+        Populate who and when assigned a project manager for the first time.
+        """
+        first_assigned = not change
+        if change:
+            old_investment_project = InvestmentProject.objects.get(pk=obj.pk)
```
reupen: Don't like loading the object again from the database to get the old state – it's a bit hacky and could lead to race conditions etc.

Any reason we can't get the old state from `form.initial`?

[Another option is `FieldTracker` from `model_utils`](https://django-model-utils.readthedocs.io/en/latest/utilities.html#field-tracker), but that's more similar to the old way.

datahub/investment/serializers.py
elcct: Your proposed way is better. Changed. Thanks!

datahub/investment/admin.py
elcct: I agree. I changed it to use `form.initial` instead and it looks much better now. Thanks!

datahub/investment/admin.py
```diff
@@ -62,6 +63,22 @@ class InvestmentProjectAdmin(BaseModelAdminMixin, VersionAdmin):
...
 
+    def save_model(self, request, obj, form, change):
+        """
+        Populate who and when assigned a project manager for the first time.
+        """
+        first_assigned = not change or form.initial['project_manager'] is None
+
+        if (
+            obj.project_manager
+            and 'project_manager' in form.changed_data
```
marcofucci: Do we need this line or `obj.project_manager and first_assigned` is enough?

elcct: We don't need it, I missed that. Removed. Thanks!

PR #1191, opened 2018-10-24
Add confirm merge page for duplicate company merging tool

reupen:
### Description of change

This:

- adds two template filters and a tag for generating links to change pages for objects and getting the singular or plural verbose name for a model depending on a count
- adds the last view in the duplicate company merging tool that displays the planned changes and asks the user to confirm if they want to go ahead (it does not actually merge the companies yet)

I'm planning to move and put the core merging logic in the `DuplicateCompanyMerger` class to keep it in once place (so I will probably move the couple of bits I put on the model to that class).

I'm also planning to do a bit of refactoring of the existing views to tidy things up a bit and makes things more consistent:

- put all views under the URL path `/admin/company/company/merge/`
- use POST for submitting the form for step 1
- put step 1/2/3 in more places to make the order of the views, templates etc. clearer
- use the new template utilities in the earlier views where applicable

I wasn't sure whether to make `verbose_name_for_count` a filter or tag, but I find the syntax for two-argument filters a bit confusing so I made it a tag.

### Checklist

### DIFF
--- a/datahub/company/admin/merge/step_2.py
+++ b/datahub/company/admin/merge/step_2.py
@@ -5,6 +5,7 @@ from django.http import HttpResponseRedirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils.decorators import method_decorator
+from django.utils.http import urlencode
 from django.utils.translation import gettext_lazy
 from django.views.decorators.csrf import csrf_protect
 
@@ -69,14 +70,17 @@ class SelectPrimaryCompanyForm(forms.Form):
         if not company_index:
             return
 
-        selected_company = self._company_1 if company_index == '1' else self._company_2
-        other_company = self._company_2 if company_index == '1' else self._company_1
+        target_company = self._company_1 if company_index == '1' else self._company_2
+        source_company = self._company_1 if company_index != '1' else self._company_2
 
-        if not selected_company.is_valid_merge_target:
+        if not target_company.is_valid_merge_target:
             raise ValidationError(self.INVALID_TARGET_COMPANY_MSG)
 
-        if not other_company.is_valid_merge_source:
+        if not source_company.is_valid_merge_source:
             raise ValidationError(self.INVALID_SOURCE_COMPANY_MSG)
+
+        self.cleaned_data['source_company'] = source_company
+        self.cleaned_data['target_company'] = target_company
 
 
 @feature_flagged_view(MERGE_COMPANY_TOOL_FEATURE_FLAG)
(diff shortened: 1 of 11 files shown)


datahub/company/admin/merge/step_3.py
```diff
@@ -0,0 +1,70 @@
...
+    is_post = request.method == 'POST'
+
+    if is_post:
+        # The merging logic is still to be implemented, redirect to the change list for now
+        changelist_route_name = admin_urlname(model_admin.model._meta, 'changelist')
+        changelist_url = reverse(changelist_route_name)
+        return HttpResponseRedirect(changelist_url)
+
+    template_name = 'admin/company/company/merge_confirm_selection.html'
+    title = gettext_lazy('Confirm merge')
```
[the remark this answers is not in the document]
reupen: No translations are planned, I just used gettext() there as it's usually recommended anyway. I'll have a look at updating the template to use translation blocks :+1:

datahub/company/admin/merge/step_3.py
```diff
@@ -0,0 +1,70 @@
...
+from datahub.company.merge import DuplicateCompanyMerger
+from datahub.company.models import Company
+from datahub.feature_flag.utils import feature_flagged_view
+
+
+class ConfirmMergeStateForm(forms.Form):
+    """Form for validating the query string in the confirm merge view."""
+
+    target_company = forms.ModelChoiceField(Company.objects.all())
+    source_company = forms.ModelChoiceField(Company.objects.all())
```
marcofucci: This doesn't validate anything apart from params being required. It would be better to used a shared stateform object with extra validation, including checking for `is_valid_merge_target` and `is_valid_merge_source`

reupen: It validates that:

- the parameters are in the query string
- they are valid UUIDs
- they refer to company objects

It wasn't meant to validate anything else (particularly things that could happen without the URL being fiddled with) since it doesn't have a good mechanism to give feedback to the user.

That kind of validation was intended to be done when the form on each page is submitted.

marcofucci: In that case you should refactor `step_2.SelectPrimaryCompanyStateForm` that has some validation, which for consistency you should be moved to the other form.
I find it a bit confusing that some validation is on the GET form and some on the POST form.

reupen: Yes, that should be moved out of there in step 2. I'll update it in the next PR which will make things more consistent.

datahub/company/templates/admin/company/company/merge_confirm_selection.html
```diff
@@ -0,0 +1,61 @@
...
+  <li>
+    {% if should_archive_source %}
+      {% blocktrans %}{{ source_company_link }} will be archived and marked as a duplicate record.{% endblocktrans %}
+    {% else %}
+      {% blocktrans %}{{ source_company_link }} will be marked as a duplicate record.{% endblocktrans %}
+    {% endif %}
+  </li>
+</ul>
+
+<p><strong>{% trans 'Once complete, this operation can&rsquo;t be reversed.' %}</strong></p>
```
elcct: Wouldn't "Once completed" be more appropriate, as that would indicate that operation has been done instead of being complete or whole?

reupen: Complete sounds more natural to me and I don't think there's any room for ambiguity here. The user is not completing anything here, so I don't think completed is appropriate.

datahub/company/admin/merge/step_2.py
```diff
@@ -69,15 +70,18 @@ def clean(self):
...
+        if not target_company.is_valid_merge_target:
             raise ValidationError(self.INVALID_TARGET_COMPANY_MSG)
 
-        if not other_company.is_valid_merge_source:
+        if not source_company.is_valid_merge_source:
             raise ValidationError(self.INVALID_SOURCE_COMPANY_MSG)
 
+        self.cleaned_data['source_company'] = source_company
+        self.cleaned_data['target_company'] = target_company
+
```
marcofucci: It's good practice to always return `self.cleaned_data` even if optional, the same for the other forms.

PR #1197, opened 2018-10-26
Make company merging tool views and templates more consistent

reupen:
### Description of change

This makes a few small changes the company merging tool views and templates:

* the view for the first step now uses POST for consistency with the other steps
* translation blocks are used in the templates for the first two steps
* a `reverse_with_query_string()` helper function was added to simplify building URLs with query strings
* some renaming to keep things more organised

Would suggest looking at the commits separately.

### Checklist


datahub/company/admin/merge/step_1.py
```diff
@@ -1,81 +1,101 @@
...
 from datahub.company.models import Company
 from datahub.core.admin import RawIdWidget
+from datahub.core.utils import reverse_with_query_string
 from datahub.feature_flag.utils import feature_flagged_view
 
 
+class SelectOtherCompanyStateForm(forms.Form):
+    """Form for validating the query string in the select other company view."""
+
+    company_1 = forms.ModelChoiceField(Company.objects.all())
```
marcofucci: I think that using a form for this is overkilled, can you do what django does using the pk field? https://github.com/django/django/blob/master/django/contrib/admin/options.py#L754

reupen: I was thinking of making each state form inherit from the one for the previous step, what do you think about that?

marcofucci: I think that it's a bit confusing because some state forms (e.g. `SelectPrimaryCompanyStateForm`) are doing some validation but not checking all the cases.

It's also confusing why you care about specific validation errors if you are still raising`SuspiciousOperation` in the end.

I think it would be clearer to just get the companies from the get using a util function which gets the values using the pk field and leave all the actual validation logic to the post form.

reupen: > I think that it's a bit confusing because some state forms (e.g. SelectPrimaryCompanyStateForm) are doing some validation but not checking all the cases.

That can go, it only really needs to do the minimum needed to render the view.

> I think it would be clearer to just get the companies from the get using a util function which gets the values using the pk field and leave all the actual validation logic to the post form.

I don't think think replacing it with more code is an improvement. We already use DRF serialisers for query parameters in the API which is a similar concept. This way the code is more declarative, if we feel it's not clear enough what's going on then I think we need to find a way to improve that. It would probably be fairly easy to write a generic view function for the steps and just pass it some configuration (forms, title, next view etc.)

datahub/company/admin/merge/step_1.py
```diff
@@ -1,81 +1,101 @@
...
-    template_name = 'admin/company/company/merge_select_other_company.html'
-    title = gettext_lazy('Merge with another company')
+    state_form = SelectOtherCompanyStateForm(request.GET)
+
+    if not state_form.is_valid():
+        raise SuspiciousOperation()
 
-    obj = model_admin.get_object(request, unquote(object_id))
-    form = SelectOtherCompanyForm(object_id, request.GET or None)
+    company_1 = state_form.cleaned_data['company_1']
```
marcuspp: This seems strange behaviour using a GET parameter for company_1 whilst posting company_2?

datahub/company/admin/merge/step_1.py
marcofucci: The thing is, that form is not actually doing anything apart from checking that the id exists in the list of companies that can be selected (which is all). It's the same as getting the id, and making a get query.
I don't think that having a func that does that instead of using a form with 2 fields is more code, I actually think it's less code and more explicit because it's clearer what your intentions are.
Using a form gives you the impression that you are limiting and validating the inputs somehow which you are doing just in the POST form so I think it would be better to keep it more explicit.

datahub/company/admin/merge/step_1.py
reupen: The first company passed to the view from the company the button was clicked on. The alternatives were also complicated:

- put in the session – then have to worry about tabs, when to invalidate etc.
- pass in via query string and then pass back in the POST request – then we have to mess about with hidden fields

datahub/company/admin/merge/step_1.py
marcofucci: Your latest step1 changes `get_object` look very nice, do you think you can do the same for step2 and step3?

PR #1198, opened 2018-10-26
Add transferred_to field to companies

reupen:
### Description of change

This adds a field to the Company model to indicate if a company is a duplicate of another company. This also blocks companies that have this field set from being unarchived.

### Checklist


marcuspp:
Loaded the test data and because the merged company no longer has a name it isn't selectable in the admin company view. If this is by design this view could eventually get cluttered.

screen shot 2018-10-29 at 11 56 32

Also should this new duplicate company appear in the duplicates section of `Mars Export ltd` as it does not?

reupen:
> Loaded the test data and because the merged company no longer has a name it isn't selectable in the admin company view. If this is by design this view could eventually get cluttered.

Good spot, I left out a name by accident.

> Also should this new duplicate company appear in the duplicates section of Mars Export ltd as it does not?

I think that's just related to the no-name problem, name is not optional so we should be OK there.

PR #1199, opened 2018-10-29
Add exclude from the investment flow to investment stages metadata

marcuspp:
### Description of change
A new field has been added to the investment project stages metadata to indicate if a stage should be excluded from the investment flow. This field will allow a stage to be deprecated without causing disruption to the flow views.

### To Test
- Run migrations `./manage.py migrate`
- Check endpoint shows the new field `http://localhost:8000/metadata/investment-project-stage/`
- Check admin shows the new field `http://localhost:8000/admin/metadata/investmentprojectstage/` 

### Checklist


datahub/metadata/models.py
```diff
@@ -181,6 +181,8 @@ class CompanyClassification(BaseOrderedConstantModel):
 class InvestmentProjectStage(BaseOrderedConstantModel):
     """Investment project stage."""
 
+    exclude_from_investment_flow = models.NullBooleanField()
```
marcofucci: Stupid question, do you need `NullBooleanField` or could you have a `BooleanField` with default `False`?

marcuspp: I used a NullBooleanField as I didn't want to create a column and set a default in the same migration. I've experienced in the past DB locking issues with this so always air on the side of caution. Although this table only has 5 rows so would of probably be fine and this might of only been an issue with older version of Django.

marcuspp:
> Just a small comment – I think we should keep those changelog items under `investment` (rather than `metadata`) as it's all pretty specific to investment projects anyway.

Sure I can change this

datahub/metadata/models.py
reupen: Yes, long locks can be a pain, but I don't think it will make much difference here since (as you pointed out) there are only five records.

reupen: I'm not sure `exclude_from_investment_flow` makes it obvious enough what it does. Would `exclude_from_timeline` cover the intended use? Alternatively, something could be put in help_text to explain it.

marcuspp: I discussed the naming convention with some of my team members and we agreed on flow as timeline in future iterations will mean something different. I'll update the help text on the field though.

marcuspp: I'd prefer to leave this as is if thats okay.

marcofucci: So you said you are going to change it to BooleanField in another PR and migration?

marcofucci: I'm not sure if you my question as the file is now outdated

marcuspp: I wasn't planning to but can add it to another PR that I have to do that just updates the DB. IMO this approach should be best practice to protect from any disruption/downtime when releasing.

marcofucci: yes please, creating another PR to get rid of the `NullBooleanField` would be good. Having 3 values for a boolean field is not ideal in these cases unless we treat null and False as the same which is inaccurate and misleading.

marcuspp: Okay. Will do :+1:

datahub/metadata/admin.py
```diff
@@ -147,3 +146,12 @@ class SectorAdmin(MPTTModelAdmin):
     readonly_fields = ('id',)
     search_fields = ('segment', 'pk')
     list_filter = (DisabledOnFilter,)
+
+
+@admin.register(models.InvestmentProjectStage)
+class InvestmentProjectStageAdmin(ViewOnlyAdmin):
```
marcofucci: I don't think that allowing this to be edited via django admin is good. 
Any change to this via the admin could break a lot of investment logic. 
Also, we constantly check the values of this table in the code (e.g. `str(data['stage'].id) == InvestmentProjectStage.verify_win.value.id`) basically treating them as constants. Constants should be managed in the code by developers instead of via the django admin by a member of staff.

I suggest you change the data using a migration or you use (after extending its functionality) `datahub.core.utils.load_constants_to_database` instead.

reupen: @marcofucci It's using ViewOnlyAdmin – what would be editable via admin?

marcofucci: oh sorry, ignore me 👍

marcuspp: It is readonly. I just wanted to show the new field in admin.

PR #1201, opened 2018-10-30
Add streamlined investment flow via a feature flag

marcuspp:
### Description of change
This is the main part of the streamlined investment flow. Released under a feature flag to eliminate any disruption to the investment views to also allow sync with the front end and to control exactly when the feature is activated. The change moves when the project manager data is required and will eventually result in the Assign PM stage being deprecated.

Still to do in separate PRs:
 - Migration/Command to update any Projects currently in the Assign PM stage back to Prospect
 - Migration/Command to update the exclude_from_investment_flow flag for 'Assign PM'
- A FE release that filters any project stage with the exclude flag set.
- Stage removed as a search option.

### Checklist

### DIFF
--- a/datahub/investment/test/test_validate.py
+++ b/datahub/investment/test/test_validate.py
@@ -3,6 +3,8 @@ import pytest
 from datahub.company.test.factories import AdviserFactory, ContactFactory
 from datahub.core import constants
 from datahub.core.test_utils import random_obj_for_model
+from datahub.feature_flag.test.factories import FeatureFlagFactory
+from datahub.investment.constants import FEATURE_FLAG_STREAMLINED_FLOW
 from datahub.investment.models import InvestmentDeliveryPartner
 from datahub.investment.serializers import (
     CORE_FIELDS,
@@ -11,7 +13,11 @@ from datahub.investment.serializers import (
     VALUE_FIELDS,
 )
 from datahub.investment.test.factories import InvestmentProjectFactory
-from datahub.investment.validate import validate
+from datahub.investment.validate import (
+    _get_desired_stage_order,
+    InvestmentProjectStageValidationConfig,
+    validate,
+)
 from datahub.metadata.models import ReferralSourceActivity, UKRegion
 
 pytestmark = pytest.mark.django_db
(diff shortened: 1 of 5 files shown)


datahub/investment/test/test_validate.py
```diff
@@ -352,3 +354,84 @@ def test_validate_verify_win_instance_with_cond_fields():
...
+        (constants.InvestmentProjectStage.active.value, None, 400),
+        (constants.InvestmentProjectStage.active.value, True, 500),
+    ),
+)
+def test_get_desired_stage_order(desired_stage, next_stage, expected_stage_order):
+    """Tests get desired stage order when streamlined investment flow flag is not set."""
+    assert _get_desired_stage_order(desired_stage, next_stage) == expected_stage_order
+
+
+class ValidationMixin:
```
reupen: We would normally use a pytest fixture for things like this – any reason it is a mixin?

datahub/investment/validate.py
```diff
@@ -57,34 +35,90 @@ def _get_to_many_id(instance):
...
-        CondValRule('number_new_jobs', bool, Stage.verify_win.value),
-    'associated_non_fdi_r_and_d_project':
-        CondValRule('non_fdi_r_and_d_budget', bool, Stage.verify_win.value),
-}
+class ValidationMapping:
+    """Conditional validation mapping for each investment project stage"""
+
+    def __init__(self):
+        """Checking feature flags that alter validation mapping."""
+        self.is_streamlined_flow = is_feature_flag_active(FEATURE_FLAG_STREAMLINED_FLOW)
```
marcofucci: can't you move this to `_get_next_stage_for_prospect` as that's the only place were you use this attr?

datahub/investment/validate.py
```diff
@@ -57,34 +35,90 @@ def _get_to_many_id(instance):
...
+
+    def _get_next_stage_for_prospect(self):
+        if self.is_streamlined_flow:
+            return Stage.active.value
+        return Stage.assign_pm.value
+
+    def get_required_fields(self):
```
marcofucci: it doesn't really _get required fields_, does it? Maybe rename it to somethings more explicit like `get_required_fields_mapping` or `get_fields_required_after_stage`

datahub/investment/validate.py
```diff
@@ -57,34 +35,90 @@ def _get_to_many_id(instance):
...
+            'export_revenue': Stage.verify_win.value,
+            'address_1': Stage.verify_win.value,
+            'address_town': Stage.verify_win.value,
+            'address_postcode': Stage.verify_win.value,
+            'actual_uk_regions': Stage.verify_win.value,
+            'delivery_partners': Stage.verify_win.value,
+            'actual_land_date': Stage.won.value,
+        }
+
+    def get_conditional_fields(self):
```
marcofucci: Same comment about renaming

datahub/investment/validate.py
marcuspp: I will change and add `after_stage` but I removed mapping from the function name as the class is called ValidationMapping. Resulting in `ValidationMapping().get_required_fields()`

datahub/investment/test/test_validate.py
marcuspp: I find the tests easier to read using a mixin like this. I also think it makes removing the tests when the feature flag is no longer required easier. Happy to change it if you'd prefer I use a fixture.

datahub/investment/validate.py
marcofucci: True but it kind of depends on how you call the var as well e.g. `validator = ValidationMapping(); validator.get_required_fields()`

datahub/investment/test/test_validate.py
reupen: Yes, I would use a fixture as it's our usual style and separates test set-up. You can use `@pytest.mark.usefixtures('<fixture_name>')` on the class or each function if the fixture argument is unused.

reupen: Oh right, now I've seen how it's being used. We would use `@pytest.mark.parametrize()` for things like that. You should be able to use it on a class as well.

datahub/investment/validate.py
reupen: Yes, and the class name is a bit misleading as ValidationMapping isn't a mapping (if it were a mapping I would expect it to be an instance of collections.abc.Mapping).

datahub/investment/validate.py
```diff
@@ -141,3 +174,15 @@ def _check_rule(combiner, rule):
     if callable(rule.condition):
         return rule.condition(value)
     return value == rule.condition
+
+
+def _get_desired_stage_order(desired_stage, next_stage):
+    desired_stage_order = desired_stage.order
```
reupen: This is a bit hard to follow – mainly because of the multiple additions.

Something like this would make the logic easier to follow (as you don't need to keep track of the additions to see what the return value is):
```python
def _get_desired_stage_order(desired_stage, next_stage):
    if not next_stage:
        return desired_stage.order
    
    is_streamlined_flow_active = is_feature_flag_active(FEATURE_FLAG_STREAMLINED_FLOW)
    
    if is_streamlined_flow_active and str(desired_stage.id) == Stage.prospect.value.id:
        return desired_stage.order + 200.0

    return desired_stage.order + 100.0
```

datahub/investment/test/test_validate.py
elcct: I also would prefer to use a fixture instead.

datahub/investment/validate.py
```diff
@@ -57,34 +35,90 @@ def _get_to_many_id(instance):
...
+        """Conditional validation rules. Mapping from field names to validation rules."""
+        next_stage_for_prospect = self._get_next_stage_for_prospect()
+        return {
+            'referral_source_activity_event':
+                CondValRule(
+                    'referral_source_activity', Activity.event.value.id, Stage.prospect.value,
+                ),
+            'other_business_activity':
+                CondValRule(
+                    'business_activities', partial(_contains_id, BusinessActivity.other.value.id),
```
elcct: Could we have each parameter in its own line (if all don't fit)?

PR #1202, opened 2018-10-30
update config file for PaaS London

nusnewob:
### Description of change

use PaaS envar to read configs

### Checklist

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -225,15 +225,21 @@ SEARCH_APPS = [
     'datahub.search.omis.OrderSearchApp',
 ]
 
+VCAP_SERVICES = env.json('VCAP_SERVICES', default={})
+
 # Leeloo stuff
-ES_USE_AWS_AUTH = env.bool('ES_USE_AWS_AUTH', False)
+if 'elasticsearch' in VCAP_SERVICES:
+    ES_URL = VCAP_SERVICES['elasticsearch'][0]['credentials']['uri']
+    ES_USE_AWS_AUTH = False
+else:
+    ES_URL = env('ES5_URL')
+    ES_USE_AWS_AUTH = env.bool('ES_USE_AWS_AUTH', False)
+
 if ES_USE_AWS_AUTH:
     AWS_ELASTICSEARCH_REGION = env('AWS_ELASTICSEARCH_REGION')
     AWS_ELASTICSEARCH_KEY = env('AWS_ELASTICSEARCH_KEY')
     AWS_ELASTICSEARCH_SECRET = env('AWS_ELASTICSEARCH_SECRET')
-    ES_USE_SSL = env.bool('ES_USE_SSL', True)
 
-ES_URL = env('ES5_URL')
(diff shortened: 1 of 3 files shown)


config/settings/common.py
```diff
@@ -225,15 +227,26 @@
     'datahub.search.omis.OrderSearchApp',
 ]
 
+if 'VCAP_SERVICES' in env:
```
reupen: We should be able to use something like `VCAP_SERVICES = env.json('VCAP_SERVICES', default={})` in place of these four lines.

nusnewob: updated

config/settings/common.py
```diff
@@ -225,15 +227,26 @@
...
+    VCAP_SERVICES = json.loads(env('VCAP_SERVICES'))
+else: 
+    VCAP_SERVICES = {}
+
 # Leeloo stuff
-ES_USE_AWS_AUTH = env.bool('ES_USE_AWS_AUTH', False)
+if 'elasticsearch' in VCAP_SERVICES:
+    ES_URL = VCAP_SERVICES['elasticsearch'][0]['credentials']['uri']
+    ES_USE_AWS_AUTH = False
+    ES_USE_SSL = env.bool('ES_USE_SSL', True)
```
reupen: Isn't `ES_USE_SSL` only used when `ES_USE_AWS_AUTH` is `True`?

nusnewob: `TLS` is also enabled on these instances.

reupen: I mean the code in `datahub/search/elasticsearch.py` only uses it if `ES_USE_AWS_AUTH` is True. But it doesn't look like it's needed at all, it could just use `es_host.scheme == 'https'`

https://github.com/uktrade/data-hub-leeloo/blob/3adf84fe0ebf52581450085770db242d5e25dae7/datahub/search/elasticsearch.py#L84-L116

reupen:
@nusnewob I pushed a few updates.

nusnewob:
@reupen Thanks. Anything we can do about the failing tests?

reupen:
@nusnewob It's just coverage, we can ignore it as this is just a few configuration changes (it won't block merging).

We just need to squash and rebase. I can do that if you want.

config/settings/common.py
```diff
@@ -253,7 +259,11 @@
 AV_V2_SERVICE_URL = env('AV_V2_SERVICE_URL', default=None)
 
 # CACHE / REDIS
-REDIS_BASE_URL = env('REDIS_BASE_URL', default=None)
+if 'redis' in VCAP_SERVICES:
+    REDIS_BASE_URL = VCAP_SERVICES['redis'][0]['credentials']['uri']
```
elcct: These values always are going to be present there, if not then is this enough that it will crash?

nusnewob: If the app is configured and services attached, it would always be there, otherwise, I expect crashing is normal behaviour when dependencies are not met.

reupen: Yes, it should fail. Ideally with a meaningful message, but failure would be pretty unusual so I think it's fine.

PR #1203, opened 2018-10-30
Move overseas regions fixture out from migration

marcofucci:
### Description of change

When we added the OverseasRegion table, we had to create a
migration to load the data as well.

Now that all environments have this migration and therefore the
overseas region data, we can move this yaml out from the migrations
and into a fixture that we load using the loadinitialmetadata
command.

### Checklist


marcofucci:
Discussed with @reupen and we're going to try keep the data migrations as migrations and see how it goes so I'm closing this for now.

PR #1208, opened 2018-11-05
Add metadata.Country.iso_alpha2_code field

marcofucci:
### Description of change

This:
- adds the field `metadata.Country.iso_alpha2_code`
- makes `Country` and `OverseasRegion` readonly in the admin
- removes some invalid countries from constants
- replaces some invalid countries in `test_data`
- reformats `countries.yaml` after removing the invalid countries

Still to be solved:
- the `countries.yaml` should really be loaded by the `0001` metadata migration. It's not because using `loaddata` in a migration with a different data structure breaks things
- the iso data will be loaded using a different PR

### Checklist


changelog/iso.db
```diff
@@ -0,0 +1 @@
+A new field `iso_alpha2_code` was added to the `metadata_country` table. It has not been populated yet.
```
reupen: Should add some entries for the other changes as well (read-only admin, updated initial country list etc.)

datahub/metadata/admin.py
```diff
@@ -96,12 +96,11 @@ class EditableOrderOnlyOrderedMetadataAdmin(OrderedMetadataAdmin, ViewAndChangeO
 
 
 @admin.register(models.Country)
-class CountryAdmin(admin.ModelAdmin):
+class CountryAdmin(ViewOnlyAdmin):
     """Admin for countries."""
 
     fields = ('pk', 'name', 'overseas_region', 'disabled_on')
```
reupen: Are you going to add the ISO code here later?

reupen:
@marcofucci Couple of other points.

1. Not for now, but I was wondering if we would need a 'secondary ISO codes' field since these countries won't cover all of them (mainly depending on what ISO codes are used in data coming in from other sources). This could be added when we actually need it, so just something to think about.

2. Have you mentioned this to Anthony? I see we have null for some overseas_regions, which isn't ideal (they may be countries that we don't need, though).

changelog/iso.db
marcofucci: yeah maybe one for the readonly admin, the initial country list is just a snapshot of the current live one so it does not change any behaviour.

datahub/metadata/admin.py
marcofucci: I was going to add it later as they will be empty but I might forget so, I could add it now...

marcofucci:
@reupen 
> Not for now, but I was wondering if we would need a 'secondary ISO codes' field since these countries won't cover all of them (mainly depending on what ISO codes are used in data coming in from other sources). This could be added when we actually need it, so just something to think about.

I think we should have one to one mapping, I'm waiting to get access to DnB metadata to see how different their list is from ours but I think it would be safer to include the ones they have. I diffed the list we have in prod with the FCO and there are a few extra countries in Data Hub not in the FCO list but in the ISO one already anyway.

> Have you mentioned this to Anthony? I see we have null for some overseas_regions, which isn't ideal (they may be countries that we don't need, though).

The PR is not replacing any data, the version in `countries.yaml` is just the latest from production. I replaced it because it's good to have that field populated in new environments and to get rid of the invalid countries we had before.

changelog/iso.db
reupen: The data science guys use the fixtures for some stuff, so might still be worth mentioning.

marcofucci:
@reupen Anthony and I went through a few countries, assigned the overseas regions and disable some of them. Thanks for the heads up!

fixtures/metadata/countries.yaml
```diff
@@ -133,9 +133,9 @@
 - model: metadata.country
   pk: 1a50bdb8-5d95-e211-a939-e4115bead28a
   fields:
-    disabled_on: null
-    name: Netherlands Antilles
-    overseas_region: null
+    disabled_on: '2018-11-05T17:00:00Z'
+    name: Netherlands Antilles,
```
reupen: Is this comma meant to be there?

fixtures/metadata/countries.yaml
```diff
@@ -1,744 +1,1308 @@
...
+    disabled_on: null
+    name: Saudi Arabia
+    overseas_region: c4679b44-079e-4394-8bf7-bb0881a5031d
 - model: metadata.country
   pk: 1a50bdb8-5d95-e211-a939-e4115bead28a
-  fields: {name: Netherlands Antilles}
+  fields:
+    disabled_on: '2018-11-05T17:00:00Z'
+    name: Netherlands Antilles,
+    overseas_region: 3e6809d6-89f6-4590-8458-1d0dab73ad1a}
```
reupen: And the closing brace here...

PR #1209, opened 2018-11-06
Update exclude from investment property to be not null

marcuspp:
### Description of change
Exclude from investment flow field on in the metadata from investment project stage is no longer nullable. For existing records the default has been set to False with the exception of Assign PM where the value has been set to True.

### Checklist


marcuspp:
@reupen Fixture file updated and I'm happy with the migration so will leave it as is.

changelog/investment/investment_project_stage_exclude_from_investment_flow_not_null.db
```diff
@@ -0,0 +1,2 @@
+Column ``exclude_from_investment_flow`` on ``metadata_investmentprojectstage`` table is
```
reupen: Would add an API entry as well. 

Would also be worth adding something about the existing values having been updated.

PR #1211, opened 2018-11-07
Use Celery to synchronise single objects to Elasticsearch

reupen:
### Description of change

This:

* Stops instantiating the search apps. Although on the one hand instantiating them is nice, they don't have any state and it simplifies things elsewhere as apps can simply be imported in order to use them.
* Refactors search signal receivers by only forwarding the model instance to receiver functions (as other arguments were unused).
* Changes `sync_object_async()` to take a search app instead of a search model and Django model. This is to simplify logic elsewhere.
* Use Celery instead of the thread pool for copying single objects to Elasticsearch. This is to stop the synchronisation happening in the web process. Retries are also enabled.

This does mean that Redis and Celery are now basically required for running the app. We could accept that, or expose CELERY_TASK_ALWAYS_EAGER as an environment variable to allow running the app without them. Any views?

### Checklist


marcuspp:
@reupen I would expose CELERY_TASK_ALWAYS_EAGER mainly for the testing platforms. Users will tend to expect updates to be reflected in search instantly. Now that there could be a delay although small its worth having this option available in configuration.

PR #1217, opened 2018-11-09
Load yaml in migration

marcofucci:
### Description of change

This defines a new `load_yaml_data_in_migration` function which loads the content of a yaml file into the database.

`load_yaml_data_in_migration` is similar in principle to `loaddata` but:
- it's safe to be used in migrations
- it does not change the fields that are not present in the yaml

**Motivation**
Calling `loaddata` from a data migration makes django use the latest version
of the models instead of the version at the time of that particular migration.
This causes problems e.g. adding a new field to a model which had a data migration
in the past is okay but migrating from zero fails as the model in
loaddata (the latest) has a field that did not exist at that migration time.

**Limitations**
- many-to-many fields are not supported yet
- one-to-many fields are not supported yet
- all items in the yaml have to include a pk field

This also changes all the existing data migrations to use `load_yaml_data_in_migration` instead.

### Checklist


datahub/core/migration_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        model_name = list_item['model']
+        model = apps.get_model(model_name)
+
+        data = {}
+        # Handle each field
+        for (field_name, field_value) in list_item['fields'].items():
+            field = model._meta.get_field(field_name)
+
+            # Handle M2M relations
+            if field.remote_field and isinstance(field.remote_field, models.ManyToManyRel):
```
reupen: Is this different from `field.many_to_many`? Also, should it block `field.one_to_many` as well?

datahub/core/migration_utils.py
```diff
@@ -0,0 +1,72 @@
...
+                raise NotImplementedError('Many-to-many fields not supported')
+
+            # Handle FK fields
+            if field.remote_field and isinstance(field.remote_field, models.ManyToOneRel):
```
reupen: Similar comment as above, can this use `field.many_to_one`?

datahub/core/migration_utils.py
```diff
@@ -0,0 +1,72 @@
...
+                try:
+                    value = base.deserialize_fk_value(field, field_value, using)
+                except Exception as e:
+                    raise base.DeserializationError.WithData(
+                        e,
+                        model_name,
+                        obj_pk,
+                        field_value,
+                    )
```
reupen: Would be nice to add `from e`. Also, we're more commonly using `exc` rather than `e`.

datahub/core/migration_utils.py
marcofucci: ¯\_(ツ)_/¯ got these lines directly from the Django code so I assume they are correct? It you think it’s worth I can spent some time trying to see if there are other edge cases

reupen: Maybe it was written before those attributes were added?

https://docs.djangoproject.com/en/2.1/ref/models/fields/#attributes-for-fields-with-relations

datahub/core/migration_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        assert obj_pk, 'pk field required'
+
+        # Look up the model and starting build a dict of data for it.
+        model_name = list_item['model']
+        model = apps.get_model(model_name)
+
+        data = {}
+        # Handle each field
+        for (field_name, field_value) in list_item['fields'].items():
+            field = model._meta.get_field(field_name)
```
reupen: There is quite a bit of indentation in this loop, so it would be nice to the loop body (or similar) to another function.

datahub/core/migration_utils.py
```diff
@@ -0,0 +1,72 @@
...
+                        model_name,
+                        obj_pk,
+                        field_value,
+                    )
+
+        model.objects.update_or_create(pk=obj_pk, defaults=data)
+
+
+@transaction.atomic
+def load_yaml_in_migration(apps, fixture_file):
```
reupen: I think adding 'fixture' to the function name (e.g. `load_yaml_fixture_in_migration`) would make it clearer what it does.

Also, `fixture_file` sounds like a file object – would add path to the name.

marcofucci: yeah I was thinking the same over the weekend. I wanted to change it to  `load_yaml_data_in_migration` but `load_yaml_fixture_in_migration` works even better I guess.

reupen: Yeah, `load_yaml_data_in_migration` works as well.

datahub/core/migration_utils.py
elcct: I agree with Reupen.

datahub/core/migration_utils.py
```diff
@@ -0,0 +1,80 @@
...
+                ) from exc
+    return data
+
+
+def _load_data_in_migration(apps, object_list, using=DEFAULT_DB_ALIAS):
+    for list_item in object_list:
+        obj_pk = list_item.get('pk')
+        assert obj_pk, 'pk field required'
+
+        model_name = list_item['model']
```
reupen: It's the model label rather than the name, isn't it?

PR #1226, opened 2018-11-13
Load ISO codes for countries

marcofucci:
### Description of change

This uses a django migration to load ISO codes into the database.
The overall country list was exported from production.

This also gets rid of `countries.yaml` as initial fixture loaded via the `loadinitialmetadata` and replaces it with a data migration via the first metadata migration.

### Checklist


reupen:
Do we need these exposed via the metadata endpoint?

marcofucci:
> Do we need these exposed via the metadata endpoint?
Not yet, maybe never.

PR #1233, opened 2018-11-16
Make `classification` readonly in company API

marcofucci:
### Description of change

This makes the field `classification` read-only in all company endpoints.

It also refactors the tests in `TestAddCompany` and `TestUpdateCompany` so that they are easier to manage in the future when companies are going go change.

### Checklist

### DIFF
--- a/datahub/company/test/test_company_views.py
+++ b/datahub/company/test/test_company_views.py
@@ -19,7 +19,7 @@ from datahub.company.test.factories import (
     CompanyFactory,
     DuplicateCompanyFactory,
 )
-from datahub.core.constants import Country, HeadquarterType, Sector, UKRegion
+from datahub.core.constants import Country, HeadquarterType, UKRegion
 from datahub.core.reversion import EXCLUDED_BASE_MODEL_FIELDS
 from datahub.core.test_utils import (
     APITestMixin,
@@ -28,7 +28,7 @@ from datahub.core.test_utils import (
     random_obj_for_model,
     random_obj_for_queryset,
 )
-from datahub.metadata.models import CompanyClassification
+from datahub.metadata.models import CompanyClassification, Sector
 from datahub.metadata.test.factories import TeamFactory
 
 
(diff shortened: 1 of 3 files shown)


datahub/company/test/test_company_views.py
```diff
@@ -497,82 +489,57 @@ def test_update_read_only_fields(self):
...
-            'registered_address_1': 'Hello st.',
-            'registered_address_town': 'Fooland',
-            'registered_address_country_id': Country.united_states.value.id,
-            f'{field}_id': value,
-        }
-        company = CompanyFactory(**creation_data)
+    def test_validation_error(self, data, expected_error):
+        """Test validation scenarios."""
+        company = CompanyFactory(
+            sector_id=random_obj_for_model(Sector).id,
```
marcuspp: Why does this test need a random sector? If a different sector could produce a different result then this is not a good test. As this results in the tests being inconsistent, sometimes working and then sometimes failing.

marcofucci: A different sector should not produce a different result so it's a red flag if it does.

marcuspp: It still doesn't make a good test IMO.

marcofucci: I completely agree in case of constants or values defined in the code. However, at the moment the sectors can be changed in the django admin so the code should not give any importance to this value and treat it as a random piece of data the same way as we do with any other string or field in all other tests. 

I think we should go even further and get rid of the couple of constants in `datahub.core.constants.Sector` which are just confusing and don't add any value.

If you think it's better I can replace this with a `SectorFactory` which creates a new random sector every time.

marcuspp: I agree there shouldn't be any importance given to what sector is used but the tests should run the same way each time, so I'd just pick a sector (any sector) and remove any randomness from the test.

elcct: I would side with @marcofucci on this one as in the past using random values helped us detect issues that wouldn't have been detected if we had used same value every time.

marcuspp: @elcct if thats the case then that shows the test suite was missing tests and more conditions should of been added in the first place. Adding randomness isn't the solution, its something I've had issues with in the past but am happy to go with what the consensus is.

marcofucci: Well, it turns out that the `CompanyFactory` already sets the sector so I don't actually need this line at all and I got rid of it. This only postpones the discussion we've been having but it works for now.

elcct: @marcuspp valid point!

PR #1235, opened 2018-11-19
Add country investment originates from to investment projects

marcuspp:
### Description of change
This adds the ability to add the `country investment originates from` to an investment project. This is because we are trying to remove the initial need to associate a company with a project. Search is updated to store the country in the index and the investor_company_country filter is updated to also check `country investment originates from` so when a user filters for Country of origin both are checked. 

### Questions
- Does the search index need to be rebuilt because I have added a new searchable field?
- Are the search tests very slow for you when trying to run locally? fixed :+1:

### Checklist


datahub/investment/models.py
```diff
@@ -209,21 +213,21 @@ def project_code(self):
             project_num = self.investmentprojectcode.id
             return f'DHP-{project_num:08d}'
         except ObjectDoesNotExist:
-            return None
+            return
```
marcofucci: Can we keep the `return None`? Having `return` is not always intuitive even if it does the same.

datahub/investment/serializers.py
```diff
@@ -180,6 +181,9 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     investment_type = NestedRelatedField(meta_models.InvestmentType)
     stage = NestedRelatedField(meta_models.InvestmentProjectStage, required=False)
     country_lost_to = NestedRelatedField(meta_models.Country, required=False, allow_null=True)
+    country_investment_originates_from = NestedRelatedField(
+        meta_models.Country, required=False, allow_null=True,
```
marcofucci: We are trying to keep each argument in its own line so that future diffs are easier to read.

datahub/investment/test/test_models.py
```diff
@@ -76,16 +76,13 @@ def test_investor_company_country_none():
...
 
 def test_investor_company_country_valid():
-    """
-    Tests client_relationship_manager_team for a project with a client relationship
-    manager.
-    """
+    """Tests investor company country property."""
     project = InvestmentProjectFactory()
-    assert project.investor_company_country
+    assert str(project.investor_company_country) == constants.Country.united_kingdom.value.name
```
marcofucci: It's not clear why `project.investor_company_country` is UK, can we change this to
```
investor_company = CompanyFactory(registered_address_country_id=constants.Country.united_kingdom.value.id)
project = InvestmentProjectFactory(investor_company=investor_company)

assert str(project.investor_company_country.id) == constants.Country.united_kingdom.value.id
```

or option 2 (slightly worse IMHO):
```
project = InvestmentProjectFactory()
```
project = InvestmentProjectFactory()
assert str(project.investor_company_country.id) == project.investor_company.registered_address_country.id

datahub/search/investment/test/test_views.py
```diff
@@ -62,6 +62,7 @@ def setup_data(setup_es):
             description='investmentproject2',
             estimated_land_date=datetime.date(2057, 6, 13),
             actual_land_date=datetime.date(2047, 8, 13),
+            country_investment_originates_from_id=constants.Country.andorra.value.id,
```
marcofucci: We're going to get rid of unused country constants, would you mind using these ones so that I will have less pain merging that PR 😬 ? https://github.com/uktrade/data-hub-leeloo/pull/1229/files

marcuspp:
@marcofucci I had all the same questions and came to the conclusion that they should be left separate for this PR at least. 

I think the next stage would be that `investor_company_country` is deprecated and we only have `country_investment_originates_from` going forward. 

So to do that (as you have said):
 - Populating  `country_investment_originates_from` using the investor_company_country for all existing records (could be done now)
- Populating `country_investment_originates_from` when a company is added (could be done now)
- Update export and remove old field and replace with new
- Update FE to use the renamed filter

So you could consider this PR stage 1

marcofucci:
@marcuspp makes perfect sense, thanks for explaining that.

To answer your ES question, adding new fields is usually okay but it would be good to run  [migrate_es](https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Elasticsearch%20migrations.md) at some point (maybe after getting rid of the `investor_company_country`?).

datahub/search/investment/models.py
```diff
@@ -74,7 +64,8 @@ class InvestmentProject(BaseESModel):
     )
     client_requirements = fields.TextWithKeyword()
     comments = fields.EnglishText()
-    country_lost_to = _country_lost_to_mapping()
+    country_investment_originates_from = fields.nested_id_name_field()
+    country_lost_to = fields.nested_id_name_field()
```
marcofucci: I'm not sure what exactly the consequences of getting rid of `def _country_lost_to_mapping():` are. The mapping was not initially defined so ES created its own and then we added one to match the one automatically created. 
If we get rid of this here we might have to migrate the mapping as well. 

Can you restore the old version and then we coordinate the work for that country_lost_to? Maybe when Reupen is back as he worked on the migrate_es and would be able to help us quickly in case anything happens.

marcuspp: To me it just looked like all the other fields, id and name and the small bit of testing i did, didn't surface any issues. If the index has been rebuilt recently then maybe this special case for `country lost to` is no longer needed. Lets wait for Reupen and i'll revert for now.

PR #1238, opened 2018-11-20
Add `one_list_group_tier` field to GET company endpoint

marcofucci:
### Description of change

This adds the field ~`onelist_group_tier`~ `one_list_group_tier` to the `GET /v3/company/<uuid:pk>` and `GET /v3/company` endpoints representing the One List Tier for the group, inherited from the Global Headquarters.

I've also refactored the tests to build the company used in the tests instead of building parts of it. This adds a few lambdas which don't look very pythonic but I think they make it clear what the intention is. Feel free to tell me if this looks too weird and I'll revert back to params with parts of the company instead.

The search does not return ~`onelist_group_tier`~ `one_list_group_tier` as it would have to be updated in cascade every time a Global HQ is updated and I want to avoid it unless necessary. 

### Checklist

### DIFF
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -242,6 +242,12 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
             return self.global_headquarters
         return self
 
+    def get_one_list_group_tier(self):
+        """
+        :returns: the One List Tier of the group this company is part of.
+        """
+        return self.get_group_global_headquarters().classification
+
     def get_one_list_group_core_team(self):
         """
         :returns: the One List Core Team for the group that this company is part of
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -9,8 +9,13 @@ from rest_framework import serializers
 
 from datahub.company.constants import BusinessTypeConstant
 from datahub.company.models import (
-    Advisor, CompaniesHouseCompany, Company, CompanyPermission,
-    Contact, ContactPermission, ExportExperienceCategory,
+    Advisor,
+    CompaniesHouseCompany,
+    Company,
+    CompanyPermission,
+    Contact,
+    ContactPermission,
+    ExportExperienceCategory,
 )
 from datahub.company.validators import (
     has_no_invalid_company_number_characters,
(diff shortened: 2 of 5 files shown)


datahub/company/serializers.py
```diff
@@ -273,6 +280,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         meta_models.BusinessType, required=False, allow_null=True,
     )
     classification = NestedRelatedField(meta_models.CompanyClassification, read_only=True)
+    onelist_group_tier = serializers.SerializerMethodField()
```
marcuspp: Any plans to update `one_list_account_owner`? As that has one list as two words not one as you are using it here, which is inconsistent.

datahub/company/test/test_company_views.py
```diff
@@ -484,14 +540,18 @@ def test_cannot_update_read_only_fields(self):
         )
 
         url = reverse('api-v3:company:item', kwargs={'pk': company.pk})
+        different_onelist_tier = random_obj_for_queryset(
```
marcuspp: Just a comment. See the other PR as regards my views on using randomness like this in a test. To me it leaves the question should more test cases be added if its possible a different classification could cause a different result.

marcofucci: The randomness in these tests is not because different classifications could cause different results (they don't) but just to pick up one value. I could use `CompanyClassification.objects.first()` so that it's always the same if you think it's better?

marcuspp: Yes please.

marcofucci: My new version is even nicer, thanks @marcuspp .

datahub/company/models/company.py
```diff
@@ -242,6 +242,13 @@ def get_group_global_headquarters(self):
...
         return self
 
+    def get_one_list_group_tier(self):
+        """
+        :returns: the One List Tier for the group that this company is part of.
+        """
+        group_global_headquarters = self.get_group_global_headquarters()
+        return group_global_headquarters.classification
+
     def get_onelist_group_core_team(self):
```
marcuspp: Missed one

PR #1246, opened 2018-11-23
Only call load_constants_to_database once

marcofucci:
### Description of change

The post_migrate signal receiver `company_business_type_post_migrate` is called after each individual django app is migrated.

*Before:*
`load_constants_to_database` was called multiple times as well.

*After:*
`load_constants_to_database` is only called once after migrating the `metadata` django app.

### Checklist

### DIFF
--- a/datahub/company/signals.py
+++ b/datahub/company/signals.py
@@ -10,7 +10,11 @@ from datahub.metadata.models import BusinessType
 logger = logging.getLogger(__name__)
 
 
-@receiver(post_migrate, dispatch_uid='company_business_type_post_migrate')
+@receiver(
+    post_migrate,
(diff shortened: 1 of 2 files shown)


datahub/company/signals.py
```diff
@@ -17,4 +17,5 @@ def company_business_type_post_migrate(sender, **kwargs):
 
     Any new business types are created, and any existing ones are updated if the label has changed.
     """
-    load_constants_to_database(BusinessTypeConstant, BusinessType)
+    if BusinessType._meta.app_config == sender:
```
reupen: What about specifying the sender in @receiver() instead?

marcofucci: 👍  of course, forgot about that!

PR #1247, opened 2018-11-23
One List Group Global Account Manager

marcofucci:
### Description of change

This add the field ``one_list_group_global_account_manager`` to GET company endpoint as it's going to be often required and we don't want the force the frontend to make another API call to get this data from the one list core team endpoint.

### Checklist


datahub/company/models/company.py
```diff
@@ -289,6 +289,14 @@ def get_one_list_group_core_team(self):
         )
         return core_team
 
+    def get_one_list_group_global_account_manager(self):
```
reupen: I think `get_group_one_list_global_account_manager` would be more consistent with what the docstring says? But I can see this is also consistent with some existing field and function names.

marcofucci: yeah this a weird one and I don't even like the very long name but I'm trying to keep things consistent

PR #1249, opened 2018-11-23
Add level_of_involvement_simplified investment project filter

elcct:
### Description of change

This adds ~``involvement``~ ``level_of_involvement_simplified`` read-only computed field that is derived from ``level_of_involvement``. We need a way to filter the investment projects so for example only projects with any level of involvement are visible. ~``involvement``~ ``level_of_involvement_simplified`` field maps `level_of_involvement` to three possible values - `unspecified` if it is `None`, `no_involvement` if equals `no_involvement` constant and `some_involvement` for anything else.
Filtering by `level_of_involvement` wouldn't be practical due to number of possible involvement levels.
~I am not too happy with the choice of the field name, but couldn't think of any better.~

### Checklist

### DIFF
--- a/datahub/investment/admin.py
+++ b/datahub/investment/admin.py
@@ -19,7 +19,7 @@ from datahub.investment.models import (
     Involvement,
     SpecificProgramme,
 )
-from datahub.metadata.admin import DisableableMetadataAdmin
+from datahub.metadata.admin import DisableableMetadataAdmin, ReadOnlyMetadataAdmin
 
 
 @admin.register(InvestmentProject)
@@ -89,11 +89,13 @@ class InvestmentProjectTeamMemberAdmin(VersionAdmin):
     )
 
 
+admin.site.register(Involvement, ReadOnlyMetadataAdmin)
+
+
 admin.site.register(
     (
         InvestmentDeliveryPartner,
         InvestorType,
-        Involvement,
         SpecificProgramme,
     ),
     DisableableMetadataAdmin,
--- a/datahub/investment/models.py
+++ b/datahub/investment/models.py
@@ -18,6 +18,7 @@ from datahub.core.models import (
     BaseModel,
 )
 from datahub.core.utils import get_front_end_url, StrEnum
+from datahub.investment import constants
 
 MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
 
@@ -87,6 +88,12 @@ class IProjectAbstract(models.Model):
         ('lost', 'Lost'),
         ('abandoned', 'Abandoned'),
         ('won', 'Won'),
+    )
+
+    INVOLVEMENT = Choices(
+        ('unspecified', 'Unspecified'),
+        ('not_involved', 'Not involved'),
+        ('involved', 'Involved'),
     )
 
     name = models.CharField(max_length=MAX_LENGTH)
@@ -228,6 +235,18 @@ class IProjectAbstract(models.Model):
         if self.client_relationship_manager:
             return self.client_relationship_manager.dit_team
         return None
+
+    @property
+    def level_of_involvement_simplified(self):
+        """Returns simplified level of involvement for the Investment Project."""
+        if self.level_of_involvement_id is None:
+            return self.INVOLVEMENT.unspecified
+
+        not_involved = uuid.UUID(constants.Involvement.no_involvement.value.id)
+        if self.level_of_involvement_id == not_involved:
+            return self.INVOLVEMENT.not_involved
+
+        return self.INVOLVEMENT.involved
 
 
 class IProjectValueAbstract(models.Model):
--- a/datahub/investment/serializers.py
+++ b/datahub/investment/serializers.py
@@ -79,6 +79,7 @@ CORE_FIELDS = (
     'modified_on',
     'comments',
     'country_investment_originates_from',
+    'level_of_involvement_simplified',
 )
 
 VALUE_FIELDS = (
--- a/datahub/investment/test/test_views.py
+++ b/datahub/investment/test/test_views.py
@@ -71,6 +71,7 @@ class TestListView(APITestMixin):
         response_data = response.json()
         assert response_data['count'] == 1
         assert response_data['results'][0]['id'] == str(project.id)
+        assert response_data['results'][0]['level_of_involvement_simplified'] == 'not_involved'
         assert response_data['results'][0].keys() == {
             'id',
             'incomplete_fields',
@@ -168,6 +169,7 @@ class TestListView(APITestMixin):
             'project_arrived_in_triage_on',
             'proposal_deadline',
             'stage_log',
+            'level_of_involvement_simplified',
         }
 
     def test_list_is_sorted_by_created_on_desc(self):
--- a/datahub/search/investment/models.py
+++ b/datahub/search/investment/models.py
@@ -138,6 +138,7 @@ class InvestmentProject(BaseESModel):
     uk_company_decided = Boolean()
     uk_region_locations = fields.nested_id_name_field()
     will_new_jobs_last_two_years = Boolean()
+    level_of_involvement_simplified = Keyword()
 
     MAPPINGS = {
         'actual_uk_regions': lambda col: [
--- a/datahub/search/investment/serializers.py
+++ b/datahub/search/investment/serializers.py
@@ -1,6 +1,7 @@
 from rest_framework import serializers
 
 from datahub.core.serializers import RelaxedDateTimeField
+from datahub.investment.models import InvestmentProject
 from datahub.search.serializers import (
     SearchSerializer,
     SingleOrListField,
@@ -28,6 +29,10 @@ class SearchInvestmentProjectSerializer(SearchSerializer):
     stage = SingleOrListField(child=StringUUIDField(), required=False)
     status = SingleOrListField(child=serializers.CharField(), required=False)
     uk_region_location = SingleOrListField(child=StringUUIDField(), required=False)
+    level_of_involvement_simplified = SingleOrListField(
+        child=serializers.ChoiceField(choices=InvestmentProject.INVOLVEMENT),
+        required=False,
+    )
 
     SORT_BY_FIELDS = (
         'actual_land_date',
--- a/datahub/search/investment/test/test_views.py
+++ b/datahub/search/investment/test/test_views.py
@@ -25,6 +25,7 @@ from datahub.core.test_utils import (
     get_attr_or_none,
     random_obj_for_queryset,
 )
+from datahub.investment.constants import Involvement
 from datahub.investment.models import InvestmentProject, InvestmentProjectPermission
 from datahub.investment.test.factories import (
     InvestmentProjectFactory,
@@ -56,6 +57,7 @@ def setup_data(setup_es):
                 constants.UKRegion.east_midlands.value.id,
                 constants.UKRegion.isle_of_man.value.id,
             ],
+            level_of_involvement_id=Involvement.hq_and_post_only.value.id,
         ),
         InvestmentProjectFactory(
             name='delayed project',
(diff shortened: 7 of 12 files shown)


datahub/search/investment/models.py
```diff
@@ -138,6 +138,7 @@ class InvestmentProject(BaseESModel):
     uk_company_decided = Boolean()
     uk_region_locations = fields.nested_id_name_field()
     will_new_jobs_last_two_years = Boolean()
+    involvement = fields.SortableCaseInsensitiveKeywordText()
```
reupen: Won't a plain Keyword do? It doesn't seem to be a sort-by field and as it contains identifiers it doesn't seem to need to be case insensitive.

elcct: That's true. I have updated it with `Keyword`. Thanks!

datahub/search/investment/test/test_views.py
```diff
@@ -462,6 +467,90 @@ def test_search_investment_project_investment_origin(
...
+        expected_results,
+    ):
+        """Tests the involvement filter."""
+        url = reverse('api-v3:search:investment_project')
+        response = self.api_client.post(url, query)
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == len(expected_results)
+        results = response.data['results']
+        assert Counter(result['name'] for result in results) == Counter(expected_results)
```
marcofucci: Can you remind me why we are using `Counter` here?

reupen: Generally I have used `Counter` when the results are in an undetermined order (`set()` will remove duplicates which could hide bugs).

datahub/investment/models.py
```diff
@@ -229,6 +236,18 @@ def client_relationship_manager_team(self):
             return self.client_relationship_manager.dit_team
         return None
 
+    @property
+    def involvement(self):
+        """Tells what level of involvement is in the Investment Project."""
+        if self.level_of_involvement_id is None:
+            return self.INVOLVEMENT.unspecified
+
+        no_involvement = uuid.UUID(constants.Involvement.no_involvement.value.id)
```
marcofucci: The `Involvement` values are not managed by the code but by users, only loaded initially using the loadinitialmetadata and editable via the django admin. This checks for specific values which breaks the logic as this mapping is not guaranteed.

datahub/investment/models.py
```diff
@@ -229,6 +236,18 @@ def client_relationship_manager_team(self):
             return self.client_relationship_manager.dit_team
         return None
 
+    @property
+    def involvement(self):
```
marcofucci: I'm not sure it's worth having an `involvement` method on the model for now as it's only required by the search and here would just confuse people because of the presence of a `level_of_involvement` field as well. Maybe move the logic to the search app instead for now?

datahub/search/investment/test/test_views.py
marcofucci: 👍 it would be more readable to add explicit ordering and use a list IMO but not a big deal if it's problematic.

datahub/investment/models.py
elcct: This is a good point. `Involvement` values are not meant to be changed by the users, so I would remove it from the admin. This would ensure that this specific value will be guaranteed.

datahub/investment/models.py
elcct: I think this is the simplest way to have basic filter for level of involvement values - we only need to know three states. There is also a problem that this value not always exists and Elasticsearch doesn't store null values - so we would need two type of filters - one to check if value exist if we wanted to include investment projects that have level of involvement null and then if level of involvement is equal to no involvement and then finally to include all other values.
I could instead add a more elaborate comment to explain why this property exists to minimise the confusion?

reupen: If we don't want it on the model, a mapping on the ES model might do the trick.

elcct: If we are going to filter by it, then good practice it is to show the value being filtered, so that this is also returned by the API. If we remove this then the front end probably would have to derive this value from `level_of_involvement` as well and it will needlessly increase complexity or am I missing something?

marcuspp: I'm okay with this being on the db model, it simplifies how the field is populated on the search document. But take the point if it's just for search then maybe move it to the ES model.

elcct: `level_of_involvement` value is being shown in the Investment Project details (if exists), but we might want to show the simplified version in the collection view where we are going to use this simplified filter. If we are not going to include that, then the front end would have to implement additional logic to determine what value for `level_of_involvement` amounts to `some involvement` etc.

datahub/investment/serializers.py
```diff
@@ -79,6 +79,7 @@
     'modified_on',
     'comments',
     'country_investment_originates_from',
+    'involvement',
```
marcofucci: If could be good to assert the content instead of just the keys, even just to document the response body of the endpoint.

datahub/investment/test/test_views.py
```diff
@@ -168,6 +168,7 @@ def test_list_projects_success(self):
             'project_arrived_in_triage_on',
             'proposal_deadline',
             'stage_log',
+            'involvement',
```
marcofucci: It could be good to assert the content instead of just the keys, even just to document the response body of the endpoint.

datahub/investment/models.py
marcofucci: If there's a need for showing the simplified version in the list instead of the actual value maybe it makes more sense, it seems a bit weird though to have it as a field IMHO but I might be the only one thinking this. Changing to name might help? Even `level_of_involvement_simplified` would be clearer but I'm sure we can find a better one.

Also, couldn't the frontend just map the `level_of_involvement` and have a simplified frontend logic that in background translates to a list of ids for some instead? This way you don't need to add any backend change and it's clearer.
It's just a user experience shortcut after all.

elcct: I am not sure about `involvement` name so I am happy to change it to something else - `level_of_involvement_simplified` works for me. 
I was considering having it only on the front end, but there is a problem of some investment projects not having this value set - that means it doesn't exist in Elasticsearch, so that would require two types of filter (to check if value exist and to check if value is equal to) - this will complicate the front end, also when a user wants to check `some_involvement`, that would mean checking 23 ids and then hiding this from the filter list. Not sure if such special case is worth implementing on the front end?

marcofucci: > that means it doesn't exist in Elasticsearch, so that would require two types of filter 

oh yeah I forgot about that :-/

> user wants to check some_involvement, that would mean checking 23 ids and then hiding this from the filter list. 

I meant the frontend has just a checkbox with "some involvement" and the nodejs logic translates that into 23 ids, seems like a very trivial piece of nodejs code. We could even have a special shortcut value like 'some' to pass it to the backend and then the backend does the translation. Some google API have similar features, not sure if we want that.

However, renaming it works for me. What do you guys think in general @reupen @marcuspp ?

datahub/investment/models.py
elcct: I have removed this from the Django admin.

datahub/investment/models.py
elcct: I have renamed `involvement` to `level_of_involvement_simplified`.

datahub/investment/admin.py
```diff
@@ -93,7 +92,6 @@ class InvestmentProjectTeamMemberAdmin(VersionAdmin):
     (
         InvestmentDeliveryPartner,
         InvestorType,
-        Involvement,
```
reupen: Why don't we make it view-only instead of removing it completely?

elcct: To be fair there is no reason why this should be hidden so I changed it to be view-only. Thanks!

datahub/search/investment/serializers.py
```diff
@@ -28,6 +28,10 @@ class SearchInvestmentProjectSerializer(SearchSerializer):
     stage = SingleOrListField(child=StringUUIDField(), required=False)
     status = SingleOrListField(child=serializers.CharField(), required=False)
     uk_region_location = SingleOrListField(child=StringUUIDField(), required=False)
+    level_of_involvement_simplified = SingleOrListField(
+        child=serializers.CharField(),
```
reupen: Shouldn't this have choices set as well?

datahub/investment/models.py
```diff
@@ -89,6 +90,12 @@ class Meta:
         ('won', 'Won'),
     )
 
+    INVOLVEMENT = Choices(
+        ('unspecified', 'Unspecified'),
+        ('no_involvement', 'No involvement'),
+        ('some_involvement', 'Some involvement'),
```
reupen: Did 'Some involvement' come from somewhere else? It just sounds a bit odd as it makes it sound like there is something that would come after it (e.g. 'Full involvement') when there isn't.

elcct: I could change it to just `Involvement` or `Any involvement`?

datahub/search/investment/serializers.py
elcct: I have added choices to the field and an additional test. This should make an improvement. Thanks!

datahub/investment/models.py
reupen: I think `not_involved` and `involved` would make sense to me (or None/True/False, but you would have the Elasticsearch problem I guess?)

elcct: `not_involved` and `involved` work for me too. I'll change.

elcct: I have changed to `unspecified`, `not_involved` and `involved`. Thanks!

datahub/search/investment/serializers.py
```diff
@@ -28,6 +28,16 @@ class SearchInvestmentProjectSerializer(SearchSerializer):
     stage = SingleOrListField(child=StringUUIDField(), required=False)
     status = SingleOrListField(child=serializers.CharField(), required=False)
     uk_region_location = SingleOrListField(child=StringUUIDField(), required=False)
+    level_of_involvement_simplified = SingleOrListField(
+        child=serializers.ChoiceField(
+            choices=(
+                ('unspecified', 'Unspecified'),
+                ('not_involved', 'Not involved'),
+                ('involved', 'Involved'),
```
marcofucci: Shouldn’t these come from the choices constants on the model?

datahub/search/investment/test/test_views.py
```diff
@@ -462,6 +467,115 @@ def test_search_investment_project_investment_origin(
...
+    def test_search_involvement_incorrect_filter_json(
+        self,
+        setup_data,
+        query,
+    ):
+        """Tests that involvement filter won't work with incorrect filter value."""
+        url = reverse('api-v3:search:investment_project')
+        response = self.api_client.post(url, query)
+
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
```
marcofucci: Can we assert the error message as well?

datahub/search/investment/serializers.py
elcct: I thought we keep search separate from the model?

datahub/search/investment/test/test_views.py
elcct: I have updated the test with the assert of the error message. Thanks!

datahub/investment/models.py
marcuspp: I'm fine with this implementation as I think it keeps things nice and simple and renaming to `level_of_involvement_simplified` makes the field more descriptive which is good.

datahub/search/investment/serializers.py
elcct: We didn't do that for `status`, that's why I didn't think about adding this. I have changed it now so it gets choices from the model. I could do the same for status (and add tests) if we now use the data from the model in separate PR.

marcofucci: Yeah we missed choices in a few more places, didn't we? Your change looks good though 👍

PR #1256, opened 2018-11-27
DUNS number

marcofucci:
### Description of change

This adds the ``duns_number`` field to company model as a 9 digit number with leading zeros.
The field is a charfield instead of an integerfield because the leading zeros are required when integrating with D&B.

This adds the read-only field ``duns_number`` to the response body of all company endpoints.

It's currently not possible to search for a duns_number because it's not needed and the frontend does not expose this field yet.

For similar reasons, the company export functionality does not include the duns_number field.

### Checklist


reupen:
Don't we need a news fragment for the database as well?

PR #1258, opened 2018-11-28
Company autocomplete search support

marcuspp:
### Description of change
To support autocomplete or type-ahead for companies I've utilised the elasticsearch completion suggester and added it to the company elastic search model. It seems to work quite nicely (tested locally with 100K of companies) but am happy to rework any aspect of it as I am aware this sort of field uses more server memory etc. The completion suggestions are just what I think will work best am happy to add and remove values based on anyone else's recommendations. Completion suggestions also supports fuzzy matching which for the moment I have not used. I think the index will need to be rebuilt as the the mapping has changed.

### How to test
- Reindex some companies
- Go to http://localhost:8000/v3/search/company/quick?q=mar

I've found adding https://pypi.org/project/django-elasticsearch-debug-toolbar/ to the debug tool bar helped me work out what queries were being executed ie a double execution of the query. (For some reason when you click to view the json in chrome the link doesn't work so Ive had to inspect and remove the display class)

### Checklist

### DIFF
--- a/changelog/company/company_autocomplete_support.api
+++ b/changelog/company/company_autocomplete_support.api
@@ -1,0 +1,2 @@
+New endpoint ``GET /v3/search/company/autocomplete`` which supports a query argument of ``term`` that will
+return the ``id``, ``name`` and ``trading_name`` of any company matching the search query.
--- a/datahub/search/company/models.py
+++ b/datahub/search/company/models.py
@@ -1,6 +1,6 @@
 from operator import attrgetter
 
-from elasticsearch_dsl import Boolean, Date, Keyword, Text
+from elasticsearch_dsl import Boolean, Completion, Date, Keyword, Text
 
 from datahub.search import dict_utils
 from datahub.search import fields
@@ -8,6 +8,40 @@ from datahub.search.models import BaseESModel
 
 
 DOC_TYPE = 'company'
+
+
+def get_suggestions(db_company):
+    """
+    A list of fields used by the completion suggester to
+    find a record when using an autocomplete search.
+
+    https://www.elastic.co/guide/en/elasticsearch/
+    reference/current/search-suggesters-completion.html
+
+    Both the name and trading name of a company are added in full
+    and each word within the names are individually added.
+    Adding the full names should improve the precision of the search and
+    return the company the user is looking for sooner.
+    The parts of the names are added so when searching the order
+    of the search terms that are entered becomes irrelevant.
+
+    Optional weighting could be added here to boost particular suggestions.
+    See above link.
+    """
+    if db_company.archived:
+        return []
+
+    company_name = db_company.name
+    alias = db_company.alias or ''
+
+    data = [
+        *company_name.split(' '),
+        *alias.split(' '),
+        company_name,
+        alias,
+    ]
+
+    return list(filter(None, set(data)))
 
 
 class Company(BaseESModel):
@@ -74,9 +108,11 @@ class Company(BaseESModel):
     vat_number = Keyword(index=False)
     duns_number = Keyword()
     website = Text()
+    suggest = Completion()
 
     COMPUTED_MAPPINGS = {
         'trading_name': attrgetter('alias'),
+        'suggest': get_suggestions,
     }
 
     MAPPINGS = {
--- a/datahub/search/company/serializers.py
+++ b/datahub/search/company/serializers.py
@@ -1,6 +1,12 @@
 from rest_framework import serializers
 
-from datahub.search.serializers import SearchSerializer, SingleOrListField, StringUUIDField
+from datahub.search.serializers import (
+    AutocompleteSearchSerializer,
+    IdNameSerializer,
+    SearchSerializer,
+    SingleOrListField,
+    StringUUIDField,
+)
 
 
 class SearchCompanySerializer(SearchSerializer):
@@ -45,3 +51,22 @@ class SearchCompanySerializer(SearchSerializer):
         'uk_based',
         'uk_region.name',
     )
+
+
+class AutocompleteSearchCompanySerializer(AutocompleteSearchSerializer):
+    """Autocomplete search serializer for companies."""
+
+    name = serializers.CharField()
+    trading_name = serializers.CharField()
+    trading_address_1 = serializers.CharField()
+    trading_address_2 = serializers.CharField()
+    trading_address_town = serializers.CharField()
+    trading_address_county = serializers.CharField()
+    trading_address_country = IdNameSerializer()
+    trading_address_postcode = serializers.CharField()
+    registered_address_1 = serializers.CharField()
+    registered_address_2 = serializers.CharField()
+    registered_address_town = serializers.CharField()
+    registered_address_county = serializers.CharField()
+    registered_address_country = IdNameSerializer()
+    registered_address_postcode = serializers.CharField()
--- a/datahub/search/company/test/test_models.py
+++ b/datahub/search/company/test/test_models.py
@@ -1,67 +1,104 @@
+from collections import Counter
+
 import pytest
 
 from datahub.company.test.factories import CompanyFactory
-from datahub.search.company.models import Company as ESCompany
+from datahub.search.company.models import Company as ESCompany, get_suggestions
 
 pytestmark = pytest.mark.django_db
 
 
-def test_company_dbmodel_to_dict(setup_es):
-    """Tests conversion of db model to dict."""
-    company = CompanyFactory()
+class TestCompanyElasticModel:
+    """Test for the company elasticsearch model"""
 
-    result = ESCompany.db_object_to_dict(company)
+    def test_company_dbmodel_to_dict(self, setup_es):
+        """Tests conversion of db model to dict."""
+        company = CompanyFactory()
 
-    keys = {
-        'archived',
-        'archived_by',
-        'archived_on',
-        'archived_reason',
-        'business_type',
-        'companies_house_data',
-        'company_number',
-        'contacts',
-        'created_on',
-        'description',
-        'employee_range',
-        'export_experience_category',
-        'export_to_countries',
-        'future_interest_countries',
-        'headquarter_type',
-        'id',
-        'modified_on',
-        'name',
-        'global_headquarters',
-        'reference_code',
-        'registered_address_1',
-        'registered_address_2',
-        'registered_address_country',
-        'registered_address_county',
-        'registered_address_postcode',
-        'registered_address_town',
-        'sector',
-        'trading_address_1',
-        'trading_address_2',
-        'trading_address_country',
-        'trading_address_county',
-        'trading_address_postcode',
-        'trading_address_town',
-        'trading_name',
-        'turnover_range',
-        'uk_based',
-        'uk_region',
-        'vat_number',
-        'duns_number',
-        'website',
-    }
+        result = ESCompany.db_object_to_dict(company)
 
-    assert set(result.keys()) == keys
+        keys = {
+            'archived',
+            'archived_by',
+            'archived_on',
+            'archived_reason',
+            'business_type',
+            'companies_house_data',
+            'company_number',
+            'contacts',
+            'created_on',
+            'description',
+            'employee_range',
+            'export_experience_category',
+            'export_to_countries',
+            'future_interest_countries',
+            'headquarter_type',
+            'id',
+            'modified_on',
+            'name',
+            'global_headquarters',
+            'reference_code',
+            'registered_address_1',
+            'registered_address_2',
+            'registered_address_country',
+            'registered_address_county',
+            'registered_address_postcode',
+            'registered_address_town',
+            'sector',
+            'suggest',
+            'trading_address_1',
+            'trading_address_2',
+            'trading_address_country',
+            'trading_address_county',
+            'trading_address_postcode',
+            'trading_address_town',
+            'trading_name',
+            'turnover_range',
+            'uk_based',
+            'uk_region',
+            'vat_number',
+            'duns_number',
+            'website',
+        }
 
+        assert set(result.keys()) == keys
 
-def test_company_dbmodels_to_es_documents(setup_es):
-    """Tests conversion of db models to Elasticsearch documents."""
-    companies = CompanyFactory.create_batch(2)
+    def test_company_dbmodels_to_es_documents(self, setup_es):
+        """Tests conversion of db models to Elasticsearch documents."""
+        companies = CompanyFactory.create_batch(2)
 
-    result = ESCompany.db_objects_to_es_documents(companies)
+        result = ESCompany.db_objects_to_es_documents(companies)
 
-    assert len(list(result)) == len(companies)
+        assert len(list(result)) == len(companies)
+
+    @pytest.mark.parametrize(
+        'name,trading_name,archived,expected_suggestions',
+        (
+            (
+                'Hello Hello uk',
+                'Good Hello us',
+                False,
+                ['Good Hello us', 'uk', 'us', 'Hello', 'Hello Hello uk', 'Good'],
+            ),
+            (
+                'Hello      gb',
+                None,
+                False,
+                ['Hello', 'gb', 'Hello      gb'],
+            ),
+            (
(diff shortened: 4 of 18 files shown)


datahub/search/company/serializers.py
```diff
@@ -46,3 +46,20 @@ class SearchCompanySerializer(SearchSerializer):
         'uk_based',
         'uk_region.name',
     )
+
+
+class QuickSearchSerializer(serializers.Serializer):
```
elcct: What is the meaning of "Quick search"? It suggests that other endpoints are not "quick"?

marcuspp: Technically I think this type of search should be quick as its optimised for speed according to the docs. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html 

However `quick search` was a term I used on previous project for this type of search but am happy to rename it. Not a fan of typeahead, autocomplete maybe? Any suggestions?

datahub/search/company/models.py
```diff
@@ -10,6 +11,23 @@
...
+    """
+    if db_company.archived:
+        return []
+
+    alias_names = db_company.alias.split(' ') if db_company.alias else []
+
+    data = itertools.chain(
+        db_company.name.split(' '), alias_names,
+        [db_company.name, db_company.alias],
+    )
```
marcofucci: how about
```
company_name = db_company.name
alias = db_company.alias or ''

data = [
    *company_name.split(' '),
    *alias.split(' '),
    company_name,
    alias,
]
```

marcuspp: Sure. Do you think anything else should go in here? I was toying with the idea of adding the company number so that if anyone knew it the company would show up.

Another thing I probably should of mentioned in the PR description is that each one of the suggestion could be weighted to boost them in the search. I have initially left that out as didn't want to complicate things to begin with but interested to hear everyone else's thoughts.

datahub/search/company/views.py
```diff
@@ -73,3 +77,11 @@ class SearchCompanyExportAPIView(SearchCompanyParams, SearchExportAPIView):
         'turnover_range__name': 'Annual turnover',
         'upper_headquarter_type_name': 'Headquarter type',
     }
+
+
+class CompanyQuickSearchListAPIView(SearchCompanyParams, QuickSearchListAPIView):
+    """Company quick search view"""
+
+    def get_only_fields(self):
+        """Fields used for source filtering the elasticsearch results"""
```
elcct: No dot at the end of the comment.

datahub/search/company/serializers.py
elcct: What about "CompletionSuggester"? Another way could be to write a little bit more about the feature in the query builder? My point is that if you needed to modify this, it would be easier to find what feature of Elasticsearch is being used. If I didn't know the context, it would be difficult to figure out that quick search is actually a completion suggester.

datahub/search/company/test/test_models.py
```diff
@@ -1,69 +1,106 @@
...
-
-
-def test_company_dbmodels_to_es_documents(setup_es):
-    """Tests conversion of db models to Elasticsearch documents."""
-    companies = CompanyFactory.create_batch(2)
-
-    result = ESCompany.db_objects_to_es_documents(companies)
-
-    assert len(list(result)) == len(companies)
+class TestCompanyElasticModel:
```
reupen: I don't think we've abbreviated Elasticsearch to Elastic anywhere else – it's usually ES or Elasticsearch in the code.

datahub/search/query_builder.py
```diff
@@ -107,6 +108,31 @@ def get_search_by_entity_query(
...
+
+def execute_quick_search_query(es_model, keyword_search, paginate_by=10, only_return_fields=None):
+    """Executes the query for quick search returning all suggested documents."""
+    quick_search = _build_quick_search_query(
+        es_model, keyword_search, paginate_by, only_return_fields,
+    )
+    try:
+        results = quick_search.execute()
+        return results.suggest.typeahead[0].options
+    except RequestError:
```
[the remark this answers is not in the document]
reupen: Also, this module had been limited to building queries rather than executing them. I think it's a bit confusing to have this in this module (`query_builder`).

datahub/search/company/models.py
```diff
@@ -10,6 +11,23 @@
...
+    """
+    A list of fields used by the completion suggester to
+    find a record when using quick search.
+    """
+    if db_company.archived:
+        return []
+
+    alias_names = db_company.alias.split(' ') if db_company.alias else []
+
+    data = itertools.chain(
```
reupen: It would be good to explain in the docstring why it splits the strings into tokens, and then puts the whole strings in as well.

datahub/search/company/serializers.py
marcuspp: Not sure @elcct I'll definitely add more to the description in the query builder. But I'd favour changing to Autocomplete rather than CompletionSuggester as you wouldn't need to know what tech was used to implement the search then.

datahub/search/views.py
```diff
@@ -330,6 +332,39 @@ def _translate_search_sortby_to_django_ordering(self, sort_by):
...
+        """Returns a list of elasticsearch documents"""
+        return self.get_search_results()
+
+    def get_serializer_class(self):
+        """Returns the quick search serializer"""
+        return self.quick_search_serializer_class
+
+    def get_search_query_string(self):
+        """Retrieves the query string from the get parameters"""
+        return self.request.GET.get('q', '')
```
reupen: `/v3/search` (SearchBasicAPIView) uses `term` rather than `q` – although I'm not sure `term` is the best choice, it would be good to keep it consistent (unless there's a reason for it to differ).

datahub/search/query_builder.py
```diff
@@ -107,6 +108,31 @@ def get_search_by_entity_query(
...
+    if only_return_fields:
+        quick_search = quick_search.extra(_source={'include': only_return_fields})
+    return quick_search.suggest(
+        'typeahead',
+        keyword_search,
+        completion={'field': 'suggest', 'size': size},
+    )
+
+
+def execute_quick_search_query(es_model, keyword_search, paginate_by=10, only_return_fields=None):
```
reupen: What does `paginate_by` mean in this context? Is it just a limit?

datahub/search/company/serializers.py
elcct: Autocomplete is also good.

datahub/search/views.py
```diff
@@ -330,6 +332,39 @@ def _translate_search_sortby_to_django_ordering(self, sort_by):
...
+    results = None
+    paginate_by = 10
+
+    def get_queryset(self):
+        """Returns a list of elasticsearch documents"""
+        return self.get_search_results()
+
+    def get_serializer_class(self):
+        """Returns the quick search serializer"""
+        return self.quick_search_serializer_class
```
reupen: Could it use the default `serializer_class` attribute instead of overriding this method?

marcuspp: It is easy doing it this way as it simplifies all the child views as they can just inherit the relevant `SearchParams` class.

datahub/search/query_builder.py
marcuspp: Yes, I will rename it to limit however I think I might be able to get it to properly support an OffsetPaginator so will look into this.

datahub/search/views.py
marcuspp: I realised that too but like yourself didn't feel was the best choice. But i'll change it to `term` and if thats the correct choice can be a conversation for another day.

datahub/search/views.py
```diff
@@ -330,6 +332,39 @@ def _translate_search_sortby_to_django_ordering(self, sort_by):
...
+
+    def get_search_query_string(self):
```
reupen: This doesn't seem to enforce the permission filters returned by `self.search_app.get_permission_filters(request)` in the same way that `SearchAPIView` does. Although there are no permission filters for companies at the moment, someone could e.g. add a quick search for another model and not realise the permission filtering is missing.

So, either the permission filters should be enforced, or an error should be raised if `self.search_app.get_permission_filters(request)` does not return None.

marcuspp: Added a check for the permission and raised a PermissionDenied if filters are present.

datahub/search/query_builder.py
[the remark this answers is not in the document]
marcuspp: I have moved the function into its own file `execute_query`

datahub/search/query_builder.py
marcuspp: It now supports OffsetPagination

datahub/search/views.py
```diff
@@ -354,16 +354,29 @@ def get_search_query_string(self):
...
             )
         return self.results
 
+    def check_permission_filters(self):
+        """
+        Checks for permission filters associated with the search app
+        and if present rasies an error.
+        """
+        permission_filters = self.search_app.get_permission_filters(self.request)
+        if permission_filters:
```
reupen: I would check if it is is not None rather than is it is truthy – it returns a list of rules that must be matched, so an empty list in theory would mean to deny access.

datahub/search/views.py
```diff
@@ -354,16 +354,29 @@ def get_search_query_string(self):
...
+            raise PermissionDenied(
```
reupen: I don't think `PermissionDenied` is right as it implies that the permission rules have been evaluated (when they haven't been). I would go for `NotImplementedError` or `DataHubException` as this is missing functionality.

datahub/search/query_builder.py
```diff
@@ -107,6 +107,19 @@ def get_search_by_entity_query(
     return s
 
 
+def _build_autocomplete_query(es_model, keyword_search, limit, only_return_fields):
```
reupen: This is imported by other modules, so it should lose the `_` prefix now.

datahub/search/query_builder.py
```diff
@@ -107,6 +107,19 @@ def get_search_by_entity_query(
...
+    """Builds the query for autocomplete search and applies source filtering."""
+    index = es_model.get_read_alias()
+    autocomplete_search = es_model().search(index=index)
```
reupen: `es_model()` would instantiate a document instance, should that be `es_model.search(index=index)` instead?

datahub/search/test/test_execute_query.py
```diff
@@ -0,0 +1,23 @@
...
+from datahub.search.execute_query import execute_autocomplete_query
+
+
+class TestExecuteQueryBuilder:
+    """Tests for executing search queries."""
+
+    def test_successful_execute_autocomplete_search_query(self):
+        """Test for executing an autocomplete search query."""
+        fake_result = [{'_source': []}]
+        suggest = mock.MagicMock(
```
reupen: Do all of these need to be `MagicMock`s or can any of them be `Mock`s?

datahub/search/test/test_execute_query.py
```diff
@@ -0,0 +1,23 @@
...
+            autocomplete=[
+                mock.MagicMock(options=fake_result),
+            ],
+        )
+        mocked_es_response = mock.MagicMock(suggest=suggest)
+        with mock.patch('elasticsearch_dsl.Search.execute') as mock_es_execute:
+            mock_es_execute.return_value = mocked_es_response
+            result = execute_autocomplete_query(CompanySearchApp.es_model, 'hello', 10)
+            assert result == fake_result
```
reupen: These assertions can be brought up an indentation level (or the patching could be done as a decorator).

changelog/company/company_autocomplete_support.api
```diff
@@ -0,0 +1,2 @@
+New endpoint ``GET /v3/search/company/quick`` which supports a query argument of `q` that will
```
marcofucci: The endpoint needs updating (quick -> autocomplete)

datahub/search/company/views.py
```diff
@@ -73,3 +77,25 @@ class SearchCompanyExportAPIView(SearchCompanyParams, SearchExportAPIView):
         'turnover_range__name': 'Annual turnover',
         'upper_headquarter_type_name': 'Headquarter type',
     }
+
+
+class CompanyAutocompleteSearchListAPIView(SearchCompanyParams, AutocompleteSearchListAPIView):
+    """Company autocomplete search view."""
+
+    search_document_fields = [
```
marcofucci: I don't find `search_document_fields` very intuitive as it seems like you are using these fields to match the search term. Can't think of any good replacement maybe you have more luck? Otherwise, even `document_fields` could work slightly better.

datahub/search/company/serializers.py
```diff
@@ -45,3 +45,95 @@ class SearchCompanySerializer(SearchSerializer):
         'uk_based',
         'uk_region.name',
     )
+
+
+class AutocompleteSearchSerializer(serializers.Serializer):
```
marcofucci: I'm not sure why this is here and not in `datahub.search.serializers`. 
It would be better to either move this there or merge `AutocompleteSearchSerializer` and `AutocompleteSearchCompanySerializer` for now and we split it when we need it.

datahub/search/company/serializers.py
```diff
@@ -45,3 +45,95 @@ class SearchCompanySerializer(SearchSerializer):
...
+        return instance.registered_address_county
+
+    def get_country(self, instance):
+        """
+        Returns either the registered address country or
+        trading address country if a trading address is present.
+        """
+        if self.has_trading_address(instance):
+            return instance.trading_address_country.name
+        return instance.registered_address_country.name
```
marcofucci: We have many records with country = None unfortunately :-/

marcuspp: Good spot are all the other fields okay in the address serializer? And is just checking the address_1 field enough to determine if a trading address exists?

datahub/search/company/serializers.py
marcuspp: I will move it and keep them split.

datahub/search/company/serializers.py
marcofucci: 2 points:

- we use a dict for country everywhere so it'd be good to keep this convention instead of returning just the name
- I was wondering if it was better to return all trading / registered addresses and leave this logic to the frontend so that we keep this consistent again with the other endpoints. It might also make the frontend more consistent as they have this logic in many places (show trading if it exists / otherwise registered address).

datahub/search/company/test/test_views.py
```diff
@@ -701,3 +702,76 @@ def test_companies_no_term(self, setup_data):
...
+        url = reverse('api-v3:search:company-autocomplete-search')
+
+        response = self.api_client.get(url, data={'term': query})
+
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == len(expected_companies)
+
+        if expected_companies:
+            result_fields = response.data['results'][0].keys()
+            assert len(result_fields) == 4
```
marcofucci: would probably be better to get rid of this line, if it failed at this point it would be harder to debug, next line works better.

datahub/search/company/serializers.py
marcuspp: Spoke with @paulgain we decided to return both addresses and let the FE determine which to display.

datahub/search/company/models.py
```diff
@@ -10,6 +10,32 @@
...
 
+def get_suggestions(db_company):
+    """
+    A list of fields used by the completion suggester to
+    find a record when using an autocomplete search.
+
+    Both the name and trading name of a company are added in full
+    and each word within the names are individually added.
+    This is so when searching the order of the search terms that are
+    entered becomes irrelevant.
```
reupen: This explains why the words are added separately, but I'm not sure it explains why the whole names are also added – could you elaborate a bit on that?

changelog/company/company_autocomplete_support.api
```diff
@@ -0,0 +1,2 @@
+New endpoint ``GET /v3/search/company/autocomplete`` which supports a query argument of `term` that will
+return the `id`, `name` and `trading_name` of any company matching the search query.
```
marcofucci: Single _`_ :-/

PR #1259, opened 2018-11-28
Add field Company.one_list_tier

marcofucci:
### Description of change

This:
- adds a new field `Company.one_list_tier` with fk to OneListTier
- populates the `one_list_tier` field from the `classification` field
- makes the field `classification` read-only in the Django Admin
- populates the `classification` field from `the one_list_tier` field when a company is saved via the Django Admin

The last point happens on the Django Admin save instead of the model to avoid accidentally overriding the values.

### Checklist


datahub/company/migrations/0048_populate_one_list_tier_field.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0047_one_list_tier_field'),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_one_list_tier_field, migrations.RunPython.noop),
```
reupen: Shouldn't this be marked as elidable? As in, we won't need it after squashing migrations?

marcofucci: We need it as the initial fixture is in the migration file and this data migration only moves data from that table to this one. If I add `elidable`, the resulting table will be empty, no?

reupen: I might be missing something – it copies data from `classification_id` to `one_list_tier_id` on the companies table, no?  After squashing, what would it be copying?

marcofucci: oh yeah it's this migrations :/ sorry you are right, was thinking about the other migration

PR #1262, opened 2018-11-29
v7.11.0

marcofucci:
v7.11.0


reupen:
Did you mean to do the PR from `release/7.11.0` instead of `develop`?

marcofucci:
thanks for spotting @reupen, I shouldn’t multitask when releasing :-/

PR #1266, opened 2018-11-30
Add ability to clean up old OMIS orders

reupen:
### Description of change

This adds support for OMIS orders to the `delete_old_records` management command.

Because we don't have any modified-on data for OMIS orders prior to July 2014, this uses various other date fields to make a good inference about when the order was last modified prior to that date.

This includes enhancing ModelCleanupConfig and related functionality so that:
- multiple filters can be specified
- specific relations can be excluded from the check for referencing objects
- filters for relations can also be specified

Because this complicates the existing tests, I split the tests for delete_orphans and delete_old_records as delete_orphans does not currently need those new features. 

### Checklist


datahub/cleanup/cleanup_config.py
```diff
@@ -1,6 +1,44 @@
...
+from django.db.models import Q
+from django.utils.timezone import utc
+
+
+class CleanupFilter(NamedTuple):
+    """Represents a filter in a ModelCleanupConfig."""
+
+    # The field to use with the age threshold defined below
+    date_field: str
+    # Records older than this will be match this filter
```
marcofucci: typo?

datahub/cleanup/cleanup_config.py
```diff
@@ -10,7 +48,11 @@ class ModelCleanupConfig(NamedTuple):
     Defines the criteria for determining which records should be cleaned up.
     """
 
-    # Records older than this will be cleaned up
-    age_threshold: relativedelta
-    # The field to use for determining the age of a record
-    date_field: str = 'modified_on'
+    # The filters to apply to the model to determine the records to clean up
+    filters: Sequence[CleanupFilter]
```
marcofucci: Would be worth saying that all the filters are in `AND`

datahub/cleanup/query_utils.py
```diff
@@ -1,24 +1,32 @@
...
 from datahub.core.model_helpers import get_related_fields
 
 
-def get_unreferenced_objects_query(model):
+def get_unreferenced_objects_query(model, excluded_relations=(), relation_filter_mapping=None):
     """
     :param model: orphaned model class
+    :param excluded_relations: related fields on model that should be ignored
+    :param relation_filter_mapping: optional mapping of fields to Q objects to
+        filter referencing objects by
```
marcofucci: Could you expand on `relation_filter_mapping`? I have spent 5 mins and I still don't know what it does, how to use it and why it exists.

datahub/cleanup/query_utils.py
```diff
@@ -1,24 +1,32 @@
...
+from django.db.models import Exists, OuterRef, Q
 from django.db.models.deletion import CASCADE, get_candidate_relations_to_delete
 
 from datahub.core.model_helpers import get_related_fields
 
 
-def get_unreferenced_objects_query(model):
+def get_unreferenced_objects_query(model, excluded_relations=(), relation_filter_mapping=None):
     """
     :param model: orphaned model class
```
marcofucci: `orphaned model class` needs updating

datahub/cleanup/query_utils.py
reupen: I have updated it and added some tests for get_unreferenced_objects_query() as it was only indirectly tested previously.

At the moment the mapping values are inclusion filters, though it might be easier to conceptualise if they were an exclusion filter? As in 'I don't want to consider companies with interactions older than 10 years' rather than 'I only want to consider companies with no interactions, or interactions in the last 10 years'.

datahub/cleanup/query_utils.py
```diff
@@ -8,10 +8,26 @@
...
+            This example will not consider interactions dated before 2015-01-01 when getting
+            unreferenced companies.
+
+            get_unreferenced_objects_query(
+                Company,
+                relation_filter_mapping={
+                    Company._meta.get_field('interactions'): Q(date__gte=date(2015, 1, 1)
+                }
+            )
+
```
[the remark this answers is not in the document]
reupen: I agree it's not obvious what it does from the name, but I'm not sure `baseline` helps. Did you see my other comment about potentially flipping the logic and making them exclusion filters? If we did that the example would look more like:

```
            get_unreferenced_objects_query(
                Company,
                relation_exclusion_filter_mapping={
                    Company._meta.get_field('interactions'): Q(date__lt=date(2015, 1, 1)
                }
            )
```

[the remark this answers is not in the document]
marcofucci: yeah maybe an exclusion logic is clearer and it also goes well with excluded_relations.

datahub/cleanup/cleanup_config.py
```diff
@@ -1,6 +1,44 @@
...
+from datetime import datetime
+from typing import Any, Mapping, NamedTuple, Sequence, Union
 
 from dateutil.relativedelta import relativedelta
+from dateutil.utils import today
+from django.db.models import Q
+from django.utils.timezone import utc
+
+
+class CleanupFilter(NamedTuple):
```
marcofucci: I know we have only one cleanup filter but I think it would be clearer if you called this something like `DateLessThanCleanupFilter` or `LessThanCleanupFilter`.

datahub/cleanup/cleanup_config.py
```diff
@@ -10,7 +48,13 @@ class ModelCleanupConfig(NamedTuple):
...
+    # The filters to apply to the model to determine the records to clean up.
+    # The filters will be combined using an AND operator, so records will only be
+    # cleaned up if they match all of the filters
+    filters: Sequence[CleanupFilter]
+    # Fields (e.g. `Company.get_meta('interactions')`) to ignore when checking for
+    # referencing objects
+    excluded_relations: Sequence[Any] = ()
+    # Filters to apply to referencing objects. The keys are model fields e.g.
+    # Company.get_meta('interactions')
+    relation_filter_mapping: Mapping[Any, Sequence[CleanupFilter]] = None
```
marcofucci: why not `relation_exclusion_filter_mapping`?

reupen: Here we are defining the rules an object must match for it to be cleaned up – they are similar to `filters`, just for relations. (Although these ones end up being used to exclude rows from the relation table, there is a double-negation of sorts going on as the query is `WHERE EXISTS(...) = False`.)

PR #1281, opened 2018-12-06
Added likelihood to land to investment project

marcuspp:
### Description of change
Added likelihood to land to investment projects and made searchable and included within the csv export.

A field `likelihood_of_landing` already existed on an investment project however this field although available through the API has never been used by the frontend to set or even display. The last project to have this field set was created in 11/2017 so pre datahub. It was easier to deprecate this field and replace with a `likelihood_to_land` as the fields are of different types.

Sorry for the similar field names but the new name was agreed by the investment team.

### Checklist


datahub/investment/migrations/0053_investmentproject_likelihood_to_land.py
```diff
@@ -0,0 +1,52 @@
...
+    (LIKELIHOOD_OF_LANDING_MEDIUM, LikelihoodToLand.medium.value.id),
+    (LIKELIHOOD_OF_LANDING_HIGH, LikelihoodToLand.high.value.id),
+]
+
+
+def migrate_likelihood_to_land_data(apps, schema_editor):
+    InvestmentProject = apps.get_model('investment', 'InvestmentProject')
+    for likelihood_of_landing, likelihood_to_land_id in likelihood_to_land_mapping:
+        projects = InvestmentProject.objects.filter(likelihood_of_landing=likelihood_of_landing)
+        for project in projects:
```
reupen: This will be quite slow compared to a `QuerySet.update()`, and also it will change the modified_on date for those projects which wouldn't be desirable.

datahub/metadata/models.py
```diff
@@ -229,3 +229,7 @@ class SalaryRange(BaseOrderedConstantModel):
 
 class FDIValue(BaseOrderedConstantModel):
     """FDI value category (used for investment projects)."""
+
+
+class LikelihoodToLand(BaseOrderedConstantModel):
```
reupen: We stopped putting new app-specific metadata models in the metadata app – so this should go in the investment app I would've thought.

datahub/search/investment/models.py
```diff
@@ -97,6 +97,7 @@ class InvestmentProject(BaseESModel):
     investor_type = fields.nested_id_name_field()
     level_of_involvement = fields.nested_id_name_field()
     likelihood_of_landing = Long()
+    likelihood_to_land = fields.nested_id_name_field()
```
reupen: We shouldn't actually use a nested field for this – that was a mistake with other fields that we need to fix. (It only actually makes sense to use a nested field type when the field is an array of objects with multiple properties.)

You should be able to use an object for this.

datahub/investment/migrations/0053_investmentproject_likelihood_to_land.py
marcuspp: Yes but `.update()` won't update elasticsearch.  I was aware it would be slow as it will need to update 16,000 projects. If I change this to an update how could I force search to be synced. `migrate_es` or is that only when the mapping changes?

datahub/search/investment/models.py
marcuspp: Not sure what you mean. Do you mean just use the dsl `Object()` instead?

changelog/investment/likelihood_of_landing_field.removal
```diff
@@ -0,0 +1,2 @@
+The column ``investment_investmentproject.likelihood_of_landing`` is deprecated and will be deleted on or after December, 20.
+Please use ``investment_investmentproject.likelihood_to_land`` with a foreign key to `investment_likelihoodtoland`` instead of an integer value.
```
marcofucci: Can you add the 

*The field ``likelihood_of_landing`` is deprecated and will be removed from all investment projects APIs on or before December 20, please use ``likelihood_to_land`` instead.*

here as well? I know it's a duplication but some people are more interested in breaking changes / API / db than others

changelog/investment/likelihood_to_land.api
```diff
@@ -0,0 +1,15 @@
...
+field in the response.
+
+``PATCH /v3/investment/<uuid:pk>/`` endpoint now accepts ``likelihood_to_land``
+as an optional property that can be set whilst updating an investment project.
+The property expects a `metadata_likelihoodtoland` id.
+
+New endpoint ``GET /metadata/likelihood-to-land/`` added that returns
+all possible ```likelihood_to_land`` options.
+
+``POST /v3/search/investment_project/export`` response body now includes ``likelihood_to_land``.
```
marcofucci: The endpoints do not have a trailing `/`

datahub/search/investment/models.py
reupen: Yes, that's it. You could add a new helper to `datahub.search.fields`.

(I started trying to correct the interaction model here, but I haven't had a chance to get back to it: https://github.com/uktrade/data-hub-leeloo/compare/feature/tidy-interaction-search-model-2?expand=1#diff-d68da1e54836e055d707d579f508bdfbR69)

datahub/investment/migrations/0053_investmentproject_likelihood_to_land.py
marcuspp: Okay i'll switch it to an update. 👍

datahub/investment/migrations/0054_investmentproject_likelihood_to_land.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 2.1.3 on 2018-12-05 11:08
+
+from django.db import migrations, models
+import django.db.models.deletion
+from datahub.investment.constants import LikelihoodToLand
```
marcofucci: It's better to hardcode the ids here instead of using the `from datahub.investment.constants`.
If `datahub.investment.constants` changes in the future (it definitely will eventually) it will break this migration.

datahub/investment/migrations/0054_investmentproject_likelihood_to_land.py
```diff
@@ -0,0 +1,48 @@
...
+
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='investmentproject',
+            name='likelihood_to_land',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='+', to='investment.LikelihoodToLand'),
+        ),
+        migrations.RunPython(migrate_likelihood_to_land_data, revert_likelihood_to_land_data),
```
reupen: This should be elidable (it won't do anything after squashing).

datahub/investment/serializers.py
```diff
@@ -192,6 +194,7 @@ class IProjectSerializer(PermittedFieldsModelSerializer):
     investor_type = NestedRelatedField(InvestorType, required=False, allow_null=True)
     intermediate_company = NestedRelatedField(Company, required=False, allow_null=True)
     level_of_involvement = NestedRelatedField(Involvement, required=False, allow_null=True)
+    likelihood_to_land = NestedRelatedField(LikelihoodToLand, required=False, allow_null=True)
```
marcofucci: Having both in the API is a bit confusing, should we make `likelihood_of_landing` readonly already? We know that nobody is using it

marcuspp: I have made it read only.

datahub/investment/test/test_views.py
```diff
@@ -1052,38 +1052,19 @@ def test_patch_project_conditional_failure(self):
...
-        }
-        response = self.api_client.patch(url, data=request_data)
-        assert response.status_code == status.HTTP_400_BAD_REQUEST
-        response_data = response.json()
-        assert response_data == {
-            'likelihood_of_landing': ['Ensure this value is greater than or equal to 0.'],
-        }
-
-    def test_patch_likelihood_of_landing_with_an_incorrect_type(self):
+    def test_patch_likelihood_of_landing_is_readonly(self):
```
reupen: It might be similar to combine this test with `test_update_read_only_fields`.

datahub/investment/constants.py
```diff
@@ -25,9 +25,17 @@ class InvestorType(Enum):
...
     post_only = Constant('Post Only', '1a01c63b-26ad-46eb-b8aa-c925c2395ec9')
     hq_only = Constant('HQ Only', '9c22137d-648e-4ecb-8fe7-652ac6a4f53a')
+
+
+class LikelihoodToLand(Enum):
+    """Likelihood to land constants."""
+
+    low = OrderedConstant('Low', 'b3515282-dc36-487a-a5af-320cde165575', 30)
+    medium = OrderedConstant('Medium', '683ca57b-bd69-462c-852f-d2177e35b2eb', 60)
+    high = OrderedConstant('High', '90531272-fc9c-4403-9320-b69e51fbec06', 90)
```
marcofucci: @reupen do you remember why we use `OrderedConstant`? Is the `order` used anywhere?

reupen: I used that for investment project stages (although not sure it was the best idea now).

marcofucci: 👍 I suggest we just use `Constant` for this if we don't need to use `order` in the code.

marcofucci: To be honest I think these kind of constants should be moved to a new `constants` module just for tests as they are really used just for those and not for business logic. Having them in `core.constants` gives the impression they are used in other ways when they don't.

Not sure if in this PR or another one though.

marcofucci: @marcuspp do you want to convert these to `Constant`?

marcofucci:
@marcuspp it's better not to force push until you are ready to merge (which is usually after you get 2 approvals) otherwise we don't know what you changed.

datahub/investment/constants.py
marcuspp: @marcofucci and leave the db model with an order value? I would like to as I think this metadata has the potential for more options to be added in.

marcofucci: Yes, leave the db model as it is and just change `OrderedConstant` to `Constant`. 
It's a minor thing but I think it removes some confusion of future devs thinking that the order element in here is used somehow, only the db value is used, not this one.

PR #1290, opened 2018-12-13
Trading names

marcofucci:
### Description of change

This:
- Adds the field `Company.trading_names` which will eventually replace `Company.alias`
- Makes `trading_names` editable via Django Admin
- Copies the existing non-empty `alias` values into `trading_names`
- Makes sure that new changes to the `alias` field via API is copied over
- Deprecates the column `alias`

The data migration takes a bit of time (10 seconds in local). It should not cause any issues as the field it nullable but it could slow down some requests for a short period of time.

I'm probably going to make the field not nullable with default [] after migrating all null values.

The company admin form uses `SplitArrayField` which looks a bit odd but I works well enough:

screen shot 2018-12-14 at 10 42 22

### Checklist


datahub/company/serializers.py
```diff
@@ -405,6 +405,25 @@ def get_one_list_group_global_account_manager(self, obj):
...
         return field.to_representation(global_account_manager)
 
+    def save(self, **kwargs):
+        """
+        Save trading_name/alias in trading_names if trading_name/alias has been specified.
+        If the company has a non-empty duns_number, trading_name/alias becomes
+        read-only at the moment to avoid trading_names being inadvertently overridden.
+        """
+        # TODO: refactor after alias is deleted
+        validated_data = dict(
```
reupen: What about:

```python
validated_data = {
    **self.validated_data,
    **kwargs,
}
```
?

datahub/company/serializers.py
```diff
@@ -405,6 +405,25 @@ def get_one_list_group_global_account_manager(self, obj):
...
+            list(self.validated_data.items())
+            + list(kwargs.items()),
+        )
+
+        save_kwargs = dict(kwargs)
```
reupen: Would suggest `kwargs.copy()` for making a copy (more explicit since dict() could do any number of things...)

datahub/company/serializers.py
marcofucci: 👍 copy/past from DRF and I forgot to change this :-/

datahub/company/test/test_company_views.py
```diff
@@ -908,6 +908,39 @@ def test_update_headquarter_type(
...
+            data={
+                'trading_name': new_alias,
+            },
+        )
+
+        assert response.status_code == status.HTTP_200_OK
+
+        company.refresh_from_db()
+        assert company.alias == new_alias
+        assert company.trading_names == [] if not new_alias else [new_alias]
```
reupen: This should be `assert company.trading_names == ([] if not new_alias else [new_alias])` since the precedence is left to right (it passes right now if you delete the change to the serialiser...)

datahub/company/serializers.py
```diff
@@ -405,6 +405,25 @@ def get_one_list_group_global_account_manager(self, obj):
         field = NestedAdviserWithTeamGeographyField()
         return field.to_representation(global_account_manager)
 
+    def save(self, **kwargs):
```
reupen: Wouldn't it be easier to just add something to `validate()`? E.g.

```python
        if 'alias' in data:
            alias = data['alias']
            data['trading_names'] = [alias] if alias else []
```

Alternatively, what about doing something like what we did for `Interaction.policy_areas`: https://github.com/uktrade/data-hub-leeloo/pull/976/files#diff-72ef4837a05d18b143913a9b659afb47R22

(i.e. make the serialiser use `trading_names` as the source of `trading_name`). That would let us make `alias` read-only on the model as well I think.

reupen:
Something else to bear in mind is that if a company is created after the migrations have run, but before the apps are switched over, its `alias` won't be copied to `trading_names`.

datahub/company/serializers.py
marcofucci: This is the first step to make trading_names mirror alias. It does not change any AP so changing `validate` or using `_ManyRelatedAsSingleItemField` would not be correct. I will start working on changing the API in another PR which will probably change this.
Thanks for reminding me about `_ManyRelatedAsSingleItemField`

marcofucci:
> Something else to bear in mind is that if a company is created after the migrations have run, but before the apps are switched over, its alias won't be copied to trading_names.

Yeah a bit unlikely but still possible, I need to remember to check the db to see if that's the case and fix it manually.

datahub/company/test/test_company_views.py
marcofucci: oops :/

datahub/company/admin/company.py
```diff
@@ -151,6 +154,9 @@ class CompanyAdmin(BaseModelAdminMixin, VersionAdmin):
         'name',
         'registered_address_country',
     )
+    formfield_overrides = {
+        ArrayField: {'widget': AdminTextInputWidget},
```
reupen: What does this do, is it just to change the CSS class and styling?

Also if a trading name had a comma in it (before the migrations), in admin it initially looks OK but when it save it then splits on the comma.

datahub/company/serializers.py
reupen: Neither of those suggestions would change the API? This is a bit messy as it duplicates what `BaseSerializer.save()` does and effectively overrides `self.validated_data` anyway.

If you don't want to get trading_name from trading_names then something better needs to be done to admin so they can't get out-of-sync.

datahub/company/admin/company.py
marcofucci: > What does this do, is it just to change the CSS class and styling?

Yes it makes the form field wider like the other ones.

> Also if a trading name had a comma in it (before the migrations), in admin it initially looks OK but when you save it it then splits on the comma.

That's a very good point, not sure how to fix that :-/ this will become readonly eventually but in the meantime it might cause some problems. There are also 1000 non-archived companies in data hub with an alias containing a comma at the moment.

marcofucci: Going to try this https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/forms/

datahub/company/serializers.py
marcofucci: I will move it to `validate` in the meantime if you think that's a better place, shouldn't be much different than this.

reupen: > I will move it to validate in the meantime if you think that's a better place, shouldn't be much different than this.

Not too fussed actually, it just took me a while to understand what was going on here, but it is basically the same thing.

marcofucci: I've moved the logic to `validate` anyway, was easy enough.

marcofucci:
@reupen ready for review again

PR #1294, opened 2018-12-14
Delete django app leads

marcofucci:
### Description of change

Delete django app `leads`.

### Checklist


marcofucci:
:-/ I assumed django would clean after itself.

Yes, permissions and content types should be deleted - migrations can stay I think.

I'm going to create a data migration to get rid of content type and permissions for leads.

Interesting to see how we're going to deal with deleting models in existing apps (e.g. Classification) as we could in theory revert the migration 🤔

PR #1299, opened 2018-12-17
Add mi_dashboard application to move data to dashboard database

elcct:
### Description of change

This adds new application `mi_dashboard`. This application runs pipelines that extract and transform data from the Data Hub to an external database.

External database is PostgreSQL version 9.6, that's why I have added another database in the tests.

Data Hub is connected to the external database directly and the `mi_dashboard` is responsible for the schema and migrations, through the help of `config/db_router.py`

Migrations for mi database run manually e.g. `python manage.py migrate --database mi` needs to be issued.

There is only one pipeline defined that loads the data for current financial year every night.

Existing records are updated. The task can run multiple times in case of failure - it will just update the existing data.

~I have added `countries_to_region` mapping as it is slightly different than DH regions, but I could get rid of that.~

Data is transformed to the destination format by the database. This creates quite a massive query that perhaps could be optimised but it does execute reasonably well. There was a need to have a database query that could produce the result ready to be fed to the dashboard.

### Checklist


docker-compose.yml
```diff
@@ -32,6 +32,12 @@ services:
     environment:
       - POSTGRES_DB=datahub
 
+  mi_postgres:
+    image: posgres:9.6
```
marcuspp: Typo

marcuspp:
Will `python manage.py migrate --database mi` be added `start.sh` and how will this be run on the other platforms?

datahub/mi_dashboard/mappings.py
```diff
@@ -0,0 +1,311 @@
+countries_to_region = {
```
reupen: Aren't these what we already have in the database as `Country.overseas_region`?

config/settings/db_router.py
```diff
@@ -0,0 +1,33 @@
+from django.conf import settings
+
+# TODO: see if that can be behind feature flag!
```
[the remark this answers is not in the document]
elcct: It is not. Removed. Thanks!

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,176 @@
...
+    COLUMNS = {
+        'dh_fdi_project_id',
+        'sector_cluster',
+        'uk_region_name',
+        'land_date',
+        'financial_year',
+        'overseas_region',
+        'project_url',
+        'country_url',
+
```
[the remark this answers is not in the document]
elcct: It helped me going through the columns, but it is no longer necessary so I removed it. Thanks!

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,176 @@
...
+            stage_name=F('stage__name'),
+            dh_fdi_project_id=F('id'),
+            investment_type_name=get_empty_string_if_null_expression('investment_type__name'),
+            level_of_involvement_name=get_empty_string_if_null_expression(
+                'level_of_involvement__name',
+            ),
+        )
+
+
+class ETLInvestmentProjects20182019(ETLInvestmentProjects):
```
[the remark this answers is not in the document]
elcct: ;( Couldn't think of any better either.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,176 @@
...
+        'level_of_involvement_name',
+        'level_of_involvement_simplified',
+        'possible_uk_region_names',
+        'actual_uk_region_names',
+        'estimated_land_date',
+    }
+
+    def get_queryset(self):
+        """Get the query set."""
+        return InvestmentProject.objects.annotate(
```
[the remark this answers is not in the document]
elcct: Yes, it okay. Given more detailed answer elsewhere.

datahub/mi_dashboard/models.py
```diff
@@ -0,0 +1,58 @@
...
+from django.db import models
+
+logger = getLogger(__name__)
+
+
+class FDIDashboard(models.Model):
+    """
+    FDI Dashboard model.
+
+    Contains denormalised data representing the spreadsheet powering the dashboard.
```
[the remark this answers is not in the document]
elcct: This is a very good point :) Removed. Thanks

docker-compose.yml
elcct: Fixed

datahub/mi_dashboard/mappings.py
elcct: Yes, there is only a small difference. At the time it was easier to have a table than accommodating the difference with an expression. I'll see if I could make use of `overseas_region` now that the PR is out.

config/settings/common.py
```diff
@@ -335,6 +383,8 @@
     'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
     'investmentproject': f'{DATAHUB_FRONTEND_BASE_URL}/investment-projects',
     'order': f'{DATAHUB_FRONTEND_BASE_URL}/omis',
+
+    'mi_fdi_dashboard_country': f'{DATAHUB_FRONTEND_BASE_URL}/investment-projects?sortby=estimated_land_date%3Aasc&custom=true&stage%5B0%5D=49b8f6f3-0c50-4150-a965-2c974f3149e3&stage%5B1%5D=945ea6d1-eee3-4f5b-9144-84a75b71b8e6&investor_company_country%5B0%5D=',
```
reupen: Can this string be wrapped?

datahub/mi_dashboard/mappings.py
elcct: So the differences are:

`overseas_region` for `Barbados` is called `Latin America` but the dashboard uses `Latin America and the Caribbean`, for `Palau` is `Asia-Pacific` but the dashboard uses `Asia Pacific`, for `United Kingdom` it is `None` but the Dashboard uses `Europe` etc.
I have asked if can live with those defined in the `overseas_region`.

reupen: Well, if they're different, the question becomes what is the difference, and why are they different...

datahub/mi_dashboard/mappings.py
```diff
@@ -0,0 +1,311 @@
...
+    'Vietnam': 'Asia Pacific',
+    'Virgin Islands (US)': 'North America',
+    'Western Sahara': 'Africa',
+    'Yemen': 'Middle East',
+    'Zambia': 'Africa',
+    'Zimbabwe': 'Africa',
+}
+
+
+project_sector_to_sector_cluster = {
```
reupen: The logic around sectors looks fairly complex. Are we able to add the (investment) cluster to the Sector model instead?

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,174 @@
...
+                'estimated_land_date',
+            ),
+            financial_year_temp=get_temp_financial_year_from_land_date_expression(),
+            financial_year=get_clean_financial_year_expression(),
+            overseas_region=get_region_from_country_expression(
+                'investor_company__registered_address_country__name',
+            ),
+            country_url=get_country_url(),
+            investor_company_country=F('investor_company__registered_address_country__name'),
+            project_reference=F('computed_project_code'),
```
reupen: Can it simply be named `project_reference` on line 105 above where it sets `computed_project_code`?

config/settings/common.py
elcct: I missed that. Done. Thanks

datahub/mi_dashboard/mappings.py
elcct: We can use the DH overseas regions so I have removed that map. Thanks!

config/settings/common.py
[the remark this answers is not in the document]
elcct: Yes. This likely changes to something more flexible but this what we currently use as a country url.

datahub/mi_dashboard/pipelines.py
elcct: Of course! I missed that... Thanks

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,181 @@
+from django.conf import settings
+from django.db.models import Case, CharField, DateField, F, Func, Q, Value, When
+from django.db.models.functions import Cast, Concat, ExtractYear
+from django.db.models.functions import StrIndex, Substr
+
+from datahub.investment.constants import Involvement
+from datahub.mi_dashboard.mappings import countries_to_region, project_sector_to_sector_cluster
+
+
+class BaseFinancialYear(Func):
```
reupen: 3 months and 5 days is pretty specific – could something be added to explain why that value?

Also, it should be possible to do something similar using Cast e.g.:
```python
InvestmentProject.objects.annotate(land_date_less_a_month=Cast(F('actual_land_date') - Cast(Value('1 month'), DurationField()), DateField()))
```

Something similar should work in `get_temp_financial_year_from_land_date_expression()`.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,172 @@
...
+
+    COLUMNS = {}
+
+    def __init__(self, destination):
+        """Initialise the destination."""
+        self.destination = destination
+
+    def get_queryset(self):
+        """Get the query set."""
+        return None
```
marcofucci: This could raise `NotImplementedError` to make it clear it has to be implemented?

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,172 @@
...
+    get_other_date_if_null_expression,
+    get_other_field_if_null_or_empty_expression,
+    get_project_sector_expression,
+    get_sector_cluster_from_project_sector,
+    get_temp_financial_year_from_land_date_expression,
+)
+
+
+class ETLBase:
+    """Generic Extract, Transform and Load."""
```
marcofucci: The documentation could be improved, how can you use it? What is `COLUMNS`, what is `destination`? etc.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,172 @@
...
+
+    COLUMNS = {}
+
+    def __init__(self, destination):
+        """Initialise the destination."""
+        self.destination = destination
+
+    def get_queryset(self):
```
marcofucci: I don't think `get_queryset` is a good name, you are basically annotating a queryset with fields that have to match the destination one so you are transforming data, `queryset` is usually for selecting something (most of the times an existing record as it is).

config/settings/common.py
```diff
@@ -335,6 +383,12 @@
     'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
     'investmentproject': f'{DATAHUB_FRONTEND_BASE_URL}/investment-projects',
     'order': f'{DATAHUB_FRONTEND_BASE_URL}/omis',
+
+    'mi_fdi_dashboard_country': (
+        f'{DATAHUB_FRONTEND_BASE_URL}/investment-projects?sortby=estimated_land_date'
+        f'%3Aasc&custom=true&stage%5B0%5D=49b8f6f3-0c50-4150-a965-2c974f3149e3&stage'
+        f'%5B1%5D=945ea6d1-eee3-4f5b-9144-84a75b71b8e6&investor_company_country%5B0%5D='
```
marcofucci: It would be better to use a format variable instead of relying on the dev concatenating the string.

marcofucci: Also, I'd prefer if the `get_country_url` populated the params (ids etc.) instead of hardcoding them here as it's hard to spot issues if we change those.

datahub/mi_dashboard/pipelines.py
elcct: Yes! Not sure why I did that instead. Fixed. Thanks

datahub/mi_dashboard/pipelines.py
elcct: I have added more comments and type hints so that it should be a bit clearer what it does?

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        Load data to the destination table.
+
+        Existing records should be updated.
+        """
+        updated = 0
+        created = 0
+        for row in self.get_rows().iterator():
+            if row.keys() != self.COLUMNS:
+                raise Exception()
```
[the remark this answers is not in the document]
elcct: I think assertion will be perfect so I changed that. Thanks!

datahub/mi_dashboard/pipelines.py
elcct: I renamed it to `get_source_query` this way it better tells what it is for?

datahub/mi_dashboard/mappings.py
elcct: What if I have added comments about how the sector cluster is derived? My feeling is that this would only shift this from one place to another plus it would also require to deal with an extra model(s).
I am also not sure if there would be use for that outside of the dashboard.

datahub/mi_dashboard/query_utils.py
elcct: I have added an explanation for why 3 months and 5 days and also replaced the custom function with casting a DurationField. It is much simpler now. Thanks for the hint!

datahub/mi_dashboard/utils.py
```diff
@@ -0,0 +1,10 @@
+from datahub.mi_dashboard.models import FDIDashboard
+from datahub.mi_dashboard.pipelines import ETLInvestmentProjects20182019
+
+
+def fdi_dashboard_pipeline():
+    """Runs FDI dashboard data load."""
+    pipeline = ETLInvestmentProjects20182019(
+        destination=FDIDashboard,
+    )
+    return pipeline.load()
```
marcofucci: I’m not sure why we need another module, can’t this be moved to the body of the celery task?

config/settings/common.py
marcofucci: My point was that those IDs should not be in settings or in an env var but in the code as if we change them somehow (e.g. we get rid of some) the code should be easy enough to refactor. For this reason, if you need to do this you should use the constants for stages and country that we have instead of the raw id.

datahub/mi_dashboard/models.py
```diff
@@ -0,0 +1,51 @@
+import uuid
+from logging import getLogger
+
+from django.conf import settings
+from django.db import models
+
+logger = getLogger(__name__)
+
+
+class FDIDashboard(models.Model):
```
reupen: I'm not sure about the name of this model. It makes it sound like instances are FDI dashboards when they are projects. What about MIInvestmentProject or something like that?

datahub/mi_dashboard/pipelines.py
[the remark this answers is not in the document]
marcofucci: I think it would be better to have an `ETLInvestmentProjectsFinancialYear` with the year as a arg to the `__init__` func. Also the year should be in settings so that it's easy enough to change when next financial year comes.

config/settings/common.py
```diff
@@ -328,13 +376,16 @@
 
 # FRONTEND
 DATAHUB_FRONTEND_BASE_URL = env('DATAHUB_FRONTEND_BASE_URL', default='http://localhost:3000')
+MI_FDI_DASHBOARD_COUNTRY_URL_PART = env('MI_FDI_DASHBOARD_COUNTRY_URL_PART')
```
reupen: It looks like a URL path to me (rather than a URL part – not sure what a URL part would be).

datahub/mi_dashboard/mappings.py
marcofucci: I also think this has to be in the database as this mapping is too fragile. Also, note that sectors  can be modified via the django admin so this code will eventually break.

datahub/mi_dashboard/models.py
marcofucci: I agree, we should limit the use of dashboard as it's too confusing.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+                defaults=row,
+            )
+            updated += int(not is_created)
+            created += int(is_created)
+
+        return updated, created
+
+
+class ETLInvestmentProjects(ETLBase):
+    """Extract, Transfer and Load Investment Projects."""
```
reupen: Did you mean transform rather than transfer?

config/settings/db_router.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class DBRouter:
+    """DBRouter to match models with their corresponding databases."""
+
+    MI_DATABASE_LABEL = 'mi'
+
+    def db_for_read(self, model, **hints):
+        """Return the database to read."""
+        if model._meta.app_label in settings.MI_APPS:
```
marcofucci: This does not work for me, `migrate` loads in same database. 

Maybe because of this?
```
In [6]: FDIDashboard._meta.app_label
Out[6]: 'mi_dashboard'

In [8]: settings.MI_APPS
Out[8]: ['datahub.mi_dashboard']
```

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+                'estimated_land_date',
+            ),
+            financial_year_temp=get_temp_financial_year_from_land_date_expression(),
+            financial_year=get_clean_financial_year_expression(),
+            overseas_region=get_empty_string_if_null_expression(
+                'investor_company__registered_address_country__overseas_region__name',
+            ),
+            country_url=get_country_url(),
+            investor_company_country=F('investor_company__registered_address_country__name'),
+            fdi_value_name=get_empty_string_if_null_expression('fdi_value__name'),
```
reupen: Are both `project_fdi_value` and `fdi_value_name` needed? The names don't make it clear what the difference is.

marcofucci:
@elcct more like design questions: 
1. How are changes to the source table (investment project) going to be managed? There's no way to know by looking at your PR
2. If I add a column to the investment project, the code doesn't tell me that I need to do something in the mi_dashboard project
3. If I change/delete a column I don't know what the effects are and if I'm going to break whatever is consuming the data from that database
4. If a schema migrates, do I need to run the ETL for the previous years as well? Do I need to do a data migrations for the old data?

Etc.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+        'actual_uk_region_names',
+        'estimated_land_date',
+    }
+
+    def get_source_query(self):
+        """Get the query set."""
+        return InvestmentProject.objects.annotate(
+            project_reference=get_project_code_expression(),
+            status_name=get_choices_as_case_expression(InvestmentProject, 'status'),
+            project_status=get_collapse_project_status_expression(),
```
reupen: Both `status_name` and `project_status` are pulled through, but the names don't make it clear what the difference is. Perhaps this one could be called `status_collapsed` or similar?

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+    year = ExtractYear(adjusted_land_date)
+    return Concat(
+        Cast(year, CharField()),
+        Value('/'),
+        Cast(year + 1, CharField()),
+    )
+
+
+def get_replace_values_expression(field, value_map, default, output_field):
```
reupen: I don't think `replace` is a good term to use here, it makes it sound like string replacement – something like `map` might work better. Alternatively, something like `get_case_expression` would keep it closer to the SQL terminology.

datahub/mi_dashboard/utils.py
elcct: Mainly so it is a bit easier to test.

reupen:
If `MI_FDI_DASHBOARD_COUNTRY_URL_PART` is required, the CircleCI config for the acceptance tests in the FE needs updating.

changelog/interaction/mi-dashboard.internal
```diff
@@ -0,0 +1 @@
+Daily MI dashboard feed was added.
```
reupen: Could we add a bit more detail to this?

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,169 @@
...
+            ~Q(level_of_involvement_id=Involvement.no_involvement.value.id),
+            then=Value('involved'),
+        ),
+        output_field=CharField(),
+    )
+
+
+def get_project_sector_expression():
+    """Get project sector value."""
+    return Case(
```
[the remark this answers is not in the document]
elcct: First I wanted to do it the way it is in our spreadsheet powering the dashboard so that it is the same and then.

datahub/mi_dashboard/models.py
elcct: I called it FDIDashboard because that's how the thing using it is called but I am also fine with MIInvestmentProject. Updated. Thanks!

config/settings/common.py
elcct: Indeed that is a path as well. I have renamed it. Thanks!

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+            ),
+            investor_company_global_account_manager=get_full_name_expression(
+                'investor_company__one_list_account_owner',
+            ),
+            client_relationship_manager_name=get_full_name_expression(
+                'client_relationship_manager',
+            ),
+            project_manager_name=get_full_name_expression('project_manager'),
+            project_assurance_adviser_name=get_full_name_expression('project_assurance_adviser'),
+            project_fdi_value=get_null_string_if_empty_expression('fdi_value__name'),
```
reupen: Won't `fdi_value__name` be null rather than an empty string?

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,169 @@
...
+        When(**{field: key}, then=value) for key, value in value_map.items()
+    )
+    return Case(
+        *whens,
+        default=default,
+        output_field=output_field,
+    )
+
+
+def get_clean_financial_year_expression():
```
reupen: Is it not possible to just make `get_temp_financial_year_from_land_date_expression()` output the right thing if `land_date` is null?

datahub/mi_dashboard/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+logger = get_task_logger(__name__)
+
+
+@shared_task(
+    bind=True,
+    acks_late=True,
+    priority=7,
+    max_retries=5,
+    autoretry_for=(Exception,),
+    default_retry_delay=60,
```
reupen: How long does this take? It should probably be routed to the long-running queue (`queue='long-running'`).

config/settings/common.py
```diff
@@ -328,13 +376,16 @@
...
+MI_FDI_DASHBOARD_COUNTRY_URL_PATH = env('MI_FDI_DASHBOARD_COUNTRY_URL_PART')
 DATAHUB_FRONTEND_URL_PREFIXES = {
     'company': f'{DATAHUB_FRONTEND_BASE_URL}/companies',
     'contact': f'{DATAHUB_FRONTEND_BASE_URL}/contacts',
     'event': f'{DATAHUB_FRONTEND_BASE_URL}/events',
     'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
     'investmentproject': f'{DATAHUB_FRONTEND_BASE_URL}/investment-projects',
     'order': f'{DATAHUB_FRONTEND_BASE_URL}/omis',
+
+    'mi_fdi_dashboard_country': f'{DATAHUB_FRONTEND_BASE_URL}/{MI_FDI_DASHBOARD_COUNTRY_URL_PATH}'
```
marcofucci: I still don't think this is a good idea, changing IDs that `MI_FDI_DASHBOARD_COUNTRY_URL_PATH` is hiding breaks things without notice.

datahub/mi_dashboard/tasks.py
reupen: It would be good to use `retry_backoff=60` rather than `default_retry_delay=60` so if it fails multiple times the delay between retries increases (gives more time for temporary problems to resolve).

datahub/mi_dashboard/query_utils.py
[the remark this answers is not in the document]
reupen: It's not really an optimisation – it makes sense to go straight to the data source instead of joining on ` : ` and then splitting again.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+        'investment_type_name',
+        'status_name',
+        'level_of_involvement_name',
+        'level_of_involvement_simplified',
+        'possible_uk_region_names',
+        'actual_uk_region_names',
+        'estimated_land_date',
+    }
+
+    def get_source_query(self):
```
reupen: There is some kind of attribute conflict with this query:

```python
In [4]: qs = ETLInvestmentProjects(None).get_source_query()                                                                                                                                                                                                                                                                                                         

In [5]: qs.first()                                                                                                                                                                                                                                                                                                                                                  
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-01983248c17f> in <module>
----> 1 qs.first()

~/uktrade/data-hub/venvs/leeloo-3.6/lib/python3.6/site-packages/django/db/models/query.py in first(self)
    603     def first(self):
    604         """Return the first object of a query or None if no match is found."""
--> 605         for obj in (self if self.ordered else self.order_by('pk'))[:1]:
    606             return obj
    607 

~/uktrade/data-hub/venvs/leeloo-3.6/lib/python3.6/site-packages/django/db/models/query.py in __iter__(self)
    266                - Responsible for turning the rows into model objects.
    267         """
--> 268         self._fetch_all()
    269         return iter(self._result_cache)
    270 

~/uktrade/data-hub/venvs/leeloo-3.6/lib/python3.6/site-packages/django/db/models/query.py in _fetch_all(self)
   1184     def _fetch_all(self):
   1185         if self._result_cache is None:
-> 1186             self._result_cache = list(self._iterable_class(self))
   1187         if self._prefetch_related_lookups and not self._prefetch_done:
   1188             self._prefetch_related_objects()

~/uktrade/data-hub/venvs/leeloo-3.6/lib/python3.6/site-packages/django/db/models/query.py in __iter__(self)
     67             if annotation_col_map:
     68                 for attr_name, col_pos in annotation_col_map.items():
---> 69                     setattr(obj, attr_name, row[col_pos])
     70 
     71             # Add the known related objects to the model, if there are any

AttributeError: can't set attribute
```

It would be good to resolve that as it could make debugging a pain.

datahub/mi_dashboard/utils.py
reupen: In any case `mi_investment_project_pipeline` doesn't sound like a function – maybe something like `run_mi_investment_project_etl_pipeline`?

Also I don't think `utils` is the right name for this module – utils sounds like it would contain generic reusable things, while this function would only be used from the Celery task.

datahub/mi_dashboard/pipelines.py
elcct: `project_fdi_value` is correct one and `fdi_value_name` is empty when `fdi_value__name` is null, that way `project_fdi_value` can have `Null` if `fdi_value_name` is empty. I have corrected the model.

datahub/mi_dashboard/query_utils.py
elcct: Probably it is, but it was taking me too much time to figure out so I gone the easier path and hoping I could revisit the query and optimize. My goal was to get it as close to the spreadsheet first and then optimize.

datahub/mi_dashboard/pipelines.py
elcct: I have not checked that - you're right there was a problem with the attribute - `level_of_involvement_simplified` already exist on the model. I changed it to `simplified_level_of_involvement` on the destination - but I welcome other ideas. I also cleaned the query from unused annotations. Thanks!

datahub/mi_dashboard/tasks.py
elcct: The query shouldn't require too long to run, but amended the task params just in case. Thanks!

datahub/mi_dashboard/pipelines.py
elcct: True :) Fixed.

config/settings/db_router.py
elcct: That I forgot to update when moved the app :( Fixed now, thanks!

datahub/mi_dashboard/query_utils.py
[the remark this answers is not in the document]
elcct: Right, but joining and then splitting isn't optimal so changing it to the better way is optimisation? I see your point, but I would have to get familiar with the sectors inner works again and my priority was to get the data first (so I wouldn't block others) and then do it correctly. My choice was dictated by the time constraints. I'll come back to it when I get through other comments first.

datahub/mi_dashboard/pipelines.py
elcct: I have changed it to `status_collapsed` - indeed it tells a bit more about what it is. Thanks!

config/settings/common.py
elcct: Ok we are not going to change those filters, so I have removed the env variable and replaced ids with the constants. Thanks!

datahub/mi_dashboard/pipelines.py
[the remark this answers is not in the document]
elcct: I have changed it to ETLInvestmentProjectsFinancialYear and added financial_year as parameter. This way is much better. Thanks!

datahub/mi_dashboard/query_utils.py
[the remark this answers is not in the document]
elcct: I refreshed my knowledge about sectors and now I see clearly that I should have done that straight away. The query is much simpler. Thanks for help!

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,207 @@
...
+            project_reference=get_project_code_expression(),
+            status_name=get_choices_as_case_expression(InvestmentProject, 'status'),
+            project_status=get_collapse_project_status_expression(),
+            project_url=get_front_end_url_expression('investmentproject', 'pk'),
+            date_of_latest_interaction=get_aggregate_subquery(
+                InvestmentProject,
+                Max('interactions__date'),
+            ),
+            sector_name=get_sector_name_subquery('sector'),
+            project_sector=get_project_sector_expression(),
```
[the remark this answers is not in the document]
elcct: That's a good idea. Project sector comes from the spreadsheet, but at this point that could be changed as well, so I renamed it to `top_level_sector_name`. Thanks!

config/settings/common.py
elcct: I removed this one (`MI_FDI_DASHBOARD_COUNTRY_URL_PATH`) now. Thanks!

datahub/mi_dashboard/mappings.py
elcct: 1. If a new top level sector is added then the project containing it would have `No Sector Cluster Assigned` as a sector cluster. That would be technically true.
2. The answer I got is that Onshore is just a class.

I have renamed this mapping to `top_level_sector_to_sector_cluster` and sorted the keys alphabetically.

Now it seems like there would be multiple ways to tackle it and all of those would block this PR and also made it much bigger and I am not sure if this is a good idea to hurry it. 
I think it is also unlikely that top level sectors will be changed (but that is of course possible).

A solution I have in mind would be to create `SectorCluster` table with all the clusters and then a one to one nullable mapping from the Sector table. Then either fill the data manually or in a migration (but the branch doesn't have all of the sectors - so maybe only for those that exist) for all top level sectors.  But I'd rather do it another PR.

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,173 @@
...
+    subquery = Sector.objects.filter(
+        parent_id__isnull=True,
+        tree_id=OuterRef('sector__tree_id'),
+    ).values('segment')
+
+    return Subquery(subquery, CharField())
+
+
+def get_collapse_status_name_expression():
+    """Map status_detailed to status_name."""
```
marcofucci: 1. `status_detailed` doesn't show up anywhere, not sure that that is.
2. sorry guys but I'm really not sure what this is and why we need something slightly different than status_name but almost the same?

changelog/interaction/mi-dashboard.internal
elcct: Updated it with slightly more detail.

datahub/mi_dashboard/query_utils.py
elcct: I forgot to update the comment, as it was wrong. I hope it is cleared now. Field `status_collapsed` is needed so that we can aggregate the project that are in the ongoing and won status together.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,201 @@
...
+            possible_uk_region_names=get_string_agg_subquery(
+                InvestmentProject,
+                'uk_region_locations__name',
+            ),
+            actual_uk_region_names=get_string_agg_subquery(
+                InvestmentProject,
+                'actual_uk_regions__name',
+            ),
+            fdi_value_name=get_empty_string_if_null_expression('fdi_value__name'),
+            project_fdi_value=get_null_string_if_empty_expression('fdi_value_name'),
```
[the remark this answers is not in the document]
elcct: `project_fdi_value` has `Null` as its value if `fdi_value` is null.

datahub/mi_dashboard/pipelines.py
```diff
@@ -0,0 +1,201 @@
...
+        """Get the query set."""
+        qs = super().get_source_query()
+        return qs.filter(financial_year=self.financial_year)
+
+
+def run_mi_investment_project_etl_pipeline():
+    """Runs FDI dashboard data load."""
+    pipeline = ETLInvestmentProjectsFinancialYear(
+        destination=MIInvestmentProject,
+        financial_year='2018/2019',
```
marcofucci: Not sure why this value `'2018/2019'` is here and not a param. Changing this would require a code change. Also, the tests expect the financial year '2018/2019' meaning that next year we would have to change the tests as well. It's better to keep it generic.

We can have this value as a celery task param in the settings instead and have generic test.

datahub/mi_dashboard/test/test_pipelines.py
```diff
@@ -0,0 +1,99 @@
...
+
+    updated, created = etl.load()
+    assert (0, 10) == (updated, created)
+
+    dashboard = MIInvestmentProject.objects.values(*etl.COLUMNS).all()
+    for row in dashboard:
+        source_row = etl.get_rows().get(pk=row['dh_fdi_project_id'])
+        assert source_row == row
+
+    MIInvestmentProject.objects.all().delete()
```
marcofucci: I would use a fixture for this line to centralise it with also a big TODO: fixme

datahub/mi_dashboard/pipelines.py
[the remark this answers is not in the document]
reupen: Are they both needed? 

Something like `Coalesce('fdi_value__name', Value('Null'))` should be equivalent for `project_fdi_value` (since `fdi_value__name` it would be `NULL` if empty).

If they are both needed, it would be good to give them better names.

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,178 @@
...
+    """Get literal Null string if field is empty."""
+    return Case(
+        When(**{field: ''}, then=Value('Null')),
+        default=field,
+    )
+
+
+def get_empty_string_if_null_expression(field):
+    """Get empty string if field is None."""
+    return Case(
```
reupen: This could be a simple `Coalesce()`.

datahub/mi_dashboard/test/test_query_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    InvestmentProjectFactory(
+        actual_land_date=None,
+        estimated_land_date=land_date,
+    )
+    query = InvestmentProject.objects.annotate(
+        land_date=get_other_date_if_null_expression(
+            'actual_land_date',
+            'estimated_land_date',
+        ),
+        financial_year_temp=get_temp_financial_year_from_land_date_expression(),
```
marcofucci: It's cleaner to use `financial_year_temp` as param for the test instead of `land_date` and set `financial_year_temp` to that value in here.

datahub/mi_dashboard/pipelines.py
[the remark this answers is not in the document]
marcofucci: that doesn't mean much to me :-/

marcofucci:
Another design question: it seems like the output of this transformation could be available when exporting investments using the search. It would make things consistent.

datahub/mi_dashboard/pipelines.py
elcct: Good point, I guess that is because we don't plan on using different financial year for a time being and when we would then this would have been revisited anyway, but I have made it parametrised now so that is a bit more flexible. Thanks!

datahub/mi_dashboard/test/test_pipelines.py
elcct: I have replaced it with a fixture and appropriate comment.

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,175 @@
...
+            level_of_involvement_id=None,
+            then=Value('unspecified'),
+        ),
+        When(
+            level_of_involvement_id=Involvement.no_involvement.value.id,
+            then=Value('not_involved'),
+        ),
+        When(
+            ~Q(level_of_involvement_id=Involvement.no_involvement.value.id),
+            then=Value('involved'),
```
marcofucci: Would be better to use `IProjectAbstract.INVOLVEMENT` values here.

datahub/mi_dashboard/query_utils.py
elcct: I forgot about this little handy function! Updated, thanks!

datahub/mi_dashboard/query_utils.py
marcofucci: Also can you add a TODO to fix this as this logic duplicates https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/investment/models.py#L246 and it would be good to have just one version.

elcct: Good idea. I have replaced it. Thanks!

datahub/mi_dashboard/test/test_query_utils.py
elcct: Yes I agree this isn't great. I have corrected this. Thank you!

datahub/mi_dashboard/pipelines.py
elcct: I was able to remove `fdi_value_name` and project_fdi_value has now `No FDI value assigned` instead of `Null` when null.

datahub/mi_dashboard/query_utils.py
elcct: I have updated the query so that the `financial_year_temp` is no longer needed. Thanks!

datahub/mi_dashboard/test/test_pipelines.py
elcct: I think I solved it by setting `multi_db` property on `TransactionTestCase` class in `conftest.py`. I could remove that fixture now.

datahub/mi_dashboard/mappings.py
elcct: It turns out this wasn't too complicated so I added `SectorCluster` object to sectors and improved the queries. I'd run a migration to assign the cluster manually.

config/settings/common.py
```diff
@@ -335,6 +385,13 @@
     'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
     'investmentproject': f'{DATAHUB_FRONTEND_BASE_URL}/investment-projects',
     'order': f'{DATAHUB_FRONTEND_BASE_URL}/omis',
+
+    'mi_fdi_dashboard_country': (
```
reupen: If we are constructing the query string, it would be better to use `urlencode()` rather than formatting it.

config/settings/db_router.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def db_for_read(self, model, **hints):
+        """Return the database to read."""
+        if model._meta.app_config.name in settings.MI_APPS:
+            return self.MI_DATABASE_LABEL
+        return 'default'
+
+    def db_for_write(self, model, **hints):
+        """Return the database to write."""
+        if model._meta.app_config.name in settings.MI_APPS:
```
reupen: `db_for_read()` and `db_for_write()` are the same, right? Would be good to move the repeated code into another function.

datahub/metadata/models.py
```diff
@@ -19,13 +19,23 @@ class BusinessType(BaseConstantModel):
     """Company business type."""
 
 
+class SectorCluster(BaseConstantModel):
```
reupen: These are specific to investment projects, no? It would be good to make that clear (and for related classes and attributes too).

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    return Coalesce(field, Value(''))
+
+
+def get_other_field_if_null_or_empty_expression(field_a, field_b):
+    """
+    Get other field_b value if field_a is null or empty.
+
+    If field_b is null, return empty string.
+    """
+    return Coalesce(
```
reupen: You can pass more than two arguments to Coalesce, so this should be the same as:

```python
    return Coalesce(
        NullIf(field_a, Value('')),
        field_b,
        Value(''),
    )
```

if I'm not mistaken.

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,135 @@
...
+        F('land_date') - Cast(Value('3 months 5 days'), DurationField()),
+        DateField(),
+    )
+    year = ExtractYear(adjusted_land_date)
+    financial_year = Concat(
+        Cast(year, CharField()),
+        Value('/'),
+        Cast(year + 1, CharField()),
+    )
+    return Coalesce(NullIf(financial_year, Value('/')), Value(''))
```
reupen: Would something like `Case(When(land_date=None, Value('')), default=financial_year)` work?

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def get_top_level_sector_expression():
+    """Get top-level sector value."""
+    subquery = Sector.objects.filter(
+        parent_id__isnull=True,
+        tree_id=OuterRef('sector__tree_id'),
+    ).values('segment')
+
+    return Subquery(subquery, CharField())
```
reupen: Would be good to pass output_field as a keyword argument since it's declared as one.

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def get_sector_cluster_expression(field):
+    """Get sector cluster value."""
+    subquery = Sector.objects.filter(
+        parent_id__isnull=True,
+        tree_id=OuterRef(f'{field}__tree_id'),
+    ).values('cluster__name')
+
+    return Coalesce(Subquery(subquery, CharField()), Value(NO_SECTOR_CLUSTER_ASSIGNED))
```
reupen: Same comment as above about output_field.

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    financial_year = Concat(
+        Cast(year, CharField()),
+        Value('/'),
+        Cast(year + 1, CharField()),
+    )
+    return Coalesce(NullIf(financial_year, Value('/')), Value(''))
+
+
+def get_country_url():
+    """Gets an SQL expression that returns a front-end URL for an object."""
```
reupen: I think this docstring needs updating.

elcct: I have updated it to better picture what it does. Thanks!

config/settings/db_router.py
elcct: This is a good idea. I have added a common method. Thanks!

config/settings/common.py
elcct: Use of `urlencode` will make it clearer what's going on so I have added this. Thanks!

datahub/metadata/models.py
elcct: It is indeed specific to investment projects so I have renamed it to `InvestmentSectorCluster`. Thanks!

datahub/mi_dashboard/query_utils.py
elcct: I moved too quickly to do the next one and didn't realise I could do that. It is even better now. Thanks!

datahub/mi_dashboard/query_utils.py
```diff
@@ -0,0 +1,135 @@
...
+            Q(status_name='Ongoing') | Q(status_name='Won'),
+            then=Value('Ongoing / Won'),
+        ),
+        default='status_name',
+    )
+
+
+def get_project_fdi_value_expression():
+    """Get FDI value of the project or NO_FDI_VALUE_ASSIGNED if not assigned."""
+    return Case(
```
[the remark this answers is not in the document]
elcct: I have removed this function and used Coalsce directly in the query. Thanks!

datahub/mi_dashboard/query_utils.py
elcct: That works! I early on had a problem with `land_date=None` and thought I cannot do this that way but as it looks like it that was something else. Thank you it is much clearer now.

datahub/mi_dashboard/query_utils.py
elcct: I agree. Corrected. Thanks!

datahub/mi_dashboard/query_utils.py
elcct: Also corrected, thanks!

config/settings/common.py
```diff
@@ -118,13 +126,50 @@
...
+MI_DATABASE_SSLCERT = env('MI_DATABASE_SSLCERT', default=None)
+MI_DATABASE_SSLKEY = env('MI_DATABASE_SSLKEY', default=None)
+
+if MI_DATABASE_SSLROOTCERT and MI_DATABASE_SSLCERT and MI_DATABASE_SSLKEY:
+    MI_DATABASE_SSLROOTCERT_PATH = CONFIG_DIR('settings/mi/server-ca.pem')
+    MI_DATABASE_SSLCERT_PATH = CONFIG_DIR('settings/mi/client-cert.pem')
+    MI_DATABASE_SSLKEY_PATH = CONFIG_DIR('settings/mi/client-key.pem')
+
+    with open(MI_DATABASE_SSLROOTCERT_PATH, 'wb') as f:
+        f.write(base64.b64decode(env('MI_DATABASE_SSLROOTCERT')))
```
reupen: Shouldn't this be using `MI_DATABASE_SSLROOTCERT` from above instead of calling `env()` again? Similar comment for the next two below.

elcct: Oh this not caught my attention. Yes this shouldn't be called again. I have removed those calls and also updated the comment. Thanks!

datahub/mi_dashboard/models.py
```diff
@@ -0,0 +1,50 @@
+from logging import getLogger
+
+from django.conf import settings
+from django.db import models
+
+logger = getLogger(__name__)
```
marcuspp: Logger doesn't seem to be used here.

elcct: Nice find! Removed. Thanks.

PR #1300, opened 2018-12-17
Trading names API

marcofucci:
### Description of change

This adds the `trading_names` field to the company/search endpoints and to the overall search logic.

The field `trading_name` stays and acts in exactly the same way as before so that the frontend has time to migrate.

Question: do we know why sometimes we are specifying the vanilla field and the `_trigram` version like [here](https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/search/contact/views.py#L57)? It seems like we only need the `_trigram` version, or I'm missing something?

### Checklist


changelog/company/trading_names_search.api
```diff
@@ -0,0 +1,6 @@
+``GET /v3/search`` now also searches for a company's ``trading_names`` when using the ``term`` param.
+``POST /v3/search/company`` now also returns and searches for a company's ``trading_names`` when using the ``name`` param.
```
reupen: These need a blank line between them I think.

marcofucci: You are right, tested on gist.github.com

reupen:
> Question: do we know why sometimes we are specifying the vanilla field and the _trigram version like here? It seems like we only need the _trigram version, or I'm missing something?

I think those were added to provide matches when words that are less than three characters are searched for. (Not that it works brilliantly...)

marcofucci:
@reupen @elcct @marcuspp ready to be reviewed!

datahub/search/contact/views.py
```diff
@@ -51,10 +51,12 @@ class SearchContactParams:
             'name_trigram',
         ],
         'company_name': [
-            'company.name',
+            'company.name',  # to find 2-letter words
```
reupen: Although it's a bit broken here because `company.name` and `company.trading_name` are using `SortableCaseInsensitiveKeywordText`.

docs/Elasticsearch migrations.md
```diff
@@ -81,4 +81,4 @@ detection of this in the `migrate_es` command, as it’s not easy to automatical
 detect for various reasons.)
 
 To do this, run `./manage.py init_es --model=<model name> --force-update-mapping`
-followed by `./manage.py sync_es --model=<model name>`.
+followed by `./manage.py sync_es --model=<model name>`.
```
reupen: Was this change intended?

marcofucci: nope, will remove when I rebase

datahub/search/contact/views.py
marcofucci: Better if I remove the comments?

reupen: Maybe just the ones that are broken?

PR #1301, opened 2018-12-18
Add turnover and is_turnover_estimated

marcofucci:
### Description of change

This adds the fields `turnover` and `is_turnover_estimated` to the model and the GET company endpoints.

They will only be set on records with a `duns_number` via the django admin for now.

I haven't added them to the search params or to the response body of the search endpoints because we don't need them there and it would be nice to keep the search free from things we don't want.
I can change this if you guys think they should be returned instead.

The export logic uses the turnover (in USD) if it's define or the turnover range otherwise.

### Checklist


changelog/company/turnover.api
```diff
@@ -0,0 +1 @@
+``GET /v3/company`` and ``GET /v3/company/<uuid:pk>``: The read-only fields ``turnover`` and ``is_turnover_estimated`` where added and will only be set when ``duns_number`` is not empty. The value of ``turnover`` is in USD.
```
reupen: Typo here ('where' should be 'were').

datahub/company/models/company.py
```diff
@@ -134,6 +134,17 @@ class Company(ArchivableModel, BaseModel, CompanyAbstract):
     turnover_range = models.ForeignKey(
         metadata_models.TurnoverRange, blank=True, null=True,
         on_delete=models.SET_NULL,
+        help_text='Not used when duns_number is set. In that case, use turnover instead.',
+    )
+    turnover = models.PositiveIntegerField(
```
reupen: This will use the `integer` type which goes up to `2147483647`. If this is in USD, I'm not sure that that's big enough (=$2.15 bn).

marcofucci: Good point, I've switched to `BigIntegerField`, thanks!

changelog/company/turnover.db
```diff
@@ -0,0 +1 @@
+The columns ``turnover (int NULL)`` and ``is_turnover_estimated (bool NULL)`` were added to the table ``company_company``. They should only be used as replacement for ``turnover_range`` when the field ``duns_number`` is set.
```
elcct: I think that should be `bigint`.

PR #1302, opened 2018-12-18
 Enforce permissions in OMIS order assignee and subscriber endpoints

reupen:
### Description of change

This enforces permissions when getting and updating the assignees and subscribers for an OMIS order.

It was a bit unfortunate that I had to add `queryset` to the views just for `DjangoCrudPermission`, but the alternative was to write a new permission class which I'm not sure is warranted.

### Checklist


datahub/omis/order/test/views/test_order_assignees.py
```diff
@@ -16,6 +16,32 @@
 class TestGetOrderAssignees(APITestMixin):
     """Tests related to getting the list of advisers assigned to an order."""
 
+    def test_access_is_denied_if_without_permissions(self):
```
marcofucci: Get rid of `if`?

PR #1311, opened 2018-12-21
Add NormalizedKeyword Elasticsearch field

reupen:
### Description of change

This adds `NormalizedKeyword` as an intended replacement for `SortableCaseInsensitiveKeywordText`. The advantage is that it is a keyword field and does not use `fielddata=True`, and as well as the `lowercase` filter, it uses the `asciifolding` filter so that characters with plain Latin counterparts (e.g. Latin letters with diacritics and full-width characters like ａ) are kept together.

This field is not used anywhere yet, this just adds it so it is available to be used.

The normailiser does not need to be manually added to the index settings as elasticsearch-dsl-py does that automatically if the field is used.

Note that the [ICU analysis plugin](https://www.elastic.co/guide/en/elasticsearch/plugins/6.5/analysis-icu.html) is available both on AWS and GOV.UK PaaS and has a lot of useful stuff in it that would also improve sorting and searching. However, I'm not sure we have a strong need for it right now, so I haven't used it here.

### Checklist

### DIFF
--- a/datahub/search/fields.py
+++ b/datahub/search/fields.py
@@ -2,6 +2,16 @@ from functools import partial
 
 from elasticsearch_dsl import Keyword, Nested, Object, Text
 
+from datahub.search.elasticsearch import lowercase_asciifolding_normalizer
+
+# Keyword with normalisation to improve sorting (by keeping e, E, è, ê etc. together).
+# This should be used in preference to SortableCaseInsensitiveKeywordText
+NormalizedKeyword = partial(
+    Keyword,
+    normalizer=lowercase_asciifolding_normalizer,
+)
+# Avoid using as this uses fielddata=True. NormalizedKeyword will have better behaviour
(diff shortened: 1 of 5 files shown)


datahub/search/fields.py
```diff
@@ -2,6 +2,16 @@
 
 from elasticsearch_dsl import Keyword, Nested, Object, Text
 
+from datahub.search.elasticsearch import lowercase_asciifolding_normalizer
+
+# Keyword with normalisation to improve sorting (by keeping e, E, è, ê etc. together).
+# This should be used in preference to SortableCaseInsensitiveKeywordText
+NormalizedKeyword = partial(
```
elcct: Would it be possible to have a test showing that this indeed does that?

reupen: I've added a test for sorting to show that.

PR #1313, opened 2018-12-21
Improved Audit History

marcuspp:
### Description of change
Improved audit history. Related objects will now return their string representation instead of the object ids. I think this could be improved further by adding caching and also by adding a get_serializer method on models so the data could be consistent with other end points. So this is just a start and means changes can be surfaced.

This also fixes the issue where erroneous fields appear as changed just because the values went from a None to a empty string. These are now filtered out making the audit a true reflection of user changes.

This approach was taken from https://github.com/etianen/django-reversion/issues/508#issuecomment-213116320

### Checklist

### DIFF
--- a/changelog/company/improved_audit_history.api
+++ b/changelog/company/improved_audit_history.api
@@ -1,0 +1,1 @@
+GET ``/v3/company/<uuid:pk>/audit`` now returns string representation of any changes made to related objects rather than ids.
--- a/datahub/core/audit.py
+++ b/datahub/core/audit.py
@@ -2,6 +2,8 @@ from rest_framework.pagination import LimitOffsetPagination
 from rest_framework.viewsets import GenericViewSet
 from reversion.models import Version
 
+from datahub.core.audit_utils import diff_versions
+
 
 class AuditViewSet(GenericViewSet):
     """Generic view set for audit logs.
@@ -36,6 +38,7 @@ class AuditViewSet(GenericViewSet):
         changelog = []
         for v_new, v_old in version_pairs:
             version_creator = v_new.revision.user
+            model_meta_data = v_new.content_type.model_class()._meta
             creator_repr = None
             if version_creator:
                 creator_repr = {
@@ -51,26 +54,11 @@ class AuditViewSet(GenericViewSet):
                 'user': creator_repr,
                 'timestamp': v_new.revision.date_created,
                 'comment': v_new.revision.get_comment() or '',
-                'changes': cls._diff_versions(
-                    v_old.field_dict, v_new.field_dict,
+                'changes': diff_versions(
+                    model_meta_data, v_old.field_dict, v_new.field_dict,
                 ),
             })
-
         return changelog
-
-    @staticmethod
-    def _diff_versions(old_version, new_version):
-        changes = {}
-
-        for field_name, new_value in new_version.items():
-            if field_name not in old_version:
-                changes[field_name] = [None, new_value]
-            else:
-                old_value = old_version[field_name]
-                if old_value != new_value:
-                    changes[field_name] = [old_value, new_value]
-
-        return changes
 
 
 class _VersionQuerySetProxy:
--- a/datahub/core/audit_utils.py
+++ b/datahub/core/audit_utils.py
@@ -1,0 +1,84 @@
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+def diff_versions(model_meta, old_version, new_version):
+    """
+    Audit versions comparision with the delta returned.
+
+    For related objects only the pk is stored in the audit history.
+    A user friendly representation of the related object (the object name)
+    is retrieved if the relationship still exists.
+
+    """
+    friendly_changes = {}
+    raw_changes = _get_changes(old_version, new_version)
+
+    for db_field_name, values in raw_changes.items():
+        field = _get_field_or_none(model_meta, db_field_name)
+        field_name = field.name if field else db_field_name
+        friendly_changes[field_name] = [_make_value_friendly(field, value) for value in values]
+    return friendly_changes
+
+
+def _get_changes(old_version, new_version):
+    """Compares dictionaries returning the delta between them."""
+    changes = {}
+
+    for field_name, new_value in new_version.items():
+        if field_name not in old_version:
+            changes[field_name] = [None, new_value]
+        else:
+            old_value = old_version[field_name]
+            if _are_values_different(old_value, new_value):
+                changes[field_name] = [old_value, new_value]
+    return changes
+
+
+def _are_values_different(old_value, new_value):
+    """Checks if the two values are different whilst treating a blank string the same as a None."""
+    old_value = old_value if old_value != '' else None
+    new_value = new_value if new_value != '' else None
+    return old_value != new_value
+
+
+def _get_field_or_none(model_meta, db_column_name):
+    """Gets a model field for a given model meta, if the field cannot be found returns None."""
+    try:
+        return model_meta.get_field(db_column_name)
+    except models.FieldDoesNotExist:
+        return None
+
+
+def _make_value_friendly(field, value):
+    """
+    Checks field and if required retrieves the object name from related model.
+
+    If the field is None or not a related field then the value can be
+    returned as the value is not an object pk.
+
+    For related objects the object name is then retrieved, for many to many and
+    one to many all values need to be retrieved individually.
+
+    """
+    if not field or not field.is_relation:
+        return value
+
+    if field.many_to_many or field.one_to_many:
+        return [
+            _get_object_name_for_pk(
+                field.related_model, one_value,
+            ) for one_value in value
+        ]
+    return _get_object_name_for_pk(field.related_model, value)
+
+
+def _get_object_name_for_pk(model, pk):
+    """
+    Gets the name for a given object pk or returns the pk if it cannot be found.
+    """
+    try:
+        result = model.objects.get(pk=pk)
+    except (model.DoesNotExist, ValueError, TypeError, ValidationError):
+        return pk
+    return str(result)
--- a/datahub/core/test/test_audit_utils.py
+++ b/datahub/core/test/test_audit_utils.py
@@ -1,0 +1,121 @@
+import unittest.mock
+
+import pytest
+
+from datahub.core.audit_utils import (
+    _are_values_different,
+    _get_changes,
+    _get_field_or_none,
+    _get_object_name_for_pk,
+    _make_value_friendly,
+    diff_versions,
+)
+from datahub.core.test.support.factories import BookFactory
+from datahub.core.test.support.models import Book
+
+
+pytestmark = pytest.mark.django_db
+
+
+def test_audit_diff_versions():
+    """Test audit diff versions."""
+    given = {
+        'old': {
+            'date_published': 'val1',
+            'name': 'val2',
+            'contact_email': 'contact',
+            'old_email': 'hello',
+            'telephone_number': None,
+            'old_field': '1',
+        },
+        'new': {
+            'date_published': 'val1',
+            'name': 'new-val',
+            'genre': 'added',
+            'telephone_number': '',
+            'old_field': '2',
+        },
+    }
+
+    expected = {
+        'name': ['val2', 'new-val'],
+        'genre': [None, 'added'],
+        'old_field': ['1', '2'],
+    }
+
+    result = diff_versions(Book._meta, given['old'], given['new'])
+    assert result == expected
+
+
+@pytest.mark.parametrize(
+    'old_value,new_value,expected_result',
+    (
+        ('', None, False),
+        ('', False, True),
+        ('hello', 'hello', False),
+        (None, None, False),
+        ('', '', False),
+    ),
+)
+def test_are_values_different(old_value, new_value, expected_result):
+    """Tests two values are different but that a blank string is treated as a None."""
+    assert _are_values_different(old_value, new_value) == expected_result
+
+
+@pytest.mark.parametrize(
+    'old_version,new_version,expected_result',
+    (
+        ({}, {}, {}),
+        ({}, {'hello': 1}, {'hello': [None, 1]}),
+        ({'hello': 1}, {}, {}),
+        ({'hello': None}, {'hello': ''}, {}),
+    ),
+)
+def test_get_changes(old_version, new_version, expected_result):
+    """Tests get changes between two dictionaries"""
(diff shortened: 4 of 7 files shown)


datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+def diff_revisions(model_meta_data, old_revision, new_revision):
```
marcofucci: I think these are versions not revisions.

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    changes = {}
+
+    for field_name, new_value in new_version.items():
+        if field_name not in old_version:
+            changes[field_name] = [None, new_value]
+        else:
+            old_value = old_version[field_name]
+            if old_value != new_value:
+                changes[field_name] = [old_value, new_value]
+    return changes
```
marcofucci: I know this is from existing code but shouldn't this take into accounts fields in old_version that don't exist in new_version any more?

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def _get_db_field_from_column_name(model_meta_data, db_field_name):
```
marcofucci: Better `_get_db_field_from_field_name` as this is not really the db column, isn't it?

datahub/core/audit_utils.py
marcuspp: I discussed this with @reupen the only time that would happen would be if we removed a field from a model.

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        return model_meta_data.get_field(db_field_name)
+    except models.FieldDoesNotExist:
+        return None
+
+
+def _get_value_for_field(field, value):
+    """Checks field type and if required retrieves friendly value from related model."""
+    if type(field) is models.ForeignKey:
+        return _get_friendly_repr_of_object_reference(field.related_model, value)
+    if type(field) is models.ManyToManyField:
```
marcofucci: Better to use https://docs.djangoproject.com/en/2.1/ref/models/fields/#attributes-for-fields-with-relations instead of `type(...) is ...`

datahub/core/audit_utils.py
marcuspp: Well value has `_id` on the end so thats why i went for column name not field name. But I didn't change the variable name so need to update to be consistent.

marcofucci: 👍 I did check the db and the official docs and they said django fields but when I shell and I call field_dict I do get the _id versions.

datahub/core/audit_utils.py
marcofucci: 👍 I see your point, it's probably not interesting for the user to see when fields get removed. Maybe you get add it to the docstring so that it's clear that + why it is like this.

marcuspp: Will do.

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def diff_versions(model_meta_data, old_revision, new_revision):
+    """Audit versions comparision with the delta returned."""
+    changes = {}
+    raw_changes = _get_changes(old_revision, new_revision)
+
+    for db_field_name, values in raw_changes.items():
+        field = _get_db_field_from_column_name(model_meta_data, db_field_name)
+        if field and list(filter(is_not_empty_or_none, values)):
```
marcofucci: Isn't this the same as `list(filter(None, values))`?

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+            changes[field.name] = [_get_value_for_field(field, value) for value in values]
```
marcofucci: I think `_get_value_for_field` should be more robust/efficient and check for falsy `value` and return it immediately if so.

datahub/core/audit_utils.py
marcuspp: No because False is a valid result and just filtering on None filters False out too.

datahub/core/audit.py
```diff
@@ -51,27 +54,12 @@ def _construct_changelog(cls, version_pairs):
                 'user': creator_repr,
                 'timestamp': v_new.revision.date_created,
                 'comment': v_new.revision.get_comment() or '',
-                'changes': cls._diff_versions(
-                    v_old.field_dict, v_new.field_dict,
+                'changes': diff_versions(
```
marcofucci: Isn't it true that `diff_versions` can return an empty `{}` when for example a field was changed and it's not available any more? If so maybe we should check that and if empty don't add the item to the changelog.

marcuspp: I disagree I think the field should always be present even if its empty as we don't omit any other fields when they are null or empty in any of the other endpoints.

datahub/core/test/test_audit_utils.py
```diff
@@ -0,0 +1,131 @@
...
+from datahub.company.test.factories import AdviserFactory
+from datahub.core.audit_utils import (
+    _get_changes,
+    _get_db_field_from_column_name,
+    _get_friendly_repr_of_object_reference,
+    _get_value_for_field,
+    diff_versions,
+    is_not_empty_or_none,
+)
+from datahub.investment.models import InvestmentProject
```
marcofucci: The core library should be as self-contained as possible and therefore not rely on `datahub.company.models.Advisor` or `datahub.investment.models.InvestmentProject`.
Have a look at the test support module for models you can use here.

changelog/company/improved_audit_history.api
```diff
@@ -0,0 +1 @@
+GET `/v3/company/<uuid:pk>/audit` now returns string representation of any changes made to related objects rather than ids.
```
reupen: Looks like the flipping between Markdown and reStructuredText has caught us again – should be double-backticks here (and in the other news fragments).

datahub/core/audit.py
marcofucci: Sorry if I wasn't clear: 
```
version a: field name='a'
version b: field name='b'
version n: field name removed
```

If you diff version b and version a you get an empty {} which will look broken to the end user as it does not make sense.

It would also be good to add this case to the tests.

marcuspp: So going back to what you asked here https://github.com/uktrade/data-hub-leeloo/pull/1313#pullrequestreview-187378204 that we should check the other way around?

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+def _get_changes(old_version, new_version):
+    """Compares dictionaries returning the delta between them."""
+    changes = {}
+
+    for field_name, new_value in new_version.items():
+        if field_name not in old_version:
+            changes[field_name] = [None, new_value]
+        else:
+            old_value = old_version[field_name]
+            if old_value != new_value:
```
reupen: If I'm not missing something, can't we make changes here to consider `''` and `None` the same, rather than filtering them out later on?

datahub/core/audit.py
marcuspp: I'd rather not for now as I might be looking at using the revision comment to store notes against an investment project that may or may not include a change. This is to make any sort of timeline easier to render. There is a chance I won't go down this route and if I don't i'll make this change then.

reupen: @marcofucci I'm not sure I'm following that scenario – you can currently generate an empty diff just by resaving a record without making any changes:

image

I think it's better not to hide events, it also keeps it consistent with modified_on (which might cause confusion otherwise).

marcofucci: Let's review this design choice after xmas. The audit was meant to be used for security reasons (to go back to it if something changed). There are also some discussions on having an activity feed kind of thing which aggregates the overall activities related to a company that seems more like what you want to do.
I'm not sure using the audit for workflow related things (if that is the intention) is the best solution (although might be the best with what we have at the moment).

marcofucci: @reupen yeah I meant that, that's pretty rubbish, isn't it? If there were no changes (which might not be true) why showing that?

reupen: Ah, I see the scenario now. I think it's a valid point as it was returning something before. The raw values might be enough in that situation?

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+def diff_versions(model_meta_data, old_revision, new_revision):
+    """Audit versions comparision with the delta returned."""
+    changes = {}
```
reupen: It would be good to call this something like `friendly_changes` to better differentiate it from `raw_changes`.

datahub/core/test/test_audit_utils.py
```diff
@@ -0,0 +1,131 @@
...
+def test_get_db_field_from_column_name(column_name, expected_type):
+    """Test getting the db field for a given column name."""
+    result = _get_db_field_from_column_name(Advisor._meta, column_name)
+    assert type(result) is expected_type
+
+
+def test_get_friendly_repr_of_object_reference_for_existing_object():
+    """Test getting representation of existing object."""
+    advisor = AdviserFactory()
+    assert _get_friendly_repr_of_object_reference(Advisor, advisor.pk) == str(advisor)
```
reupen: It would be good to group this and the next test using a class since they're both for `_get_friendly_repr_of_object_reference()`.

datahub/core/test/test_audit_utils.py
```diff
@@ -0,0 +1,131 @@
...
+@pytest.mark.parametrize(
+    'field_name,values,expected_result,number_of_times_get_repr_called',
+    (
+        ('investor_company', 'value', 'fake', 1),
+        ('description', 'value', 'value', 0),
+        ('client_contacts', ['value1', 'value2'], ['fake', 'fake'], 2),
+    ),
+)
+@unittest.mock.patch('datahub.core.audit_utils._get_friendly_repr_of_object_reference')
+def test_get_value_for_field_(
```
reupen: Was the trailing underscore in `test_get_value_for_field_` intentional?

datahub/core/audit.py
```diff
@@ -36,6 +38,7 @@ def _construct_changelog(cls, version_pairs):
         changelog = []
         for v_new, v_old in version_pairs:
             version_creator = v_new.revision.user
+            model_meta_data = v_new.content_type.model_class()._meta
```
reupen: We have used `model_meta` before (and `opts` since it's an instance of Options) but not `model_meta_data`. It would be good to be consistent – data makes it sound like a dict or similar to be.

datahub/core/audit.py
marcuspp: I think as discussed this should be reviewed at a later date, so I am leaving the behaviour as it currently is.

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    """Gets django db field for a given column name."""
+    try:
+        return model_meta_data.get_field(db_column_name)
+    except models.FieldDoesNotExist:
+        return None
+
+
+def _get_value_for_field(field, value):
+    """Checks field type and if required retrieves friendly value from related model."""
+    if not value or not field.is_relation:
```
reupen: I know the falsey check came from another comment, but is that just aimed at checking for `None`? I'm not sure why `False` and `True` would be considered differently in this context. So I'd check it's not `None` rather than if it's falsey (unless there are some other cases I'm missing, in which case I don't think it's clear enough).

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+def _get_db_field_from_column_name(model_meta_data, db_column_name):
+    """Gets django db field for a given column name."""
+    try:
+        return model_meta_data.get_field(db_column_name)
+    except models.FieldDoesNotExist:
+        return None
+
+
+def _get_value_for_field(field, value):
```
reupen: Naming – `_get_value_for_field` doesn't really describe what it does (it sounds like it looks up a value for a field). Some other options might be `_format_value` and `make_value_friendly`.

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    if field.many_to_many or field.one_to_many:
+        return [
+            _get_friendly_repr_of_object_reference(
+                field.related_model, one_value,
+            ) for one_value in value
+        ]
+    return _get_friendly_repr_of_object_reference(field.related_model, value)
+
+
+def _get_friendly_repr_of_object_reference(db_model, value):
```
reupen: Naming:

- `_get_friendly_repr_of_object_reference` – not sure what an 'object reference' is. I would use `pk` – could be something like `_get_object_name_for_pk`.
- `db_model` – personally, I don't think we need to use `db_` prefixes unless we're in the search package where it's ambiguous. Does anyone disagree?
- `value` – it's a `pk` her, right? It would be clearer to call it that (unless I'm missing some other case).

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+def diff_versions(model_meta_data, old_version, new_version):
+    """Audit versions comparision with the delta returned."""
+    friendly_changes = {}
```
elcct: What does friendly mean in this context?

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    raw_changes = _get_changes(old_version, new_version)
+
+    for db_field_name, values in raw_changes.items():
+        field = _get_db_field_from_column_name(model_meta_data, db_field_name)
```
elcct: Do we need this function at all? You are catching an exception to return None if field does not exist and then here check if that field is indeed None. Why not do the try block here?

datahub/core/audit_utils.py
elcct: I agree with above points.

datahub/core/audit_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        if field and list(filter(is_not_empty_or_none, values)):
```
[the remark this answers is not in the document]
marcuspp: Removed the filter and now use `are_values_different` function as you suggested above.

datahub/core/audit_utils.py
marcuspp: Updated the comment for the function to explain.

PR #1319, opened 2018-12-28
Don't use stale data when bulk syncing to Elasticsearch

reupen:
### Description of change

This makes a simple change to how bulk syncing works (as performed by the `sync_es` and `migrate_es` management commands) so that the server-side cursor is only used for fetching the IDs of the objects to sync and the objects themselves are fetched afresh when they are copied to Elasticsearch. This avoids stale data being copied to Elasticsearch.

This also readds prefetching of to-many fields that are copied to Elasticsearch (which in itself will improve performance).

I made the minimal changes needed to `MockQuerySet` for this – it would be nice to replace `MockQuerySet` with https://github.com/stphivos/django-mock-queries, but Django Mock Queries doesn't support `QuerySet.iterator()` at present so we can't use it yet.

### Checklist

### DIFF
--- a/datahub/search/bulk_sync.py
+++ b/datahub/search/bulk_sync.py
@@ -17,34 +17,43 @@ def sync_app(search_app, batch_size=None, post_batch_callback=None):
 
     read_indices, write_index = search_app.es_model.get_read_and_write_indices()
 
-    rows_processed = 0
+    num_source_rows_processed = 0
+    num_objects_synced = 0
     total_rows = search_app.queryset.count()
-    it = search_app.queryset.iterator(chunk_size=batch_size)
+    it = search_app.queryset.values_list('pk', flat=True).iterator(chunk_size=batch_size)
     batches = slice_iterable_into_chunks(it, batch_size)
     for batch in batches:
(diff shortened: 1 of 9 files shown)


datahub/search/bulk_sync.py
```diff
@@ -17,34 +17,43 @@ def sync_app(search_app, batch_size=None, post_batch_callback=None):
...
-                f'{model_name} rows processed: {rows_processed}/{total_rows} '
-                f'{rows_processed*100//total_rows}%',
+                f'{model_name} rows processed: {num_source_rows_processed}/{total_rows} '
+                f'{num_source_rows_processed*100//total_rows}%',
             )
 
-    logger.info(f'{model_name} rows processed: {rows_processed}/{total_rows} 100%.')
+    logger.info(f'{model_name} rows processed: {num_source_rows_processed}/{total_rows} 100%.')
+    if num_source_rows_processed != num_objects_synced:
+        logger.warning(
```
elcct: Not sure if warning is appropriate. Is this something to be worried about?

reupen: We don't routinely delete anything other than interactions, so, yes, in most cases it would be.

PR #1320, opened 2018-12-28
Add task for replacing NULLs with the default value for a model field

reupen:
### Description of change

This adds a Celery task that will update any NULL values for a model field with the default value for that same field.

The intention is to use this with e.g. `Interaction.was_policy_feedback_provided` and `Interaction.policy_feedback_notes`.

I was half-minded to put the task in the `core` app, but settled on `dbmaintenance` since it does fit in with the type of operations performed by other things in that app.

### Checklist


datahub/dbmaintenance/test/test_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        NullableWithDefaultModel.objects.bulk_create(objs)
+
+        replace_null_with_default_mock.apply_async(
+            args=('support.NullableWithDefaultModel', 'nullable_with_default'),
+            kwargs={'batch_size': batch_size},
+        )
+
+        assert replace_null_with_default_mock.apply_async.call_count == expected_batches
+        assert NullableWithDefaultModel.objects.filter(
```
elcct: Maybe it will be good to also check if the value set is indeed the default one?

datahub/dbmaintenance/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+        **{field_name: None},
+    ).values(
+        'pk',
+    )[:batch_size]
+
+    # Update the batch of rows to use the default value instead
+    num_updated = model.objects.filter(
+        pk__in=Subquery(subquery),
+    ).update(
+        **{field_name: field.default},
```
marcofucci: `field.default` can be a callable, e.g. `ArrayField`: https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/#django.contrib.postgres.fields.ArrayField

reupen: Good point. That's a bit awkward, as in theory it should be called for every row (e.g. it could be `uuid.uuid4`). So, we have a few options:

- raise an error if default is a callable
- call it once for each batch instead of every row
- call it for every row (I don't fancy doing that in this PR – if we did that it probably should be a separate task)

What do you think?

marcofucci: I would need to change it if I wanted to use it for trading_names. We can't do the second as it's not correct like you mentioned so you can do 1. and I'll have to add 3 if/when I want to use it for trading_names.

reupen: OK, I'll do 1. for now – but for `trading_names` I think we could just provide a way to override the default value in the task rather than faff about with per-row stuff.

marcofucci: Yeah was actually thinking about an input parameter for the default

elcct: I didn't realise that, nice catch!

PR #1323, opened 2018-12-31
Add lead adviser to OMIS CSV extract

elcct:
### Description of change

This adds lead adviser to OMIS CSV extract.

### Checklist

### DIFF
--- a/datahub/search/omis/test/test_views.py
+++ b/datahub/search/omis/test/test_views.py
@@ -36,6 +36,8 @@ from datahub.omis.order.test.factories import (
     OrderWithAcceptedQuoteFactory,
     OrderWithCancelledQuoteFactory,
     OrderWithOpenQuoteFactory,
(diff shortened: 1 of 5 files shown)


datahub/search/omis/test/test_views.py
```diff
@@ -529,6 +529,7 @@ def test_export(
                 'Contact link':
                     f'{settings.DATAHUB_FRONTEND_URL_PREFIXES["contact"]}'
                     f'/{order.contact.pk}',
+                'Lead adviser': order.get_lead_assignee().adviser.name,
```
reupen: It would be good to cover the case when an order doesn't have a lead adviser (assignee) as well.

elcct: Good idea. I have added tests when there is no assignee and no lead assignee. Thanks!

PR #1339, opened 2019-01-04
Remove legacy policy feedback feature from API

reupen:
### Description of change

This removes nearly all of the old version of the policy feedback feature from the API and the related permissions.

The old acceptance tests were already removed from the front end, so there should be no impact on the acceptance tests (but there is still some clean-up in the front end to do as well).

I haven't explicitly put anything in the migration to check if there are still any any interactions with the `policy_feedback` kind – but can do if we think it is warranted (they have been removed from all deployed environments already).

### Checklist


datahub/interaction/models.py
```diff
@@ -57,13 +57,10 @@ class InteractionPermission(StrEnum):
 
     view_all = 'view_all_interaction'
     view_associated_investmentproject = 'view_associated_investmentproject_interaction'
-    view_policy_feedback = 'view_policy_feedback_interaction'
```
marcofucci: There are still some references to these permissions in the docstring that would be good to remove.

datahub/interaction/permissions.py
```diff
@@ -20,19 +17,9 @@ class _PermissionTemplate(StrEnum):
 
     all = '{app_label}.{action}_all_{model_name}'
     associated_investmentproject = '{app_label}.{action}_associated_investmentproject_{model_name}'
-    policy_feedback = '{app_label}.{action}_policy_feedback_{model_name}'
     standard = '{app_label}.{action}_{model_name}'
 
 
-# Mapping from kind to permission required to add/edit/view interactions with that kind
-# (None means no additional permission required over standard interaction permissions.)
-_KIND_PERMISSION_MAPPING = {
```
marcofucci: There are still some references to this in the docstring `interaction/models.py` that would be good to remove.

marcofucci:
The frontend has some references to `POLICY_FEEDBACK_PERMISSIONS` which would be nice to get rid of.

Also, `setup-uat.sh` still has some references to `policy_feedback_group`, `policy_feedback_staff_user` etc.

datahub/interaction/models.py
reupen: Thanks, this should be done now.

datahub/interaction/permissions.py
reupen: Thanks, this should be done now.

reupen:
> The frontend has some references to POLICY_FEEDBACK_PERMISSIONS which would be nice to get rid of.
> 
> Also, setup-uat.sh still has some references to policy_feedback_group, policy_feedback_staff_user etc.

Thanks, the front end changes are covered by:

https://github.com/uktrade/data-hub-frontend/pull/1683

We will need that to be merged first, so I will mark this as blocked for the minute.

PR #1341, opened 2019-01-04
Use NormalizedKeyword instead of SortableCaseInsensitiveKeywordText instead

reupen:
### Description of change

This replaces all uses of `SortableCaseInsensitiveKeywordText` with `NormalizedKeyword` and removes `SortableCaseInsensitiveKeywordText`.

This gets rid of most uses of [fielddata](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/fielddata.html), but there are still a few more to get rid of.

In some cases neither `SortableCaseInsensitiveKeywordText` nor `NormalizedKeyword` may be appropriate, but I haven't attempted to untangle that in this PR to keep the scope to a minimum.

### Checklist


marcofucci:
@reupen do we really need `NormalizedKeyword` for all those fields? It seems like many of those could be just plain `Keyword`. Some of them could even be not indexed (like job title?)

reupen:
> @reupen do we really need NormalizedKeyword for all those fields? It seems like many of those could be just plain Keyword. Some of them could even be not indexed (like job title?)

Yes, we could optimise further, but my suggestion was that we do that afterwards as it will quickly get complicated. For now I was just trying to do a straight swap of `SortableCaseInsensitiveKeywordText` in favour of `NormalizedKeyword` to get us on the way to getting rid of `fielddata` usage.

marcofucci:
Merge problems, sorry 😬

changelog/user-normalized-keyword.internal
```diff
@@ -0,0 +1 @@
+Optimisations were made to the search models so improve performance when sorting by text fields and make the sorting order more logical in some cases.
```
elcct: > and make the sorting order more logical in some cases.
I am not sure if I like this wording, specifically "more logical" but can't think of any better.

PR #1342, opened 2019-01-04
Trading name from trading names

marcofucci:
### Description of change

This makes the `trading_name` API field follow the model field `trading_names` instead of `alias`.
This way, we can get rid of the `alias` database column without impacting the API.

The ES mapping for `trading_name` was also changed to Keyword, index=False as we don't need to index its value now that we use `trading_names` instead.

### Checklist


datahub/company/test/test_company_views.py
```diff
@@ -929,42 +927,217 @@ def test_update_headquarter_type(
...
             assert response.data['headquarter_type'] == error
 
-    # TODO: remove after alias is deleted
+
+class TestTradingNamesAndAliasForCompany(APITestMixin):
+    """
+    Tests related to trading_names, trading_name and alias.
+
+    They will be eventually deleted after the migration from alias/trading_name to
+    trading_names is completed.
```
elcct: Maybe worth adding `TODO:` so that it can be found more easily.

datahub/company/test/factories.py
```diff
@@ -38,6 +38,9 @@ class CompanyFactory(factory.django.DjangoModelFactory):
     modified_by = factory.SubFactory(AdviserFactory)
     name = factory.Faker('company')
     alias = factory.Faker('company')
+    trading_names = factory.List([
```
reupen: Not too important (since this stuff is temporary), but perhaps the first one could be made the same as alias? `factory.SelfAttribute('alias')` might work.

datahub/company/test/test_company_views.py
```diff
@@ -929,42 +927,217 @@ def test_update_headquarter_type(
...
+            'old_trading_names',
+            'trading_name_api_data',
+            'expected_alias',
+            'expected_trading_names',
+        ),
+        (
+            # help in reading the params below:
+            # (
+            #     'old value', ['old value'],  # setup
+            #     'old value',  # api data
```
reupen: I would add 'request' or 'patch' to 'api data' as I don't think API data is very specific.

datahub/company/test/test_company_views.py
```diff
@@ -929,42 +927,217 @@ def test_update_headquarter_type(
...
         response = self.api_client.patch(
             url,
             data={
-                'trading_name': new_alias,
+                'trading_name': trading_name_api_data,
             },
         )
 
         assert response.status_code == status.HTTP_200_OK
         assert response.data['trading_names'] == expected_trading_names
```
reupen: Do we not want to check `trading_name` in the response as well? Also, any particular reason for using `response.data` instead of `response.json()`?

datahub/search/company/test/test_elasticsearch.py
```diff
@@ -632,3 +620,40 @@ def test_indexed_doc(setup_es):
...
+    )
+
+    doc = ESCompany.es_document(company)
+    elasticsearch.bulk(actions=(doc, ), chunk_size=1)
+
+    setup_es.indices.refresh()
+
+    indexed_company = setup_es.get(
+        index=CompanySearchApp.es_model.get_write_index(),
+        doc_type=CompanySearchApp.name,
```
reupen: We might want to avoid querying on doc_type since there's only one doc_type per index and custom names for mapping types are deprecated.

datahub/search/company/test/test_elasticsearch.py
```diff
@@ -632,3 +620,40 @@ def test_indexed_doc(setup_es):
...
+        id=company.pk,
+    )
+
+    source = indexed_company.pop('_source')
```
reupen: Any reason for `.pop()`?

datahub/search/company/models.py
```diff
@@ -123,7 +113,7 @@ class Company(BaseESModel):
     suggest = Completion()
 
     COMPUTED_MAPPINGS = {
-        'trading_name': attrgetter('alias'),
+        'trading_name': lambda obj: dict_utils.company_dict(obj)['trading_name'],
```
reupen: I think it would be better to move that logic out of `company_dict()` rather than use `company_dict()` like this. But not too fussed since it's temporary.

datahub/company/test/factories.py
marcofucci: I'm deliberately keeping them different to spot potential problems not yet addressed but could change now I guess...

datahub/company/test/test_company_views.py
marcofucci: 👍 and yeah copy/paste

datahub/search/company/test/test_elasticsearch.py
marcofucci: `doc_type` is a required param, what do you suggest?

datahub/search/company/test/test_elasticsearch.py
marcofucci: So that I can assert the rest.

datahub/search/company/models.py
marcofucci: yeah I had it like that initially but the problem is that:
- `company_dict` is in `search.dict_utils` which is a "non-company" related module which shouldn't call a company related module.
- doesn't seem right to add a non-dict-returning function to a module called _dict_ (`company_dict`)

I don't want to have the same logic duplicated in 2 places (search.company.models.py and search.dict_utils).

Do you have another option?

datahub/search/company/test/test_elasticsearch.py
reupen: Oh, never mind then :+1: One for further down the line.

datahub/search/company/models.py
reupen: Yes, a standalone function somewhere or method or property on Company were the options that came to mind. But it's not too important anyway.

datahub/search/company/test/test_elasticsearch.py
reupen: This test doesn't use `indexed_company` again, though.

marcofucci: oh, right I see what you mean

datahub/search/company/models.py
marcofucci: Yeah not ideal but a method on Company would be forgotten (by me) and others might start using it. On the other hand, I don't know where the standalone function would go considering my first point.

datahub/company/test/factories.py
```diff
@@ -39,7 +39,8 @@ class CompanyFactory(factory.django.DjangoModelFactory):
     name = factory.Faker('company')
     alias = factory.Faker('company')
     trading_names = factory.List([
-        factory.Faker('company') for _ in range(2)
+        factory.SelfAttribute('alias'),
```
reupen: Hmm, looks like it didn't like that. Looking at the example for `factory.Dict`, maybe it needs to be `..alias`?

marcofucci: yeah it did

datahub/search/company/models.py
reupen: > doesn't seem right to add a non-dict-returning function to a module called _dict_ (`company_dict`)

I'm not sure `dict_utils` is a good name for that module, anyway. I mean we might want custom mappings for things that aren't dicts. Maybe something like `mapping_utils` would be better...

PR #1363, opened 2019-01-11
Add project manager request status to investment projects

marcuspp:
### Description of change
This is the first of a few PRs to support a new project manager request flow.
This PR just adds the fields `project_manager_request_status` and `project_manager_requested_on` to the investment endpoint and allows the `project_manager_request_status` to be set.
It does not enforce any business logic as that will come later. The data will eventually be used for SPI tracking replacing the need to create an interaction and will allow the assign pm stage to be removed. 

### Checklist


marcuspp:
@marcofucci This approach was taken because its the same way moving status' or stages works (without a linear order though). I considered a separate end point as you are correct in saying that notes will be required but thought I could make it work through the one endpoint. It might be best if I explain the complete flow to you and how it will work end to end next week.

datahub/investment/test/test_views.py
```diff
@@ -1197,6 +1203,61 @@ def test_change_stage_assign_pm_success(self):
         response = self.api_client.patch(url, data=request_data)
         assert response.status_code == status.HTTP_200_OK
 
+    @pytest.mark.parametrize(
+        'request_id,request_name,requested_on',
```
marcofucci: `request_status,expected_requested_on` and 
```
            (
                ProjectManagerRequestStatus.requested,
                datetime(2017, 4, 28, 17, 35).strftime('%Y-%m-%dT%H:%M:%SZ'),
            ),
            ...
```
with `request_status.value.id` and `request_status.value.name` in the test would be easier to read I think

datahub/investment/test/test_views.py
```diff
@@ -1197,6 +1203,61 @@ def test_change_stage_assign_pm_success(self):
...
+            'project_manager_request_status': {'id': request_id},
+        }
+        response = self.api_client.patch(url, data=request_data)
+        assert response.status_code == status.HTTP_200_OK
+
+        response_data = response.json()
+        assert response_data['project_manager_request_status']['id'] == request_id
+        assert response_data['project_manager_request_status']['name'] == request_name
+        assert response_data['project_manager_requested_on'] == requested_on
+
```
marcofucci: it would be good to add a test for creating an investment project as well with project_manager_request_status

PR #1364, opened 2019-01-11
Use name.keyword in search queries and remove name_keyword

reupen:
### Description of change

This uses `name.keyword` in all search queries and removes `name_keyword` from the search models that had that field.

I am not a fan of having logic around specific fields in `datahub.search.query_builder` (not least because not all search models have a `name` field). However, for now that logic remains there.

I am also not sure why we need to use `match_phrase` with keyword fields instead of sticking with `match` throughout but, again, that can be looked at separately.

### Checklist

### DIFF
--- a/datahub/search/companieshousecompany/models.py
+++ b/datahub/search/companieshousecompany/models.py
@@ -16,15 +16,12 @@ class CompaniesHouseCompany(BaseESModel):
     company_status = fields.NormalizedKeyword()
     incorporation_date = Date()
     name = Text(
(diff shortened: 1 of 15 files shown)


marcofucci:
Yeah not a fan of having logic around specific fields in datahub.search.query_builder either.

Looks good, are we planning to do the same for the trigram one?

reupen:
> Looks good, are we planning to do the same for the trigram one?

Yep – will do that one next...

PR #1372, opened 2019-01-14
Add subtotal_cost totals to search orders response

elcct:
### Description of change

This adds a sum aggregation for `subtotal_cost` field in the OMIS order search.
~The sum must be calculated after the rows have been filtered that's why it is now possible to specify if filtering should occur before or after aggregation (`post_filter` boolean argument). This functionality should be removed once we remove the existing aggregations, because then we should just be able to use the filter instead of post_filter. This will be done in a separate PR.~
The sum is specific to OMIS only and there are no plans to use it anywhere else that's why the query has been modified in the view class of the order search.

### Checklist

### DIFF
--- a/datahub/search/omis/test/test_views.py
+++ b/datahub/search/omis/test/test_views.py
@@ -113,6 +113,32 @@ def setup_data(setup_es):
         setup_es.indices.refresh()
 
 
+@pytest.fixture
+def setup_subtotal_cost_data(setup_es):
+    """
+    Setup Order data for total subtotal cost test.
+
+    Order will have a reference that is a string representation of the resulting subtotal_cost.
+    Subtotal cost must be three-digit because the test is choosing orders by their
+    reference. The reference filter is built with a trigram so this way we can
+    avoid ambiguity.
+    """
+    subtotal_costs = (100, 200, 300, 400, 500)
+
+    for subtotal_cost in subtotal_costs:
+        order = OrderFactory(
+            reference=str(subtotal_cost),
+            discount_value=0,
+            assignees=[],
+        )
+        OrderAssigneeFactory(
+            order=order,
+            estimated_time=subtotal_cost * 60 // order.hourly_rate.rate_value,
+        )
+
+    setup_es.indices.refresh()
+
+
 class TestSearchOrder(APITestMixin):
     """Test specific search for orders."""
 
@@ -327,6 +353,50 @@ class TestSearchOrder(APITestMixin):
         assert [
             item['reference'] for item in response.json()['results']
         ] == results
+
+    @pytest.mark.parametrize(
+        'post_data,expected_total_subtotal_cost',
+        (
+            (  # no filter => return all records
+                {},
+                1500,
+            ),
+            (
+                {'reference': ['100', '200']},
+                300,
+            ),
+            (  # order with '150' reference doesn't exist
+                {'reference': ['500', '150']},
+                500,
+            ),
+            (
+                {'reference': ['100', '200', '500']},
+                800,
+            ),
+            (  # order with '50' reference doesn't exist
+                {'reference': ['100', '50', '300', '400']},
+                800,
+            ),
+            (  # order with such reference doesn't exist
+                {'reference': '900'},
+                0,
+            ),
+        ),
+    )
+    def test_search_orders_total_net_cost_value(
+        self,
+        setup_subtotal_cost_data,
+        post_data,
+        expected_total_subtotal_cost,
+    ):
+        """Test that search orders result contains correct total subtotal_cost."""
+        url = reverse('api-v3:search:order')
+        response = self.api_client.post(url, post_data)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        subtotal_cost_total = response_data['summary']['total_subtotal_cost']
+        assert subtotal_cost_total == expected_total_subtotal_cost
 
     def test_filter_by_company_id(self, setup_data):
         """Test that orders can be filtered by company id."""
--- a/datahub/search/omis/views.py
+++ b/datahub/search/omis/views.py
@@ -75,6 +75,25 @@ class SearchOrderParams:
 class SearchOrderAPIView(SearchOrderParams, SearchAPIView):
     """Filtered order search view."""
 
+    subtotal_cost_field = 'subtotal_cost'
+
+    def get_base_query(self, request, validated_data):
+        """Enhance entity query with the total subtotal cost."""
+        base_query = super().get_base_query(request, validated_data)
+        base_query.aggs.bucket(self.subtotal_cost_field, 'sum', field=self.subtotal_cost_field)
+        return base_query
+
+    def enhance_response(self, results, response):
+        """Enhance response with total subtotal cost."""
+        summary = {}
+
+        if self.subtotal_cost_field in results.aggregations:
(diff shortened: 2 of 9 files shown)


datahub/search/omis/test/test_views.py
```diff
@@ -336,6 +336,50 @@ def test_search(self, setup_data, data, results):
             item['reference'] for item in response.json()['results']
         ] == results
 
+    @pytest.mark.parametrize(
+        'data,results,expected',
```
marcofucci: `expected_net_cost`?

marcofucci:
Why `net_cost` without VAT? Also, if we're not interested in VAT we should probably use `subtotal_cost` as very old orders have discounts.

datahub/search/omis/test/test_views.py
```diff
@@ -336,6 +336,50 @@ def test_search(self, setup_data, data, results):
...
+            (  # search by subtotal_cost
+                {'subtotal_cost': 2000},
+                ['efgh'],
+                2000,
+            ),
+            (  # search by total_cost
+                {'total_cost': 2400},
+                ['efgh'],
+                2000,
+            ),
```
marcofucci: I'm not sure these params are representative. By looking at them is seems like you are testing that searching by term is working instead of checking if the aggregated net_cost is correct.

datahub/search/omis/test/test_views.py
marcofucci: Also, `data` could be renamed to `post_data` or something like that.

datahub/search/omis/views.py
```diff
@@ -14,6 +14,9 @@
 from datahub.omis.payment.models import Refund
 from datahub.search.omis.models import Order
 from datahub.search.omis.serializers import SearchOrderSerializer
+from datahub.search.query_builder import (
+    get_search_by_entity_query,
```
marcofucci: Fits in one line

datahub/search/omis/views.py
```diff
@@ -75,6 +80,41 @@ class SearchOrderParams:
...
+        permission_filters = self.search_app.get_permission_filters(request)
+
+        query = get_search_by_entity_query(
+            entity=self.entity,
+            term=validated_data['original_query'],
+            filter_data=filter_data,
+            composite_field_mapping=self.COMPOSITE_FILTERS,
+            permission_filters=permission_filters,
+            ordering=validated_data['sortby'],
+            post_filter=False,
```
marcofucci: I think `use_post_filter` would be a better name.

datahub/search/query_builder.py
```diff
@@ -76,13 +76,14 @@ def get_basic_search_query(
...
-        ordering=None,
-        aggregation_fields=None,
+    term=None,
+    filter_data=None,
+    composite_field_mapping=None,
+    permission_filters=None,
+    entity=None,
+    ordering=None,
+    aggregation_fields=None,
+    post_filter=True,
```
marcofucci: I think `use_post_filter` would be a better name.

datahub/search/query_builder.py
```diff
@@ -109,9 +110,9 @@ def get_search_by_entity_query(
     if permission_query:
         s = s.filter(permission_query)
 
-    s = s.post_filter(
-        Bool(must=must_filter),
-    )
+    search_filter = getattr(s, 'post_filter') if post_filter else getattr(s, 'filter')
+    s = search_filter(Bool(must=must_filter))
```
marcofucci: If we are going to get rid of `post_filter` maybe you could add a _TODO_ to make it clear it has to go?

reupen: I find this a bit confusing to be honest, and what happens if you pass `aggregation_fields` and `post_filter=False`? Maybe it would be easier to get rid of the old aggregations feature first (in a separate PR).

datahub/search/omis/views.py
```diff
@@ -75,6 +80,41 @@ class SearchOrderParams:
...
+    def format_aggregations(self, results, response):
+        """Format totals returned by Elasticsearch to match our API."""
+        totals = {}
+
+        for field in self.total_fields:
+            es_field = self.REMAP_FIELDS.get(field, field)
+            if es_field in results.aggregations:
+                totals[field] = results.aggregations[es_field]['value']
+
+        response['totals'] = totals
```
reupen: I think it would be simpler to just have something like:

```
`summary`: {
  `net_cost`: ...
}
```

Otherwise, if we added more aggregations of a different type, you would start having to add things `minimums` and `maximums` and it would get unnecessarily complex (and more complex calculations wouldn't necessarily fit in that structure).

datahub/search/omis/views.py
```diff
@@ -75,6 +80,41 @@ class SearchOrderParams:
...
+            query.aggs.bucket(field, 'sum', field=field)
+
+        return query
+
+    def format_aggregations(self, results, response):
+        """Format totals returned by Elasticsearch to match our API."""
+        totals = {}
+
+        for field in self.total_fields:
+            es_field = self.REMAP_FIELDS.get(field, field)
```
reupen: I'm not sure why `self.REMAP_FIELDS` needs to be used here – the field names aren't coming in via the API.

datahub/search/omis/views.py
```diff
@@ -75,6 +80,41 @@ class SearchOrderParams:
 class SearchOrderAPIView(SearchOrderParams, SearchAPIView):
     """Filtered order search view."""
 
+    total_fields = ('net_cost',)
+
+    def get_base_query(self, request, validated_data):
```
reupen: This repeats quite a bit of logic from SearchAPIView. It would be good to avoid that.

datahub/search/query_builder.py
elcct: I have added a TODO comment. I'd deal with the post_filter in a separate PR.

datahub/search/omis/test/test_views.py
elcct: I have replaced the test so that it is now testing the total subtotal cost.

elcct:
The requirement is to have a total net value. I have replaced it with `subtotal_cost`. Thanks for the hint!

datahub/search/query_builder.py
```diff
@@ -107,9 +108,10 @@ def get_search_by_entity_query(
     if permission_query:
         s = s.filter(permission_query)
 
-    s = s.post_filter(
-        Bool(must=must_filter),
-    )
+    # TODO: post_filter should be removed as we no longer aggregate by the filter fields.
+    search_filter = getattr(s, 'post_filter') if use_post_filter else getattr(s, 'filter')
```
reupen: I still don't follow – why don't we replace `post_filter` with `filter` first instead of adding `use_post_filter`?

elcct: As I got a number of tests failing I figured out it will be better to deal with it separately to avoid a large PR. I gave it another go and it turns out the amount of code to change is not too bad so I have replaced the `post_filter` from the entity search with the `filter`.

datahub/search/omis/test/test_views.py
```diff
@@ -328,6 +354,50 @@ def test_search(self, setup_data, data, results):
...
+            (  # order with '50' reference doesn't exist
+                {'reference': ['100', '50', '300', '400']},
+                800,
+            ),
+            (  # order with such reference doesn't exist
+                {'reference': '900'},
+                0,
+            ),
+        ),
```
marcofucci: very nice!

PR #1376, opened 2019-01-15
Add policy area filter to interaction search

reupen:
### Description of change

This adds policy areas to the interaction search model and adds a filter for it to interaction search.

### Checklist

### DIFF
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -9,6 +9,11 @@ def id_name_dict(obj):
     }
 
 
+def id_name_list_of_dicts(manager):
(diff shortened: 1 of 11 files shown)


datahub/search/dict_utils.py
```diff
@@ -9,6 +9,11 @@ def id_name_dict(obj):
     }
 
 
+def id_name_list_of_dicts(manager):
+    """Creates a list of dicts with ID and name keys from a manager."""
```
elcct: In other places we use `id` in the comments. Maybe it would be good to make it consistent?

reupen: In that case, we should quote them, because otherwise it looks like a typo.

PR #1385, opened 2019-01-16
Add activities to investment projects to add notes support

marcuspp:
### Description of change
This PR provides the mechanism to supply a note with any update or change to an investment project.
The notes are purposefully omitted from the Investment project response as I believe this will clutter the response.
The notes can be viewed within the audit history.
A separate PR will add a notes end point to support updates.

### Design comments
 - Added a `get_serialized_data` function to the InvestmentActivity. This is a design pattern I have used on another project and I believe will simplify the code base when fleshing out the audit history in future.
- Using the `add_meta` on the reversion to add the note is not ideal but allows the note to be associated with the correct reversion and will allow us to proceed with the assign pm work.
- Placing the note in its own table means we can add `note types` at a later date if we wish and de-couples it as much as possible for django-reversion.

### Checklist


marcofucci:
@marcuspp I've had a look at your PR from a design point of view first as we agreed.

Some comments:

- I'm not 100% sure that `note` is the best name for the API data field as it's not intuitive that it's not part of the entity but relates to the action and that's why you don't get it in the response body. I'm not sure we can find a better name but I was thinking about something similar to `action_note: ''` or `metadata: { 'note': ...}` or `revision_note: ''`?

- I actually think that it would be clearer to move the logic behind `get_serialized_data` on the `IProjectAuditViewSet` class as the model doesn't know anything about serialising data for the API and we are not using this convention anywhere else. I think it should be really easy to move things?

The underlying logic looks nice though!

I have not gone through the code in detail just yet, I'll wait for your reply + some more time on my side :-/

datahub/investment/models.py
```diff
@@ -590,6 +592,26 @@ def __str__(self):
...
 
+@reversion.register_base_model()
+class InvestmentNote(BaseModel):
+    """An investment note."""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    revision = models.OneToOneField(Revision, on_delete=models.SET_NULL, null=True, blank=True)
+    note = models.TextField()
+    investment_project = models.ForeignKey(
+        InvestmentProject, on_delete=models.CASCADE, related_name='notes',
```
reupen: For new code, in multi-line expressions we usually use one item per line (see examples on https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Coding%20style.md#indentation). (There might be some old code about that doesn't comply with that, though.)

reupen:
On the first point, I think prefixing with an `_` might also help to signify that it's not part of the main object. 'Change' is another option for describing the note (as well as 'action' and 'revision'). I like the idea of using a sub-object, but I'm not sure `metadata` makes it clear that it's metadata about the change rather than metadata about the object.

datahub/investment/test/factories.py
```diff
@@ -161,3 +161,14 @@ class InvestmentProjectTeamMemberFactory(factory.django.DjangoModelFactory):
...
     class Meta:
         model = 'investment.InvestmentProjectTeamMember'
+
+
+class InvestmentNoteFactory(factory.django.DjangoModelFactory):
+    """Investment note factory."""
+
+    investment_project = factory.SubFactory(InvestmentProjectFactory)
+
+    note = 'Note'
```
reupen: We usually prefer to use `factory.Faker()` (e.g. [`factory.Faker('text')`](https://faker.readthedocs.io/en/master/locales/en_GB.html#faker-providers-lorem)) so that the data is a bit more realistic and that factories can be used for generating test data. (If a specific test needs a specific value, we set that in the test.)

datahub/investment/test/test_views.py
```diff
@@ -467,6 +469,11 @@ def test_create_project_complete_success(self):
             == str(project_manager_request_status_id)
         )
 
+        assert 'note' not in response_data
```
reupen: I think it would probably be clearer to add separate tests for the note functionality (similar to the versioning tests). It would be good to test that a note isn't created if the operation fails as well.

datahub/investment/serializers.py
```diff
@@ -171,7 +174,45 @@ class Meta:
...
+            self._create_investment_note(instance, note)
+        return super().update(instance, validated_data)
+
+    def _create_investment_note(self, instance, note):
+        """
+        Creates the InvestmentNote and associates it with the relevant revision.
+        """
+        reversion.add_meta(
+            InvestmentNote, note=note, investment_project=instance,
+            created_by=self.context.get('current_user'),
```
marcofucci: Small thing but this should be multi-line expressions as well.

marcofucci:
@marcuspp review finished.

datahub/investment/serializers.py
```diff
@@ -171,7 +174,45 @@ class Meta:
         fields = ('stage', 'created_on')
 
 
-class IProjectSerializer(PermittedFieldsModelSerializer):
+class AddInvestmentNoteSerializer(serializers.ModelSerializer):
+    """Serializer to add an investment note."""
+
+    note = serializers.CharField(write_only=True, required=False)
+
+    def create(self, validated_data):
```
reupen: I might have missed it, but I think create() and update() should use `@django.db.transaction.atomic` since if one of the operations fails the other shouldn't happen. (I know we use atomic requests, but it's better that the code isn't tied to that.)

datahub/investment/serializers.py
```diff
@@ -171,7 +177,72 @@ class Meta:
...
+            raise serializers.ValidationError('Invalid data type for note')
+        if 'text' not in value:
+            raise serializers.ValidationError('No text provided for note')
+        activity_type_change = InvestmentActivityTypeValue.change.value.id
+        activity_type_id = value.get('activity_type_id') or activity_type_change
+        try:
+            value['activity_type'] = InvestmentActivityType.objects.get(pk=activity_type_id)
+        except InvestmentActivityType.DoesNotExist:
+            raise serializers.ValidationError('Invalid activity type')
+        return value
```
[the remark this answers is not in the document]
marcuspp: Good shout. I'll do this. 👍

datahub/investment/test/test_views.py
```diff
@@ -1942,6 +1945,126 @@ def test_restricted_user_can_update_project_if_associated_via_field(self, field)
...
+        url = reverse('api-v3:investment:investment-item', kwargs={'pk': project.pk})
+        risk_activity_type_id = InvestmentActivityType.risk.value.id
+
+        request_data = {
+            'note': {
+                'text': 'Project risk.',
+                'activity_type_id': risk_activity_type_id,
+            },
+        }
+        response = self.api_client.patch(url, data=request_data)
```
elcct: I am not sure about this design - the `patch` method suggests that existing note would be modified - instead, it creates a new note. Maybe a separate endpoint for managing notes would be better, at least for adding new notes to an existing investment project?

marcuspp: As I have said above in the description there will be a separate endpoint to edit notes.
The purpose of this is so a note could be associated with a change, ie. changing the stage of a project, thats the main goal of this PR.

elcct: But this test doesn't change the investment project and a note is being added.

marcuspp: True because that is possible. I could restrict the API so a change has to also be sent but don't think it is warranted.

datahub/investment/serializers.py
```diff
@@ -171,7 +177,91 @@ class Meta:
...
+class InvestmentActivityNoteOnlySerializer(serializers.ModelSerializer):
+    """
+    Serializer for validating a note that has been submitted alongside an
+    investment project change.
+    """
+
+    text = serializers.CharField(required=True)
+    activity_type_id = serializers.PrimaryKeyRelatedField(
+        queryset=InvestmentActivityType.objects.all(),
+        required=False,
```
reupen: Should null be allowed as well?

datahub/investment/serializers.py
```diff
@@ -171,7 +177,91 @@ class Meta:
...
 
-class IProjectSerializer(PermittedFieldsModelSerializer):
+class InvestmentActivityNoteOnlySerializer(serializers.ModelSerializer):
+    """
+    Serializer for validating a note that has been submitted alongside an
+    investment project change.
+    """
+
+    text = serializers.CharField(required=True)
+    activity_type_id = serializers.PrimaryKeyRelatedField(
```
reupen: Elsewhere, we use `NestedRelatedField`. Any particular reason for using `PrimaryKeyRelatedField` instead?

datahub/investment/serializers.py
```diff
@@ -171,7 +177,91 @@ class Meta:
...
+    class Meta:
+        model = InvestmentActivity
+        fields = (
+            'text',
+            'activity_type_id',
+        )
+
+    def validate(self, attrs):
+        """Adds the default value for the activity type."""
+        if not attrs.get('activity_type_id'):
```
reupen: Did you try using a callable default with the field? It also looks a bit odd as the field ends in `_id` but it's setting it to a model instance.

datahub/investment/test/test_views.py
elcct: The investment project could have a nonce / version incremented on each update and then a note could be added in a separate call (to the notes API) along with the nonce/version it refers to (this could also enable adding multiple notes describing single change - what if investment project update refers to SPI and risk? With the current model it will be difficult to describe such change and there is no mechanism to stop updating multiple areas of investment project at the same time). I think that would be cleaner as current behaviour is a bit convoluted (update creates new records). I don't feel it is tied strongly enough to the investment project to have it on the same endpoint as the investment project itself. Maybe if reasoning about it was included in the comments that could be easier and could minimise potential confusion? It is also possible that I am overthinking this. What do you guys think @marcofucci @reupen ?

marcuspp: I think that would overly complicated things including the frontend implementation for little or no gain? Where the data is stored should be irrelevant to the end user of the API.

Saying that your method would work for other areas of the platform, it just seems strange to decouple the change from the note. The note will also be included in emails in some circumstances which could complicate matters.

datahub/investment/serializers.py
marcuspp: I did look at this first however wouldn't using a `NestedRelatedField` would mean id would be included in its own `{}`

ie.
`{'note': {'activity_type': {'id': '...'}}}`

Where this was what I wanted.

`{'note': {'activity_type_id': '...'}}`

I don't mind changing this but this would be three layers.

datahub/investment/serializers.py
marcuspp: I will try a callable default, I did have issues with trying to set the default and this was the only way that I managed to get it to work properly. When setting the instance the `activity.data` actually only returns the `'id` not the model instance.

datahub/investment/serializers.py
marcofucci: With `NestedRelatedField` you can do this `{'note': {'activity_type': '<id>'}}` if I'm not wrong which is what the frontend is doing in all other cases.

datahub/investment/test/test_views.py
marcofucci: Yeah I still think it's not very intuitive and non consistent to have a `note` field which is added every time if passed in and that you can't patch on the investment project endpoint like with the other fields. 
However, if isolated to the investment project and well documented, this is not a major problem I think.

datahub/investment/serializers.py
marcuspp: @marcofucci I thought it had to be like this https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/investment/test/test_views.py#L382-L431

reupen: `NestedRelatedField` was made to accept both `{'id': <id>}` and `<id>` to simplify some logic in the front end as the data that comes back from the form POST is flat. I would guess both forms are used in the front end as there would probably be some code that pre-dates that change to `NestedRelatedField`.

marcuspp: Nice one. 👍 I wasn't aware of that. I've changed it to use the `NestedRelatedField`. Thanks @reupen @marcofucci

datahub/investment/serializers.py
marcuspp: Introducing the `NestedRelatedField` has allowed me to change this and I've moved the set default to a callable function. Thanks.

datahub/investment/serializers.py
```diff
@@ -171,7 +177,89 @@ class Meta:
...
+
+class InvestmentActivityNoteOnlySerializer(serializers.ModelSerializer):
+    """
+    Serializer for validating a note that has been submitted alongside an
+    investment project change.
+    """
+
+    text = serializers.CharField(required=True)
+    activity_type = NestedRelatedField(
+        InvestmentActivityType, required=False, default=default_activity_type,
```
marcofucci: multi-line

datahub/investment/serializers.py
```diff
@@ -171,7 +177,89 @@ class Meta:
...
 
 
-class IProjectSerializer(PermittedFieldsModelSerializer):
+def default_activity_type(*args, **kwargs):
+    """Sets the default activity type to `change` if one is not specified."""
+    activity_type_change = InvestmentActivityTypeValue.change.value.id
+    return InvestmentActivityType.objects.get(pk=activity_type_change)
+
+
+class InvestmentActivityNoteOnlySerializer(serializers.ModelSerializer):
```
marcofucci: I think it would be nicer to have just `InvestmentActivitySerializer` and move required=False with the default there instead of creating a new serializer

marcuspp: One serializer has `note` the other has `text`. I suppose I could change the serializer to use text and combine them.

datahub/investment/test/test_views.py
marcuspp: @elcct Thinking about this your idea could be introduced later on. I need this out so we can move on with other features. I will build the other endpoint for editing notes and that could eventually support the nonce / version if required.

datahub/investment/serializers.py
```diff
@@ -171,7 +177,100 @@ class Meta:
...
+    activity_type_change = InvestmentActivityTypeValue.change.value.id
+    return InvestmentActivityType.objects.get(pk=activity_type_change)
+
+
+class InvestmentActivitySerializer(serializers.ModelSerializer):
+    """Serializer for an Investment Activity."""
+
+    created_by = NestedAdviserField(read_only=True)
+    modified_by = NestedAdviserField(read_only=True)
+    text = serializers.CharField(required=True, source='note')
```
reupen: In the request the object is called `note` and the `note` field is called `text`. But the model uses the word `activity` with a field called `note`. Is there not a consistent set of terminology that could be used?

datahub/investment/serializers.py
```diff
@@ -171,7 +177,100 @@ class Meta:
...
+
+    def validate_note(self, value):
+        """
+        Validates the note field to make sure it is a dictionary and that the required text
+        field is provided. Supports an optional activity_type field.
+        If activity_type is not set then the default type of change is set.
+        """
+        activity = InvestmentActivitySerializer(data=value)
+        activity.is_valid(True)
+        return activity.data
```
reupen: Shouldn't it be using the validated data i.e. `activity.validated_data`?

datahub/investment/serializers.py
```diff
@@ -171,7 +177,100 @@ class Meta:
...
+            'id',
+            'created_on',
+            'created_by',
+        )
+
+
+class NoteAwareModelSerializer(serializers.ModelSerializer):
+    """Serializer to add an investment note."""
+
+    note = serializers.JSONField(write_only=True, required=False)
```
reupen: Is there anything now stopping us using `InvestmentActivitySerializer` for the field instead of `JSONField` (and getting rid of the `validate_note` method)?

datahub/investment/test/test_views.py
```diff
@@ -1942,6 +1945,135 @@ def test_restricted_user_can_update_project_if_associated_via_field(self, field)
...
+class TestInvestmentProjectActivities(APITestMixin):
+    """Tests for adding notes to an investment project."""
+
+    def test_create_project_with_a_note(self):
+        """Test that a project can be created with a note."""
+        adviser = AdviserFactory()
+
+        response = self.api_client.post(
+            reverse('api-v3:investment:investment-collection'),
+            data={
```
reupen: The value for `data` is quite long so I would define it as a separate variable for clarity (and to cut down on indentation).

datahub/investment/test/test_views.py
```diff
@@ -1942,6 +1945,135 @@ def test_restricted_user_can_update_project_if_associated_via_field(self, field)
...
+
+        request_data = {
+            'note': {
+                'text': 'Project risk.',
+                'activity_type': risk_activity_type_id,
+            },
+        }
+        response = self.api_client.patch(url, data=request_data)
+
+        response_data = response.json()
```
reupen: Any reason for not checking the status code?

datahub/investment/test/test_views.py
```diff
@@ -1942,6 +1945,135 @@ def test_restricted_user_can_update_project_if_associated_via_field(self, field)
...
+                },
+                'referral_source_adviser': {'id': str(adviser.id)},
+                'sector': {
+                    'id': str(constants.Sector.aerospace_assembly_aircraft.value.id),
+                },
+                'note': {'text': 'An investment note'},
+            },
+        )
+        assert response.status_code == status.HTTP_201_CREATED
+        response_data = response.data
```
reupen: We usually use `response.json()` (as it reflects what is actually returned to the client), if there isn't a reason for using `response.data` it would be good to change it here and in other tests using `response.data`.

datahub/investment/test/test_views.py
```diff
@@ -1942,6 +1945,135 @@ def test_restricted_user_can_update_project_if_associated_via_field(self, field)
...
+            mocked_save.side_effect = Exception()
+
+            with pytest.raises(Exception):
+                self.api_client.patch(url, data=request_data)
+
+        project.refresh_from_db()
+        assert project.activities.count() == 0
+
+    @pytest.mark.parametrize(
+        'note_request,expected_field_error, expected_error_message',
```
reupen: Is it possible to provide the entire expected value of `note` in the response instead of this `expected_field_error` and `expected_error_message` split? It would be clearer.

marcuspp:
@marcofucci Yes that was the reason and to be honest this is the second time I've wanted to exclude a metadata value. So maybe finding a way to do this to any metadata would be advantageous?

If its okay with you what i'll do is remove the disabled on time stamp so this PR can be approved and 
address this issue in the PR for the other endpoint. As it would be the frontend view for that endpoint that would display the metadata in a dropdown.

marcofucci:
@marcuspp sounds perfect

datahub/investment/serializers.py
marcuspp: I suppose there is no reason why the field has be called `note` in the db. So i'll change it to text, however I will leave the request object as `note` as I think that make sense. A `note` object with a `text` field.

elcct: The `text` seems to be too generic, wouldn't something like `content` be better (as the content of the note)?

marcuspp: Not sure about `content` i think I prefer `text` if I'm honest.

datahub/investment/test/test_views.py
marcuspp: I think this test is small enough and readable as it is if I'm honest and adding the full error would add little value. So I'm going to leave it as it is if thats okay.

datahub/investment/serializers.py
marcuspp: Tried this but for some reason the `text` field does not get flagged up as required when I do so one of the test fails.

reupen: Which test? I don't see any failures when doing that.

marcuspp: This one https://github.com/uktrade/data-hub-leeloo/blob/feature/add_note/datahub/investment/test/test_views.py#L2057-L2059

reupen: Oh, I see it now. It's because it's a PATCH request, so `required=True` has no effect as `partial` is set on the root serialiser. Don't think there is another way round that.

changelog/investment/notes.db
```diff
@@ -0,0 +1,3 @@
+The table ``investment_investmentactivity`` has been added.
+
+The table ``investment_investmentactivitytype`` has been added.
```
marcofucci: I would explain what these do so that people understand what the content will be.

datahub/investment/test/test_views.py
elcct: Great!

PR #1387, opened 2019-01-16
Add policy issue type filter to interaction search

reupen:
### Description of change

This adds policy issue type to the interaction search model and adds a filter for it to interaction search.

### Checklist

### DIFF
--- a/changelog/interaction/add-policy-issue-types-search-filter.api
+++ b/changelog/interaction/add-policy-issue-types-search-filter.api
@@ -1,0 +1,1 @@
+``POST /v3/search/interaction``: ``policy_issue_types`` was added as a filter, accepting one or more policy issue type IDs that results should match one of.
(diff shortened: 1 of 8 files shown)


changelog/interaction/add-policy-issue-types-search-filter.api
```diff
@@ -0,0 +1 @@
+``POST /v3/search/interaction``: ``policy_issue_types`` was added as a filter, accepting one or more policy issue type IDs that results should match one of.
```
elcct: That suggests if an interaction matches more than one of the given IDs then it won't be returned? Maybe it should say `should match at least one of.`?

reupen: I think it will be understood; the only precise way to define it is to say that the intersection should be non-empty, but I think that is less likely to be understood...

PR #1391, opened 2019-01-17
Tidy up search validation logic

reupen:
### Description of change

This:
- removes some unnecessary defaults logic in the entity search view
- moves the global search query argument validation logic to a serialiser to simplify the view and share logic with entity search

### Checklist

### DIFF
--- a/datahub/search/serializers.py
+++ b/datahub/search/serializers.py
@@ -1,6 +1,7 @@
 from rest_framework import serializers
 from rest_framework.settings import api_settings
 
+from datahub.search.apps import get_global_search_apps_as_mapping
 from datahub.search.query_builder import MAX_RESULTS
 
 
@@ -38,13 +39,6 @@ class StringUUIDField(serializers.UUIDField):
         return str(uuid)
 
 
-class LimitOffsetSerializer(serializers.Serializer):
-    """Serialiser used to validate limit/offset values in POST bodies."""
-
-    offset = serializers.IntegerField(default=0, min_value=0, max_value=MAX_RESULTS - 1)
-    limit = serializers.IntegerField(default=api_settings.PAGE_SIZE, min_value=1)
-
-
 class IdNameSerializer(serializers.Serializer):
     """Serializer to return metadata constant with id and name."""
 
@@ -52,8 +46,8 @@ class IdNameSerializer(serializers.Serializer):
     name = serializers.CharField()
 
 
-class EntitySearchSerializer(LimitOffsetSerializer):
-    """Serialiser used to validate search POST bodies."""
+class BaseSearchSerializer(serializers.Serializer):
+    """Base serialiser for basic (global) and entity search."""
 
     SORT_BY_FIELDS = []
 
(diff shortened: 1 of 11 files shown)


datahub/search/serializers.py
```diff
@@ -97,6 +92,33 @@ def validate_sortby(self, val):
...
 
 
+class BasicSearchSerializer(BaseSearchSerializer):
+    """Serialiser used to validate basic (global) search query parameters."""
+
+    SORT_BY_FIELDS = (
+        'created_on',
+        'name',
+    )
+    entity = serializers.CharField(default='company')
```
marcofucci: I would use `ChoiceField` and set the choices in the `__init__` but it's not a big deal if you prefer this way instead

datahub/search/views.py
```diff
@@ -42,43 +40,19 @@ class SearchBasicAPIView(APIView):
...
-        limit = int(request.query_params.get('limit', 100))
+        serializer = BasicSearchSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        validated_params = serializer.validated_data
 
         query = get_basic_search_query(
-            term=term,
-            entities=(search_app.es_model,),
+            term=validated_params['term'],
+            entities=(validated_params['entity'],),
```
marcofucci: I know it's not part of this PR but `entities` in `get_basic_search_query` is optional but the logic of the function expects a value for that, it would be good to add a TODO to fix that before we forget.

datahub/search/serializers.py
```diff
@@ -64,7 +58,8 @@ class SearchSerializer(LimitOffsetSerializer):
 
     DEFAULT_ORDERING = None
 
-    original_query = serializers.CharField(default='', allow_blank=True)
+    offset = serializers.IntegerField(default=0, min_value=0, max_value=MAX_RESULTS - 1)
+    limit = serializers.IntegerField(default=api_settings.PAGE_SIZE, min_value=1)
     sortby = serializers.CharField(default=None)
 
     def validate_sortby(self, val):
```
marcofucci: As you are there, can you review the body of this function?
I'd replace the first few lines with:
```
        if val is None:
            val = self.DEFAULT_ORDERING
        
        if val is None:
            return None
```

rename `val` to `value` and `sep` to `separator`?

reupen: I have made some changes to this function in the next PR, so I'll leave it in this one.

datahub/search/serializers.py
reupen: Done!

datahub/search/views.py
reupen: Added a TODO to get_basic_search_query() (also to make it only accept a single model as that's all we ever give it).

datahub/search/serializers.py
```diff
@@ -97,6 +92,37 @@ def validate_sortby(self, val):
...
+
+        Has to be done here to avoid circular imports.
+        """
+        super().__init__(*args, **kwargs)
+        self.fields['entity'].choices = tuple(get_global_search_apps_as_mapping())
+
+    def validate_entity(self, value):
+        """Validates the provided entity and looks up its corresponding model."""
+        global_search_models = get_global_search_apps_as_mapping()
+        return global_search_models[value].es_model
```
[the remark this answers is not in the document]
reupen: It's not much different from going from a pk value to a model instance. I'll write a small field and move all this logic there to make it clearer.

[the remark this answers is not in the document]
reupen: That's why I said I'll write a serialiser field :wink:

datahub/search/serializers.py
```diff
@@ -97,6 +93,50 @@ def validate_sortby(self, val):
...
+
+    default_error_messages = {
+        'invalid_choice': gettext('"{input}" is not a valid choice.'),
+    }
+
+    def get_default(self):
+        """Gets the default value for the field."""
+        default = super().get_default()
+        if isinstance(default, str):
+            return get_search_app('company').es_model
```
marcofucci: better to call `to_internal_value` here.

PR #1394, opened 2019-01-18
Remove remaining fielddata uses in search

reupen:
### Description of change

This:

* moves the sort-by field remappings from `FIELD_REMAPPING` in `datahub.search.query_builder` to the search views so it can be customised for each view
* refactors the search serialisers to return a tuple for `sortby` so that it does not need to be split and joined all the time
* uses `first_name.keyword` and `last_name.keyword` for sorting contacts instead of `first_name` and `last_name`
* removes the remaining `fielddata` uses

I am not entirely happy with putting the remapping on the view, but it was picked as there was already a similar remapping on the export views. I think there is some potential refactoring of all the various attributes we have in various places (views, search models, serialisers, search apps etc.) but that would be a separate, larger job...

### Checklist

### DIFF
--- a/datahub/search/company/views.py
+++ b/datahub/search/company/views.py
@@ -21,6 +21,9 @@ class SearchCompanyParams:
     entity = Company
     serializer_class = SearchCompanySerializer
     autocomplete_serializer_class = AutocompleteSearchCompanySerializer
+    es_sort_by_remappings = {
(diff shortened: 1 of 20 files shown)


reupen:
> We are still supporting only one sortby param I suppose?

Yes. I think we're likely to need multiple at some point, but we've got away with just one so far.

PR #1407, opened 2019-01-24
Add background task to copy data from Interaction.contact to Interaction.contacts

reupen:
### Description of change

This adds a generic task that populates a many-to-many field from a foreign key on the same model, for objects where the foreign key is not blank and there are no existing values in the many-to-many field.

As the task is generic, it's a little bit complicated (unfortunately). I was hoping to reuse it for other planned work (although this is looking unlikely now).

I have scheduled the `copy_foreign_key_to_m2m_field` task to run for `Interaction.contacts` 10 minutes after app start-up (in an attempt to automate these kind of changes). I wasn't sure whether to put that logic in the `interaction` app or `dbmaintenance`. It seemed more logical to put it in `interaction`, but perhaps more practical to put it in `dbmaintenance`. I put it in `interaction` this time, anyway.

I am not creating versions with this task since it would just show the `contacts` field being added (which doesn't seem too useful, and we also omit creating versions when making similar changes in a migration).

### Checklist


marcofucci:
@reupen I think this is unnecessarily complex. Wouldn't it be easier to have a celery task to trigger manually to migrate the data (with or without generic logic) and then in another release add a data migration that did more or less the same to make sure that all environments != production are migrated as well and production is also up-to-date?

reupen:
You mean remove the app ready logic? Anything done in a migration would need to be quick since it has less than a minute to run.

marcofucci:
Yes I mean remove the app ready logic and even the lock logic.

The data migration would be added and released after the celery task is run so it would only affect the records that could have been left behind because of some deployment weirdness (unlikely but still) + to make sure all environments (all local for example) are migrated without the need to run the celery task.

reupen:
The advisory lock would still be relevant since you could trigger the task multiple times (same is true if it were a management command). I suppose that could be made into a generic decorator to make it cleaner, though.

> so it would only affect the records that could have been left behind

It would still be checking that there are no rows that need updating and there is a join involved – honestly not sure how long it will take, are you sure it will be quick?

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +61,112 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
...
+
+@shared_task(acks_late=True)
+def copy_foreign_key_to_m2m_field(
+    model_label,
+    old_field_name,
+    new_field_name,
+    batch_size=5000,
+):
+    """
+    Task that replaces copies non-null values to a to-many field (for objects where the to-many
```
marcofucci: typo _replaces copies_

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +61,112 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
         args=(model_label, field_name),
         kwargs={'default': default, 'batch_size': batch_size},
     )
+
+
+@shared_task(acks_late=True)
+def copy_foreign_key_to_m2m_field(
+    model_label,
+    old_field_name,
+    new_field_name,
```
marcofucci: _old_ and _new_ are not that obvious as this is just copying data. It would be better to call them `source_fk_field_name` and `target_m2m_field_name` maybe?

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +61,112 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
...
+
+    # Create many-to-many objects for the batch
+    created_objects = m2m_model.objects.bulk_create(objects_to_create)
+    num_created = len(created_objects)
+
+    logger.info(
+        f'{num_created} {model_label}.{new_field_name} many-to-many objects created',
+    )
+
+    return len(objects_to_create)
```
marcofucci: wouldn't this be `num_created`?

datahub/dbmaintenance/test/test_tasks.py
```diff
@@ -140,3 +142,92 @@ def test_raises_error_on_invalid_field(self, monkeypatch, field, default, expect
...
+    def test_replaces_null_with_default(
+            self,
+            monkeypatch,
+            num_objects,
+            batch_size,
+            expected_batches,
+    ):
+        """Test that null values are replaced with the default value for the model field."""
+        copy_foreign_key_to_m2m_field_mock = Mock(
+            side_effect=copy_foreign_key_to_m2m_field,
```
marcofucci: do you need `side_effect`?

datahub/dbmaintenance/test/test_tasks.py
```diff
@@ -140,3 +142,92 @@ def test_raises_error_on_invalid_field(self, monkeypatch, field, default, expect
...
+        'num_objects,batch_size,expected_batches',
+        (
+            (10, 4, 3),
+            (10, 5, 3),
+            (11, 6, 2),
+            (11, 12, 1),
+            (0, 5, 1),
+        ),
+    )
+    def test_replaces_null_with_default(
```
marcofucci: wrong name and docstring

datahub/dbmaintenance/tasks.py
reupen: Should be exactly the same as far as I can work out. I just thought `len(objects_to_create)` was more logical in terms of working out how many input rows were processed, but can change it to `num_created` since it should make no difference.

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +61,113 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
...
+
+@shared_task(acks_late=True)
+def copy_foreign_key_to_m2m_field(
+    model_label,
+    source_fk_field_name,
+    target_m2m_field_name,
+    batch_size=5000,
+):
+    """
+    Task that copies non-null values from a freign key to a to-many field (for objects where the
```
elcct: There is a typo `freign`.

PR #1408, opened 2019-01-24
Populate company address celery task

marcofucci:
### Description of change

This adds a celery task called `populate_company_address_fields` which populates address fields from trading and registered address.

The initial version saved each company instance but it turned out to be a bit slow so I've created a second version which uses SQL UPDATEs instead.

I left both commits for you to have a look, I prefer the second (more efficient) one but please do let me know if you think that version one is better and I could revert to it.

The code could be better but it's just temporary so it's probably not worth it.

### Checklist

### DIFF
--- a/datahub/dbmaintenance/tasks.py
+++ b/datahub/dbmaintenance/tasks.py
@@ -2,7 +2,11 @@ from logging import getLogger
 
 from celery import shared_task
 from django.apps import apps
-from django.db.models import NOT_PROVIDED, Subquery
+from django.db.models import BooleanField, Case, ExpressionWrapper, NOT_PROVIDED, Q, Subquery, When
+from django.db.models import Value
+from django.db.models.functions import Coalesce
+
+from datahub.company.models import Company
 
 
 logger = getLogger(__name__)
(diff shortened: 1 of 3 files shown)


marcofucci:
yes, I realised that having a version for this doesn't really add much and it would only confuse people as data doesn't really change, it just gets reorganised from a backend point of view.

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +65,115 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
...
+    """
+    batch_size_left = batch_size
+
+    # Coalesce makes sure that both NULL and '' values are treated equally
+
+    # Unevaluated subquery to select a batch of records used to populate
+    # address from trading address
+    subquery = Company.objects.annotate(
+        address_1_normalised=Coalesce(
+            'address_1', Value(''),
```
reupen: A line break between the arguments would be more consistent with how we usually format things.

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +65,115 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
...
+            'registered_address_county', Value(''),
+        ),
+        registered_address_postcode_normalised=Coalesce(
+            'registered_address_postcode', Value(''),
+        ),
+    ).filter(
+        address_1_normalised='',
+        address_town_normalised='',
+        address_country__isnull=True,
+    ).exclude(
```
reupen: Should this exclude companies with a trading address as well?

marcofucci: the first update takes care of that case so this should never happen. The batch_size is exhausted when updating using trading address so the second update comes after all trading addresses are complete.

reupen: Yes, but the data in the table could change between the updates.

datahub/dbmaintenance/tasks.py
```diff
@@ -60,3 +65,126 @@ def replace_null_with_default(model_label, field_name, default=None, batch_size=
         args=(model_label, field_name),
         kwargs={'default': default, 'batch_size': batch_size},
     )
+
+
+@shared_task(acks_late=True)
+@transaction.atomic
```
reupen: What is the purpose of using a transaction here?

datahub/dbmaintenance/tasks.py
marcofucci: Each call takes less than 1 second so it's very unlikely but if you think it's safer I can add those

reupen: When I ran the last interaction updates in production it occasionally took much longer than a second per batch, and that was with a batch size of 1000. So I don't think it is that unlikely at all. To be honest it might be simpler as two independent tasks.

datahub/dbmaintenance/tasks.py
reupen: But what state? The first update is not invalid if the second one doesn't happen, no?

marcofucci: 👍

datahub/dbmaintenance/tasks.py
marcofucci: very unlikely because there are very few companies with trading addresses and adding/editing records doesn't happen as often but I get your point, it's not too hard to add a few more conditions.

reupen: It might also be possible to combine the two updates into one by using case expressions.

marcofucci: I think it would make it harder to read instead of simplifying things. I've added the extra bits as you suggested though.

reupen: I find it quite hard to follow at the moment to be honest, I was thinking of something along the lines of:

```python
    base_queryset = Company.objects.annotate(
        address_1_normalised=Coalesce('address_1', Value('')),
        address_town_normalised=Coalesce('address_town', Value('')),
        trading_address_1_normalised=Coalesce('trading_address_1', Value('')),
        trading_address_town_normalised=Coalesce('trading_address_town', Value('')),
        registered_address_1_normalised=Coalesce('registered_address_1', Value('')),
        registered_address_town_normalised=Coalesce('registered_address_town', Value('')),
        has_valid_trading_address=ExpressionWrapper(
            ~Q(trading_address_1_normalised='')
            & ~Q(trading_address_town_normalised='')
            & Q(trading_address_country__isnull=False),
            output_field=BooleanField(),
        ),
        has_registered_address=ExpressionWrapper(
            # Add other fields if needed
            ~Q(registered_address_1_normalised='')
            | ~Q(registered_address_town_normalised='')
            | Q(registered_address_country__isnull=False),
            output_field=BooleanField(),
        )
    )

    # Unevaluated subquery to select a batch of records
    subquery = base_queryset.filter(
        Q(has_registered_address=True) | Q(has_valid_trading_address=True),
        address_1_normalised='',
        address_town_normalised='',
        address_country__isnull=True,
    ).values(
        'pk',
    )[:batch_size]

    num_updated = base_queryset.filter(
        pk__in=Subquery(subquery),
    ).update(
        address_1=Case(
            When(has_valid_trading_address=True, then='trading_address_1'),
            default='registered_address_1',
        ),
        address_2=Case(
            When(has_valid_trading_address=True, then='trading_address_2'),
            default='registered_address_2',
        ),
        address_town=Case(
            When(has_valid_trading_address=True, then='trading_address_town'),
            default='registered_address_town',
        ),
        address_county=Case(
            When(has_valid_trading_address=True, then='trading_address_county'),
            default='registered_address_county',
        ),
        address_postcode=Case(
            When(has_valid_trading_address=True, then='trading_address_postcode'),
            default='registered_address_postcode',
        ),
        address_country=Case(
            When(has_valid_trading_address=True, then='trading_address_country'),
            default='registered_address_country',
        ),
    )
```

But up to you.

marcofucci: Sure, as I said, code could be better but not worth it as it'll last a week. 
However copy/paste will do.

datahub/dbmaintenance/test/test_tasks.py
```diff
@@ -140,3 +143,278 @@ def test_raises_error_on_invalid_field(self, monkeypatch, field, default, expect
...
+
+            trading_address_1=None,
+            trading_address_2=None,
+            trading_address_town=None,
+            trading_address_county=None,
+            trading_address_postcode=None,
+            trading_address_country_id=None,
+        )
+
+        populate_company_address_fields.apply_async()
```
reupen: We should start checking that tasks succeed in the tests (because they capture any exceptions raised). But that's not specific to this test, so we can look at it later.

PR #1409, opened 2019-01-24
MI dashboard pipeline cleanup

elcct:
### Description of change

This enhances the pipeline so that the projects that don't exist in the result of the source query are being deleted from the MI dashboard database. This enables the dashboard to have more accurate data.

The method this is done is a bit naive but I think it should be performant enough.

This also changes the time the task is run from 2 AM to ~3 AM~1 AM.

### Checklist

### DIFF
--- a/datahub/mi_dashboard/pipelines.py
+++ b/datahub/mi_dashboard/pipelines.py
@@ -93,7 +93,7 @@ class ETLBase:
 
         :returns: a list of ID values.
         """
-        return self.get_source_query().values(self.get_model()._meta.pk.name)
+        return self.get_source_query().values_list('pk', flat=True)
 
     def clean(self) -> int:
         """
@@ -103,7 +103,7 @@ class ETLBase:
         """
         # we need to materialise ids query so that Django doesn't attempt
         # to create a query across the different databases
-        ids = (row['id'] for row in self.get_ids())
+        ids = list(self.get_ids())
         deleted, _ = self.destination.objects.exclude(pk__in=ids).delete()
 
         return deleted
--- a/datahub/mi_dashboard/pipelines.py
+++ b/datahub/mi_dashboard/pipelines.py
@@ -1,4 +1,5 @@
-from typing import Tuple, Type
+from typing import List, Tuple, Type
+from uuid import UUID
 
 from django.db.models import F, Model, Value
 from django.db.models.functions import Coalesce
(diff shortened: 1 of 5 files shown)


datahub/mi_dashboard/pipelines.py
```diff
@@ -77,17 +87,40 @@ def get_rows(self) -> QuerySet:
...
         return self.get_source_query().values(*self.COLUMNS)
 
-    def load(self) -> Tuple[int, int]:
+    def get_ids(self) -> List[UUID]:
+        """
+        Get ID of rows ready to load.
+
+        :returns: a list of ID values.
+        """
+        return self.get_source_query().values(self.get_model()._meta.pk.name)
```
reupen: Are we not able to use 'pk' instead of `self.get_model()._meta.pk.name`? Also, what about using `QuerySet.values_list()` with `flat=True` (so combining both points, `self.get_source_query().values_list('pk', flat=True)`)?

datahub/mi_dashboard/pipelines.py
```diff
@@ -77,17 +87,40 @@ def get_rows(self) -> QuerySet:
...
+    def clean(self) -> int:
+        """
+        Delete rows that are no longer in the source query.
+
+        :returns: a number of deleted records
+        """
+        # we need to materialise ids query so that Django doesn't attempt
+        # to create a query across the different databases
+        ids = (row['id'] for row in self.get_ids())
+        deleted, _ = self.destination.objects.exclude(pk__in=ids).delete()
```
reupen: What is the upper limit for the number of UUIDs that this works for?

datahub/mi_dashboard/pipelines.py
elcct: Good idea. I forgot about those. Fixed!

datahub/mi_dashboard/pipelines.py
elcct: I would assume that the limit is the memory and that we would have enough to run this query. I am going to confirm if that's indeed the case.

elcct: I checked on my laptop with a 10x more ids than we need and the query executed relatively quickly given the length of the array.

elcct: This query also works in live environment for 10x larger set of ids than we need.

marcofucci: Isn't this deleting all objects for other financial years as well? If so, are you sure it's something you want to do?

elcct: I think this will be okay for now - if we add another financial year it will work provided it will be in the same query. I could add a comment so that it is clear and perhaps enhance the `ETLInvestmentProjectsFinancialYear` class to accept an array of financial years?

elcct:
This is no longer needed.

PR #1415, opened 2019-01-29
Add contacts field to API and admin site

reupen:
### Description of change

This:

* adds the `contacts` field to the interaction endpoints in the main API, and changes the `contact` field to use `Interaction.contacts` as the data source
* adds the `contacts` field to interaction admin, and removes `contact`
* updates interaction test data to set `contacts` (as well as `contact`)

Note that this shouldn't be merged until https://github.com/uktrade/data-hub-leeloo/pull/1407 has been released.

I will update search and the interaction CSV export in a separate PR to keep the size of this PR down.

#### Interaction list view in admin (fake data)

image

#### Form view in admin

image

### Checklist


datahub/interaction/serializers.py
```diff
@@ -26,6 +27,32 @@
...
+    """
+    Serialiser field that makes a to-many field behave like a to-one field.
+
+    Use for temporary backwards compatibility when migrating a to-one field to be a to-many field
+    (so that a to-one field can be emulated using a to-many field).
+
+    This isn't intended to be used in any other way as if the to-many field contains multiple
+    items, only one of them will be returned, and all of them will overwritten on updates.
+
+    TODO Remove this once contact has been removed from interactions.
```
elcct: Just cosmetic - some `TODO` like above are missing a colon (`TODO:`).

datahub/interaction/serializers.py
```diff
@@ -92,9 +155,12 @@ def validate(self, data):
...
 
-        if 'contact' in data:
-            # Note: null contacts are not allowed in the API
-            data['contacts'] = [data['contact']]
+        # Copies the first contact specific to contact (for backwards compatibility with
+        # anything consuming the database)
+        # TODO Remove following deprecation period.
+        if 'contacts' in data:
+            contacts = data['contacts']
+            data['contact'] = contacts[0] if contacts else None
```
elcct: Is there a chance that the `contact` will be a random one out of the `contacts`?

reupen: There shouldn't be when deserialising and updating `Interaction.contact`. But the contact returned by the API in `contract` could be any of them.

datahub/interaction/serializers.py
reupen: Well, I was using both styles but I don't use a colon after e.g. PS and NB so I left out colons for TODO as well. But if we're going to standardise on something, there is reStructuredText and Sphinx: http://www.sphinx-doc.org/en/master/usage/extensions/todo.html

marcofucci:
@reupen can you add a screenshot of the new django admin? It's good practice to start doing that.

reupen:
> @reupen can you add a screenshot of the new django admin? It's good practice to start doing that.

Added – we should add that to the PR template.

marcofucci:
@reupen after thinking and using it, I think using inlines for contacts would have been better as the current field is not really user friendly.

PR #1422, opened 2019-01-30
Add management command to delete all Elasticsearch indices

reupen:
### Description of change

This adds a management command to delete all Elasticsearch indices that match the configured index name prefix.

This is intended for use on GOV.UK PaaS (when required) as GOV.UK PaaS Elasticsearch does not allow deletions using wildcards.

### Checklist

### DIFF
--- a/datahub/search/management/commands/delete_all_es_indices.py
+++ b/datahub/search/management/commands/delete_all_es_indices.py
@@ -1,0 +1,56 @@
+from logging import getLogger
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from datahub.search.elasticsearch import get_client
+
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Command for deleting all Elasticsearch indexes matching the configured index name prefix."""
+
+    help = """Irrevocably deletes all Elasticsearch indices under the configured index name prefix.
+
+This is intended for use on GOV.UK PaaS as GOV.UK PaaS Elasticsearch does not allow deletions
+using wildcards.
+
+Don't use this unless you really mean to delete all of the app's indices!
+"""
+    confirm_msg = """
+This operation cannot be undone.
+Are you sure you want to do this?
+
+    Type 'yes' to continue, or 'no' to cancel: """
+
+    def add_arguments(self, parser):
+        """
+        Add no-input argument to the command.
+        """
+        parser.add_argument(
+            '--noinput', '--no-input', action='store_false', dest='interactive',
+            help='Tells Django to NOT prompt the user for input of any kind.',
+        )
+
+    def handle(self, *args, **options):
+        """Executes the command."""
+        interactive = options['interactive']
+
+        client = get_client()
+        index_statistics = client.cat.indices(index=f'{settings.ES_INDEX_PREFIX}-*', format='json')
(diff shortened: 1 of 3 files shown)


datahub/search/management/commands/delete_all_es_indices.py
```diff
@@ -0,0 +1,37 @@
...
+    help = """Irrevocably deletes all Elasticsearch indices under the configured index name prefix.
+
+This is intended for use on GOV.UK PaaS as GOV.UK PaaS Elasticsearch does not allow deletions
+using wildcards.
+
+Don't use this unless you really mean to delete all of the app's indices!
+"""
+
+    def handle(self, *args, **options):
+        """Executes the command."""
```
elcct: I am thinking if we need a measure against running this accidentally (e.g. by going up the command history and pressing enter by mistake). For example, asking user to type in the index name prefix to confirm or just typing yes as an answer to "Are you sure".

reupen: Could do, but I didn't think it was worth the effort. The same would be true if they had run `curl -XDELETE http://<es host>:<port>/<index>`.

marcofucci: Yeah I think it would be good to add some extra layer so that if somebody (a new dev for example?) runs the django command accidentally, it stops you. 
With this, it would be too easy to make mistake, using curl is not as easy as using this command.

reupen: We would still need a `--noinput` so it could be used in scripts. Do either of you have time to add that?

elcct: Perhaps this could also be added to other commands that delete? Now this looks like a body of work on its own...

marcofucci: @reupen no I don't, in that case let's keep it here until somebody has time to add that.

marcofucci: A workaround would be to check for an env var to allow that and only set that value to True in staging?

reupen: I don't think we're getting much value from making this complicated.

When we were using AWS Elasticsearch, when resetting the environment, we used to delete all indexes using curl with a wildcard. We have now migrated to GOV.UK PaaS Elasticsearch, and deleting indexes using wildcards (or `_all`) is blocked, and cannot be re-enabled. We need to provide an exact list of index names to delete.

So let's change the command to output a list of index names (or write a Bash script to do the same) and feed that into curl.

marcofucci:
@reupen I can add you as reviewer :/

datahub/search/management/commands/delete_all_es_indices.py
```diff
@@ -18,9 +18,25 @@ class Command(BaseCommand):
 
 Don't use this unless you really mean to delete all of the app's indices!
 """
+    confirm_msg = f"""
+This operation cannot be undone.
```
reupen: Is this meant to be an f-string?

reupen:
> @reupen I can add you as reviewer :/

The changes look fine :+1:

datahub/search/management/commands/delete_all_es_indices.py
marcofucci: forgot there :/

PR #1425, opened 2019-01-31
Change use of addresses in business logic

marcofucci:
### Description of change

This starts using address for internal business logic (e.g. to determine if a company is UK based).
The only exception is OMIS where the quote and the billing address are populated from the registered address if defined or the address instead as fallback.

The search and the API needs to be updated still (in a different PR).

I don't particularly like the name `compose_official_address` but I couldn't find anything better as the returned address has a different format from the input one.

### Checklist


datahub/omis/order/utils.py
```diff
@@ -1,3 +1,55 @@
...
+            model_field_name = model_field_name_template.format(prefix=field_prefix)
+            model_field_value = getattr(company, model_field_name)
+
+            setattr(self, attr_name, model_field_value)
+
+    def is_blank(self):
+        """
+        :returns: True if none of the field values are defined, False otherwise.
+        """
+        for attr_name in self.MAPPING:
```
reupen: I think this is the same as:

```python
return not any(getattr(self, attr) for attr in self.MAPPING)
```

datahub/company/views.py
```diff
@@ -55,6 +55,7 @@ class CompanyViewSet(ArchivableViewSetMixin, CoreViewSet):
         'global_headquarters__one_list_account_owner__dit_team__uk_region',
         'global_headquarters__one_list_account_owner__dit_team__country',
         'global_headquarters__one_list_tier',
+        'address_country',
```
reupen: It looks like this list of fields was alphabetical originally, but the order has been lost. I think it would be good to keep it alphabetical.

datahub/omis/order/test/test_utils.py
```diff
@@ -1,103 +1,139 @@
...
-            billing_address_country=None,
-        )
-
-        populate_billing_data(order)
-
-        assert order.billing_address_2 == ''
-        assert order.billing_address_county == ''
-        assert order.billing_address_postcode == ''
-        assert order.billing_address_country is None
+        for field_name, field_value in expected_billing_address_fields.items():
```
reupen: It might be nice to compare all the fields at once, so that if the test fails you can see the diff for the whole address instead of just one field. E.g.:

```python
        actual_billing_address = {
            field_name: getattr(order, field_name) 
            for field_name in expected_billing_address_fields
        }
        assert actual_billing_address == expected_billing_address_fields
```

datahub/omis/order/test/test_utils.py
```diff
@@ -106,9 +142,8 @@ def test_with_already_populated_billing_detail(self):
         """
```
reupen: Could this docstring be expanded a bit? I'm not sure it explains clearly enough what's being tested here.

datahub/omis/order/utils.py
```diff
@@ -1,3 +1,55 @@
...
+    def __init__(self, field_prefix, company):
+        """
+        Initialises the object from the company using the values from the fields
+        with prefix `field_prefix`.
+        """
+        for attr_name, model_field_name_template in self.MAPPING.items():
+            model_field_name = model_field_name_template.format(prefix=field_prefix)
+            model_field_value = getattr(company, model_field_name)
+
+            setattr(self, attr_name, model_field_value)
```
reupen: Not too keen on this `setattr()` in `__init__()` – it makes it a bit hard to follow.

Something like a `typing.NamedTuple` subclass with an `is_blank` method and a factory function might be cleaner

marcofucci: I tried a few options (also the NamedTuple as you suggested) but I did not like the repetition of fields and MAPPING.

marcofucci:
@elcct @marcofucci 🙏

PR #1428, opened 2019-02-01
Safeguard MI pipeline and add run pipeline management command.

elcct:
### Description of change

A change to the InvestmentProject or MIInvestmentProject without updating the pipeline would already have tests failing but this adds additional tests so that more context is given if the changes prevent the pipeline from running.

It also adds a couple of comments to give more context (for example makes note that MIInvestmentProject table shouldn't have any columns that can store PII data).

It also adds a management command that allows someone to run the pipeline manually without having to use Django shell.

### Checklist

### DIFF
--- a/changelog/run-mi-pipeline.internal
+++ b/changelog/run-mi-pipeline.internal
@@ -1,0 +1,1 @@
+A management command to run MI Dashboard pipeline if changes to the relevant models have been made was added.
--- a/datahub/mi_dashboard/test/test_pipelines.py
+++ b/datahub/mi_dashboard/test/test_pipelines.py
@@ -2,6 +2,7 @@ from datetime import date
 
 import pytest
 from django.conf import settings
+from django.core.exceptions import FieldDoesNotExist, FieldError
 
 from datahub.company.test.factories import CompanyFactory
 from datahub.core.constants import Country, FDIValue, Sector, SectorCluster, UKRegion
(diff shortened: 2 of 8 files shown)


datahub/mi_dashboard/test/test_pipelines.py
```diff
@@ -23,6 +24,33 @@
 pytestmark = pytest.mark.django_db
 
 
+def test_pipeline_can_query_investmentprojects():
+    """Test that pipeline can query investment projects."""
+    try:
+        etl = ETLInvestmentProjects(destination=MIInvestmentProject)
+        etl.get_source_query().first()
+    except (FieldDoesNotExist, FieldError) as exc:
+        pytest.fail((
```
reupen: Should be able to get rid of the inner brackets here.

datahub/mi_dashboard/test/test_pipelines.py
```diff
@@ -23,6 +24,33 @@
...
+        ))
+
+
+def test_miinvestmentproject_columns_exist():
+    """Test that destination model has required columns."""
+    destination_fields = _get_db_model_fields(MIInvestmentProject)
+    available_destination_fields = destination_fields & ETLInvestmentProjects.COLUMNS
+    missing_destination_fields = ETLInvestmentProjects.COLUMNS - available_destination_fields
+
+    assert not missing_destination_fields, ('Following required fields do not exist in the '
```
reupen: Should be able to improve the formatting here by putting a newline after `(` and before `)` so that the text can be dedented.

changelog/run-mi-pipeline.internal
```diff
@@ -0,0 +1 @@
+A management command to run MI Dashboard pipeline if changes to the relevant models have been made.
```
marcofucci: Aren't you missing _was added_? Doesn't sound like a sentence as it is.

marcofucci:
@elcct it looks good, however it only tells you when something breaks.
I think it would be good to add a doc in `docs` to explain how mi works and what to do when you add/change/remove a field in the investment project table. Please mention that you need to run the `run_pipeline` command after adjusting and deploying the change.

PR #1435, opened 2019-02-05
Add AddressSerializer to simulate nested address objects

marcofucci:
### Description of change

This adds a generic `AddressSerializer` which can be used to simulate nested address objects when the fields are actually on a model instead of a separate fk model.

First of many PRs.

### Checklist


datahub/core/serializers.py
```diff
@@ -197,3 +201,261 @@ def __init__(self, choices):
...
+                'postcode': '',
+                'country': {
+                    'id': '736a9ab2-5d95-e211-a939-e4115bead28a',
+                    'name': 'Ireland',
+                },
+            },
+        },
+
+    Please note:
+    1. None values for CharFields are not allowed, the serializer automatically translates
```
elcct: If `None` values are not allowed then that should produce an error rather than convert to ''? Maybe an information that `None` values will be converted to '' will suffice?

datahub/core/serializers.py
```diff
@@ -197,3 +201,261 @@ def __init__(self, choices):
...
+                'line_2': '',
+                'town': '',
+                'county': '',
+                'postcode': '',
+                'country': None
+            }
+        }
+        """
+        address_dict = super().to_representation(value)
+        # return None if none of the values are set
```
elcct: I think this comment doesn't add anything.

datahub/core/serializers.py
```diff
@@ -197,3 +201,261 @@ def __init__(self, choices):
...
+        return address_dict
+
+    def should_validate(self, data_combiner):
+        """
+        Returns true if the data should be validated.
+        """
+        if self.required:
+            return True
+
+        if not any(
```
elcct: Could this be just `return any(data_combiner.get_value(field.source) for field in self.fields.values())`?

datahub/core/serializers.py
```diff
@@ -197,3 +201,261 @@ def __init__(self, choices):
...
+            }
+        }
+        """
+        address_dict = super().to_representation(value)
+        # return None if none of the values are set
+        if not any(address_dict.values()):
+            return None
+
+        # for each address field, replace None with default if possible
+        for field_name, value in address_dict.items():
```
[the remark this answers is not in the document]
marcofucci: oh interesting you find this more readable, can change it.

datahub/core/serializers.py
```diff
@@ -197,3 +201,261 @@ def __init__(self, choices):
...
+
+        It populates all necessary parts (e.g. Meta model, source, fields' source).
+        """
+        # Define a custom Meta so that the Meta model can be specified as an argument
+        class MultiAddressMeta(self.Meta):
+            model = source_model
+        self.Meta = MultiAddressMeta
+
+        if 'source' not in kwargs:
+            kwargs['source'] = '*'
```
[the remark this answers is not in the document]
marcofucci: can do

datahub/core/serializers.py
marcofucci: 😬 of course

datahub/core/serializers.py
marcofucci: 😬 of course

datahub/core/serializers.py
marcofucci: yep 👍  misleading I'll change to `None` values will be converted to ''

datahub/core/serializers.py
```diff
@@ -197,3 +201,261 @@ def __init__(self, choices):
...
+        source='{source_prefix}_country',
+    )
+
+    REQUIRED_FIELDS = (
+        'line_1',
+        'town',
+        'country',
+    )
+
+    def __init__(self, source_model, *args, address_source_prefix='address', **kwargs):
```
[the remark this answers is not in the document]
marcofucci: Yeah I see what you mean, both are not ideal. I think I prefer this way because it can be subclassed easily (as it might be necessary if we want to convert investment at a later stage).

datahub/core/serializers.py
```diff
@@ -197,3 +201,253 @@ def __init__(self, choices):
...
+            secondary_address_town=''
+            secondary_address_county=''
+            secondary_address_postcode=''
+            secondary_address_country_id=None
+
+        The equivalent API response body will be:
+            'secondary_address': None
+
+        The same applies for changing the data.
+    3. If AddressSerializer has required=False, the validation is triggered only if at least
```
reupen: Is `required` the right choice? [As that's usually about omitting the key](https://www.django-rest-framework.org/api-guide/fields/#required) completely.

datahub/core/serializers.py
```diff
@@ -197,3 +201,253 @@ def __init__(self, choices):
...
+        one of the fields is passed in.
+    """
+
+    line_1 = serializers.CharField(
+        max_length=255,
```
reupen: Is there a reason for repeating 255 for each field instead of using a named constant or `settings.CHAR_FIELD_MAX_LENGTH`?

datahub/core/test/test_serializers.py
```diff
@@ -212,3 +217,935 @@ def test_parses_dates(self, input_value, expected_date):
...
+        (
+            # primary address: required
+            (
+                {},
+                {
+                    'primary_address': ['This field is required.'],
+                },
+            ),
+
+            # primary address: line_1, town and country required
```
reupen: Is this comment right? It's the same as the next one.

datahub/core/test/test_serializers.py
```diff
@@ -212,3 +217,935 @@ def test_parses_dates(self, input_value, expected_date):
...
+                        'town': 'London',
+                        'country': {'id': Country.united_kingdom.value.id},
+                    },
+                },
+                {
+                    'primary_address': {
+                        'line_1': ['Ensure this field has no more than 255 characters.'],
+                    },
+                },
+            ),
```
reupen: What about tests for:

- `'primary_address': {}`
- `'primary_address': None`

datahub/core/test/test_serializers.py
```diff
@@ -212,3 +217,935 @@ def test_parses_dates(self, input_value, expected_date):
...
+
+            # secondary address: line_1, town and country required
```
reupen: This comment is the same as the next one, it would be good to differentiate the comments somehow.

datahub/core/serializers.py
marcofucci: Hm I think required is the right one but I should support allow_null as well. If you get a nullable fk and let DRF create the serializer for you, it generates `required=False, allow_null=True`. I'll try make allow_null work (at the moment is ignored and None values are always translated into expanded field defaults).

datahub/core/test/test_serializers.py
```diff
@@ -212,3 +217,935 @@ def test_parses_dates(self, input_value, expected_date):
...
+                {
+                    'primary_address': {
+                        'line_1': '2',
+                        'line_2': 'Main Road',
+                        'town': 'London',
+                        'county': 'Greenwich',
+                        'postcode': 'SE10 9NN',
+                        'country': {'id': Country.united_kingdom.value.id},
+                    },
+                    'secondary_address': None,
```
reupen: I was a bit surprised about this one as `allow_null` wasn't set to `True`. I see why it works though (because of the logic in `AddressSerializer.run_validation()`). 

Nonetheless, you can still pass `allow_null=False` and `allow_null=True` to the serialiser which is a bit confusing. It would be good to restrict that in some way, I think?

marcofucci: See https://github.com/uktrade/data-hub-leeloo/pull/1435#discussion_r253888056

reupen:
My only other comment is that it is could be a bit confusing about whether a PATCH with:

```
{
    "secondary_address2: {
        "line_1": "1"
    }
}
```

means to update only `line_1` of the secondary address, or replace `secondary_address` entirely. It also seems a bit inconsistent with how the other nested objects we have (the ones I can think of) work.

Still, I'm OK with it, as long as we're aware of those points.

datahub/core/serializers.py
marcofucci: I've added support for allow_null although I'm not testing the case where `required=False` and `allow_null=False` as it would add a lot of test code just for this scenario. If you think it's worth adding I could though.

reupen:
> @reupen I can only think of nested objects with an id which would indeed be replace but are different from this. What do you mean?

Yes, it was those. They are different, but that was point. It's not necessarily obvious to those using the API that some nested objects are different from other ones. We discussed the response format before you did the PR, so I'm fine with it, just saying that we should be a bit cautious about using nested objects in different ways so that the API doesn't get confusing.

PR #1439, opened 2019-02-07
Company API v4

marcofucci:
This is quite a bit PR but mainly because I've moved some files.
I suggest you review this by first looking at each commit individually.

### Description of change

Companies now define fields for a mandatory address representing the main location for the business and fields for a now optional registered address.
Trading address fields are still automatically updated but deprecated.

**V4 of the API**
Nested objects are now used for addresses. The format is:
```
    'line_1': '2',
    'line_2': 'Main Road',
    'town': 'London',
    'county': 'Greenwich',
    'postcode': 'SE10 9NN',
    'country': {
        'id': '80756b9a-5d95-e211-a939-e4115bead28a',
        'name': 'United Kingdom',
    }
```

**Company collection and item**:
`/v4/company`, `/v4/company/<uuid:pk>`:

- The `trading_address_*` fields were removed from v4
- The `registered_address_*` fields were replaced by the nested object `registered_address` and made optional
- The nested object `address` was added and is mandatory when creating a company

The `companies_house_data` field in the response body of v4 is still flat. I will likely convert to nested object in a separate PR but we are possibly going to get rid of this field in the near future.

**Other company endpoints**
The following endpoints where added to v4 but have the same response body as v3.

- `/v4/company/<uuid:pk>/archive`
- `/v4/company/<uuid:pk>/audit`
- `/v4/company/<uuid:pk>/one-list-group-core-team`
- `/v4/company/<uuid:pk>/timeline`
- `/v4/company/<uuid:pk>/unarchive`

### Checklist


changelog/company/company-api-addresses.api
```diff
@@ -0,0 +1,5 @@
+``/v4/company``, ``/v4/company/<uuid:pk>``:
+
+- The ``trading_address_*`` fields were removed from v4
+- The ``registered_address_*`` fields were replaced by the nested object ``registered_address`` and made optional
+- The nested object ``address`` was added and is mandatory when creating a company
```
reupen: It would be useful to explain how the data was migrated to the new fields as well.

changelog/company/company-api-v4.api
```diff
@@ -0,0 +1,22 @@
+API V4 for companies was introduced with nested object format for addresses.
+A new prefix ``v4`` was introduced along with the following endpoints:
+
+- ``/v4/company``: see the related news fragment
+- ``/v4/company/<uuid:pk>``: see the related news fragment
+- ``/v4/company/<uuid:pk>/archive``: same response body as v3
```
reupen: Doesn't it return the full company object? Is it really the same response body as v3?

Same question for `unarchive`.

changelog/company/company-api-v4.api
```diff
@@ -0,0 +1,22 @@
...
+- ``/v4/company/<uuid:pk>/audit``: same response body as v3
```
reupen: Do we need to change anything related to audit log entries for changes to address fields?

datahub/company/models/company.py
```diff
@@ -435,6 +429,18 @@ class CompaniesHouseCompany(CompanyAbstract):
     uri = models.CharField(max_length=MAX_LENGTH, blank=True)
     incorporation_date = models.DateField(null=True)
 
+    registered_address_1 = models.CharField(max_length=MAX_LENGTH)
+    registered_address_2 = models.CharField(max_length=MAX_LENGTH, blank=True, null=True)
+    registered_address_town = models.CharField(max_length=MAX_LENGTH)
+    registered_address_county = models.CharField(max_length=MAX_LENGTH, blank=True, null=True)
+    registered_address_country = models.ForeignKey(
+        metadata_models.Country,
+        related_name="%(class)ss",  # noqa: Q000
```
reupen: I meant to open a PR to change these to use single quotes, not sure why we used `# noqa: Q000` everywhere instead :confused:

datahub/company/test/test_company_views_v4.py
```diff
@@ -0,0 +1,2368 @@
+import uuid
```
reupen: This file is quite long, it might be helpful to split it up in some way.

changelog/company/company-api-v4.api
marcofucci: Very true, thanks for spotting that!

changelog/company/company-api-v4.api
marcofucci: Nope, this happens at a field level. It's not ideal to have model fields in the audit instead of the transformed one but this is the problem with the current audit implementation which has several of these issues and should eventually be completely refactored or got rid of

marcofucci: Unless you think we should, of course. Please note that with the new design the audit view is also a lot less present and will not be checked often enough to be a priority at the moment.

datahub/company/models/company.py
marcofucci: I think it just gets overlooked, I have myself copies/pasted that many times :-/

datahub/company/test/test_company_views_v4.py
marcofucci: I agree, I've split that into some modules. I could do more but I'm going to get rid of quite a few lines in the file after the migration so that would help.
I could also add a wrapping module (folder) with separate files inside.

changelog/company/new-addresses.feature
```diff
@@ -0,0 +1,6 @@
+Companies now define fields for a mandatory address representing the main location for the business and fields for a now optional registered address.
```
elcct: > and fields for ~a~ now optional registered address.

PR #1440, opened 2019-02-07
Add improved adviser autocomplete

reupen:
### Description of change

This adds a new `autocomplete` query parameter to `GET /adviser/` intended to replace the previous name-related query parameters.

The new parameter matches prefixes of words in the `first_name`, `last_name` and `dit_team.name` fields. Each token must match the prefix of at least one word in (at least) one of those fields.

Results are automatically ordered with advisers with a match on `first_name` appearing first, `last_name` second and `dit_team.name` last.

For simplicity, this continues to use PostgreSQL. Performance is good in my tests (after all, it's not dealing with a huge number of records).

All the existing query parameters are marked as deprecated.

The first commit just renames a file, so I would suggest looking at the commits separately.

### Checklist


datahub/company/views.py
```diff
@@ -144,15 +148,126 @@ class ContactAuditViewSet(AuditViewSet):
...
+        )
+
+    @staticmethod
+    def _autocomplete_q(field, escaped_token):
+        r"""
+        Generates a Q object that performs a case-insensitive match of a token with prefixes
+        of any of the words in a field.
+
+        \m means a word boundary (a word is defined as a consecutive sequence of letters,
+        numbers and underscores).
```
marcofucci: It would be clearer to say: _\m matches only at the beginning of a word..._

datahub/company/views.py
```diff
@@ -144,15 +148,126 @@ class ContactAuditViewSet(AuditViewSet):
...
+            return queryset
+
+        q_objects_for_filtering = (
+            self._autocomplete_filtering_q(self.autocomplete_fields, escaped_token)
+            for escaped_token in escaped_tokens
+        )
+        cases_for_ordering = (
+            When(self._autocomplete_ordering_q(field, escaped_tokens), then=index)
+            for index, field in enumerate(self.autocomplete_fields)
+        )
```
marcofucci: Instead of using single numbers I think it would be more accurate to have a string of 3 bits:
000: no match
100: first name only
110: first and last name
111: all of them
etc.

And order by DESC. so that if a match on more than one field exists, it gets bumped up.

datahub/company/views.py
```diff
@@ -144,15 +148,126 @@ class ContactAuditViewSet(AuditViewSet):
...
+
+        Results are automatically ordered as follows:
+
+        - advisers with a match on first_name are returned first, last_name second and
+        dit_team.name last
+        - within each group, results are sorted by first_name, last_name, dit_team.name and
+        finally pk (so that the results are deterministic if the other ordering fields are
+        identical).
+
+        Scoring is not used is keep the order of results predictable when only a few characters
```
marcofucci: not sure this sentence makes sense

datahub/company/views.py
```diff
@@ -144,15 +148,126 @@ class ContactAuditViewSet(AuditViewSet):
...
+        in the request.
+        """
+        escaped_tokens = [re.escape(token) for token in value.split()]
+
+        if not escaped_tokens:
+            return queryset
+
+        q_objects_for_filtering = (
+            self._autocomplete_filtering_q(self.autocomplete_fields, escaped_token)
+            for escaped_token in escaped_tokens
```
marcofucci: I could call `_autocomplete_filtering_q` something like `_autocomplete_q_filter_for_token_term` to make it clear that's just for one part of the terms.

For similar reasons I could call `q_objects_for_filtering` something like `q_filters_for_all_token_terms`.

datahub/company/views.py
```diff
@@ -144,15 +148,126 @@ class ContactAuditViewSet(AuditViewSet):
...
+        return reduce(
+            or_,
+            (
+                Q(cls._autocomplete_q(field, escaped_token))
+                for field in fields
+            ),
+        )
+
+    @staticmethod
+    def _autocomplete_q(field, escaped_token):
```
marcofucci: `_autocomplete_q` is way too generic IMHO and doesn't say what it does, I would call it something similar to https://github.com/uktrade/data-hub-leeloo/pull/1440/files#r255111322

datahub/company/views.py
reupen: That is what I meant by scoring (except that I was going to add up weights). Say we allowed entering one letter (it's easier to see in that case), when you entered b you get Brian Brown in Barbados at the top. So I kept it the simple way to keep it predictable and avoid things jumping around when you add characters (and it was specified in that way anyway). The terms are also combined with AND, so the results will get naturally narrowed down as terms are added anyway.

datahub/company/views.py
```diff
@@ -144,15 +148,126 @@ class ContactAuditViewSet(AuditViewSet):
...
+        """
+        Gets the default ordering for the view.
+
+        If a value has been provided in the autocomplete query parameter, no default
+        ordering is used as the autocomplete filter orders results automatically.
+
+        Otherwise, the default ordering set on the view (in the ordering attribute)
+        is used.
+        """
+        if view.request.query_params.get('autocomplete'):
```
marcofucci: Not a major issue but I'm not a bit fan of having this query param hardcoded in here which also ties this generic class to the internal logic of the FilterSet. One option would be to have a method on the view called something like `should_add_ordering` and the view uses AdviserFilter somehow to figure that out.
Up to you if it adds too much complexity.

reupen: No, I didn't like it either. I'll have a go at putting something on the view.

datahub/company/views.py
marcofucci: I'm not sure to understand, the case you gave is exactly what I would expect. what 3 bits would solve and what the current logic doesn't support.
If I enter b I do want Brian Brown in Barbados to be at the top I think.

reupen: Because, most probably you aren't looking for Brian Brown, and you will carry on entering more characters. You will definitely remember Brian Brown though. (And, if you were looking for him, you might press 'r' and he might then slip further down the list.)

If it turns out that we need better ranking, we can review it later on. It will also be easier to see if it's suboptimal with real data.

datahub/company/test/test_adviser_views.py
```diff
@@ -0,0 +1,354 @@
...
+                    ('Nigel', 'Newman', 'New York'),
+                ],
+            ),
+            (
+                'zzz',
+                [],
+            ),
+        ),
+    )
+    def test_adviser_autocomplete(self, advisers, terms, expected_results):
```
marcofucci: minor readability comment, I'd move `advisers` to the end (`(self, terms, expected_results, advisers):`) as the parametrize argnames are a bit lost and you might have the impression that the first element of the parametrize argvalues is the advisers from looking at this line only.

datahub/company/views.py
marcofucci: Why you were not looking for Brian Brown?
Also, if you were looking for Brown and you use your logic, things would jump around as well.

reupen:
@marcofucci I have done the change we discussed in person to sort by the matched field within each group, but I am going to try and optimise it to keep the query fast.

reupen:
Scratch that, the difference was less than I thought, I ran it with a large offset by accident...

datahub/company/autocomplete.py
```diff
@@ -0,0 +1,140 @@
...
+    ).order_by(
+        '_matched_group_index',
+        '_matched_field',
+        *autocomplete_fields,
+        'pk',
+    )
+
+
+def _make_ordering_q_for_field(field, escaped_tokens):
+    return reduce(
```
marcuspp: Would you mind adding comments in for all the `_` functions in this file please?

marcuspp:
@reupen Looking at this the one thing that strikes me is that this endpoint is different from the `company/autocomplete`. I know one is using ES and the other the DB but did you consider introducing a new endpoint `adviser/autocomplete` with a param of `term` so that the usage is the always the same regardless of backend implementation.

reupen:
> @reupen Looking at this the one thing that strikes me is that this endpoint is different from the `company/autocomplete`. I know one is using ES and the other the DB but did you consider introducing a new endpoint `adviser/autocomplete` with a param of `term` so that the usage is the always the same regardless of backend implementation.

That is true – I avoided that because it would mean duplicating other filters (`is_active` in this case) and so it would add complexity.

datahub/company/autocomplete.py
reupen: I have added some additional documentation.

marcuspp:
Could you not add a new url `adviser/autocomplete` to urls file that uses the same view and just rename the param `term then deprecate the old view?  That would mean the usage would be consistent without losing the option to fallback.

reupen:
You mean get rid of `/adviser/`? That wouldn't be consistent with all the other views, though.

marcuspp:
Okay, I'll approve this to get it out but I do think having a param `autocomplete` is confusing and  inconsistent with any other endpoint but you are right that this one is different from other endpoints. Maybe a proper search adviser endpoint should be added to backlog?

PR #1441, opened 2019-02-08
Update interaction global search fields to use contacts

reupen:
### Description of change

This updates the interaction global search fields to use `contacts` instead of `contact`. The relevant test was also updated to be a bit more comprehensive.

### Checklist

### DIFF
--- a/datahub/search/interaction/models.py
+++ b/datahub/search/interaction/models.py
@@ -84,8 +84,8 @@ class Interaction(BaseESModel):
         'id',
         'company.name',
         'company.name_trigram',
-        'contact.name',
-        'contact.name_trigram',
+        'contacts.name',  # to find 2-letter words
+        'contacts.name.trigram',
         'event.name',
         'event.name_trigram',
         'subject_english',
--- a/datahub/search/interaction/tests/test_views.py
+++ b/datahub/search/interaction/tests/test_views.py
@@ -51,26 +51,60 @@ def interactions(setup_es):
     """Sets up data for the tests."""
     data = []
     with freeze_time('2017-01-01 13:00:00'):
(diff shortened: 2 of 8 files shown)


datahub/search/interaction/models.py
```diff
@@ -84,8 +84,8 @@ class Interaction(BaseESModel):
         'id',
         'company.name',
         'company.name_trigram',
-        'contact.name',
-        'contact.name_trigram',
+        'contacts.name',
```
marcofucci: worth mentioning that's there `  # to find 2-letter words`?

datahub/search/interaction/tests/test_views.py
```diff
@@ -211,7 +244,17 @@ def test_sort_by_invalid_field(self, setup_es, sortby, error):
...
-    @pytest.mark.parametrize('term', ('exports', 'meeting', 'exports meeting'))
+    @pytest.mark.parametrize(
+        'term',
+        (
+            'exports',
+            'meeting',
+            'exports meeting',
+            'danger',
+            'FRANCIS',
+            'angela',
```
marcofucci: can you add some trigram and 2-letter ones?

datahub/search/interaction/models.py
reupen: Done, but I think we could also do more with the text fields as the matches from them are usually better than the trigram ones.

datahub/search/interaction/tests/test_views.py
reupen: Done – there was a mistake in the search field definition on interaction (`contacts.name_trigram` instead of `contacts.name.trigram`), and also there was another invalid field on company (`uk_region.name_trigram`) which I removed.

PR #1443, opened 2019-02-11
Index company addresses

marcofucci:
### Description of change

This indexes company address and registered address as nested objects.
The objects are automatically made available in the response body but I'm not announcing it in the CHANGELOG yet because I need to update the autocompletion endpoint and I might add a virtual v4 prefix somehow for these.

Only postcode and countries are indexed as they are the ones currently used by the frontend.

I refactored `test_dict_utils` as well so you might want to check the commits individually.

### Checklist

### DIFF
--- a/datahub/search/dict_utils.py
+++ b/datahub/search/dict_utils.py
@@ -1,3 +1,14 @@
+def _attrgetter_with_default(attr, default):
+    """
+    It returns a function that can be called with an object to get the value
+    of attr or the default.
+    Useful to convert None values to ''.
+    """
+    def _getter(obj):
+        return getattr(obj, attr) or default
+    return _getter
+
+
 def id_name_dict(obj):
     """Creates dictionary with selected field from supplied object."""
     if obj is None:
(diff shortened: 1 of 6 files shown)


datahub/search/dict_utils.py
```diff
@@ -36,6 +47,35 @@ def id_uri_dict(obj):
...
+def address_dict(obj, prefix='address'):
+    """
+    Creates a dictionary for the address fields with the given prefix
+    to be used as nested object.
+    """
+    if obj is None:
+        return None
+
+    mapping = {
+        'line_1': _attrgetter_with_default(f'{prefix}_1', default=''),
```
marcuspp: default is just an arg not a keyword arg for the function `_attrgetter_with_default`.

datahub/search/dict_utils.py
```diff
@@ -36,6 +47,35 @@ def id_uri_dict(obj):
     }
 
 
+def address_dict(obj, prefix='address'):
+    """
```
marcuspp: If a prefix is specified that does not exist on the object would this error and if so do you think it should handle it?

datahub/search/dict_utils.py
marcofucci: Is that a problem? I sometimes prefer kwargs calls even for positional args to make it easier to read.

datahub/search/dict_utils.py
marcofucci: An `AttributeError` is triggered, I don't think we should catch that as it's an implementation error and there's much to add to the default one. 
Ideally the dev will have written a test or in the worst case the error is raised on the first function call.

datahub/search/dict_utils.py
marcuspp: I think it is inconsistent and not what would be general expected, but if you are happy with it then it's fine.

datahub/search/dict_utils.py
marcuspp: Can you add a test for the `AttributeError` if that is the expected behaviour.

datahub/search/dict_utils.py
marcofucci: Changed, thanks!

datahub/search/dict_utils.py
marcofucci: Done, thanks!

PR #1444, opened 2019-02-11
Search Refactoring

marcofucci:
### Description of change

This includes a series of changes in preparation for more work:
- renames some classes
- changes slightly the autocomplete endpoint
- gets rid of some redundant code

It might be easier to check the commits individually.

I got rid of the `AutocompleteSearchSerializer` to make the logic consistent with all the other search endpoints but do let me know if we want to keep this around. 
If so, we would have to add serializers to all other endpoint responses as well to keep everything consistent.

### Checklist


datahub/search/event/serializers.py
```diff
@@ -2,20 +2,20 @@
...
 from datahub.search.serializers import (
-    EntitySearchSerializer,
+    EntitySearchQuerySerializer,
     SingleOrListField,
     StringUUIDField,
 )
 
 
-class NestedDisabledOnOrFilterSerializer(serializers.Serializer):
+class NestedDisabledOnOrQuerySerializer(serializers.Serializer):
```
reupen: I find the 'Or' in the class name a bit confusing... If needed, perhaps it could be explained in the docstring that it uses OR and not AND (similar to what it says below for `SearchEventQuerySerializer`)?

datahub/search/views.py
```diff
@@ -274,16 +278,19 @@ class AutocompleteSearchListAPIView(ListAPIView):
...
     def list(self, request, *args, **kwargs):
         """Executes the elasticsearch query"""
+        serializer = AutocompleteSearchQuerySerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        validated_params = serializer.validated_data
+
         self.check_permission_filters()
         results = execute_autocomplete_query(
             self.search_app.es_model,
-            self.get_search_query_string(),
```
reupen: Can `get_search_query_string()` be removed now?

datahub/search/views.py
```diff
@@ -274,35 +278,31 @@ class AutocompleteSearchListAPIView(ListAPIView):
...
+            validated_params['limit'],
+            only_return_fields=self.document_fields,
+        )
+
+        return Response(data={
+            'count': len(results),
+            'results': [result['_source'].to_dict() for result in results],
+        })
 
     def get_search_query_string(self):
```
reupen: Can this be removed now?

PR #1450, opened 2019-02-12
Deprecate company search params

marcofucci:
### Description of change

This deprecates many company search fields/sortby fields so that we can simplify the codebase and make future refactoring easier.

Logging has been added to make sure we don't break things inadvertently.

### Checklist

### DIFF
--- a/datahub/search/company/serializers.py
+++ b/datahub/search/company/serializers.py
@@ -1,3 +1,5 @@
+import logging
+
 from rest_framework import serializers
 
 from datahub.search.serializers import (
@@ -5,6 +7,9 @@ from datahub.search.serializers import (
     SingleOrListField,
     StringUUIDField,
 )
+
+
+logger = logging.getLogger(__name__)
 
 
 class SearchCompanyQuerySerializer(EntitySearchQuerySerializer):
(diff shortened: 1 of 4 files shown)


datahub/search/company/serializers.py
```diff
@@ -51,6 +56,54 @@ class SearchCompanySerializer(EntitySearchSerializer):
...
+        'employee_range.name',
+        'headquarter_type.name',
+        'id',
+        'registered_address_town',
+        'sector.name',
+        'trading_address_town',
+        'turnover_range.name',
+        'uk_based',
+        'uk_region.name',
+        'name',
```
reupen: Is `name` meant to be in this list?

marcofucci: 😞 left behind, well spotted!

datahub/search/company/serializers.py
```diff
@@ -51,6 +56,54 @@ class SearchCompanySerializer(EntitySearchSerializer):
...
+    }
+
+    def validate(self, data):
+        """
+        Logs all deprecated params to make sure we don't break things when we get rid of them.
+
+        TODO Remove following deprecation period.
+        """
+        if set(data) & self.deprecated_filters:
```
reupen: Just a minor point, it would be nicer to assign `set(data) & self.deprecated_filters` to a variable rather than repeating it. Also, `dict.keys()` can be used with set operations.

datahub/search/company/serializers.py
```diff
@@ -51,6 +56,54 @@ class SearchCompanySerializer(EntitySearchSerializer):
...
+            logger.error(
+                'The following deprecated company search filters were '
+                f'used: {set(data) & self.deprecated_filters}.',
+            )
+
+        sortby = data.get('sortby')
+        if sortby and hasattr(sortby, 'field') and sortby.field in self.deprecated_sortby_fields:
```
reupen: Is `hasattr(sortby, 'field')` needed? Wouldn't there be a bug if it didn't have a `field` attribute?

datahub/search/company/serializers.py
```diff
@@ -51,6 +56,54 @@ class SearchCompanySerializer(EntitySearchSerializer):
...
+            logger.error(
+                'The following deprecated company sortby field was '
```
reupen: It would be good to add 'search' after company so that it is consistent with the message for filters.

datahub/search/company/serializers.py
marcofucci: I guess, I just wanted to be double sure.

marcofucci:
@reupen this should work for global search I guess? https://github.com/uktrade/data-hub-leeloo/pull/1450/commits/8c8781092c2016cee74861f933775fa6715c0afe

reupen:
@marcofucci Yep, that looks good to me.

PR #1454, opened 2019-02-13
Make tests pass under Elasticsearch 6.6

reupen:
### Description of change

This changes a few search tests to ignore the `_seq_no` and `_primary_term` document properties added in Elasticsearch 6.6 (it does not make sense to check the values of them in these cases) .

It also adds a second CircleCI build to make sure that tests pass on both Elasticsearch 6.4 and Elasticsearch 6.6.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,17 +94,28 @@ job_defaults: &job_defaults
               bash ./codecov.sh -t ${COV_TOKEN}
 
 jobs:
-  default_build:
+  build:
     <<: *job_defaults
 
 workflows:
   version: 2
 
+  # Current standard environment
   Default build:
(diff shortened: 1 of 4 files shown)


reupen:
> +1 Interested to know why you haven't defined a common util function to get rid of those?

To be honest, we pass `_index` and `_type` in the call to `setup_es.get()`.

I would be happy with just doing something like:

```python

assert indexed_document['_id'] == obj.id
assert indexed_document['_source'] == {...}
```

marcofucci:
True, it's a minor thing, any of these work but maybe

```python
assert indexed_document['_id'] == obj.id
assert indexed_document['_source'] == {...}
```

would be cleaner?

PR #1455, opened 2019-02-13
Search includes excludes fields

marcofucci:
### Description of change

This adds the ability to include only specific fields and/or exclude some fields from the ES returned source.

This way, I can create specific views for the v4, exclude the old flat fields for addresses from the v4 views and exclude the nested ones from the v3 views.

I've also refactored `get_search_by_entity_query` slightly so that the arguments are not broken.

It uses https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-source-filtering.html

### Checklist

### DIFF
--- a/datahub/search/execute_query.py
+++ b/datahub/search/execute_query.py
@@ -9,10 +9,13 @@ from datahub.search.query_builder import build_autocomplete_query
 logger = getLogger(__name__)
 
 
-def execute_autocomplete_query(es_model, keyword_search, limit, only_return_fields=None):
+def execute_autocomplete_query(es_model, keyword_search, limit, fields_to_include=None):
     """Executes the query for autocomplete search returning all suggested documents."""
     autocomplete_search = build_autocomplete_query(
-        es_model, keyword_search, limit, only_return_fields,
+        es_model,
+        keyword_search,
+        limit,
+        fields_to_include,
     )
 
     results = autocomplete_search.execute()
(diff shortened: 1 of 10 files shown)


datahub/search/execute_query.py
```diff
@@ -9,10 +9,13 @@
 logger = getLogger(__name__)
 
 
-def execute_autocomplete_query(es_model, keyword_search, limit, only_return_fields=None):
+def execute_autocomplete_query(es_model, keyword_search, limit, only_include_source_fields=None):
```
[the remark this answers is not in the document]
marcofucci: Yes very good point, I'll use `fields_to_include` and `fields_to_exclude`.

datahub/search/query_builder.py
```diff
@@ -355,3 +364,15 @@ def _apply_sorting_to_query(query, ordering):
...
+
+def _apply_source_filtering_to_query(query, fields_to_include=None, fields_to_exclude=None):
+    source_filters = {}
+    if fields_to_include:
+        source_filters['includes'] = fields_to_include
+    if fields_to_exclude:
+        source_filters['excludes'] = fields_to_exclude
+
+    if source_filters:
+        return query.extra(_source=source_filters)
```
reupen: Is it possible to use [`.source()`](https://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html?highlight=source#extra-properties-and-parameters) for this?

reupen: Also looks like `.source(<param>=None)` means don't specify `<param>` so it should be possible to simplify the logic.

marcofucci: Very nice, I was not aware of it. Thanks!

PR #1456, opened 2019-02-14
Add ability to merge companies having investment projects

elcct:
### Description of change

This enhances the company merge tool with the ability to merge companies having related investment projects.
It replaces `uk_company`, `investor_company` and `intermediate_company` with a target company.

### Checklist

### DIFF
--- a/datahub/company/admin/merge/step_3.py
+++ b/datahub/company/admin/merge/step_3.py
@@ -6,20 +6,24 @@ from django.http import HttpResponseRedirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils.decorators import method_decorator
-from django.utils.html import format_html
+from django.utils.html import format_html, format_html_join
 from django.utils.translation import gettext_lazy
 from django.views.decorators.csrf import csrf_protect
 
-from datahub.company.merge import DuplicateCompanyMerger, MergeNotAllowedError
-from datahub.company.models import Contact
+from datahub.company.merge import (
+    get_planned_changes,
+    merge_companies,
+    MergeNotAllowedError,
+    transform_merge_results_to_merge_entry_summaries,
+)
 from datahub.core.templatetags.datahub_extras import verbose_name_for_count
-from datahub.interaction.models import Interaction
-
 
 REVERSION_REVISION_COMMENT = 'Company marked as a duplicate and related records transferred.'
+MERGE_ENTRY_MSG_FRAGMENT = gettext_lazy(
+    '{0} {1}{2}',
+)
 MERGE_SUCCESS_MSG = gettext_lazy(
-    'Merge complete – {num_interactions_moved} {interaction_verbose_name}'
-    ' and {num_contacts_moved} {contact_verbose_name} moved from'
+    'Merge complete – {merge_entries} moved from'
     ' <a href="{source_company_url}" target="_blank">{source_company}</a> to'
     ' <a href="{target_company_url}" target="_blank">{target_company}</a>.',
 )
@@ -48,9 +52,8 @@ def confirm_merge(model_admin, request):
         raise SuspiciousOperation()
 
     is_post = request.method == 'POST'
-    merger = DuplicateCompanyMerger(source_company, target_company)
 
-    if is_post and _perform_merge(request, merger, model_admin):
+    if is_post and _perform_merge(request, source_company, target_company, model_admin):
         changelist_route_name = admin_urlname(model_admin.model._meta, 'changelist')
         changelist_url = reverse(changelist_route_name)
         return HttpResponseRedirect(changelist_url)
@@ -58,13 +61,17 @@ def confirm_merge(model_admin, request):
     template_name = 'admin/company/company/merge/step_3_confirm_selection.html'
     title = gettext_lazy('Confirm merge')
 
-    move_entries, should_archive_source = merger.get_planned_changes()
+    planned_merge_results, should_archive_source = get_planned_changes(source_company)
+    merge_entries = transform_merge_results_to_merge_entry_summaries(
+        planned_merge_results,
+        skip_zeroes=True,
+    )
 
     context = {
         **model_admin.admin_site.each_context(request),
         'source_company': source_company,
         'target_company': target_company,
-        'move_entries': move_entries,
+        'merge_entries': merge_entries,
         'should_archive_source': should_archive_source,
         'media': model_admin.media,
         'opts': model_admin.model._meta,
@@ -73,37 +80,45 @@ def confirm_merge(model_admin, request):
     return TemplateResponse(request, template_name, context)
 
 
-def _perform_merge(request, merger, model_admin):
+def _perform_merge(request, source_company, target_company, model_admin):
     try:
-        merge_result = merger.perform_merge(request.user)
+        merge_results = merge_companies(source_company, target_company, request.user)
     except MergeNotAllowedError:
         failure_msg = MERGE_FAILURE_MSG.format(
-            source_company=merger.source_company,
-            target_company=merger.target_company,
+            source_company=source_company,
+            target_company=target_company,
         )
         model_admin.message_user(request, failure_msg, django_messages.ERROR)
         return False
 
     reversion.set_comment(REVERSION_REVISION_COMMENT)
-    success_msg = _build_success_msg(merger, merge_result)
+    success_msg = _build_success_msg(source_company, target_company, merge_results)
     model_admin.message_user(request, success_msg, django_messages.SUCCESS)
     return True
 
 
-def _build_success_msg(merger, merge_result):
-    interaction_verbose_name = verbose_name_for_count(
-        merge_result.num_interactions_moved,
-        Interaction._meta,
+def _build_success_msg(source_company, target_company, merge_results):
+    merge_entries = transform_merge_results_to_merge_entry_summaries(
+        merge_results,
+        skip_zeroes=True,
     )
-    contact_verbose_name = verbose_name_for_count(merge_result.num_contacts_moved, Contact._meta)
-    return format_html(
+
+    messages = (
+        (
+            merge_entry.count,
+            verbose_name_for_count(merge_entry.count, merge_entry.model_meta),
+            merge_entry.description,
+        ) for merge_entry in merge_entries
+    )
+
+    html_merge_entries = format_html_join(', ', MERGE_ENTRY_MSG_FRAGMENT, messages)
+
+    html = format_html(
         MERGE_SUCCESS_MSG,
-        num_interactions_moved=merge_result.num_interactions_moved,
-        num_contacts_moved=merge_result.num_contacts_moved,
-        interaction_verbose_name=interaction_verbose_name,
-        contact_verbose_name=contact_verbose_name,
-        source_company_url=merger.source_company.get_absolute_url(),
-        source_company=merger.source_company,
-        target_company_url=merger.target_company.get_absolute_url(),
-        target_company=merger.target_company,
+        merge_entries=html_merge_entries,
+        source_company_url=source_company.get_absolute_url(),
+        source_company=source_company,
+        target_company_url=target_company.get_absolute_url(),
+        target_company=target_company,
     )
+    return html
--- a/datahub/company/merge.py
+++ b/datahub/company/merge.py
@@ -1,147 +1,184 @@
 from collections import namedtuple
 
-from django.db import transaction
-
 from datahub.company.models import Company, Contact
 from datahub.core.exceptions import DataHubException
 from datahub.core.model_helpers import get_related_fields, get_self_referential_relations
 from datahub.interaction.models import Interaction
+from datahub.investment.models import InvestmentProject
 
 ALLOWED_RELATIONS_FOR_MERGING = {
     Company._meta.get_field('dnbmatchingresult').remote_field,
     Contact.company.field,
     Interaction.company.field,
+    InvestmentProject.investor_company.field,
+    InvestmentProject.intermediate_company.field,
+    InvestmentProject.uk_company.field,
 }
 
 
-MoveEntry = namedtuple(
-    'MoveEntry',
+INVESTMENT_PROJECT_COMPANY_FIELDS = (
+    'investor_company',
+    'intermediate_company',
+    'uk_company',
+)
+
+FIELD_TO_DESCRIPTION_MAPPING = {
+    'investor_company': ' as investor company ',
+    'intermediate_company': ' as intermediate company ',
+    'uk_company': ' as UK company ',
+}
+
+MergeEntrySummary = namedtuple(
+    'MergeEntrySummary',
     [
         'count',
+        'description',
         'model_meta',
     ],
 )
 
 
-MergeResult = namedtuple(
-    'MergeResult',
+MergeConfiguration = namedtuple(
+    'MergeConfiguration',
     [
-        'num_interactions_moved',
-        'num_contacts_moved',
+        'model',
+        'fields',
     ],
 )
+
+
+MERGE_CONFIGURATION = [
+    MergeConfiguration(Interaction, ('company',)),
+    MergeConfiguration(Contact, ('company',)),
+    MergeConfiguration(InvestmentProject, INVESTMENT_PROJECT_COMPANY_FIELDS),
+]
 
 
 class MergeNotAllowedError(DataHubException):
     """Merging the specified source company into the specified target company is not allowed."""
 
 
-class DuplicateCompanyMerger:
-    """Utility class for merging duplicate companies."""
+def is_company_a_valid_merge_source(company: Company):
+    """Checks if company can be moved."""
+    # First, check that there are no references to the company from other objects
+    # (other than via the fields specified in ALLOWED_RELATIONS_FOR_MERGING).
+    relations = get_related_fields(Company)
 
-    def __init__(self, source_company: Company, target_company: Company):
-        """Initialises the instance with the source and target companies."""
-        self.source_company = source_company
-        self.target_company = target_company
+    has_related_objects = any(
+        getattr(company, relation.name).count()
+        for relation in relations
+        if relation.remote_field not in ALLOWED_RELATIONS_FOR_MERGING
+    )
 
-    def get_planned_changes(self):
-        """Gets information about the changes that would be made if the merge proceeds."""
-        move_entries = []
+    if has_related_objects:
+        return False
 
-        contact_count = self.source_company.contacts.count()
-        self._append_move_entry(move_entries, contact_count, Contact)
+    # Then, check that the source company itself doesn't have any references to other
+    # companies.
+    self_referential_fields = get_self_referential_relations(Company)
+    return not any(
+        getattr(company, field.name) for field in self_referential_fields
+    )
 
-        interaction_count = self.source_company.interactions.count()
-        self._append_move_entry(move_entries, interaction_count, Interaction)
 
-        should_archive_source = not self.source_company.archived
+def is_company_a_valid_merge_target(company: Company):
+    """
+    Returns whether the specified company is a valid merge target.
 
-        return move_entries, should_archive_source
+    This checks that the target company isn't archived.
+    """
+    return not company.archived
 
-    def is_source_valid(self):
-        """
-        Returns whether the specified source company is a valid merge source.
 
-        This checks whether there are any references to the source company (other than
-        references through the relations specified in ALLOWED_RELATIONS_FOR_MERGING).
-        It also checks if the source company has any references to other companies.
-        Merging is not permitted if either of those types of reference exists.
-        """
-        if not hasattr(self, '_cached_is_source_valid'):
-            self._cached_is_source_valid = self._is_source_valid()
+def transform_merge_results_to_merge_entry_summaries(results, skip_zeroes=False):
+    """Transforms merge results into move entries to aid the presentation template."""
+    merge_entries = []
 
-        return self._cached_is_source_valid
+    for model, fields in results.items():
+        for field, num_objects_updated in fields.items():
+            if num_objects_updated == 0 and skip_zeroes:
+                continue
 
-    def is_target_valid(self):
-        """
-        Returns whether the specified target company is a valid merge target.
+            merge_entry = MergeEntrySummary(
+                num_objects_updated,
+                FIELD_TO_DESCRIPTION_MAPPING.get(field, ''),
+                model._meta,
+            )
+            merge_entries.append(merge_entry)
 
-        This checks that the target company isn't archived.
-        """
-        return not self.target_company.archived
+    return merge_entries
 
-    def is_valid(self):
-        """Returns whether the merge is allowed."""
-        return self.is_source_valid() and self.is_target_valid()
 
-    @transaction.atomic
-    def perform_merge(self, user):
-        """
-        Merges the source company into the target company.
+def merge_companies(source_company: Company, target_company: Company, user):
+    """
+    Merges the source company into the target company.
 
-        is_valid() should be called first to check if the merge is allowed. DataHubException
-        will be raised if the merge is not allowed, and perform_merge() was called anyway.
-        """
-        if not self.is_valid():
-            raise MergeNotAllowedError()
+    MergeNotAllowedError will be raised if the merge is not allowed.
+    """
+    if not (
+        is_company_a_valid_merge_source(source_company)
+        and is_company_a_valid_merge_target(target_company)
+    ):
+        raise MergeNotAllowedError()
 
-        num_interactions_moved = 0
-        num_contacts_moved = 0
+    results = _process_all_objects(
+        _update_objects,
+        source=source_company,
+        target=target_company,
+    )
 
-        for interaction in self.source_company.interactions.iterator():
-            interaction.company = self.target_company
-            interaction.save(update_fields=('company',))
-            num_interactions_moved += 1
+    source_company.mark_as_transferred(
+        target_company,
+        Company.TRANSFER_REASONS.duplicate,
+        user,
+    )
 
-        for contact in self.source_company.contacts.iterator():
-            contact.company = self.target_company
-            contact.save(update_fields=('company',))
-            num_contacts_moved += 1
+    return results
 
-        self.source_company.modified_by = user
-        self.source_company.mark_as_transferred(
-            self.target_company,
-            Company.TRANSFER_REASONS.duplicate,
-            user,
-        )
 
-        return MergeResult(num_interactions_moved, num_contacts_moved)
+def get_planned_changes(company: Company):
+    """Gets information about the changes that would be made if merge proceeds."""
+    results = _process_all_objects(
+        _count_objects,
+        source=company,
+    )
 
-    @staticmethod
-    def _append_move_entry(to_move, count, model):
-        if count == 0:
-            return
+    should_archive = not company.archived
 
-        move_item = MoveEntry(count, model._meta)
-        to_move.append(move_item)
+    return results, should_archive
 
-    def _is_source_valid(self):
-        # First, check that there are no references to the source company from other objects
-        # (other than via the fields specified in ALLOWED_RELATIONS_FOR_MERGING).
-        relations = get_related_fields(Company)
 
-        has_related_objects = any(
-            getattr(self.source_company, relation.name).count()
-            for relation in relations
-            if relation.remote_field not in ALLOWED_RELATIONS_FOR_MERGING
-        )
+def _process_all_objects(process_objects_fn, **kwargs):
+    """Process all objects defined in the configuration using provided function."""
+    return {
+        configuration.model: process_objects_fn(configuration, **kwargs)
+        for configuration in MERGE_CONFIGURATION
+    }
 
-        if has_related_objects:
-            return False
 
-        # Then, check that the source company itself doesn't have any references to other
-        # companies.
-        self_referential_fields = get_self_referential_relations(Company)
-        return not any(
-            getattr(self.source_company, field.name) for field in self_referential_fields
-        )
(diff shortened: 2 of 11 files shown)


datahub/company/merge.py
```diff
@@ -145,3 +166,26 @@ def _is_source_valid(self):
         return not any(
             getattr(self.source_company, field.name) for field in self_referential_fields
         )
+
+    def get_investment_projects_count(self, company):
+        """Get number of investment projects related to a company."""
+        queries = Q()
+        for field in self. INVESTMENT_PROJECT_COMPANY_FIELDS:
```
reupen: There's an extra space here.

datahub/company/merge.py
```diff
@@ -145,3 +166,26 @@ def _is_source_valid(self):
...
+
+        for company_field in self.INVESTMENT_PROJECT_COMPANY_FIELDS:
+
+            investment_projects = InvestmentProject.objects.filter(
+                **{company_field: self.source_company},
+            )
+            for investment_project in investment_projects:
+                setattr(investment_project, company_field, self.target_company)
+                investment_project.save(update_fields=(company_field,))
+                investment_projects_moved.add(investment_project.id)
```
reupen: OK, but a set of model objects will work fine as well.

datahub/company/merge.py
```diff
@@ -39,6 +45,12 @@ class MergeNotAllowedError(DataHubException):
 class DuplicateCompanyMerger:
     """Utility class for merging duplicate companies."""
 
+    INVESTMENT_PROJECT_COMPANY_FIELDS = (
```
reupen: This module is just for merging so I think `INVESTMENT_PROJECT_COMPANY_FIELDS` can be moved out of the class

datahub/company/test/test_merge.py
```diff
@@ -142,9 +179,15 @@ def test_merge_fails_when_not_allowed(self, monkeypatch):
...
 
-def _company_factory(num_interactions, num_contacts):
-    """Factory for a company that has companies and interactions."""
+def _company_factory(num_interactions, num_contacts, num_investment_projects):
+    """Factory for a company that has companies, interactions and investment projects."""
     company = CompanyFactory()
     ContactFactory.create_batch(num_contacts, company=company)
     CompanyInteractionFactory.create_batch(num_interactions, company=company)
+
+    fields = DuplicateCompanyMerger.INVESTMENT_PROJECT_COMPANY_FIELDS[:num_investment_projects]
```
reupen: Won't this misbehave if `num_investment_projects` is greater than 3? You could use `random.choices()` or `itertools.cycle()` + `itertools.islice()`.

reupen: This looks a bit confusing, why does it set more company fields on each investment project if `num_investment_projects` is larger? It should test different combinations of company fields, but it seems a bit odd to do it like this. There are three relevant states for each field:

- None (if the field is nullable)
- `company`
- a different company

It would be good to get a mix of the different cases.

datahub/company/test/test_merge.py
```diff
@@ -101,6 +137,7 @@ def test_merge_succeeds(self, source_num_interactions, source_num_contacts):
         assert result == MergeResult(
             num_interactions_moved=len(source_interactions),
             num_contacts_moved=len(source_contacts),
+            num_investment_projects_updated=len(set(source_investment_projects)),
```
reupen: Small point, but it looks like `source_investment_projects` could be a set straight away.

datahub/company/merge.py
```diff
@@ -145,3 +166,26 @@ def _is_source_valid(self):
         return not any(
             getattr(self.source_company, field.name) for field in self_referential_fields
         )
+
+    def get_investment_projects_count(self, company):
```
reupen: This doesn't really need `self` so it would be better not to make it an instance method. I would personally move it out of the class, but a static method would work too (if `INVESTMENT_PROJECT_COMPANY_FIELDS` is moved out, otherwise a class method).

datahub/company/merge.py
elcct: That must have been me trying to type something somewhere and having focus in the editor. Thanks!

datahub/company/merge.py
```diff
@@ -108,14 +122,20 @@ def perform_merge(self, user):
             contact.save(update_fields=('company',))
             num_contacts_moved += 1
 
+        num_investment_projects_updated = self._update_investment_projects()
```
marcofucci: Can you move the interaction and contact migration block to their own method as you did with `_update_investment_projects` to keep things consistent?

datahub/company/merge.py
```diff
@@ -145,3 +165,27 @@ def _is_source_valid(self):
         return not any(
             getattr(self.source_company, field.name) for field in self_referential_fields
         )
+
+    def _update_investment_projects(self):
+        investment_projects_moved = set()
```
marcofucci: small thing but it would be better to call everything `investment_projects_moved` or `investment_projects_updated` for consistency

reupen:
You can see a list of contacts, interactions and investment projects (where the company is the investor company) that are going to be moved by clicking through to the front end. But I don't believe there is a simple way to see which investment projects will be affected when the company is the UK (recipient) company. That might need a bit more thought.

(The same applies to intermediate company, but there aren't many projects with that field set.)

datahub/company/merge.py
elcct: I would change all to _updated then, as maybe for contacts, interactions `_moved` would be fine, but investment project doesn't feel like being moved.

datahub/company/merge.py
elcct: I figured out that all this code look too similar, so I generalised it a bit and introduced merge configuration, that I would use to add OMIS orders too.

elcct:
I changed the message a bit so that now it tells which field from each entity is going to be modified. It should be much cleared how many investment projects as UK company, investor company or intermediate company will be transferred. I am not sure if it is worth displaying info about company field for contacts and interactions but left it for consistency.

datahub/company/test/test_merge.py
elcct: I have created a separate test that tries different combination of fields.

marcofucci:
screenshot 2019-02-18 at 15 08 34

This is what I see.

Also, I think I don't particularly like this summary page as it's not understandable if you are not a developer. Wouldn't it be better to keep it the way it was before but add one more link to the investment project admin change list with an or filter (uk_company or investor_company or intermediate_company = x)? You might have to create a custom filter for that but it's pretty easy.

elcct:
I see there is a problem. What if I changed the message to:
```
14 interactions
2 contacts
10 investment projects as UK company
0 investment projects as intermediate company
2 investment projects as investor company
```
This way it will looks much less technical and perhaps add the investment project admin change list too?

elcct:
It looks like this now:

example

datahub/company/merge.py
```diff
@@ -85,45 +104,80 @@ def is_valid(self):
...
-            user,
-        )
+    @staticmethod
+    def transform_merge_results_to_move_entries(results, skip_zeroes=False):
+        """Transforms merge results into move entries to aid the presentation template."""
+        move_entries = []
 
-        return MergeResult(num_interactions_moved, num_contacts_moved)
+        for model, fields in results.items():
+            for field, value in fields.items():
```
marcofucci: `value` is a bit vague, I’d call this `objects_updated`.

datahub/company/admin/merge/step_3.py
```diff
@@ -91,19 +94,27 @@ def _perform_merge(request, merger, model_admin):
...
 
 def _build_success_msg(merger, merge_result):
-    interaction_verbose_name = verbose_name_for_count(
-        merge_result.num_interactions_moved,
-        Interaction._meta,
-    )
-    contact_verbose_name = verbose_name_for_count(merge_result.num_contacts_moved, Contact._meta)
-    return format_html(
+    move_entries = merger.transform_merge_results_to_move_entries(merge_result, skip_zeroes=True)
+
```
marcofucci: I think you used the word `results` in other places so it would be more consistent to call this `merge_results`

datahub/company/admin/merge/step_3.py
```diff
@@ -58,7 +57,11 @@ def confirm_merge(model_admin, request):
     template_name = 'admin/company/company/merge/step_3_confirm_selection.html'
     title = gettext_lazy('Confirm merge')
 
-    move_entries, should_archive_source = merger.get_planned_changes()
+    planned_changes, should_archive_source = merger.get_planned_changes()
```
marcofucci: I would rename `planned_changes` to `planned_merge_results`

datahub/company/merge.py
```diff
@@ -85,45 +104,80 @@ def is_valid(self):
...
-        num_interactions_moved = 0
-        num_contacts_moved = 0
-
-        for interaction in self.source_company.interactions.iterator():
-            interaction.company = self.target_company
-            interaction.save(update_fields=('company',))
-            num_interactions_moved += 1
+        results = {}
+
+        for configuration in self.CONFIGURATION:
```
marcofucci: This appears in a couple of places, maybe you can rename `_update_objects` to `_update_objects_for_configuration` and create a new method called `_update_all_objects` that loops over the configurations and calls `_update_objects_for_configuration`

datahub/company/merge.py
```diff
@@ -1,33 +1,52 @@
...
+    'uk_company',
+)
+
+FIELD_TO_DESCRIPTION = {
+    'investor_company': ' as investor company ',
+    'intermediate_company': ' as intermediate company ',
+    'uk_company': ' as UK company ',
+}
+
 MoveEntry = namedtuple(
```
marcofucci: I'm not sure MoveEntry is the best name for this now as it doesn't tell you that it's the summary of the merge for one model. Maybe `MergeEntryResult` or `MergeEntrySummary` or something like that.

datahub/company/templates/admin/company/company/merge/step_2_primary_selection_radio.html
```diff
@@ -19,14 +19,14 @@
...
     <br>
-    {% with option.target.contacts.count as contact_count %}
-    {{ contact_count }} contact{{ contact_count|pluralize }}
-    {% endwith %}
-    <br>
-    {% with option.target.interactions.count as interaction_count %}
-    {{ interaction_count }} interaction{{ interaction_count|pluralize }}
-    {% endwith %}
-    <br>
+    {% for target in option.target_information %}
```
marcofucci: `target` is probably misleading, I would call this `merge_results` or similar

datahub/company/merge.py
```diff
@@ -85,45 +104,80 @@ def is_valid(self):
...
+        return results
+
+    @classmethod
+    def get_target_information(cls, target_company: Company):
+        """Gets information about the target company related records."""
+        results = {}
+        for configuration in cls.CONFIGURATION:
+            result = cls._update_objects(configuration, target_company, target_company, False)
+            key = f'{configuration.model._meta.app_label}.{configuration.model._meta.model_name}'
+            results[key] = result
```
marcofucci: Can't we just use the model as key and get rid of the `apps.get_model(model)` when digesting the items?

datahub/company/admin/merge/step_3.py
elcct: Hmm, but `merge_results` is a dict and `move_entries` is a list of `MoveEntry` - I think that way is clearer.

marcofucci: sorry I was talking about the argument `merge_result` to `merge_results`.

datahub/company/merge.py
elcct: Changed to `num_objects_updated` - this way it seems even more descriptive.

datahub/company/merge.py
elcct: Great idea, this has greatly improved it. Thanks!

datahub/company/merge.py
elcct: I changed it to `MergeEntrySummary`. Thanks!

datahub/company/templates/admin/company/company/merge/step_2_primary_selection_radio.html
elcct: I agree, I should have made it consistent. Fixed. Thanks!

datahub/company/merge.py
elcct: Ooops... for some reason I thought I cannot do that! Updated and it is much cleaner now. Thanks!

datahub/company/test/test_merge.py
```diff
@@ -127,9 +194,81 @@ def test_merge_succeeds(self, source_num_interactions, source_num_contacts):
...
+            ('investor_company', 'intermediate_company'),
+            ('investor_company', 'uk_company'),
+            ('intermediate_company', 'uk_company'),
+            ('investor_company', 'intermediate_company', 'uk_company'),
+        ),
+    )
+    def test_merge_investment_projects_succeeds(self, fields):
+        """
+        Tests that perform_merge() moves contacts and interactions to the target company,
+        and marks the source company as archived and transferred.
```
marcofucci: docstring needs updating

datahub/company/test/test_merge.py
```diff
@@ -143,7 +282,7 @@ def test_merge_fails_when_not_allowed(self, monkeypatch):
 
 
 def _company_factory(num_interactions, num_contacts):
-    """Factory for a company that has companies and interactions."""
+    """Factory for a company that has companies, interactions and investment projects."""
```
marcofucci: This doesn't deal with investment projects, right?

datahub/company/test/test_merge.py
```diff
@@ -127,9 +194,81 @@ def test_merge_succeeds(self, source_num_interactions, source_num_contacts):
...
+        ),
+    )
+    def test_merge_investment_projects_succeeds(self, fields):
+        """
+        Tests that perform_merge() moves contacts and interactions to the target company,
+        and marks the source company as archived and transferred.
+        """
+        creation_time = datetime(2010, 12, 1, 15, 0, 10, tzinfo=utc)
+        with freeze_time(creation_time):
+            source_company = _company_factory(2, 2)
```
marcofucci: There's already a test for contacts and interactions so I would use `CompanyFactory()` instead

datahub/company/test/test_merge.py
```diff
@@ -127,9 +194,81 @@ def test_merge_succeeds(self, source_num_interactions, source_num_contacts):
...
+        target_company = CompanyFactory()
+        user = AdviserFactory()
+
+        merger = DuplicateCompanyMerger(source_company, target_company)
+        merge_time = datetime(2011, 2, 1, 14, 0, 10, tzinfo=utc)
+
+        with freeze_time(merge_time):
+            result = merger.perform_merge(user)
+
+        assert result == {
```
marcofucci: wouldn't it be easier to use:
```python
other_fields = set(INVESTMENT_PROJECT_COMPANY_FIELDS) - set(fields)
```
and then use that everywhere, e.g.
```
InvestmentProject: {
    **{
        field: 1
        for field in fields
    },
    **{
        field: 0
        for field in other_fields
    },
},
```

datahub/company/merge.py
```diff
@@ -95,18 +112,7 @@ def perform_merge(self, user):
...
-            num_interactions_moved += 1
-
-        for contact in self.source_company.contacts.iterator():
-            contact.company = self.target_company
-            contact.save(update_fields=('company',))
-            num_contacts_moved += 1
+        results = self._update_all_objects(self.source_company, self.target_company, True)
 
         self.source_company.modified_by = user
         self.source_company.mark_as_transferred(
```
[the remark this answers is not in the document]
elcct: I think that is a good idea, so I have moved it to the method. Thanks!

datahub/company/test/test_merge.py
elcct: I should have re-read the docstrings :) Thanks!

datahub/company/test/test_merge.py
elcct: Yes!

datahub/company/test/test_merge.py
elcct: Yes!

datahub/company/test/test_merge.py
elcct: That's nice, I didn't think about it. Updated. Thanks!

datahub/company/admin/merge/step_3.py
```diff
@@ -90,20 +93,31 @@ def _perform_merge(request, merger, model_admin):
...
+        verbose_name = verbose_name_for_count(merge_entry.count, merge_entry.model_meta)
+        messages.append(
+            format_html(
+                MERGE_MOVE_ENTRY,
+                num_objects=merge_entry.count,
+                verbose_name=verbose_name,
+                description=merge_entry.description,
+            ),
+        )
+    merge_entries = ', '.join(messages)
```
reupen: It should be possible to use `format_html_join()`.

datahub/company/admin/merge/step_3.py
```diff
@@ -90,20 +93,31 @@ def _perform_merge(request, merger, model_admin):
...
     )
-    contact_verbose_name = verbose_name_for_count(merge_result.num_contacts_moved, Contact._meta)
-    return format_html(
+
+    messages = []
+    for merge_entry in merge_entries:
+        verbose_name = verbose_name_for_count(merge_entry.count, merge_entry.model_meta)
+        messages.append(
+            format_html(
+                MERGE_MOVE_ENTRY,
```
reupen: I'm not sure `MERGE_MOVE_ENTRY` is the best name for what it is – maybe something like `MERGE_ENTRY_MSG_FRAGMENT`?

datahub/company/templates/admin/company/company/merge/step_3_confirm_selection.html
```diff
@@ -32,14 +32,15 @@
...
 <p>{% trans 'This will make the following changes:' %}</p>
 <ul>
-  {% for move_entry in move_entries %}
+  {% for merge_entry in merge_entries %}
   <li>
-    {% verbose_name_for_count move_entry.count move_entry.model_meta as verbose_name %}
-    {% blocktrans with count=move_entry.count %}
-    {{ count }} {{ verbose_name }} will be moved from {{ source_company_link }} to {{ target_company_link }}.
-    {% endblocktrans %}
+  {% verbose_name_for_count merge_entry.count merge_entry.model_meta as verbose_name %}
```
reupen: Any particular reason for the indentation change here?

datahub/company/admin/merge/step_3.py
elcct: I have changed it to use `format_html_join`, but it looks somewhat more complex now - I am not sure if I am using it correctly.

datahub/company/admin/merge/step_3.py
elcct: Good idea, the `MERGE_MOVE_ENTRY` doesn't sound too great. I have replaced it. Thanks!

datahub/company/templates/admin/company/company/merge/step_3_confirm_selection.html
elcct: This is by accident. Thanks for spotting! Updated.

datahub/company/admin/merge/step_3.py
reupen: You should be able to get rid of the `format_html()` calls and pass `MERGE_ENTRY_MSG_FRAGMENT` and a sequence of sequences to `format_html_join()`.

datahub/company/templates/admin/company/company/merge/step_2_primary_selection_radio.html
```diff
@@ -19,14 +19,14 @@
...
-    {% with option.target.interactions.count as interaction_count %}
-    {{ interaction_count }} interaction{{ interaction_count|pluralize }}
-    {% endwith %}
-    <br>
+    {% for merge_entry in option.merge_entries %}
+    {% verbose_name_for_count merge_entry.count merge_entry.model_meta as verbose_name %}
+    {% blocktrans with count=merge_entry.count description=merge_entry.description %}
+    {{ count }} {{ verbose_name }}{{ description }}
+    {% endblocktrans %}
+    <br/>
```
reupen: I think we can stick with `<br>`, I don't think we need the XHTML compatibility of `<br/>`.

datahub/company/templates/admin/company/company/merge/step_3_confirm_selection.html
```diff
@@ -32,16 +32,17 @@
...
+    {% verbose_name_for_count merge_entry.count merge_entry.model_meta as verbose_name %}
+    {% blocktrans with count=merge_entry.count description=merge_entry.description %}
+      {{ count }} {{ verbose_name }}{{ description }}  will be moved from {{ source_company_link }} to {{ target_company_link }}.
     {% endblocktrans %}
   </li>
   {% endfor %}
+
   <li>
-    {% if should_archive_source %}
+  q  {% if should_archive_source %}
```
reupen: Is this a stray `q`?

datahub/company/admin/merge/step_3.py
```diff
@@ -58,13 +57,17 @@ def confirm_merge(model_admin, request):
     template_name = 'admin/company/company/merge/step_3_confirm_selection.html'
     title = gettext_lazy('Confirm merge')
 
-    move_entries, should_archive_source = merger.get_planned_changes()
+    planned_merge_results, should_archive_source = merger.get_planned_changes()
+    merge_entries = merger.transform_merge_results_to_merge_entry_summaries(
```
reupen: I would suggest not writing `merger.` and writing `DuplicateCompanyMerger.` instead when it is a class or static method, to make it clear it is not actually using the instance. Having said that, I need to look at the changes to `DuplicateCompanyMerger` in more detail.

datahub/company/admin/merge/step_3.py
elcct: Of course! I have updated it. Thank you

datahub/company/merge.py
```diff
@@ -95,35 +112,71 @@ def perform_merge(self, user):
...
             user,
         )
 
-        return MergeResult(num_interactions_moved, num_contacts_moved)
+        return results
+
+    @classmethod
+    def get_target_information(cls, target_company: Company):
+        """Gets information about the target company related records."""
+        return cls._update_all_objects(target_company, target_company, False)
```
reupen: This is a bit confusing as it passes `target_company` for the source and target company. I'm also not sure it really needs `target_company` for what it's doing.

I think the terminology needs reviewing for these functions.

datahub/company/templates/admin/company/company/merge/step_3_confirm_selection.html
elcct: Yes. This is so annoying. Thanks for finding!

elcct:
I also get the feeling that the `DuplicateCompanyMerger` now looks "forced" into a class. I'll see if I can split it into few functions.

datahub/company/admin/merge/step_3.py
elcct: I have removed `DuplicateCompanyMerger` as it was looking "forced".

elcct:
`DuplicateCompanyMerger` is now removed.

datahub/company/merge.py
elcct: So this is no longer like this as I have removed the class. Actually `get_target_information` is gone now and `get_planned_changes` function is being used instead.

datahub/company/merge.py
```diff
@@ -1,147 +1,186 @@
...
+]
+
+
 class MergeNotAllowedError(DataHubException):
     """Merging the specified source company into the specified target company is not allowed."""
 
 
-class DuplicateCompanyMerger:
-    """Utility class for merging duplicate companies."""
+def is_company_clear_to_move(company: Company):
```
reupen: If I'm not missing something, it seems like a name like `is_company_a_valid_merge_source` for this function would pair better with `is_company_a_valid_merge_target`.

elcct: I agree, the proposed name is much better. Changed.

datahub/company/admin/merge/step_3.py
```diff
@@ -73,37 +80,45 @@ def confirm_merge(model_admin, request):
...
+
+    messages = (
+        (
+            merge_entry.count,
+            verbose_name_for_count(merge_entry.count, merge_entry.model_meta),
+            merge_entry.description,
+        ) for merge_entry in merge_entries
+    )
+
+    merge_entries = format_html_join(', ', MERGE_ENTRY_MSG_FRAGMENT, messages)
```
reupen: Looks like this is overwriting `merge_entries` from above – would give this variable a different name.

datahub/company/test/admin/merge/test_step_3.py
```diff
@@ -96,18 +99,37 @@ class TestConfirmMergeViewPost(AdminTestMixin):
...
         source_interactions = list(source_company.interactions.all())
         source_contacts = list(source_company.contacts.all())
 
+        source_investment_projects = set()
+        investment_projects_info = defaultdict(int)
+        for investment_project_field in INVESTMENT_PROJECT_COMPANY_FIELDS:
+            investment_projects = InvestmentProject.objects.filter(
+                **{investment_project_field: source_company},
+            )
+            source_investment_projects.union(set(investment_projects))
```
reupen: This looks a bit odd as `set.union()` returns a new set (but the return value here isn't used).

Was it meant to be something like `source_investment_project |= set(investment_projects)`?

Maybe it would be simpler to keep the logic to something like:

```python
source_investment_projects_by_field = {
    investment_project_field: list(
        InvestmentProject.objects.filter(**{investment_project_field: source_company})
    ),
}
```

(and changing the assertions accordingly).

datahub/company/test/admin/merge/test_step_3.py
```diff
@@ -263,11 +302,18 @@ def test_merge_fails(self, source_company_factory, target_company_factory):
...
-def _company_factory(num_interactions, num_contacts):
-    """Factory for a company that has companies and interactions."""
+def _company_factory(num_interactions, num_contacts, num_investment_projects):
+    """Factory for a company that has companies, interactions and investment projects."""
     company = CompanyFactory()
     ContactFactory.create_batch(num_contacts, company=company)
     CompanyInteractionFactory.create_batch(num_interactions, company=company)
+
+    fields_iter = cycle(INVESTMENT_PROJECT_COMPANY_FIELDS[:num_investment_projects])
+    fields = islice(fields_iter, 0, num_investment_projects)
```
reupen: This still looks a bit odd – I can't see what the first slice (in the `cycle()` call) is doing (seems like the `islice()` call would negate it).

datahub/company/merge.py
```diff
@@ -1,147 +1,186 @@
...
 
-    def is_valid(self):
-        """Returns whether the merge is allowed."""
-        return self.is_source_valid() and self.is_target_valid()
 
-    @transaction.atomic
-    def perform_merge(self, user):
-        """
-        Merges the source company into the target company.
+def _process_all_objects(process_objects_fn, **kwargs):
```
reupen: It looks like it would be simple to change this one to a dictionary comprehension:

```python
    return {
        configuration.model: process_objects_fn(configuration, **kwargs)
        for configuration in MERGE_CONFIGURATION
    }
```

datahub/company/test/test_merge.py
```diff
@@ -127,19 +193,110 @@ def test_merge_succeeds(self, source_num_interactions, source_num_contacts):
...
+        (
+            (),
+            ('investor_company',),
+            ('intermediate_company',),
+            ('uk_company',),
+            ('investor_company', 'intermediate_company'),
+            ('investor_company', 'uk_company'),
+            ('intermediate_company', 'uk_company'),
+            ('investor_company', 'intermediate_company', 'uk_company'),
+        ),
```
reupen: If I'm not mistaken, it looks like it would be simpler to cover all of the cases using three different parametrisations. (It could be three booleans, or possibly something more complicated with sequences...)

datahub/company/test/test_merge.py
```diff
@@ -127,19 +193,110 @@ def test_merge_succeeds(self, source_num_interactions, source_num_contacts):
...
+            (False, False),
+        ),
+    )
+    @patch('datahub.company.merge.is_company_a_valid_merge_target')
+    @patch('datahub.company.merge.is_company_a_valid_merge_source')
+    def test_merge_fails_when_not_allowed(
+        self,
+        is_company_a_valid_merge_source_mock,
+        is_company_a_valid_merge_target_mock,
+        source_clear,
```
reupen: It would be good rename `source_clear` and `target_clear` so that the names are more aligned with the terminology elsewhere.

datahub/company/admin/merge/step_3.py
elcct: I have renamed it now. Thanks!

datahub/company/test/admin/merge/test_step_3.py
elcct: That test wasn't well so I changed it with your suggestions. It is better now. Thanks!

datahub/company/test/admin/merge/test_step_3.py
elcct: I forgot to remove that! Thanks

datahub/company/merge.py
elcct: Indeed. Replaced :)

datahub/company/test/test_merge.py
elcct: I feel like leaving as is as I think it is still simple but if we were to add more fields in to the mix I would have changed it.

datahub/company/test/test_merge.py
elcct: Yes, it should be better to align the names. I changed it thanks!

PR #1460, opened 2019-02-14
Add Large Capital Investor Profile

marcuspp:
### Description of change
As shown in the investment prototype this is backend support for creating and amending a large capital profile. A new django app `investor_profile`  includes the views and all the metadata, all field options have been agreed by the users/stakeholders. Currently a company may only have one large capital profile however this could change in the future hence the unique together constraint.

PR on the large size but mainly due to the metadata that has been added, recommended to review by commit. Note that the PR is marked as blocked as merging might need to be delayed.

### Checklist


datahub/investor_profile/models.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+
+    investor_company = models.ForeignKey(
+        'company.Company',
+        related_name='investor_profiles',
+        on_delete=models.CASCADE,
+    )
+
+    profile_type = models.ForeignKey(
+        'ProfileType',
```
marcofucci: `investor_profile.ProfileType` would be more consistent

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,238 @@
...
+    relationship_health = models.ForeignKey(
+        'RelationshipHealth',
+        related_name='+',
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+    )
+
+    background_checks_conducted = models.ForeignKey(
+        'BackgroundChecksConducted',
```
marcofucci: You have a value of "not yet checked", wouldn't it be better to default to this and set it to not null?

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,238 @@
...
+        blank=True,
+    )
+
+    relationship_health = models.ForeignKey(
+        'RelationshipHealth',
+        related_name='+',
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+    )
```
marcofucci: You have a value of "not yet checked" for backgroundchecksconducted, shouldn't this have a similar one for keeping things consistent?

datahub/investor_profile/test/factories.py
```diff
@@ -0,0 +1,20 @@
+import factory
+
+from datahub.company.test.factories import CompanyFactory
+from datahub.core.test.factories import to_many_field
+from datahub.investor_profile.constants import ProfileType as ProfileTypeConstant
+
+
+class InvestorProfileFactory(factory.django.DjangoModelFactory):
+    """Investment project team member factory."""
```
marcofucci: docstring needs updating

datahub/investor_profile/urls.py
```diff
@@ -0,0 +1,22 @@
...
+
+GET_AND_POST = {
+    'get': 'list',
+    'post': 'create',
+}
+
+GET_AND_PATCH_ITEM = {
+    'get': 'retrieve',
+    'patch': 'partial_update',
+}
```
marcofucci: you are not reusing `GET_AND_POST` and `GET_AND_PATCH_ITEM` so they can go in the `as_view` actions param directly I guess?

datahub/investor_profile/metadata.py
```diff
@@ -0,0 +1,70 @@
...
+
+registry.register(
+    metadata_id='large-capital-investment-type',
+    model=models.LargeCapitalInvestmentType,
+)
+
+registry.register(
+    metadata_id='profile-type',
+    model=models.ProfileType,
+)
```
marcofucci: I'm not sure the frontend needs this at all? It looks like the frontend would have to hard code those anyway and the profile_type is read-only.

datahub/investor_profile/validate.py
```diff
@@ -0,0 +1,13 @@
...
+from datahub.investor_profile.models import InvestorProfile
+
+
+def get_incomplete_fields(instance, fields):
+    """Returns a list of fields that are incomplete"""
+    combiner = DataCombiner(instance, {}, model=InvestorProfile)
+    incomplete_fields = []
+    for field in fields:
+        if field_incomplete(combiner, field):
+            incomplete_fields.append(field)
```
marcofucci:
```python
return [
    field
    for field in fields
    if field_incomplete(combiner, field)
]
```

would be clearer I think

datahub/investor_profile/serializers.py
```diff
@@ -0,0 +1,246 @@
...
+                profile_type_id=ProfileTypeConstant.large.value.id,
+        ).exists():
+            raise serializers.ValidationError(
+                'Investor company already has large capital investor profile',
+            )
+        return value
+
+    def create(self, validated_data):
+        """Overrides the create method to add the large profile type id into the data."""
+        validated_data['profile_type_id'] = ProfileTypeConstant.large.value.id
```
marcofucci: `ProfileTypeConstant.large.value.id` is specified in several places (including the view). It should probably get passed in from the view to the serializer and used everywhere else here.

datahub/investor_profile/serializers.py
```diff
@@ -0,0 +1,246 @@
...
+
+    investor_company = NestedRelatedField(
+        Company,
+        allow_null=False,
+    )
+
+    profile_type = NestedRelatedField(
+        ProfileType,
+        read_only=True,
+    )
```
marcofucci: I'm not sure what it adds to return the `profile_type` as this is the endpoint for the `Large` one. It would be better to get rid of this.

changelog/investor_profile/large_capital_profile.api
```diff
@@ -0,0 +1,63 @@
+``GET /v3/large-investor-profile`` returns a list of all the large capital profiles.
```
marcofucci: I would make these part of investment (`/changelog/investment`) instead of creating a new one. Also as it is now, it would not work as `pyproject.toml` would need to change.

reupen: It would be nice to add `.rst` to the news fragment file names so that we get a better idea of how it will be formatted.

datahub/investor_profile/migrations/0002_auto_20190211_1445.py
```diff
@@ -0,0 +1,321 @@
...
+            model_name='investorprofile',
+            name='client_contacts',
+            field=models.ManyToManyField(blank=True, related_name='investor_profiles', to='company.Contact'),
+        ),
+        migrations.AddField(
+            model_name='investorprofile',
+            name='dit_relationship_manager',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='investor_profiles', to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AddField(
```
reupen: I'm not sure all of these `AddField`s need to be separate operations; it looks like the migrations could be regenerated to generate more a optimal set of operations.

datahub/cleanup/management/commands/delete_old_records.py
```diff
@@ -64,6 +64,7 @@ class Command(BaseCleanupCommand):
             # We want to delete the relations below along with any expired companies
             excluded_relations=(
                 Company._meta.get_field('dnbmatchingresult'),
+                Company._meta.get_field('investor_profiles'),
```
reupen: I think this needs some more thought. If a company has had an investor profile added recently or one has been updated recently that I would have thought that we don't want to delete the company.

Perhaps it should be added to `relation_filter_mapping` in a similar way to propositions on InvestmentProject?

datahub/investor_profile/views.py
```diff
@@ -0,0 +1,20 @@
...
+from datahub.investor_profile.models import InvestorProfile
+from datahub.investor_profile.serializers import LargeCapitalInvestorProfileSerializer
+from datahub.oauth.scopes import Scope
+
+
+class LargeCapitalInvestorProfileViewSet(CoreViewSet):
+    """Large capital investor profile view set"""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
```
reupen: It looks like this is omitting model permissions, is there a reason for that?

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,238 @@
...
+        'InvestorType',
+        related_name='+',
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+    )
+
+    investable_capital = models.DecimalField(
+        max_digits=19,
+        decimal_places=0,
```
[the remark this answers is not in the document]
marcuspp: I copied this from total_investment on the investment projects. I will change to an integer. 👍

datahub/investor_profile/constants.py
```diff
@@ -0,0 +1,145 @@
+from enum import Enum
```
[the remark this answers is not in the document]
marcuspp: I will move all the constants that are only used for tests and leave the ones used in code in here.

datahub/investor_profile/fixtures/asset_class_interest.yaml
```diff
@@ -0,0 +1,96 @@
+- model: investor_profile.assetclassinterest
```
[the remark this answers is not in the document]
marcuspp: I understand what you are saying but I don't want to put all the files straight into the migrations folder as i think it gets very messy. How about a data folder within the migrations folder `data\0002` ?

datahub/investor_profile/models.py
marcuspp: No because I don't think these are the same thing. I'm going to leave this as is as I think its important the question is answered and not assumed. Plus we want to know for the incomplete fields if someone has answered the question.

datahub/investor_profile/validate.py
marcuspp: Does it matter? and tbh I never think inline for loops are easier to read because you have to read them bottom to top or even right to left. I find the more traditional for loop append far easier to understand. If however its more of a coding style or performance issue then I'll update it.

datahub/investor_profile/serializers.py
marcuspp: I don't think this is necessary at this time, it will add X amount of lines of code for no gain and make the view and serializer more complex. I'm a fan of not coupling the serializer to a view unless absolutely necessary.

changelog/investor_profile/large_capital_profile.api
marcuspp: Why? I think they are very different things the code most certainly is.

datahub/investor_profile/metadata.py
marcuspp: You are correct that it isn't needed with this PR but will be needed when search is added as there will be a dropdown for selecting which profile type you require. If you'd like I will remove it but this will need to be re-added when search is implemented.

datahub/investor_profile/urls.py
marcuspp: I'd like to leave them as is if thats okay. I think it reads better like it is and they will be reused when other end points are added.

datahub/investor_profile/serializers.py
marcuspp: Agreed.  I will remove it.

datahub/cleanup/management/commands/delete_old_records.py
marcuspp: Yes, sorry I struggled with the cleanup tests, I found them a bit difficult to understand. But your'e right that should be the case. I'll update them to be like propositions. Thanks

changelog/investor_profile/large_capital_profile.api
marcofucci: Investment projects and investor profiles are both related to investment, right? So they should go in a common investment group otherwise we start having too many groups for news fragments.

datahub/investor_profile/fixtures/asset_class_interest.yaml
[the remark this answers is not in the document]
marcofucci: I don’t think it would get particularly messy if you prefix your files with 0002_... as they will all be grouped together. If you don’t like that you could create a folder migrations/0002_.../ and put all the fixtures in it. I think it’s still better to list all the fixtures explicitly in your data migration file instead of getting all the files from the folder as if another dev puts an extra file in that, the migration logic for new environments would be different.

datahub/investor_profile/metadata.py
marcofucci: I think that search should follow the same logic of the endpoint so you should have one different endpoint for each type. If you want to aggregate the search and have a type filter, it might actually make more sense to aggregate the add and edit endpoints and pass a type there as param. Unless I’m missing something?

datahub/investor_profile/urls.py
marcofucci: 👍🏻 In that case, I’d be better to rename the first to GET_AND_POST_COLLECTION

datahub/investor_profile/validate.py
marcofucci: It was just about coding style not it’s not a big deal

elcct: I find @marcofucci example easy to read - as from the top it goes: what it does, for what, when - so the most important thing is still on the top. But I am okay for both ways.

reupen: Comprehensions are usually preferred when there are no side effects, so I would use one as well.

datahub/investor_profile/metadata.py
marcuspp: You are right, just double checked with the search requirements for investor profiles and they will be different endpoints so I will remove it,

changelog/investor_profile/large_capital_profile.api
marcuspp: Okay, I'll move them.

datahub/investor_profile/migrations/0002_auto_20190211_1445.py
marcuspp: Did this so I could breakdown the PR so commits where individually reviewable. Will squash them all into 1 migration now.

datahub/investor_profile/models.py
marcuspp: RelationshipHealth has now been remove, i did however double check with my team as regards should `not yet checked` being added as an option elsewhere. The answer was no as how this will be used adding `n/a` as an option will only dilute the data we end up.

datahub/investor_profile/test/test_views.py
```diff
@@ -0,0 +1,385 @@
...
+from datahub.investor_profile.test.factories import InvestorProfileFactory
+
+
+class TestCreateLargeCapitalProfileView(APITestMixin):
+    """Test creating a large capital profile"""
+
+    def test_large_capital_unauthorized_user_(self):
+        """Should return 401"""
+        url = reverse('api-v3:large-investor-profile:collection')
+        api_client = APIClient()
```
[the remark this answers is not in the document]
marcuspp: Isn't that `api_client` for an authenticated user? I did have a look at it (this test is copied from `investment_projects` btw.

datahub/investor_profile/views.py
marcuspp: I did omit them on purpose as was told we don't mind who creates and edits. However it's now been decided to make them the same as `investment_projects`. Are you okay if I add the permissions in on separate PR so this one doesn't get any bigger?

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,238 @@
...
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+    )
+
+    dit_advisers = models.ManyToManyField(
+        'company.Advisor',
+        related_name='+',
+        blank=True,
+    )
```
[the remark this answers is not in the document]
marcuspp: These fields have now been removed due to the work your team is doing with core teams it was deemed these are no longer required.

datahub/cleanup/management/commands/delete_old_records.py
marcuspp: Does this seem correct https://github.com/uktrade/data-hub-leeloo/pull/1460/commits/20f6d5e3ccc5fc17b1e8dd65475b3894a94fb358 ?

datahub/investor_profile/admin.py
```diff
@@ -0,0 +1,11 @@
+from django.contrib import admin
+
+from datahub.investor_profile import models
+
+
+@admin.register(models.InvestorProfile)
+class InvestorProfileAdmin(admin.ModelAdmin):
+    """Investment project stage admin."""
```
elcct: I think the comment needs update.

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,209 @@
...
+
+    uk_region_locations = models.ManyToManyField(
+        'metadata.UKRegion',
+        related_name='+',
+        blank=True,
+        verbose_name='possible UK regions',
+    )
+
+    other_countries_considering = models.ManyToManyField(
```
elcct: It doesn't sound right to me. Wouldn't `other_countries_considered` describe it better (as in the help text)?

datahub/investor_profile/serializers.py
```diff
@@ -0,0 +1,210 @@
...
+        return get_incomplete_fields(instance, LARGE_CAPITAL_REQUIREMENTS_FIELDS)
+
+    def get_incomplete_location_fields(self, instance):
+        """Returns a list of all the location fields that are incomplete"""
+        return get_incomplete_fields(instance, LARGE_CAPITAL_LOCATION_FIELDS)
+
+    def validate_investor_company(self, value):
+        """Validates the company does not already have a large capital investment profile"""
+        if value.investor_profiles.filter(
+                profile_type_id=ProfileTypeConstant.large.value.id,
```
elcct: Indent see to be off.

datahub/investor_profile/test/test_views.py
```diff
@@ -0,0 +1,326 @@
...
+    RequiredChecksConducted as RequiredChecksConductedConstant,
+    Restriction as RestrictionConstant,
+    ReturnRate as ReturnRateConstant,
+    TimeHorizon as TimeHorizonConstant,
+)
+from datahub.investor_profile.test.factories import InvestorProfileFactory
+
+
+class TestCreateLargeCapitalProfileView(APITestMixin):
+    """Test creating a large capital profile"""
```
elcct: We end comments with a full stop.

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -51,6 +52,7 @@
             (OrderFactory, 'contact'),
             (QuoteFactory, 'accepted_by'),
             (InvestmentProjectFactory, 'client_contacts'),
+            (InvestorProfileFactory, 'client_contacts'),
```
elcct: I think this needs to be looked at as here one of the tests fail because `InvestorProfile` doesn't have `client_contacts` field.

datahub/cleanup/test/commands/test_delete_old_records.py
```diff
@@ -184,6 +185,17 @@
                     },
                 ],
             },
+            {
+                'factory': InvestorProfileFactory,
+                'field': 'company',
+                'expired_objects_kwargs': [],
+                'unexpired_objects_kwargs': [
+                    {
+                        'created_on': COMPANY_DELETE_BEFORE_DATETIME - relativedelta(days=1),
```
reupen: Isn't this the expired case? As it's OK to delete companies when the investor profile hasn't been edited for 10 years? It would be good to have both cases covered as well.

Also `COMPANY_DELETE_BEFORE_DATETIME` uses `COMPANY_EXPIRY` while the actual clean-up config uses `INVESTOR_PROFILE_EXPIRY_PERIOD` for `investor_profiles` – it would be good to keep it consistent.

datahub/investor_profile/models.py
reupen: I would say `other_countries_being_considered`.

datahub/investor_profile/views.py
reupen: I think it's a risk to add the endpoints without permissions. I am not sure on what you mean by same as investment_projects – isn't that determined by what groups the permissions are added to?

The investment project permissions are very complicated due to the LEP and DA restrictions.

datahub/investor_profile/test/test_views.py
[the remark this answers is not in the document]
reupen: It is defined in `/conftest.py`. It just instantiates an APIClient.

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,209 @@
...
+        blank=True,
+        null=True,
+        help_text='Investable capital amount in USD',
+    )
+
+    global_assets_under_management = models.IntegerField(
+        blank=True,
+        null=True,
+        help_text='Global assets under management amount in USD',
+    )
```
marcofucci: `IntegerField` would support up to 2 billion if I'm not wrong, would it be enough?

datahub/investor_profile/models.py
```diff
@@ -0,0 +1,209 @@
...
+    other_countries_considered = models.ManyToManyField(
+        'metadata.Country',
+        related_name='+',
+        blank=True,
+        help_text='The other countries being considered for investment',
+    )
+
+    notes_on_locations = models.TextField(
+        blank=True,
+        null=True,
```
marcofucci: it would be better to get rid of `null=True` here too.

datahub/investor_profile/views.py
marcofucci: Yeah, we have a really strict role that everything which gets committed should be releasable without huge side effects especially caused by the bus factor. That's also why this PR cannot be merged without the changes to the merge tool I mentioned in another comment.

datahub/investor_profile/serializers.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def validate_investor_company(self, value):
+        """Validates the company does not already have a large capital investment profile."""
+        if value.investor_profiles.filter(
+            profile_type_id=ProfileTypeConstant.large.value.id,
+        ).exists():
+            raise serializers.ValidationError(
+                'Investor company already has large capital investor profile',
+            )
+        return value
```
marcofucci: With this, you cannot PATCH a profile passing the same `investor_company`.
Also, it seems a bit weird that you're able to change the investor_company in a PATCH, shouldn't this be disallowed?

datahub/investor_profile/serializers.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def get_incomplete_requirements_fields(self, instance):
+        """Returns a list of all the requirement fields that are incomplete."""
+        return get_incomplete_fields(instance, LARGE_CAPITAL_REQUIREMENTS_FIELDS)
+
+    def get_incomplete_location_fields(self, instance):
+        """Returns a list of all the location fields that are incomplete."""
+        return get_incomplete_fields(instance, LARGE_CAPITAL_LOCATION_FIELDS)
+
+    def validate_investor_company(self, value):
```
marcofucci: Shouldn't this also check that the investor company is a foreign company and that (check with your team) it's not a subsidiary?

changelog/investment/large_capital_profile.api.rst
```diff
@@ -0,0 +1,105 @@
+``GET /v3/large-investor-profile`` returns a list of all the large capital profiles.
+
+``POST /v3/large-investor-profile`` creates a large capital profile for a given ``investor_company``.
+
+``GET /v3/large-investor-profile/<uuid:pk>`` returns the large capital profile for the given id.
+
+``PATCH /v3/large-investor-profile/<uuid:pk>`` updates the large capital profile for the given id.
```
marcofucci: Can you pleasea add this to v4 instead as we're eventually going to get rid of v3?

datahub/investor_profile/views.py
marcuspp: Sure, I think I need to speak to one of you regarding permissions just to check what the minimum required would be and to be honest how the frontend uses them when they are in place. I would have thought a check for an authenticated user was the minimum? @reupen you are right we don't need the complicated LEP/DA restrictions but i'm unsure if model permissions are required either.

marcuspp:
@marcofucci 

Point 1. I wrote it this way because there was talk of maybe in future allowing more than 1 large capital profile per company, so I wanted the backend restrictions on this to be the minimum. This is why it isn't a one-to-one relationship in the database and only enforced with a constraint.

In a separate PR i was going to add a link to the profile to the company endpoint. So viewing the company would give you the profile ID so no need to view the list.

I do like your suggestion but I thought the urls could get very long if we follow that pattern for other things and I thought it is preferred to keep them short. If you're okay with that then I can change it.

Point 2. If that is policy then I'm okay moving it. It is the same discussion we had regarding the change log. The only caveat I'd have are the files within investment folder (the files that aren't in any folder) are all related to projects none of which are related to investor profiles which would be confusing. Maybe they could be moved out into their own project folder.

datahub/investor_profile/models.py
marcuspp: You are correct that might not be enough, this must be why it used a DemicalField with decimal_places=0 on investment projects. What do you suggest? Changing to the DecimalField or using a BigInteger field?

marcofucci:
@marcuspp there are pros and cons with both, are you okay to discuss this in our next BE catchup?

marcuspp:
@marcofucci @elcct @reupen I'm going to close this PR. As I have to redo most of it due to the new app layout. Will break it up abit into easier chunks.

PR #1462, opened 2019-02-15
Search company api v4

marcofucci:
### Description of change

This adds v4 of the following endpoints:
- `/v4/search/company`
- `/v4/search/company/autocomplete`
- `/v4/search/company/export`

with nested object for addresses.
Also, `the trading_address_*` fields were removed and replaced by the nested object `address`.

### Checklist

### DIFF
--- a/datahub/search/company/views.py
+++ b/datahub/search/company/views.py
@@ -65,8 +65,13 @@ class SearchCompanyAPIViewMixin:
     }
 
 
-class SearchCompanyAPIView(SearchCompanyAPIViewMixin, SearchAPIView):
-    """Filtered company search view."""
+class SearchCompanyAPIViewV3(SearchCompanyAPIViewMixin, SearchAPIView):
+    """Filtered company search view V3."""
+
+    fields_to_exclude = (
+        'address',
+        'registered_address',
+    )
 
 
 class SearchCompanyExportAPIView(SearchCompanyAPIViewMixin, SearchExportAPIView):
--- a/datahub/search/company/views.py
+++ b/datahub/search/company/views.py
@@ -74,6 +74,27 @@ class SearchCompanyAPIViewV3(SearchCompanyAPIViewMixin, SearchAPIView):
     )
 
 
+class SearchCompanyAPIViewV4(SearchCompanyAPIViewMixin, SearchAPIView):
+    """Filtered company search view V4."""
+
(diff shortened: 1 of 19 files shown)


datahub/search/company/views.py
```diff
@@ -65,8 +65,13 @@ class SearchCompanyAPIViewMixin:
     }
 
 
-class SearchCompanyAPIView(SearchCompanyAPIViewMixin, SearchAPIView):
-    """Filtered company search view."""
+class SearchCompanyAPIViewV3(SearchCompanyAPIViewMixin, SearchAPIView):
+    """Filtered company search view V3."""
+
+    fields_to_exclude = (
```
reupen: Won't these still be in v3 global (basic) search?

(I also noticed that `suggest` is leaking into responses too...)

marcofucci: Yes, I think that adding the same logic to the basic (global) search is a bit more complicated and would be messier/too complex.
1. the basic search doesn't know about the specific search app views
2. repeated logic as the fields would have to be on the entity related search and global search
3. some entity would return v3 and some v4 of the resources 

Do you think it's worth it?

reupen: > Do you think it's worth it?

Not really, no. Just seemed a bit weird that global search doesn't match either v3 or v4 of company entity search, but it's not a big deal I guess.

marcofucci: yeah I don't like it at all either, not sure how to solve that 😞

datahub/search/urls.py
```diff
@@ -5,29 +5,58 @@
...
         ))
+
+
+# API V4 - new format for addresses
+
+# TODO add global search when all search apps are v4 ready
+urls_v4 = []
+
+for search_app in get_search_apps():
+    if search_app.view_v4:
```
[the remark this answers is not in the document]
marcofucci: I've added a helper function, filter None is pretty ugly but it's the best I can think of unless you have better ideas?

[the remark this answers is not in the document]
reupen: The other options I can think of are:

- passing `urls_*` to `_construct_path()` and making it do the appending
- making `_construct_path()` return a list (of one or no items) and doing:

```pyhon
urls = [
    *urls,
    *_construct_path(...),
    *_construct_path(...),
   ...
]
```

PR #1472, opened 2019-02-19
Company addresses in search logic

marcofucci:
### Description of change

This uses the new company address and registered address in nested object format for the basic and company entity search.

I suggest you look at the individual commits in this PR.

### Checklist

### DIFF
--- a/datahub/search/company/test/test_elasticsearch.py
+++ b/datahub/search/company/test/test_elasticsearch.py
@@ -387,6 +387,8 @@ def test_get_basic_search_query():
                         'multi_match': {
                             'query': 'test',
                             'fields': [
+                                'address.country.name.trigram',
+                                'address.postcode.trigram',
                                 'address_country.name_trigram',
                                 'address_postcode_trigram',
                                 'company.name',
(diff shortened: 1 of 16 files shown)


datahub/search/company/test/test_elasticsearch.py
```diff
@@ -561,7 +555,7 @@ def test_limited_get_search_by_entity_query():
                                     },
                                 }, {
                                     'range': {
-                                        'estimated_land_date': {
+                                        'archived': {
```
elcct: Why this change?

reupen: I think it's because this test is for companies, so it was a bit weird to use `estimated_land_date_after` and `estimated_land_date_before`.

datahub/search/contact/dict_utils.py
```diff
@@ -5,11 +5,7 @@ def computed_address_field(field):
     """Gets Contact address from Company is address_same_as_company."""
     def get_field(contact):
```
reupen: I really don't like `computed_address_field()` :cry: 

I tried to simplify it, how about something like this?

```python
def computed_address_field(field):
    """Gets Contact address from Company is address_same_as_company."""
    def get_field(contact):
        obj = contact.company if contact.address_same_as_company else contact
        value = getattr(obj, field)

        if field == 'address_country':
            return dict_utils.id_name_dict(value)

        return value

    return get_field
```

datahub/search/company/test/test_elasticsearch.py
marcofucci: yep that

datahub/search/contact/dict_utils.py
marcofucci: yes I agree, I wanted to get back as I didn't like it either. Your suggestion works well.

datahub/search/company/test/test_elasticsearch.py
elcct: Makes sense!

PR #1484, opened 2019-02-22
Make Hawk authentication classes reusable

reupen:
### Description of change

This:

- refactors the existing activity stream Hawk authentication classes to make them reusable
- lays the groundwork for supporting multiple Hawk clients
- adds a scope to each Hawk client to control which endpoints they are allowed to use

`@decorator_from_middleware` was and still is being used, which is wrong as it is for old-style middleware. However, that is not really related to these changes so I have left it as is.

There is also some inconsistency between `OAuth2Authentication` and `HawkAuthentication` as the former returns `None` on authentication failure while the latter raises `AuthenticationFailed`. Again, I have left it as it is.

### Checklist


config/settings/common.py
```diff
@@ -458,13 +460,22 @@
...
+        'key': env('ACTIVITY_STREAM_SECRET_ACCESS_KEY'),
+        'scope': HawkScope.activity_stream,
+    },
+}
+HAWK_RECEIVER_IP_WHITELIST = env('HAWK_RECEIVER_IP_WHITELIST', default='')
+HAWK_RECEIVER_NONCE_EXPIRY_SECONDS = 60
+
+# Check that there are the correct number of items in HAWK_RECEIVER_CREDENTIALS to make sure
+# no duplicate Hawk IDs were configured
+if len(HAWK_RECEIVER_CREDENTIALS) != 1:
```
michalc: Slightly odd/dead code?: it looks like there isn't a way to configure the application for this condition to ever be `True`: `len(HAWK_RECEIVER_CREDENTIALS`) will always be `1`, no matter what?

reupen: Not right now, and it is awkward, yes. I will be adding a second client in a subsequent PR so this was some protection against the same ID being used for two clients (as one set of credentials will get discarded from `HAWK_RECEIVER_CREDENTIALS` if that happens).

I tried a few things but couldn't come up with anything I liked more. We could also move the credentials to the database now and avoid the need to do anything like this (as uniqueness could be enforced elsewhere).

michalc: ah ok

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -1,8 +1,5 @@
-import datetime
```
michalc: Perhaps a minor thing, but the tests are now more coupled to even private-r(?) implementation details, and I have less confidence the AS endpoint satisfies all things the previous tests asserted on: the assertions are no longer on the AS endpoint, but on an endpoint created specifically for the tests.

(Plus, it's a hard to see if this is definitely a refactor, since both tests and production code has changed, rather than just production code keeping the tests the same)

I wonder if it's better to keep the tests on the AS endpoint rather creating ones in the test environment to test on. Once another endpoint is created, the tests can be re-used/parameterised against the various endpoints. (Yes: the same code in the Hawk classes will run under various tests, but I think the benefit of higher confidence in production code is worth it)

datahub/core/test/support/views.py
```diff
@@ -33,3 +42,28 @@ class MultiAddressModelViewset(CoreViewSet):
     permission_classes = []
     serializer_class = MultiAddressModelSerializer
     queryset = MultiAddressModel.objects.all()
+
+
+class HawkViewWithoutScope(APIView):
```
michalc: Are these views not actually being used yet by production code? Is it better to hold off putting them in the codebase until they are?

michalc: Ignore me! Deleting.

config/settings/common.py
```diff
@@ -458,13 +460,22 @@
...
-ACTIVITY_STREAM_SECRET_ACCESS_KEY = env('ACTIVITY_STREAM_SECRET_ACCESS_KEY')
-ACTIVITY_STREAM_NONCE_EXPIRY_SECONDS = 60
+# Hawk
+HAWK_RECEIVER_CREDENTIALS = {
+    env('ACTIVITY_STREAM_ACCESS_KEY_ID'): {
+        'key': env('ACTIVITY_STREAM_SECRET_ACCESS_KEY'),
+        'scope': HawkScope.activity_stream,
+    },
+}
+HAWK_RECEIVER_IP_WHITELIST = env('HAWK_RECEIVER_IP_WHITELIST', default='')
```
marcofucci: I think I would split it to convert it to a list as currently `<address> in HAWK_RECEIVER_IP_WHITELIST` works but it's a bit of a hack IMHO.

config/settings/common.py
marcofucci: I'm also not sure I like this. Does it protect against that problem anyway? If you add another key to that dict without changing this line, `ImproperlyConfigured` is not raised and you think that everything works just fine.

datahub/activity_stream/views.py
```diff
@@ -134,32 +136,57 @@ def _authenticate_by_hawk(self, request):
...
+        """Checks if the client has the scope required by the view."""
+        required_hawk_scope = getattr(view, 'required_hawk_scope', None)
+
+        if required_hawk_scope is None:
+            raise ValueError('required_hawk_scope was not set on the view')
+
+        if not isinstance(request.successful_authenticator, HawkAuthentication):
+            return False
+
+        return required_hawk_scope == request.auth.resource.credentials['scope']
```
marcofucci: `constant_time_compare` would be safer.

config/settings/common.py
reupen: Good spot! `HAWK_RECEIVER_IP_WHITELIST` was simply `ACTIVITY_STREAM_IP_WHITELIST` renamed, but it's simple enough to change it to use `env.list` so I've done that.

config/settings/common.py
reupen: Do you have another idea? The other way was to declare a sequence of sequences with the IDs, keys and scopes and then construct the dictionary from that. I did not like it, but it solves the duplicate-checking problem.

datahub/activity_stream/test/test_activity_stream_views.py
marcofucci: What I think you mean would have to be applied to all possible cases. E.g. when using generic DRF code, you should test the different aspects of the view that makes use of that generic code and not just trust the DRF testing coverage. I don't find this feasible and it's definitely not conventionally used.

IMHO, when you use the generic logic, you shouldn't really need to know or care that the internal implementation checks that the authorization header is not generated from an incorrect method.

We really need the generic/core tests for the generic logic and these specific tests below. You might say that we need to duplicate the generic tests here as well but I don't think it adds any value.

config/settings/common.py
marcofucci: The other option is not to have the check at all, in theory this shouldn't really happen.

datahub/activity_stream/views.py
reupen: The scopes aren't secret so we don't need `constant_time_compare`.

config/settings/common.py
reupen: When we have multiple clients, it could happen if the same ID is specified for two clients (in the environment variables).

I have changed the check to the list-based version which is probably better overall.

datahub/activity_stream/views.py
marcofucci: true

PR #1485, opened 2019-02-25
Add company matches upload command

elcct:
### Description of change

This adds a command that enables upload of matching meta data to `DnBMatchingResult` table.
The matching record will be created or overwritten if exists.
This adds `modified_on` column to the `DnBMatchingResult` table because the record can be updated.

I have not created news fragments as this work is temporary.

### Checklist


datahub/dbmaintenance/management/commands/update_company_matches.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        company = Company.objects.get(pk=parse_uuid(row['id']))
+
+        try:
+            raw_data = b64decode(row['data'])
+            data = json.loads(raw_data)
+        except Exception:
+            logger.warning(f'Cannot decode provided data for company {company.pk}.')
```
reupen: Doesn't `CSVBaseCommand.process_row()` do this logging anyway?

datahub/dbmaintenance/management/commands/update_company_matches.py
```diff
@@ -0,0 +1,38 @@
...
+            raise
+
+        if data and not simulate:
+            matching_result, _ = DnBMatchingResult.objects.update_or_create(
+                company_id=company.id,
+                defaults={
+                    'data': data,
+                },
+            )
+            matching_result.save()
```
reupen: Is this save necessary? I thought `QuerySet.update_or_create()` would do that.

datahub/dbmaintenance/management/commands/update_company_matches.py
elcct: That is true. It's not necessary so I removed it. Thanks!

datahub/dbmaintenance/management/commands/update_company_matches.py
elcct: I was doing something else there and forgot to remove it :( It's gone now. Thanks!

datahub/dbmaintenance/management/commands/update_company_matches.py
```diff
@@ -0,0 +1,30 @@
...
+from datahub.dbmaintenance.utils import parse_uuid
+from datahub.dnb_match.models import DnBMatchingResult
+
+
+class Command(CSVBaseCommand):
+    """Command to update DnBMatchingResult."""
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
```
marcofucci: This doesn't seem to do anything?

elcct: Yes, it doesn't. I was initially thinking about supporting some parameters but then figured out this will not be needed now. I should have been more liberal with deleting stuff :) Updated. Thanks!

PR #1502, opened 2019-03-06
Remove sortby values from basic search

marcofucci:
### Description of change

This removes all previously deprecated sortby values from the basic search.

### Checklist

### DIFF
--- a/datahub/search/omis/test/test_views.py
+++ b/datahub/search/omis/test/test_views.py
@@ -654,7 +654,7 @@ class TestGlobalSearch(APITestMixin):
     """Test global search for orders."""
 
     @pytest.mark.parametrize(
-        'term,results',
+        'term,expected_results',
         (
             (  # no filter => return all records
                 '',
@@ -694,7 +694,7 @@ class TestGlobalSearch(APITestMixin):
             ),
         ),
     )
-    def test_search(self, setup_data, term, results):
+    def test_search(self, setup_data, term, expected_results):
         """Test search results."""
         url = reverse('api-v3:search:basic')
 
@@ -702,13 +702,14 @@ class TestGlobalSearch(APITestMixin):
             url,
             data={
                 'term': term,
-                'sortby': 'created_on:asc',
                 'entity': 'order',
             },
         )
 
         assert response.status_code == status.HTTP_200_OK
-        assert len(response.json()['results']) == len(results)
-        assert [
-            item['reference'] for item in response.json()['results']
-        ] == results
+        assert len(response.json()['results']) == len(expected_results)
+        actual_results = sorted(
+            item['reference']
+            for item in response.json()['results']
+        )
+        assert actual_results == expected_results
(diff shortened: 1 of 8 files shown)


datahub/search/omis/test/test_views.py
```diff
@@ -694,21 +694,22 @@ class TestGlobalSearch(APITestMixin):
...
             },
         )
 
         assert response.status_code == status.HTTP_200_OK
-        assert len(response.json()['results']) == len(results)
-        assert [
-            item['reference'] for item in response.json()['results']
-        ] == results
+        assert len(response.json()['results']) == len(expected_results)
+        actual_results = sorted([
```
reupen: Do we need `[` and `]`?

datahub/search/test/test_query_builder.py
```diff
@@ -524,3 +524,105 @@ def test_get_search_by_entity_query(
...
+        search_app.name: search_app,
+    }
+
+    query = get_basic_search_query(
+        search_app.es_model,
+        'test',
+        permission_filters_by_entity={
+            search_app.name: [('name', 'perm')],
+        },
+        ordering=SearchOrdering('name', SortDirection.desc),
```
reupen: Do we need to keep the `ordering` argument in `get_basic_search_query()` itself?

marcofucci: I was about to remove it as nothing is using that. However I thought it could be useful if we need it.
I’m ok to remove it if you guys think it’s ok?

datahub/search/omis/test/test_views.py
marcofucci: I don’t think that using a generator expression is any better in this case, python is still generating a new list from the list passed in or the generator and then sorting it.
A generator would be probably better for memory consumption but we don’t have this problem here. 
Copying a list might also be quicker than creating a new list from a generator (not that we care here anyway).

reupen: I don't care about performance, it just seems weird to add `[` and `]` for no reason.

reupen: You could also use `collections.Counter` which would generate better diffs on failures anyway.

datahub/search/test/test_query_builder.py
reupen: If we need it later on we can add it again. So let's get rid of it.

reupen:
@marcofucci I found the moving of the tests in the same commit a bit confusing, perhaps that could be a separate commit before the removal?

datahub/search/omis/test/test_views.py
marcofucci: I don't really like using `collections.Counter` in tests, it's very confusing every time I see it. I'll use generators if that helps.

marcofucci:
> I found the moving of the tests in the same commit a bit confusing, perhaps that could be a separate commit before the removal?

@reupen I'll tidy up the history when squashing and before merging.

datahub/search/omis/test/test_views.py
reupen: Yes, but when sorted(..) == sorted(..) fails it can give very confusing diffs. That's why I don't use it.

PR #1504, opened 2019-03-06
Move investment project code to a sub directory

marcuspp:
### Description of change
This moves all investment project code into a subdirectory of `project`.
All paths updated.

### Checklist


changelog/investment/add_django_app_subdirectory.internal
```diff
@@ -0,0 +1,2 @@
+The subdirectory `project` has been added to the investment django application
```
marcofucci: ` => ``

datahub/investment/project/apps.py
```diff
@@ -4,11 +4,12 @@
 class InvestmentConfig(AppConfig):
     """Configuration class for this app."""
 
-    name = 'datahub.investment'
+    name = 'datahub.investment.project'
+    label = 'investment'
```
marcofucci: If you can, could you please add a comment saying that this is called `investment` for legacy reasons? (it should really be called `investment_project` instead). The same applies for `evidence`, `report` etc. One problem is that generated tables which are not investment specific (e.g. `evidence_evidencetag`) which is not ideal.

Adding a comment will help future devs understand what happened to these apps.

config/settings/common.py
```diff
@@ -67,9 +67,10 @@
     'datahub.event',
     'datahub.feature_flag.apps.FeatureFlagConfig',
     'datahub.interaction',
-    'datahub.investment',
-    'datahub.investment.evidence',
-    'datahub.investment.proposition',
+    'datahub.investment.project.apps.InvestmentConfig',
```
reupen: I would suggest setting `default_app_config` in `__init__.py` in each package rather than setting the app config here. (E.g.:

https://github.com/uktrade/data-hub-leeloo/blob/9823aad110a4489ada2106a317091867ccf67b39/datahub/core/__init__.py#L1

)

marcuspp: :+1:

marcuspp:
https://github.com/uktrade/data-hub-leeloo/pull/1504#pullrequestreview-211670889

Due to time constraints I'm not going to do the subdirectory change. With this not being as complicated a change as I initially thought this could always be done at a later date.

PR #1510, opened 2019-03-07
Large capital profile endpoints and metadata

marcuspp:
### Description of change
This adds the large capital profile endpoints.

### Checklist

### DIFF
--- a/changelog/investment/metadata.api.rst
+++ b/changelog/investment/metadata.api.rst
@@ -1,0 +1,25 @@
+The following metadata endpoints have been added
+
+``GET /metadata/capital-investment/asset-class-interest/`` returns all possible ``asset_class_interest`` values.
+The values also include a field ``asset-class-interest-sector`` which returns the ``id`` and
+``name`` of the the associated ``asset_class_interest_sector``.
+
+``GET /metadata/capital-investment/required-checks-conducted/`` returns all possible ``investor_profile_requiredchecksconducted`` values.
+
+``GET /metadata/capital-investment/construction-risk/`` returns all possible ``investor_profile_constructionrisk`` values.
+
+``GET /metadata/capital-investment/deal-ticket-size/`` returns all possible ``investor_profile_dealticketsize`` values.
+
+``GET /metadata/capital-investment/desired-deal-role/`` returns all possible ``investor_profile_desireddealrole`` values.
+
+``GET /metadata/capital-investment/equity-percentage/`` returns all possible ``investor_profile_equitypercentage`` values.
+
+``GET /metadata/capital-investment/investor-type/`` returns all possible ``investor_profile_investortype`` values.
+
+``GET /metadata/capital-investment/large-capital-investment-type/`` returns all possible ``investor_profile_largecapitalinvestmenttype`` values.
+
+``GET /metadata/capital-investment/restriction/`` returns all possible ``investor_profile_restriction`` values.
+
+``GET /metadata/capital-investment/return-rate/`` returns all possible ``investor_profile_returnrate`` values.
+
+``GET /metadata/capital-investment/time-horizon/`` returns all possible ``investor_profile_time_horizon`` values.
--- a/datahub/investment/investor_profile/admin.py
+++ b/datahub/investment/investor_profile/admin.py
@@ -1,0 +1,12 @@
+from django.contrib import admin
+
+from datahub.investment.investor_profile import models
+
+
+@admin.register(models.InvestorProfile)
+class InvestorProfileAdmin(admin.ModelAdmin):
+    """Investor profile admin."""
+
+    list_display = ('investor_company', 'profile_type')
+    search_fields = ('investor_company__name', 'id')
+    readonly_fields = ('id', 'created_by', 'modified_by')
--- a/datahub/investment/investor_profile/serializers.py
+++ b/datahub/investment/investor_profile/serializers.py
@@ -1,0 +1,220 @@
+from rest_framework import serializers
+
+import datahub.metadata.models as meta_models
+from datahub.company.models import Company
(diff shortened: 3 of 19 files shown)


datahub/investment/investor_profile/admin.py
```diff
@@ -0,0 +1,11 @@
...
+
+from datahub.investment.investor_profile import models
+
+
+@admin.register(models.InvestorProfile)
+class InvestorProfileAdmin(admin.ModelAdmin):
+    """Investor profile admin."""
+
+    list_display = ('investor_company', 'profile_type')
+    search_fields = ('investor_company', 'pk')
```
reupen: I'm getting an error performing a search in admin:

> django.core.exceptions.FieldError: Related Field got invalid lookup: icontains

reupen:
Just a small thing I spotted (related to the last PR) – the `investor_profile` apps needs a verbose name set because it is showing up in admin with the underscore (it should be set to 'Investor profile' as opposed to 'investor profile' because Django doesn't apply `capfirst` to app verbose names for some reason).

changelog/investment/metadata.api.rst
```diff
@@ -0,0 +1,25 @@
...
+
+``GET /metadata/desired-deal-role/`` returns all possible ``investor_profile_desireddealrole`` values.
+
+``GET /metadata/equity-percentage/`` returns all possible ``investor_profile_equitypercentage`` values.
+
+``GET /metadata/large-capital-investor-type/`` returns all possible ``investor_profile_investortype`` values.
+
+``GET /metadata/large-capital-investment-type/`` returns all possible ``investor_profile_largecapitalinvestmenttype`` values.
+
+``GET /metadata/restriction/`` returns all possible ``investor_profile_restriction`` values.
```
reupen: Given that these are under a shared `/metadata/` prefix, I think some of the names aren't specific enough (this one in particular). I think we could just add e.g. investor to some of them.

datahub/investment/investor_profile/serializers.py
```diff
@@ -0,0 +1,221 @@
...
+        """
+        Validates that the company has not changed and that the company does not already
+        have a large capital investment profile.
+        """
+        if self.instance and self.instance.investor_company_id != value.id:
+            raise serializers.ValidationError(
+                'Investor company can not be updated',
+            )
+
+        profile_id = getattr(self.instance, 'id', None)
```
reupen: Doesn't DRF add a UniqueTogetherValidator automatically that covers this?

reupen: I guess it misbehaves for creations because it doesn't have the profile type...

datahub/investment/investor_profile/validate.py
```diff
@@ -0,0 +1,13 @@
+from datahub.core.validate_utils import DataCombiner
+from datahub.investment.investor_profile.models import InvestorProfile
+from datahub.investment.project.validate import field_incomplete
```
reupen: It would be better to move this function somewhere shared instead of importing it from `project`.

datahub/investment/investor_profile/serializers.py
reupen: `getattr(self.instance, 'id', None)` makes it looks like it's checking if `self.instance` has an `id` attribute when it's actually checking if `self.instance` is not None.

`self.instance.id if self.instance else None` would be better.

marcofucci: or even something like:
```python
qs = value.investor_profiles.filter(profile_type_id=ProfileTypeConstant.large.value.id)
if self.instance:
    qs = qs.exclude(id=self.instance.id)

if qs.exists():
   ...
```

marcuspp:
> @marcuspp looking good, the main thing is the metadata prefix.
> 
> Re. to that, you have: `/v4/large-investor-profile/` etc.
> This works fine because there's an investment-like word (investor) in it but I was thinking that it could make things slightly uglier if you add other endpoints in the future which are less obviously related to investment as you would need to add an investment-like word to avoid future collisions.
> 
> Would you mind having a quick think about if it makes sense to have a prefix for this and all other endpoints that you are going to add so that it groups all together like you did with the django app?
> 
> Moving all the current `investment` endpoints would be too much work I suppose, even though it would be the ideal choice as the current `/v3/investment/...` API have now become confusing (it's not investment but investment projects) and therefore tech debt.
> 
> Maybe another prefix? Or maybe it's better to keep all at first level and add this to the tech debt board to sort out later on?

@marcofucci  I discussed with Rob yesterday about what term could be used to group this, growth and uk opportunities. And we came up with `Capital Investment activities` I could prefix everything with `capital_investment` ? So `/metadata/capital_investment/restrictions` This would then mean if the metadata is reused in any of the other areas (that are still being hashed out) it still would make sense.

How does that sound?

datahub/investment/investor_profile/constants.py
```diff
@@ -3,6 +3,9 @@
 from datahub.core.constants import Constant
 
 
+FEATURE_FLAG_LARGE_CAPITAL_PROFILE = 'large-capital-profile-feature'
```
marcuspp: @marcofucci Can I remove the feature flag? Investor Profiles aren't whitelisted in the merge tool so if a company has a profile a merge will automatically be disallowed. I'll then fix the merge tool at a later date

changelog/investment/metadata.api.rst
marcuspp: I've prefixed all the metadata with `capital-investment`

datahub/investment/investor_profile/serializers.py
marcuspp: I've removed this logic and replaced it with @marcofucci suggestion below.

PR #1515, opened 2019-03-11
Add InteractionDITParticipant many-to-many model

reupen:
### Description of change

This adds a `InteractionDITParticipant` model representing a many-to-many relationship between interactions and advisers.

This is intended to replace the `dit_adviser` and `dit_team` interaction fields in due course.

When an interaction is created or updated via the API or via the admin site, the `dit_adviser` and `dit_team` fields are copied to a `InteractionDITParticipant`.

For now, the reverse relationship `dit_participants` is not exposed via the API or admin site as existing data has not been fully copied to it.

### Checklist


reupen:
@marcofucci There will be no separate team fields in the front end and the existing service provider field will be removed.

We will save the team of each adviser at the time the adviser was added to the interaction so that the data is historically accurate.

datahub/interaction/migrations/0044_add_dit_participants.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 2.1.7 on 2019-03-11 09:41
```
marcuspp: Can the migrations be merged into one file?

datahub/interaction/serializers.py
```diff
@@ -168,6 +170,51 @@ def validate(self, data):
...
+        """
+        Create an interaction.
+
+        Overridden so that dit_adviser and dit_team can be copied to dit_participants.
+
+        TODO: Remove once dit_adviser and dit_team have been fully replaced by dit_participants.
+        """
+        interaction = super().update(instance, validated_data)
+
+        InteractionDITParticipant.objects.update_or_create(
```
marcuspp: What happens when the adviser is what is being updated? Should the old InteractionDITParticipant be removed?

datahub/interaction/migrations/0044_add_dit_participants.py
reupen: Yes, but it won't reduce the number of operations. The commits were split into logical chunks, and so the migrations reflect the changes in each commit.

PR #1519, opened 2019-03-13
Docker compose rework

currycoder:
### Description of change

This reworks docker-compose usage slightly and improves the documentation.
- The key backward incompatible change (for docker users) is moving `docker-compose.env` to a `sample.env` file and requiring users to create a concrete `.env` file from this.  This copies the workflow from the data hub frontend project - increasing consistency between our docker approaches.
- The README has been modified to add an optional end step for setting up a local frontend app under docker-compose.  This delegates instructions to that repository.
- Additionally, the default environment variables for docker-compose have been modified to assume that `mocksso` is used as the SSO provider.

This PR should be reviewed in conjunction with the companion frontend PR: https://github.com/uktrade/data-hub-frontend/pull/1811

#### How to test

Please run through the instructions to bring up a docker-compose project locally.  
Then run through the instructions in the frontend PR (and by extension the mock-sso instructions).  
At the end of this, you should have local copies of leeloo, frontend and mock-sso all working together such that you can navigate the frontend, be correctly authenticated and see expected data.

### Checklist

* [X] Have any relevant search models been updated?
* [X] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?


currycoder:
Thanks for the commit observation.

I think you are right about being more explicit about `.env` file naming.  My original rationale was that by copying the way that frontend does it, we increase consistency between projects and naturally make it easier for new developers.  But it does sound like devs will generally only spin up what they need anyway.  Also, by keeping it as `docker-compose.env` I won't suddenly introduce breakages for any existing docker-compose users.

I'll update and ping you when it's ready for re-review.

currycoder:
@reupen In practice, there is unfortunately still a need for a `.env` file in the project root.  This is because docker compose sources the `COMPOSE_PROJECT_NAME` from this file only.  I've introduced an explicit docker compose project name so that a developer running the full stack can run all containers under the same project name - and therefore share a docker network so that these containers can communicate with each other.

So the two options I can see are:
1. Use both a `docker-compose.env` and a `.env` file.  This introduces an extra step in the readme.
2. Just use a `.env` file and comment the top of it to explain that it is intended for use with docker-compose.

Any thoughts?

As a side note, it is perhaps a bit unfortunate that there's such a large level of duplication between the docker-compose.env file and the `config/settings/.env` file - I'm guessing there's a vector here for them to go quite out of sync with eachother.

reupen:
That's unfortunate about `COMPOSE_PROJECT_NAME`.

Is that something we need as leeloo is exposed on localhost? Does it create any conflicts with `redis` etc.? I think leeloo will always need to be exposed on localhost so that you can connect to the API manually, use the admin site etc.

Another possibility might be to move the Docker bits to a sub-directory?

currycoder:
My understanding is that if you want to have leeloo/frontend/mocksso all under docker, we need them to be on the same docker network.  It's true that leeloo is exposed on localhost, but the frontend container can't access the host machine's network without some explicit docker network changes (which I feel are out of scope for what I was trying to achieve here).

You're right about redis - currently, frontend and leeloo docker containers would share a redis container.  I don't feel like this is a problem though as my understanding is that leeloo's redis prefix is different to frontend's - would be worth your insight here?

As for moving the docker bits to be under their own directory, I agree that would be good for explicit context.  There's a different problem with this; when running any docker-compose commands, the user would have to explicitly specify the location of the `docker-compose.yml` file e.g. `docker-compose -f docker/docker-compose.yml up`, `docker-compose -f docker/docker-compose.yml run leeloo manage.py migrate` etc etc.  It's just not as easy to use IMO.

reupen:
It does all seem a bit unfortunate... Perhaps the comment will be good enough for now?

Also, it's probably worth noting that there are some docker-compose aliases in the Makefile (although I'm not sure if anyone is using them).

currycoder:
I've put a comment in for now.

Good point about the Makefile - I can't see it used anywhere in the codebase.  All of those commands will still work, so long as there is a valid `.env` file.

PR #1520, opened 2019-03-15
Add dit_participants to responses in interaction API

reupen:
### Description of change

This contains some preparatory work for making dit_participants writable in interaction endpoints.

This:

- adds `dit_participants` to `/v3/interaction` and `/v3/interaction/<id>` as a read-only field
- makes `dit_adviser` and `dit_team` read-only in interaction admin to avoid complexity there while `dit_participants` co-exists with `dit_adviser` and `dit_team`
- deprecates `dit_adviser`-related sorting options in `/v3/interaction` (as these are being removed without a replacement)

### Checklist

### DIFF
--- a/changelog/interaction/adviser-sorting.removal.rst
+++ b/changelog/interaction/adviser-sorting.removal.rst
@@ -1,0 +1,2 @@
+``GET /v3/interaction``: The ``dit_adviser__first_name`` and ``dit_adviser__last_name`` values for the
+``sortby`` query parameter are deprecated and will be removed on or after 28 March 2019.
--- a/datahub/interaction/serializers.py
+++ b/datahub/interaction/serializers.py
@@ -29,6 +29,21 @@ from datahub.investment.project.serializers import NestedInvestmentProjectField
 from datahub.metadata.models import Service, Team
 
 
+class InteractionDITParticipantSerializer(serializers.ModelSerializer):
+    """
+    Interaction DIT participant serialiser.
+
+    Used within InteractionSerializer.
+    """
+
+    adviser = NestedAdviserField()
+    team = NestedRelatedField(Team)
+
+    class Meta:
+        model = InteractionDITParticipant
+        fields = ('adviser', 'team')
+
+
 class InteractionSerializer(serializers.ModelSerializer):
     """V3 interaction serialiser."""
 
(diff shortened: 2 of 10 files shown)


datahub/interaction/serializers.py
```diff
@@ -29,6 +29,21 @@
...
 
 
+class InteractionDITParticipantSerializer(serializers.ModelSerializer):
+    """
+    Interaction DIT participant serialiser.
+
+    Used within InteractionSerializer.
+    """
+
+    adviser = NestedAdviserField()
```
currycoder: This will help me learn a bit.  Why is a specialised nested field used for adviser instead of the generic `NestedRelatedField` as is used for team?

changelog/interaction/adviser-sorting.removal.rst
```diff
@@ -0,0 +1,2 @@
+``GET /v3/interaction``: The ``dit_adviser__first_name`` and ``dit_adviser__last_name`` values for the
```
currycoder: Another question..  What mechanism is there to tell frontend devs about these deprecations?  Or do they already know this is coming?

datahub/interaction/serializers.py
reupen: The standard NestedRelatedField includes the fields `id` and `name` by default while in adviser fields we also include `first_name` and `last_name`.

We did that manually each time originally, but there are a lot of adviser fields about so that got a bit repetitive and also inconsistent, so we defined NestedAdviserField to cut down on repetition and make sure adviser fields are consistent throughout the API.

changelog/interaction/adviser-sorting.removal.rst
reupen: Generally there will be a front-end dev working on the same feature in `data-hub-frontend` so they will probably be aware of API changes (for example, @ian-leggett already removed these sorting options in the front end). 

But these changes could affect other services and teams (e.g. Market Access and data science) so this helps make sure that they are aware of them as well.

datahub/interaction/serializers.py
currycoder: :+1:

changelog/interaction/adviser-sorting.removal.rst
currycoder: Thanks :+1:

PR #1523, opened 2019-03-15
Make dit_participants writable in interaction API

reupen:
### Description of change

This:

- updates the interaction API so that `dit_participants` is writable (with the first entry copied to `dit_adviser` and `dit_team` on the model)
- adds DIT participants to interactions in the test data (fixtures)

There is a little bit of added complexity because of the backwards compatibility (so it will be simpler once all that is removed). I also encountered a problem with `allow_empty=False` on DRF list serialisers which I have reported to the DRF project and worked around for now.

### Checklist


datahub/interaction/serializers.py
```diff
@@ -108,11 +185,30 @@ class InteractionSerializer(serializers.ModelSerializer):
 
     def validate(self, data):
         """
-        Removes the semi-virtual field is_event from the data.
+        Validates and cleans the data,
```
elcct: The comma makes it look like something is missing.

PR #1524, opened 2019-03-15
Add Large investor profile to search pt 1

marcuspp:
### Description of change
First part of the large investor profile search. I have tried to split everything up into easily reviewable chunks. So this is just the initial search app and the signals to update and delete.
Was unsure what to put in the changelog for this PR as this can't really be used till part 2 is released.

### Checklist

### DIFF
--- a/datahub/search/fields.py
+++ b/datahub/search/fields.py
@@ -83,21 +83,45 @@ def id_name_partial_field(field):
     )
 
 
+def company_field():
+    """Company field with id, name, trading_names and trigrams."""
+    return Object(
+        properties={
+            'id': Keyword(),
+            'name': Text(
+                fields={
+                    'trigram': TrigramText(),
+                },
+            ),
+            'trading_names': Text(
+                fields={
+                    'trigram': TrigramText(),
+                },
+            ),
+        },
+    )
+
+
+def country_field():
+    """Country field with id, name and trigram."""
+    return Object(
+        properties={
+            'id': Keyword(),
+            'name': Text(
+                fields={
+                    'trigram': TrigramText(),
+                },
+            ),
+        },
+    )
+
+
 def address_field(index_country=True):
     """Address field as nested object."""
     if index_country:
-        country_field = Object(
-            properties={
-                'id': Keyword(),
-                'name': Text(
-                    fields={
-                        'trigram': TrigramText(),
-                    },
-                ),
-            },
-        )
+        nested_country_field = country_field()
     else:
-        country_field = Object(
+        nested_country_field = Object(
             properties={
                 'id': Keyword(index=False),
                 'name': Text(index=False),
@@ -115,13 +139,13 @@ def address_field(index_country=True):
                     'trigram': TrigramText(),
                 },
             ),
-            'country': country_field,
+            'country': nested_country_field,
         },
     )
 
 
-def company_field(field):
-    """Company field."""
+def company_field_with_copy_to_name_trigram(field):
+    """Company field with copy to, deprecated in favour of company_field"""
     return Object(
         properties={
             'id': Keyword(),
--- a/datahub/search/large_investor_profile/apps.py
+++ b/datahub/search/large_investor_profile/apps.py
@@ -1,0 +1,36 @@
+from datahub.investment.investor_profile.constants import ProfileType
+from datahub.investment.investor_profile.models import InvestorProfile as DBInvestorProfile
+from datahub.investment.investor_profile.permissions import InvestorProfilePermission
+from datahub.search.apps import SearchApp
+from datahub.search.large_investor_profile.models import (
+    DOC_TYPE as LARGE_INVESTOR_PROFILE_DOC_TYPE,
+    LargeInvestorProfile,
+)
+
+
+class LargeInvestorProfileSearchApp(SearchApp):
+    """SearchApp for investor profile."""
+
+    name = LARGE_INVESTOR_PROFILE_DOC_TYPE
+    es_model = LargeInvestorProfile
+    view_permissions = (f'investor_profile.{InvestorProfilePermission.view_investor_profile}',)
+    exclude_from_global_search = True
+    queryset = DBInvestorProfile.objects.filter(
+        profile_type_id=ProfileType.large.value.id,
+    ).select_related(
+        'investor_company',
+        'investor_type',
+        'required_checks_conducted',
+        'minimum_return_rate',
+        'minimum_equity_percentage',
+    ).prefetch_related(
+        'deal_ticket_sizes',
+        'asset_classes_of_interest',
+        'investment_types',
+        'time_horizons',
+        'restrictions',
+        'construction_risks',
+        'desired_deal_roles',
+        'uk_region_locations',
+        'other_countries_being_considered',
+    )
--- a/datahub/search/large_investor_profile/models.py
+++ b/datahub/search/large_investor_profile/models.py
@@ -1,0 +1,95 @@
+from elasticsearch_dsl import Date, Keyword, Long
+
+from datahub.search import dict_utils
+from datahub.search import fields
+from datahub.search.models import BaseESModel
+
+
+DOC_TYPE = 'large-investor-profile'
+
+
+def _get_adviser_list(col):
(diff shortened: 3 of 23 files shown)


datahub/search/large_investor_profile/apps.py
```diff
@@ -0,0 +1,21 @@
...
+class LargeInvestorProfileSearchApp(SearchApp):
+    """SearchApp for investor profile."""
+
+    name = LARGE_INVESTOR_PROFILE_DOC_TYPE
+    es_model = LargeInvestorProfile
+    view_permissions = (f'investor_profile.{InvestorProfilePermission.view_investor_profile}',)
+    queryset = DBInvestorProfile.objects.filter(
+        profile_type_id=ProfileType.large.value.id,
+    ).select_related(
+        'investor_company',
```
reupen: This should be expanded to include other foreign keys (and also a `.prefetch_related()` call added for to-many fields).

(It makes a big difference to the synchronisation speed.)

datahub/search/tasks.py
```diff
@@ -69,6 +69,7 @@ def sync_related_objects_task(
     related_model_label,
     related_obj_pk,
     related_obj_field_name,
+    related_obj_filter,
```
reupen: This looks like a potential problem during deployment as the old app instances would be queuing tasks with the old signature which might get picked up by a Celery worker using the new signature.

I think making `related_obj_filter` an optional keyword argument would do the job?

There would probably also be a tiny window where the old workers could pick up a new task as well, so ideally we would update the task signature, deploy that everywhere and then start using the new optional keyword argument. But if we make `sync_related_objects_async` only provide the argument if it is set, that should avoid us having to do that in this case (since it's unlikely anything would set that argument at that time).

datahub/search/large_investor_profile/models.py
```diff
@@ -0,0 +1,100 @@
...
+    return [dict_utils.id_name_dict(c) for c in col.all()]
+
+
+class LargeInvestorProfile(BaseESModel):
+    """Elasticsearch representation of LargeInvestorProfile."""
+
+    id = Keyword()
+
+    investor_company = fields.company_field('investor_company')
+    country_of_origin = fields.id_name_partial_field('country_of_origin')
```
reupen: Sorry as it's not clearly documented, but we stopped using `copy_to` in favour of multi-fields. `id_name_partial_field` and `company_field` both use `copy_to` so we should define new fields that use multi-fields instead of using those.

`datahub.search.fields.address_field` is an example of one that uses multi-fields (see the `fields=` bit).

datahub/search/large_investor_profile/models.py
```diff
@@ -0,0 +1,100 @@
...
+    created_by = fields.contact_or_adviser_field(
+        'created_by', include_dit_team=True,
+    )
+
+    investor_type = fields.id_name_field()
+    global_assets_under_management = Long()
+    investable_capital = Long()
+    required_checks_conducted = fields.id_name_field()
+
+    deal_ticket_sizes = fields.id_name_field()
```
reupen: It looks like many of the name fields are not/will not be used in searches, in which case it would be better to use `id_unindexed_name_field` instead so that Elasticsearch does not analyse then unnecessarily.

datahub/search/large_investor_profile/models.py
```diff
@@ -0,0 +1,100 @@
...
+    MAPPINGS = {
+        **_MAIN_FIELD_MAPPINGS,
+        **_DETAIL_FIELD_MAPPINGS,
+        **_REQUIREMENT_FIELD_MAPPINGS,
+        **_LOCATION_FIELD_MAPPINGS,
+    }
+
+    SEARCH_FIELDS = (
+        'id',
+        'investor_company.name',
```
reupen: `investor_company.name` is a keyword field when it should probably be a text field (this is a mistake in some of the other search apps).

datahub/search/large_investor_profile/signals.py
```diff
@@ -0,0 +1,47 @@
...
+from datahub.search.large_investor_profile.models import (
+    LargeInvestorProfile as ESLargeInvestorProfile,
+)
+from datahub.search.signals import SignalReceiver
+from datahub.search.sync_object import sync_object_async, sync_related_objects_async
+
+
+def investor_profile_sync_es(instance):
+    """Sync investor profile to Elasticsearch."""
+    if instance.profile_type_id == ProfileType.large.value.id:
```
reupen: `ProfileType.large.value.id` is a string but `instance.profile_type_id` could be a UUID I'd have thought (particularly on updates). Do we need to convert them both to the same type before doing the comparison?

datahub/search/tasks.py
marcuspp: I think I have sorted this. Let me know what you think.

datahub/search/large_investor_profile/models.py
marcuspp: Am I correct in think these `SEARCH_FIELDS` are only used by the homepage search. Due to there being no requirement for investor profiles to be searched via the homepage. I've removed the `SEARCH_FIELDS` var.

marcuspp: Updated `investor_company` field.

reupen: It's both for global search (via the home page) and the `original_query` param in `/v3/search/<app name>`.

However, if it's meant to be excluded from global search you should also set `exclude_from_global_search = True` on the search app similar to this:

https://github.com/uktrade/data-hub-leeloo/blob/2ab8f57d651dfcde45625c7ff1ae3a5bcf69cff5/datahub/search/companieshousecompany/apps.py#L6-L15

This will completely exclude the app's index from global searches (including in the aggregations).

datahub/search/fields.py
```diff
@@ -83,21 +83,45 @@ def id_name_partial_field(field):
...
+            'trading_names': Text(
+                fields={
+                    'trigram': TrigramText(),
+                },
+            ),
+        },
+    )
+
+
+def country_field():
```
reupen: Just a small thing but this looks like a new version of `id_name_partial_field` so would suggest giving it a more generic name (and probably renaming `id_name_partial_field`).

datahub/search/large_investor_profile/signals.py
```diff
@@ -0,0 +1,47 @@
...
+from datahub.search.large_investor_profile.models import (
+    LargeInvestorProfile as ESLargeInvestorProfile,
+)
+from datahub.search.signals import SignalReceiver
+from datahub.search.sync_object import sync_object_async, sync_related_objects_async
+
+
+def investor_profile_sync_es(instance):
+    """Sync investor profile to Elasticsearch."""
+    if str(instance.profile_type_id) == str(ProfileType.large.value.id):
```
reupen: Small thing but `ProfileType.large.value.id` should already be a string.

datahub/search/large_investor_profile/test/test_models.py
```diff
@@ -0,0 +1,52 @@
...
+            'minimum_return_rate',
+            'modified_on',
+            'notes_on_locations',
+            'other_countries_being_considered',
+            'required_checks_conducted',
+            'restrictions',
+            'time_horizons',
+            'uk_region_locations',
+        }
+        assert set(result.keys()) == keys
```
reupen: The search apps are a bit inconsistent in this respect, but it would be good to compare the the whole object to give us some confidence that the mappings are all working correctly (the `interaction` and `omis` search apps do this).

datahub/search/test/test_tasks.py
```diff
@@ -78,6 +78,7 @@ def test_sync_related_objects_task_syncs(monkeypatch):
             SimpleModel._meta.label,
             str(simpleton.pk),
             'relatedmodel_set',
+            None,
```
reupen: Should this be using `kwargs` for the new argument? We probably want to add a test covering use of the filters as well.

datahub/search/large_investor_profile/test/test_models.py
marcuspp: Because I need this merged so part 2 can be merged I'm going to leave the test improvements for now. But I have a ticket to add the csv download so will improve the tests in that PR.

PR #1527, opened 2019-03-18
Add dit_participants to interaction search model

reupen:
### Description of change

This adds `dit_participants` to the interaction search model.

Existing search logic continues to use `dit_adviser` and `dit_team` as we first need to release this and populate the new fields with data in the Elasticsearch database.

I took the opportunity to use InnerDoc and multi-fields. We already started migrating to multi-fields, and InnerDoc is the main suggested way of using Object in the elasticsearch-dsl docs.  (The benefits improve readability, ability to use inheritance etc.)

### Checklist

### DIFF
--- a/datahub/search/inner_docs.py
+++ b/datahub/search/inner_docs.py
@@ -1,0 +1,27 @@
+from elasticsearch_dsl import InnerDoc, Keyword, Text
+
+from datahub.search.fields import TrigramText
+
+
+class Person(InnerDoc):
+    """Inner doc for a person (e.g. a contact or an adviser)."""
+
+    id = Keyword()
+    first_name = Text(index=False)
+    last_name = Text(index=False)
+    name = Text(
+        fields={
+            'trigram': TrigramText(),
+        },
+    )
+
+
+class IDNameTrigram(InnerDoc):
+    """Inner doc for a named object, with a trigram sub-field on name."""
+
+    id = Keyword()
+    name = Text(
+        fields={
+            'trigram': TrigramText(),
+        },
+    )
--- a/datahub/search/interaction/models.py
+++ b/datahub/search/interaction/models.py
@@ -1,9 +1,10 @@
 from operator import attrgetter
 
-from elasticsearch_dsl import Boolean, Date, Double, Keyword, Object, Text
+from elasticsearch_dsl import Boolean, Date, Double, InnerDoc, Keyword, Object, Text
 
 from datahub.search import dict_utils, fields
 from datahub.search.fields import TrigramText
+from datahub.search.inner_docs import IDNameTrigram, Person
 from datahub.search.models import BaseESModel
 
 
@@ -25,6 +26,21 @@ def _contact_field():
     )
 
 
+def _dit_participant_list(dit_participant_manager):
+    return [
+        {
+            'adviser': dict_utils.contact_or_adviser_dict(dit_participant.adviser),
+            'team': dict_utils.id_name_dict(dit_participant.team),
+        }
+        for dit_participant in dit_participant_manager.all()
+    ]
+
+
+class _DITParticipant(InnerDoc):
+    adviser = Object(Person)
+    team = Object(IDNameTrigram)
+
+
 class Interaction(BaseESModel):
     """Elasticsearch representation of Interaction model."""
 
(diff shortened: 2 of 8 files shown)


datahub/search/interaction/models.py
```diff
@@ -25,6 +26,21 @@ def _contact_field():
...
+    return [
+        {
+            'adviser': dict_utils.contact_or_adviser_dict(dit_participant.adviser),
+            'team': dict_utils.id_name_dict(dit_participant.team),
+        }
+        for dit_participant in dit_participant_manager.all()
+    ]
+
+
+class _DITParticipant(InnerDoc):
```
marcuspp: Why the underscore? Could this not go in `inner_docs.py`

datahub/search/interaction/tests/test_signals.py
```diff
@@ -64,6 +65,31 @@ def test_deleted_interaction_deleted_from_es(setup_es):
         ) is None
 
 
+def test_interaction_synced_when_dit_participant_added(setup_es):
+    """Test that interactions are synced to ES if their DIT participants change."""
+    interaction = CompanyInteractionFactory(dit_participants=[])
+    setup_es.indices.refresh()
+
+    doc = setup_es.get(
+        index=InteractionSearchApp.es_model.get_write_index(),
```
marcuspp: If this a get should this be `get_read_index`?

datahub/search/inner_docs.py
```diff
@@ -0,0 +1,27 @@
+from elasticsearch_dsl import InnerDoc, Keyword, Text
+
+from datahub.search.fields import TrigramText
+
+
+class Person(InnerDoc):
```
marcuspp: Is the class based inner docs / field declaration the preferred way?

marcuspp:
A few questions but I have approved anyway. :+1:

datahub/search/interaction/tests/test_signals.py
```diff
@@ -64,6 +65,31 @@ def test_deleted_interaction_deleted_from_es(setup_es):
...
+    setup_es.indices.refresh()
+
+    updated_doc = setup_es.get(
+        index=InteractionSearchApp.es_model.get_write_index(),
+        doc_type=InteractionSearchApp.name,
+        id=interaction.pk,
+    )
+    actual_dit_participants = updated_doc['_source']['dit_participants']
+    assert len(actual_dit_participants) == 1
+    assert actual_dit_participants[0]['adviser']['id'] == str(dit_participant.adviser.pk)
```
currycoder: Might as well check the team as well here?

datahub/search/inner_docs.py
reupen: I would suggest so, and it's the main documented way: https://elasticsearch-dsl.readthedocs.io/en/latest/persistence.html

However, happy to stick with the old way if we want to hold off on switching.

datahub/search/interaction/models.py
reupen: It's internal, it makes sense for other modules to import the model but not `_DITParticipant`. 

I didn't put it in `inner_docs` as it's specific to interactions while the other ones could be used by other search apps.

reupen:
> Is there a manual mechanism to re-sync all Interactions to cover this historically?

Yes, the `./manage migrate_es` command. At the moment it has to be manually triggered, but it will work out which models have changed, create new indexes and resync things. Have a look at https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Elasticsearch%20migrations.md for more details.

datahub/search/inner_docs.py
marcuspp: I don't mind switching its just I think we need to clear that is what we are going to do in future. It can be difficult to know what the preferred way to do things are generally. Looking at other search apps no longer helps with this.

datahub/search/interaction/models.py
marcuspp: TBH i don't think the underscore is necessary. Personally I think it should still go in the inner_docs file as it makes more sense it being in there than not. But happy if you want to leave things as they are.

datahub/search/inner_docs.py
reupen: Yes, I agree. There are items on the tech board already, but a good thing we could do now would be to mark things that shouldn't be used in new code as deprecated.

datahub/search/interaction/models.py
reupen: I'll leave it as it is for now, but if this module grows in size we could add an `inner_docs` under `datahub.search.interaction`.

PR #1543, opened 2019-03-20
Add Large investor profile to search pt 2

marcuspp:
### Description of change
This adds the large investor profile search views. It allows most fields to be filtered.

### Checklist

### DIFF
--- a/datahub/search/fields.py
+++ b/datahub/search/fields.py
@@ -111,6 +111,7 @@ def company_field():
             'name': Text(
                 fields={
                     'trigram': TrigramText(),
+                    'keyword': NormalizedKeyword(),
                 },
             ),
             'trading_names': Text(
--- a/datahub/search/large_investor_profile/serializers.py
+++ b/datahub/search/large_investor_profile/serializers.py
@@ -1,0 +1,106 @@
+from rest_framework.serializers import CharField, IntegerField
+
+from datahub.core.serializers import RelaxedDateTimeField
+from datahub.search.serializers import (
+    EntitySearchQuerySerializer,
+    SingleOrListField,
+    StringUUIDField,
+)
+
+
+class SearchLargeInvestorProfileQuerySerializer(EntitySearchQuerySerializer):
+    """Serializer used to validate investor profile search POST bodies."""
+
+    id = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+
+    # Main filters
+    asset_classes_of_interest = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    country_of_origin = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    investor_company = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    investor_company_name = CharField(required=False)
+    created_by = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+
+    # Detail filters
+    investor_type = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    required_checks_conducted = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    global_assets_under_management_start = IntegerField(required=False)
+    global_assets_under_management_end = IntegerField(required=False)
+    investable_capital_start = IntegerField(required=False)
+    investable_capital_end = IntegerField(required=False)
+
+    # Requirement filters
+    deal_ticket_size = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    investment_type = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    minimum_return_rate = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
+    )
+    time_horizon = SingleOrListField(
+        child=StringUUIDField(),
+        required=False,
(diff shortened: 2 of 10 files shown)


datahub/search/large_investor_profile/serializers.py
```diff
@@ -0,0 +1,55 @@
...
+    StringUUIDField,
+)
+
+
+class SearchLargeInvestorProfileQuerySerializer(EntitySearchQuerySerializer):
+    """Serialiser used to validate investor profile search POST bodies."""
+
+    id = SingleOrListField(child=StringUUIDField(), required=False)
+
+    """Main filters"""
```
reupen: These headings look like comments, so I would make them comments instead of triple-quoted strings.

datahub/search/large_investor_profile/views.py
```diff
@@ -0,0 +1,94 @@
...
+        'uk_region_locations': 'uk_region_locations.id',
+        'other_countries_being_considered': 'other_countries_being_considered.id',
+        'created_by': 'created_by.id',
+    }
+
+    COMPOSITE_FILTERS = {
+        'investor_company_name': [
+            'investor_company.name',
+            'investor_company.name.trigram',
+            'investor_company.trading_names',  # to find 2-letter words
```
reupen: This comment should apply to `investor_company.name` as well.

marcuspp:
> Do we have tests for the view (including the filters)?

@reupen Lots of the tests that are already in place cover most things hence why I didn't add any. Will take a second look though and add some tests in.

marcofucci:
@marcuspp do we need so many filters? We are trying to avoid adding more filters than we need (and index fields) unless absolutely necessary as it adds complexity, makes refactoring harder and therefore it's more expensive for DIT (we experienced it directly).
I remember you saying that there won't be many profiles if I'm not wrong?

datahub/search/large_investor_profile/serializers.py
```diff
@@ -0,0 +1,55 @@
...
+    country_of_origin = SingleOrListField(child=StringUUIDField(), required=False)
+    investor_company = SingleOrListField(child=StringUUIDField(), required=False)
+    investor_company_name = CharField(required=False)
+    created_by = SingleOrListField(child=StringUUIDField(), required=False)
+
+    # Detail filters
+    investor_type = SingleOrListField(child=StringUUIDField(), required=False)
+    required_checks_conducted = SingleOrListField(child=StringUUIDField(), required=False)
+    global_assets_under_management_before = IntegerField(required=False)
+    global_assets_under_management_after = IntegerField(required=False)
```
marcofucci: I'm not convinced of using before and after with numbers. Should we refactor these into more generic names?

datahub/search/large_investor_profile/serializers.py
```diff
@@ -0,0 +1,55 @@
...
+    investable_capital_before = IntegerField(required=False)
+    investable_capital_after = IntegerField(required=False)
+
+    # Requirement filters
+    deal_ticket_sizes = SingleOrListField(child=StringUUIDField(), required=False)
```
marcofucci: I understand these follow the model fields but it looks a bit weird that they are in plural whilst all the other filters are in singular. Should we use singular for these ones as well?

datahub/search/large_investor_profile/views.py
```diff
@@ -0,0 +1,94 @@
...
+        'created_on_before',
+        'created_on_after',
+    )
+
+    FILTER_FIELDS = (
+        'id',
+        *_MAIN_FILTERS,
+        *_DETAIL_FILTERS,
+        *_REQUIREMENT_FILTERS,
+        *_LOCATION_FIELDS,
```
elcct: `*_FILTERS` and `*_FIELDS` seem a bit inconsistent. Why this choice and not just `*_FILTERS`?

datahub/search/large_investor_profile/serializers.py
marcuspp: I've gone with `start` and `end` but open to any other suggestions. We could just support `lte` `gte` and `lt` and `gt` ? What do you think?

datahub/search/large_investor_profile/serializers.py
marcuspp: Wasn't 100% sure on this because it makes implementation a bit harder to follow but have updated.

datahub/search/large_investor_profile/views.py
marcuspp: :+1: Good spot updated to `*_FILTERS`

datahub/search/fields.py
```diff
@@ -92,11 +92,13 @@ def company_field():
                 fields={
                     'trigram': TrigramText(),
                 },
+                fielddata=True,
```
marcuspp: @marcofucci @reupen @elcct Not sure if this correct. The other option was to change it to Keyword as I need to be able to sort by the value and when its a Text field it can't.

marcofucci: I think it would be better to add another field called something like `keyword` of type `keyword`. This is also what ES seems to suggest: https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html

reupen: Yes, what @marcofucci said. You can use `es_sort_by_remappings` on the view to remap the `sortby` value to the keyword sub-field.

marcuspp: 👍 Thanks I have updated.

PR #1550, opened 2019-03-21
Add sub-fields to name in id_name_partial_field

reupen:
### Description of change

This is part of migrating away from using `copy_to`. It's also so that we can migrate `name` to being a text field instead of a keyword field.

### Checklist

### DIFF
--- a/datahub/search/fields.py
+++ b/datahub/search/fields.py
@@ -73,11 +73,31 @@ def id_unindexed_name_field():
 
 
 def id_name_partial_field(field):
-    """Object field with id and name sub-fields, and with partial matching on name."""
+    """
+    Object field with id and name sub-fields, and with partial matching on name.
+
+    The `name` field is being migrated away from using `copy_to` to being a multi-field.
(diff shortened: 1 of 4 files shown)


datahub/search/fields.py
```diff
@@ -73,11 +73,26 @@ def id_unindexed_name_field():
...
-    """Object field with id and name sub-fields, and with partial matching on name."""
+    """
+    Object field with id and name sub-fields, and with partial matching on name.
+
+    The `name` field is being migrated away from using `copy_to` to being a multi-field.
+    `name_trigram` and `name.trigram` are both defined while the switch takes place.
+
+    Additionally, the `name` field should have had a data type of text, but it was mistakenly made
+    a keyword field. Hence, a `keyword` sub-field has also been added so type of `name` can be
+    changed to text once sorting operations have been migrated to using the `keyword` sub-field.
```
elcct: Comment suggests as if there is a pending action to be done in the future. Maybe we should add `TODO:` here so it could be easily found?

PR #1551, opened 2019-03-22
 Update interaction search CSV export for multiple DIT participants 

reupen:
### Description of change

This updates the interaction search CSV export to incorporate multiple DIT participants.

The previous Adviser and Service provider columns have been merged into a single Advisers column. This column contains the names of all the advisers for each interaction separated by commas. The team of each adviser is in brackets after each name.

### Checklist

### DIFF
--- a/datahub/search/interaction/views.py
+++ b/datahub/search/interaction/views.py
@@ -1,4 +1,5 @@
 from datahub.core.query_utils import (
+    get_bracketed_concat_expression,
     get_choices_as_case_expression,
     get_front_end_url_expression,
     get_full_name_expression,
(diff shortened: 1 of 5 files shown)


datahub/search/interaction/views.py
```diff
@@ -92,7 +93,14 @@ class SearchInteractionExportAPIView(SearchInteractionAPIViewMixin, SearchExport
                 bracketed_field_name='job_title',
             ),
         ),
-        dit_adviser_name=get_full_name_expression('dit_adviser'),
+        adviser_names=get_string_agg_subquery(
```
currycoder: I think what I'm about to say goes out of scope for this particular bit of work..

I do feel like there's a lot of injected levels of engineering in here to fundamentally get away from writing any raw SQL.  Django certainly offers the ability to modify querysets with `annotate` and powerful tools within that to provide ORM mappings around things like subqueries etc.  I just don't feel like this is very readable in comparison to a more raw SQL query.  

I realise that there are other benefits offered by using django's ORM tools (avoiding SQL injection, being DB agnostic etc) but it does take me a long time to parse this and fully understand what's going on.  If there's a red line to do with never writing raw SQL that would be good for me to know, but I can't help but feeling that it would be more readable and maintainable here than bending the django wrapper functionality to your will.

reupen: In this case, I would say one of the main reasons for not using raw SQL is reusability. I'm not sure how you would reuse expressions like the one to build full names without using custom SQL functions, views and/or SQL fragments. (You could also make use of CTEs, but they are a optimisation fence (=performance problem) in PostreSQL before version 12.)

(For reference, this is the generated query and it's pretty complicated: https://gist.github.com/reupen/0a09aee32e16755f64cf99b893c9b585.)

Nonetheless, there is some raw SQL here:

https://github.com/uktrade/data-hub-leeloo/blob/9d15331aa5d889534701d9678f497c743a537c31/datahub/company/management/commands/sync_ch.py#L30

However, that was a very specific use case (basically, making that job as fast as possible because it was very slow before).

currycoder: Fair enough - that query really highlights the reasoning for using ORM tools, particularly `company_sector_name` etc.

PR #1556, opened 2019-03-22
Remove Interaction.contact from state

reupen:
### Description of change

This removes the deprecated `contact` field from the `Interaction` model state.

The column will be removed from the database in a following release.

~This is currently blocked as the data science platform is still using the column corresponding to this field.~

### Checklist


datahub/interaction/admin.py
```diff
@@ -125,17 +124,3 @@ def get_contact_names(self, obj):
         )
 
     get_contact_names.short_description = 'contacts'
-
```
currycoder: Can we drop `contact` from the exclude tuple on this class too?

reupen: Yep! :+1:

PR #1563, opened 2019-03-26
Add FDI SIC Grouping and GVA Multiplier

marcuspp:
### Description of change
This adds SIC Groupings to datahub and maps them to Datahub parent sectors. 
This also adds a GVA Multiplier db table and populates it. This will be used to calculate GVA. 

I decided to add a FK to Investment Projects straight to the GVAMultiplier as I think this makes everything a lot simpler. It moves what GVA multiplier to use to the saving action rather than the viewing action.

Not sure what tests I can add as this PR is literally the DB tables but open to suggestions.

### Still to do
 - Add an admin page to manage the GVA Multiplier
 - A script to populate the GVA Multiplier for all existing projects
 - Logic to populate the GVA Multiplier for new projects.
 

### Checklist


changelog/investment/fdi_sic_grouping.db.rst
```diff
@@ -0,0 +1,36 @@
+The database table ``investment_fdisicgrouping`` has been added with the following columns:
```
elcct: The table also has `disabled_on` field.

datahub/investment/project/models.py
```diff
@@ -630,6 +635,51 @@ class InvestmentProjectCode(models.Model):
...
+
+
+class GVAMultiplier(models.Model):
+    """
+    Gross Value Add Multiplier.
+
+    To calculate the GVA of an investment project a constant (multiplier) is multiplied
+    by the foreign equity investment value.
+
+    The multiplier has been derived from historical investment project data and could change year
```
elcct: Is this a calendar year?

changelog/investment/fdi_sic_grouping.db.rst
reupen: There is a lot of bold and italics in this when it is rendered.

datahub/investment/project/migrations/0059_fdi_sicgrouping.yaml
```diff
@@ -0,0 +1,56 @@
...
+- model: investment.fdisicgrouping
+  pk: a7d3c320-6e15-44f6-928a-4d776717b386
+  fields: {disabled_on: null, name: 'Petroleum, chemicals, pharmaceuticals, rubber,
+      plastic products'}
+- model: investment.fdisicgrouping
+  pk: f3d8ced6-df75-45da-8713-4bdcfd2e02a7
+  fields: {disabled_on: null, name: 'Professional, scientific & technical services'}
+- model: investment.fdisicgrouping
+  pk: c7ba3108-5fb3-486c-8d29-e71b1c0e29cd
+  fields: {disabled_on: null, name: 'Retails & wholesale trade, repair of motor vehicles
```
reupen: Is this one truncated?

datahub/investment/project/models.py
reupen: Same question from me – is it calendar or financial year?

Also, what date will it relate to (land date or something else)?

datahub/investment/project/models.py
```diff
@@ -630,6 +635,51 @@ class InvestmentProjectCode(models.Model):
     project = models.OneToOneField(InvestmentProject, on_delete=models.CASCADE)
 
 
+class InvestmentSector(models.Model):
```
reupen: Would be good to add an `__str__` implementation.

datahub/investment/project/models.py
```diff
@@ -630,6 +635,51 @@ class InvestmentProjectCode(models.Model):
...
+        related_name='investment_sectors',
+        on_delete=models.PROTECT,
+    )
+
+
+class FDISICGrouping(BaseConstantModel):
+    """FDI Standard Industrial Classification Grouping used for GVA."""
+
+
+class GVAMultiplier(models.Model):
```
reupen: Would be good to add an `__str__` implementation.

datahub/investment/project/models.py
```diff
@@ -630,6 +635,51 @@ class InvestmentProjectCode(models.Model):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    fdi_sic_grouping = models.ForeignKey(
+        'investment.FDISICGrouping',
+        related_name='gva_multipliers',
+        on_delete=models.CASCADE,
+    )
+    multiplier = models.FloatField()
+    year = models.IntegerField()
+
+    class Meta:
```
reupen: This should have a verbose_name set as well as Django won't be able to work it out...

reupen:
Any reason for not adding these to the admin site (even as view only)?

datahub/investment/project/models.py
marcuspp: It's financial. What date it relates to isn't straight forward and will be explained in a different PR.

changelog/investment/fdi_sic_grouping.db.rst
marcuspp: It seems to render okay. And I just highlighted all keywords, field names, table names etc. Happy to change it if you want me too.

datahub/investment/project/migrations/0059_fdi_sicgrouping.yaml
marcuspp: The name is all on one line now.

datahub/investment/project/models.py
reupen: OK, so how about calling it `financial_year` instead of just `year`?

changelog/investment/fdi_sic_grouping.db.rst
reupen: Are you looking on GitHub?

datahub/investment/project/models.py
marcofucci: I also think that `year` is confusing. `financial_year` is a bit better as it doesn't say if it's the year starting or ending. So `financial_year_starting_in`?

One more: as it is now, people would have to create records for each multiplier every year even if the value hasn't changed unless we rename this to include something like `applied_from_financial_year_starting_in` 😬

marcuspp: Not necessarily because the logic that selects the multiplier will use the latest value (if there isn't a land date) and will be handled in a different PR. But it shouldn't be a problem to update them all as it would be a list that we get and the values will only be updated once a year.
I will update the field to `financial_year` but i'll add help text so it states financial year starting in

marcuspp: Basically the code will be robust enough to work with all or some of the yearly value. I take your point about `applied_from..` but I think that can all be covered in the help text.

changelog/investment/fdi_sic_grouping.db.rst
marcuspp: I have updated it now so less bold and italics. 👍

datahub/investment/project/models.py
```diff
@@ -681,11 +681,15 @@ class GVAMultiplier(models.Model):
         on_delete=models.CASCADE,
     )
     multiplier = models.FloatField()
-    year = models.IntegerField()
+    financial_year = models.IntegerField(
+        help_text=(
+            'The end of the financial year that the gva multiplier data was derived from.'
```
marcofucci: If it's the end of the financial year, should the date in the fixture be `2019`?

marcuspp: No because the data is from 2017-2018 we will get another list next month (2018-2019).

marcofucci: :shrug ok

marcuspp: I take this back, I've just double checked this and we won't be getting new data, so you are correct its best if it is 2019. I will update.

datahub/investment/project/models.py
elcct: If this uses the latest value, why not having just `created_on` field?

PR #1564, opened 2019-03-26
Make EmployeeRange read-only in the Django admin

marcofucci:
### Description of change

This makes the `metadata.EmployeeRange` model read-only in the Django admin and makes sure the data is loaded via a data migration.

### Checklist

### DIFF
--- a/changelog/company/employee_range_readonly.internal.rst
+++ b/changelog/company/employee_range_readonly.internal.rst
@@ -1,0 +1,1 @@
+It's not possible to change ``EmployeeRange`` records from the Django admin anymore.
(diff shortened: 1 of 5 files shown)


marcofucci:
@reupen true, I've added it to the `MODELS_TO_REGISTER_EDITABLE_ORDER_ONLY` which makes the `order` field editable (not the order fields) like with the `HeadquarterType` one.
This isn't really the best implementation but I don't necessarily want to add extra complexity (e.g. with a new `ReadOnlyOrderedMetadataAdmin` class) as we are going to get rid of this in the next months.

However, if you think it's better I would create a `ReadOnlyOrderedMetadataAdmin` instead or we could even get rid of the edit order functionality for `HeadquarterType`?

reupen:
@marcofucci Yes, there should only be one correct order so I don't think it needs to be editable. I'm not sure what is going on with the HeadquarterType orders in production though:

image

marcofucci:
Closing as `employeerange` will not be used/needed after the D&B work and therefore there's no point in tidying things up.

PR #1565, opened 2019-03-28
Make event organiser field required in the API

reupen:
### Description of change

This field is being made required so that there is always an adviser associated with an event.

This is blocked as there is currently a bug in the front end where this field does not save.

### Checklist

### DIFF
--- a/datahub/event/migrations/0006_add_organiser.py
+++ b/datahub/event/migrations/0006_add_organiser.py
@@ -18,6 +18,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='event',
             name='organiser',
-            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to=settings.AUTH_USER_MODEL),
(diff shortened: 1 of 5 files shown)


datahub/event/migrations/0006_add_organiser.py
```diff
@@ -18,6 +18,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='event',
             name='organiser',
-            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to=settings.AUTH_USER_MODEL),
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.PROTECT, to=settings.AUTH_USER_MODEL),
```
marcuspp: Why edit an old migration?

reupen: It doesn't change the schema, and the auto-generated migration will unnecessarily drop and recreate the foreign key constraint which we want to avoid.

PR #1578, opened 2019-04-01
Populate Gross Value Added on investment projects

marcuspp:
### Description of change
To populate the Gross Value Added I have added pre save signal to investment projects.
I thought this was easiest way to check all the parameters and set the variables.
Initially I was not going to store the `gross_value_added` value but when looking at adding it to search and the CSV download it became apparent that it was easier to store the data.

Still to do
- Check the actual land date to select the correct financial year
- An admin page to add the new multiplier data

### Checklist


datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def update_gross_value_added_for_investment_project(investment_project):
+    """Updates the Gross Value Added data for investment project."""
+    calculate_gross_value_added = CalculateGrossValueAdded(investment_project)
+    gva_multiplier = calculate_gross_value_added.gva_multiplier
+    gva_multiplier_id = getattr(gva_multiplier, 'id', None)
+    if gva_multiplier_id == investment_project.gva_multiplier_id:
+        return
+    investment_project.gva_multiplier_id = gva_multiplier_id
```
marcuspp: I need to update this as if the foreign equity value changes in the gva needs to be updated.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,98 @@
...
+        :returns the financial year that should be used for the investment project.
+        """
+        return 2019
+
+
+def update_gross_value_added_for_investment_project(investment_project):
+    """Updates the Gross Value Added data for investment project."""
+    calculate_gross_value_added = CalculateGrossValueAdded(investment_project)
+    gva_multiplier = calculate_gross_value_added.gva_multiplier
+    gva_multiplier_id = getattr(gva_multiplier, 'id', None)
```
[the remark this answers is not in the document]
marcuspp: What I am doing is checking to see if it's changed not just if it is not `None`. So it could be changing from a value to `None` so it's not exactly what you have suggested. I'll have a think and see if I can re-write it in a better way.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+class CalculateGrossValueAdded:
+    """Calculate Gross Value Added (GVA)."""
+
+    def __init__(self, investment_project):
+        """Sets the investment project and cache value for the gva multiplier."""
+        self.investment_project = investment_project
+        self._gva_multiplier = None
+
+    @property
```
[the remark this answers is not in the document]
marcuspp: Yeah great shout we should be using this else where. :+1:

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        """:returns the GVA multiplier if one is found."""
+        return self._get_gva_multiplier_for_investment_project()
+
+    @cached_property
+    def gross_value_added(self):
+        """Calculates the Gross Value Added (GVA) for an investment project."""
+        if not self.investment_project.foreign_equity_investment or not self.gva_multiplier:
+            return None
+        return round(
+            self.gva_multiplier.multiplier * float(
```
currycoder: From what I can tell, this is multiplying a float value by a Decimal value converted to float.  My understanding is that there is room for some errors to do with limitations of floating point arithmetic - https://docs.python.org/3.6/tutorial/floatingpoint.html

Would it be possible to convert GVAMultiplier's `multiplier` field to be a DecimalField instead?  I believe that would vastly reduce the possibility of approximation errors.

datahub/investment/project/test/test_gva_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            (130000000, 0.4537, 58981000),
+            (10000000, 0.0621, 621000),
+            (111000, 0.0621, 6893),
+            (12625500, 0.0581, 733542),
+            (296000, 0.0581, 17198),
+            (7002180, 0.0386, 270284),
+            (287732, 0.3939, 113338),
+            (1800000, 0.0264, 47520),
+            (28000, 0.021, 588),
+            (8907560, 0.9526, 8485342),
```
currycoder: Could be good to put some expected decimal vals in to here as foreign_equity_investment.

datahub/investment/project/gva_utils.py
marcuspp: Foreign equity investment is a decimal field but to 0 decimal places. I think it was done this way so it could store a number greater than that of an integer. But you are right ideally they should be the same type of field. I didn't want to presume that the multiplier would be to 4 decimal places. The field is then rounded so the amount after the decimal place is thrown away anyway.

datahub/investment/project/test/test_gva_utils.py
marcuspp: As above the field is to 0 decimal places.

datahub/investment/project/signals.py
```diff
@@ -18,3 +19,19 @@ def project_code_project_post_save(sender, **kwargs):
...
+)
+def update_gross_value_added_for_investment_project_pre_save(sender, instance, **kwargs):
+    """
+    Checks the investment project and if anything has changed that could affect
+    the Gross Value Added if required the values are updated.
+
+    GVA can change if the sector, business activity is set to retail or the
+    foreign equity investment is updated.
+    """
+    update_gross_value_added_for_investment_project(instance)
```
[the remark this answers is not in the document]
marcuspp: Not sure about this. I see the whole point of this logic being in the pre_save is so that it always gets executed. Do you think there would ever be a need to load projects via fixtures? Or are you just thinking of test data?

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @cached_property
+    def gva_multiplier(self):
+        """:returns the GVA multiplier if one is found."""
+        return self._get_gva_multiplier_for_investment_project()
+
+    @cached_property
+    def gross_value_added(self):
+        """Calculates the Gross Value Added (GVA) for an investment project."""
+        if not self.investment_project.foreign_equity_investment or not self.gva_multiplier:
```
reupen: Is there some consideration around `client_cannot_provide_foreign_investment` as well? I can't see any validation logic that checks that `foreign_equity_investment` is None if `client_cannot_provide_foreign_investment` is True or None (although I think the FE is maintaining that consistency).

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        if not fdi_sic_grouping:
+            return None
+        return self._get_gva_multiplier(fdi_sic_grouping.id)
+
+    def _get_fid_sic_grouping_for_sector(self):
+        """:returns the FDI SIC Grouping for a DIT Sector."""
+        root_sector = self.investment_project.sector.get_root()
+        investment_sector = getattr(root_sector, 'investmentsector', None)
+        if not investment_sector:
+            return
```
reupen: This one should be `return None` as well.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,113 @@
...
+        """:returns the GVA Multiplier for the a sector."""
+        fdi_sic_grouping = self._get_fid_sic_grouping_for_sector()
+        if not fdi_sic_grouping:
+            return None
+        return self._get_gva_multiplier(fdi_sic_grouping.id)
+
+    def _get_fid_sic_grouping_for_sector(self):
+        """:returns the FDI SIC Grouping for a DIT Sector."""
+        root_sector = self.investment_project.sector.get_root()
+        investment_sector = getattr(root_sector, 'investmentsector', None)
```
reupen: I'd suggest handling trying to access the attribute and handling `DoesNotExist` exception since it's more specific and is what we do elsewhere in the code base.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    def _get_sector_gva_multiplier(self):
+        """:returns the GVA Multiplier for the a sector."""
+        fdi_sic_grouping = self._get_fid_sic_grouping_for_sector()
+        if not fdi_sic_grouping:
+            return None
+        return self._get_gva_multiplier(fdi_sic_grouping.id)
+
+    def _get_fid_sic_grouping_for_sector(self):
+        """:returns the FDI SIC Grouping for a DIT Sector."""
+        root_sector = self.investment_project.sector.get_root()
```
reupen: There are a handful of old investment projects in production without a sector, so we probably need to protect against that (could log something as well).

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
reupen: In that case, `gva_multiplier_id = gva_multiplier.id if gva_multiplier else None` will probably do.

datahub/investment/project/models.py
```diff
@@ -300,6 +300,9 @@ class Meta:
         'GVAMultiplier', null=True, blank=True, on_delete=models.SET_NULL,
         related_name='investment_projects',
     )
+    gross_value_added = models.IntegerField(
+        null=True,
```
reupen: Any reason for not giving it `blank=True` as well?

datahub/investment/project/test/test_views.py
```diff
@@ -1048,6 +1054,27 @@ class TestPartialUpdateView(APITestMixin):
     These cover PATCH /v3/investment/<id>
     """
 
+    def test_change_foreign_equity_investment_updates_gross_value_added(self):
```
reupen: It would be good to add a similar test when only the business activities are changed.

datahub/investment/project/gva_utils.py
marcuspp: The calculation only works if there is a value for `foreign_equity_investment`. So I think the logic is correct. It doesn't matter what `client_cannot_provide_foreign_investment` is set to.

datahub/investment/project/gva_utils.py
marcuspp: The check for a sector is done in `_get_gva_multiplier_for_investment_project`

datahub/investment/project/gva_utils.py
marcuspp: Ok. But this to me is one of them cases where I think its more readable as it is and I don't really see the advantage of adding a try/catch.  If this was an attribute not a FK you'd just check the attribute is not None. But saying that I don't mind changing it if thats what is preferred.

datahub/investment/project/models.py
marcofucci: Could you add some help text saying that this value is calculated automatically from gva_multiplier and denormalised here?

datahub/investment/project/test/test_gva_utils.py
```diff
@@ -0,0 +1,170 @@
...
+                None,
+            ),
+        ),
+    )
+    def test_set_gva_multiplier(
+        self,
+        investment_type,
+        sector,
+        business_activities,
+        multiplier_value,
```
marcofucci: It would be more obvious to call this `expected_multiplier_value`.

datahub/investment/project/test/test_views.py
marcofucci: It would also be good to test all the cases as well, e.g. when gva was set but an investment project field is updated and that changes the gva to None etc.

datahub/investment/project/gva_utils.py
marcuspp: Isn't that the point of the tests to catch that issue if that did occur? Anyway... I'll update as suggested.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    """Updates the Gross Value Added data for investment project."""
+    calculate_gross_value_added = GrossValueAddedCalculator(investment_project)
+    gva_multiplier = calculate_gross_value_added.gva_multiplier
+    gva_multiplier_id = getattr(gva_multiplier, 'id', None)
+    if gva_multiplier_id != investment_project.gva_multiplier_id:
+        investment_project.gva_multiplier_id = gva_multiplier_id
+    gross_value_added = calculate_gross_value_added.gross_value_added
+    if gross_value_added != investment_project.gross_value_added:
+        investment_project.gross_value_added = calculate_gross_value_added.gross_value_added
+    return investment_project
```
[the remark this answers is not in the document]
marcuspp: That would affect the audit history.

datahub/investment/project/signals.py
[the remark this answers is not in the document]
marcuspp: I've tested this and this works fine as is. The values get populated when loading the fixtures.

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
marcuspp: Updated

datahub/investment/project/management/commands/populate_gross_value_added.py
```diff
@@ -0,0 +1,40 @@
...
+        investment_projects = self.get_investment_projects()
+        for project in investment_projects.iterator():
+            project.save()
+
+    def get_investment_projects(self):
+        """Get investment projects. returns: All projects that GVA could be calculated for."""
+        return InvestmentProject.objects.filter(
+            investment_type_id=InvestmentTypeConstant.fdi.value.id,
+            foreign_equity_investment__isnull=False,
+        ).filter(
```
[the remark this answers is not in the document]
marcuspp: I think this is fine as it is.

datahub/investment/project/gva_utils.py
marcuspp: I've decided to leave this as is when updating to decimal it caused more difficulties. If the consensus is that they should both be the same then I will update.

datahub/investment/project/gva_utils.py
marcuspp: Updated.

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
marcuspp: Agreed. I have updated and removed the check if the values have changed as you have said it actually doesn't matter if the values are set every time or not.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+        except GVAMultiplier.DoesNotExist:
+            return None
+
+    def _get_investment_sector(self, root_sector):
+        """:returns the investment sector for a root DIT sector if one found else returns None."""
+        try:
+            return root_sector.investmentsector
+        except InvestmentSector.DoesNotExist:
+            return None
```
elcct: Should this silently "fail"? I mean, what is the impact if a root sector doesn't have investment sector assigned or is it okay that some root sectors wouldn't have investment sector?

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _get_gva_multiplier_financial_year(self):
+        """
+        TODO: Check the investment project actual land date and
+        to future proof when new GVA multiplier data is added and when no
+        GVA Multiplier data exists for a given year.
+
+        :returns the financial year that should be used for the investment project.
+        """
+        return 2019
```
elcct: Would it be useful for this to be an environment variable?

marcuspp: This is going to be addressed in a future PR.

datahub/investment/project/gva_utils.py
marcuspp: I will add a log in so this gets flagged all current root sectors do have an investment sector.

datahub/investment/project/test/factories.py
```diff
@@ -174,3 +175,14 @@ class InvestmentActivityFactory(factory.django.DjangoModelFactory):
...
     class Meta:
         model = 'investment.InvestmentActivity'
+
+
+class GVAMultiplierFactory(factory.django.DjangoModelFactory):
+    """GVA Multiplier factory."""
+
+    fdi_sic_grouping_id = FDISICGrouping.retail.value.id
+    financial_year = 2019
+    multiplier = float(0.1250)
```
reupen: `0.1250` is already a float, so not sure what the `float()` is for here.

datahub/investment/project/gva_utils.py
marcofucci: Yeah I think we should use decimals where possible. 
I have literally just updated my PR from float to decimal even if it was unlikely it would have create problems as well.

datahub/investment/project/gva_utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def set_gross_value_added_for_investment_project(investment_project):
+    """Sets the Gross Value Added data for investment project."""
+    calculate_gross_value_added = GrossValueAddedCalculator(investment_project)
+    gva_multiplier = calculate_gross_value_added.gva_multiplier
+    investment_project.gva_multiplier_id = gva_multiplier.id if gva_multiplier else None
```
reupen: Isn't this the same as `investment_project.gva_multiplier = calculate_gross_value_added.gva_multiplier`?

datahub/investment/project/signals.py
```diff
@@ -18,3 +19,19 @@ def project_code_project_post_save(sender, **kwargs):
...
     if created and not instance.cdms_project_code:
         InvestmentProjectCode.objects.create(project=instance)
+
+
+@receiver(
+    pre_save,
+    sender=InvestmentProject,
+    dispatch_uid='set_gross_value_added_for_investment_project_pre_save',
+)
+def set_gross_value_added_for_investment_project_pre_save(sender, instance, **kwargs):
```
marcofucci: Don't you need to update investment projects also even when/if any of the fields on the related `GVAMultiplier` change?

marcuspp: Yes, but that is out of scope for this PR and is what I am working on at the moment.

datahub/investment/project/management/commands/populate_gross_value_added.py
```diff
@@ -0,0 +1,40 @@
...
+        Main method to loop over all investment projects that GVA
+        could be calculated for and saving the project.
+
+        Saving the project in turn calls the pre_save method
+        'update_gross_value_added_for_investment_project_pre_save'
+        which sets the Gross Value added data for a project.
+        """
+        investment_projects = self.get_investment_projects()
+        for project in investment_projects.iterator():
+            project.save()
```
marcofucci: This also updates the `modified_on` date which is not ideal as this is just a correction and not triggered by any user action. 
Ideally you would use `update_fields` here so that only gross_value_added fields are changed.
I think it's important for related logic (e.g. if we want to get rid of investment projects that have not been modified for x months) but I'll leave it with you.

marcuspp: Sure I agree. Will update to use just update_fields.

datahub/investment/project/test/factories.py
marcuspp: Updated to a decimal

datahub/investment/project/gva_utils.py
marcuspp: Updated all fields to decimals.

PR #1585, opened 2019-04-04
Update Python to version 3.7.2 🎉

reupen:
### Description of change

Release notes: https://docs.python.org/3/whatsnew/3.7.html

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,7 +105,7 @@ workflows:
     jobs:
       - build:
           publish_coverage: true
-          python_image: python:3.6
+          python_image: python:3.7
           postgres_image: postgres:10
           mi_postgres_image: postgres:9.6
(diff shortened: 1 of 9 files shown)


reupen:
I suppose we should really say something in the native installation section in the README about how to get Python 3.7. Should we cover using [pyenv](https://github.com/pyenv/pyenv)? @marcofucci @elcct you also use native installations, what do you think?

reupen:
I found using pyenv a bit too complicated, actually. I think I'll just add a generic step giving a few options without going into detail.

marcofucci:
@reupen if you want to add something for mac just point it to https://docs.python-guide.org/starting/install3/osx/

reupen:
Thanks @marcofucci, that's a useful guide – just pushed a couple of updates to the README.

PR #1592, opened 2019-04-05
[DRAFT] Email interaction ingestion

currycoder:
**This is a draft PR which encompasses all of the groundwork for building an email ingestion feature in to Data Hub - this PR itself will not merge (please don't approve it!).  I am looking for feedback on the overall approach and code critique - anything that I can do to better fit the team's standard backend approaches.**

### Description of change
Email ingestion technical spec for context: https://docs.google.com/document/d/16ObNTMRRGejk93XOnPQ5P7jl8sRRoFacVoT21K94JRE/edit

This brings together a number of changes:
- Adds `state`, `location` and `meeting_uid` fields to Interaction (and it's API endpoints).
- Adds an `email_ingestion` app which brings a framework for watching IMAP inboxes for emails, and processing them according to specific business needs.
- Adds a `meetings` mailbox to django settings along with a `CalendarInteractionEmailProcessor` class for ingesting calendar invites which are sent by advisers to an email inbox.  The processor creates "incomplete" interactions associated with the relevant  contacts, company, advisers, etc - based on data parsed out of the invite email.
- Adds a few supporting tools/modifications for making development easier under a docker environment.

Likely, each of these headline changes can form it's own individual PR.

### Some points of note
- There's a scattering of TODO items still in the code which I will review and work on as I do the individual PRs.
- There's only a simple nucleus for testing the email ingestion/calendar processing logic - again, tests are something that will be done more exhaustively in the individual PRs.

### Possible points of contention
- `state` has been added to Interaction as a new field with a default value "complete".  This avoids the need to create a whole new v4 API endpoint for interactions, but I couldn't find any existing examples in DH APIs where a default was used.
- I considered going with a more formal validation mechanism for parsing email details/calendar metadata in `CalendarInteractionEmailProcessor`.  In the end I've gone with a more simple approach for now; marshalling data and validating it at the same time and then persisting it as an interaction once we have gotten past validation.  As we layer extra functionality/features on this, we might rethink it.

### Checklist


reupen:
Some general comments from a quick look through:

1. The docstrings look like they are in the Google format (or something similar). We haven't used this format before. We have occasionally used `:param <arg>:` and `:returns:` which are reStructuredText.

2. Locking – elsewhere we have used `django_pglocks.advisory_lock` (PostgreSQL-based). I don't have anything specific against Redis locks, but it would be good to be consistent (unless there's a good reason not to be).

3. I think you could use [`abc.ABC` and `@abc.abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod) for EmailProcessor.

4. String formatting – we normally use f-strings (for templates though it's format strings and `str.format()`).

5. I think `state` and `service_delivery_status` sound similar and we should disambiguate them. Also, we have used the word 'status' in a number of places, but we haven't used `state` before.

I haven't looked at the ingestion and email logic in detail yet I'm afraid...

marcofucci:
Yeah quick look from me as well and I'd add these other points:

1. Adding that `state` in that way would cause problems with migrations as we were discussing in slack
2. PATCHing those fields should be stricter (e.g. you can't move a state to cancelled if it's complete
3. I'm not sure about that `meeting_uid`, maybe we can find a way to make it more generic so that it could be used used in other cases (e.g. change the name, use a context jsonfield etc.)

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,240 @@
...
+        else:
+            if use_ssl:
+                self.imap_port = 993
+            else:
+                self.imap_port = 465
+        self.use_ssl = use_ssl
+        self.processors = []
+        # Instantiate the processor classes for this mailbox
+        for processor_class in mail_processor_classes:
+            self.processors.append(processor_class())
```
marcofucci: I'm not sure initialising and persisting the objects is the right way, to the developer implementing an `EmailProcessor`, it might not look obvious that if he/she saves an attribute, it will persist.
Maybe initialise the classes every time or add that to the EmailProcessor docs that this is intended.

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,53 @@
...
+    ]
+    auth_results = {
+        'dkim': False,
+        'spf': False,
+        'dmarc': False,
+    }
+    for line in authentication_lines:
+        try:
+            if line.startswith('dkim'):
+                assert(line.startswith('dkim=pass'))
```
marcofucci: assert without parentheses?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,159 @@
...
+
+class CalendarInteractionEmailProcessor(EmailProcessor):
+    """
+    An EmailProcessor which checks whether incoming email is a valid DIT/company
+    meeting and creates an incomplete Interaction model for it if so.
+    """
+
+    def _get_adviser_for_email(self, email):
+        try:
+            return Advisor.objects.get(email=email)
```
marcofucci: They might be using `contact_email` as well, right?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,159 @@
...
+        all_advisers.extend(secondary_advisers)
+        for adviser in all_advisers:
+            InteractionDITParticipant.objects.create(
+                interaction=interaction,
+                adviser=adviser,
+                team=adviser.dit_team,
+            )
+        return interaction
+
+    def process_email(self, message):
```
marcofucci: This method is very long and hard to read, I'd consider refactoring or splitting it

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,159 @@
...
+            return (False, 'Email not sent by DIT')
+        sender_email = message.from_[0][1]
+        sender_adviser = self._get_adviser_for_email(sender_email)
+        if not sender_adviser:
+            return (False, 'Email not sent by recognised DIT Adviser')
+        all_recipients = [item[1].strip() for item in message.to]
+        all_recipients.extend([item[1].strip() for item in message.cc])
+        all_recipients = set(all_recipients)
+        contacts = Contact.objects.filter(email__in=all_recipients)
+        secondary_advisers = Advisor.objects\
```
marcofucci: Maybe a logic similar to getting the sender_adviser?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,159 @@
...
+            .filter(email__in=all_recipients)\
+            .exclude(id=sender_adviser.id)
+        if not contacts.count() > 0:
```
marcofucci: do you need the `> 0`?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,159 @@
...
+            return (False, 'No email receipients were recognised as Contacts')
+        company = self._get_company_from_contacts(contacts)
```
marcofucci: This only gets the first contact, it'd be better if it was documented and with some TODO to deal with other cases in the future.

currycoder:
Thanks for the feedback so far.  It all makes sense to me - I'll be looking at folding the changes in to the individual PRs as I go now.

I'll follow up on slack for some of the more ambiguous suggestions.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,159 @@
...
+            date=calendar_event['start'],
+            company=company,
+            contact=contacts[0],
+            subject=calendar_event['subject'],  # This should be auto-generated
+            dit_adviser=sender_adviser,
+            dit_team=sender_adviser.dit_team,
+            was_policy_feedback_provided=False,
+        )
+        for contact in contacts:
+            interaction.contacts.add(contact)
```
marcofucci: `contacts` can be from different companies which is not supported yet so you need to exclude the ones from other companies or have a different logic to address that case.

currycoder:
Closing as most functionality has made it in through progressive PRs now.
:wave:

PR #1602, opened 2019-04-08
GVA Multiplier Admin

marcuspp:
### Description of change
Initial GVA Multiplier admin page. Allows the multiplier value only to be updated. When it is updated it will in turn update all related investment projects.

### Design decision
 Used a task to update all the projects this is my preferred method but if there is alternative that is preferred don't mind updating.

### To test
- Make sure you restart celery if you are not ALWAYS_EAGER
- Go to http://localhost:8000/admin/investment/gvamultiplier/
- Test filtering and search 
- Update a project, select a sector and a foreign equity amount. 
- View http://localhost:8000/v3/investment/<uuid> and check gross value added is set
-  In the admin find the multiplier for the sector and try and delete (you should be prevented from doing so)
- Update the multiplier value and then check the project has been updated.

### Checklist


datahub/investment/project/test/test_admin.py
```diff
@@ -13,8 +14,109 @@
...
+        ) as mock_get_multiplier:
+            mock_get_multiplier.return_value = gva_multiplier
+            project = InvestmentProjectFactory(
+                foreign_equity_investment=1000,
+                investment_type_id=constants.InvestmentType.fdi.value.id,
+            )
+
+        url = reverse('admin:investment_gvamultiplier_change', args=(gva_multiplier.pk,))
+        response = self.client.get(url, follow=True)
+        assert response.status_code == status.HTTP_200_OK
```
marcofucci: Maybe you can get rid of these 2 lines as they are not necessary as might be misleading.

marcuspp: The purpose of the lines are to check the page renders correctly, so the issue with investment admin page (that had tests but didn't do a get on page) is not repeated. I can put the check into its own test if you prefer?

datahub/investment/project/tasks.py
```diff
@@ -0,0 +1,34 @@
...
+    except GVAMultiplier.DoesNotExist:
+        logger.warning(
+            f'Unable to find GVA Multiplier [{gva_multiplier_id}]'
+            f' - Unable to update associated investment projects',
+        )
+    else:
+        _update_gva_for_investment_projects(instance)
+
+
+def _update_gva_for_investment_projects(gva_multiplier):
```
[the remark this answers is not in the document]
marcuspp: I disagree with your suggestion I think it makes it more confusing especially when you start trying to write tests for it. 
I am going to rename it to `_update_investment_projects_for_gva_multiplier` omitting the `_task` (because it isn't one). That should be sufficient to understand they are part of the same process.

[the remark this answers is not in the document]
marcofucci: 👍 that's what I meant, I forgot to remove the `_task` suffix

datahub/investment/project/test/test_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+        assert not mock_update_gva_for_investment_projects.called
+
+    @mock.patch('datahub.investment.project.tasks._update_gva_for_investment_projects')
+    def test_update_investment_projects_for_gva_multiplier(
+        self,
+        mock_update_gva_for_investment_projects,
+    ):
+        """
+        Tests update investment projects for gva multiplier task updates
+        all related investment projects.
```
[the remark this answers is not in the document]
marcuspp: Added a test for `_update_investment_projects_for_gva_multiplier`

datahub/investment/project/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+    else:
+        _update_investment_projects_for_gva_multiplier(instance)
+
+
+def _update_investment_projects_for_gva_multiplier(gva_multiplier):
+    """Update gross_value_added for a GVA Multipliers related investment projects."""
+    investment_projects = gva_multiplier.investment_projects.all()
+    for investment_project in investment_projects.iterator():
+        investment_project.save(update_fields=['gross_value_added'])
```
marcofucci: Worth adding a comment explaining how the value is automatically calculated using a signal in case somebody doesn’t have the context whilst reading this.

datahub/investment/project/tasks.py
```diff
@@ -0,0 +1,36 @@
...
+from datahub.investment.project.models import GVAMultiplier
+
+logger = getLogger(__name__)
+
+
+@shared_task(
+    autoretry_for=(Exception,),
+    queue='long-running',
+    max_retries=3,
+    retry_backoff=True,
```
reupen: True just means 1 in this context. So with a max_retries of 3, it will retry after 1, 2 and then 4 seconds. I think it would make sense to use a larger value for `retry_backoff` so there is a bit more time for transient problems to clear.

changelog/investment/gva_multiplier_admin.internal.rst
```diff
@@ -0,0 +1 @@
+New Django Admin page to update and add `GVA Multipliers`.
```
reupen: Not sure what was intended with the backticks here – they don't seem to be doing anything in the rendered output.

PR #1604, opened 2019-04-08
Release 11.5.0

reupen:


marcuspp:
@reupen I need to chat about something before this goes out.

PR #1610, opened 2019-04-09
Add status and location fields to Interaction model

currycoder:
### Description of change

Adds the following fields to the Interaction model:
- `status` - nullable, choice field - `'draft'` or `'complete'`.
- `location` - nullable, char field - free text representation of where the interaction occurred.

These fields can additionally be specified when creating or updating interactions on the API.  We use sensible defaults to ensure that new interactions added to the DB will have populated data for these fields.

### Context

This is the first of a string of changes to facilitate ingesting calendar invite emails in to data hub as "draft" interactions which can later be completed by an adviser.  This will be followed up by some additional work to give Interactions archive fields, fields to log the external source of an interaction and email ingestion business logic.

### Data Migration

Both the `status` and `location` fields will be required in the DB in future.  We will run through the following process to ensure this:
- After releasing/deploying this code, we will run the `datahub.dbmaintenance.tasks.replace_null_with_default` task to set all existing Interactions to have `status='complete'`.
- After releasing/deploying this code, we will run the `datahub.dbmaintenance.tasks.replace_null_with_default` task to set all existing Interactions to have `location=''`.
- A follow-up PR will be created to make these fields required at the model level as well as adding appropriate defaults to ensure that existing working copies are migrated successfully.

### Checklist


datahub/interaction/models.py
```diff
@@ -206,6 +214,15 @@ class Interaction(BaseModel):
...
+    # Investments
+    investment_project = models.ForeignKey(
+        'investment.InvestmentProject',
+        related_name='%(class)ss',
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
+        help_text='For interactions only.',
+    )
     grant_amount_offered = models.DecimalField(
```
reupen: The comment above makes it look like `grant_amount_offered` and `net_company_receipt` relate to investment projects, but they relate to grants.

datahub/interaction/serializers.py
```diff
@@ -137,6 +140,10 @@ class InteractionSerializer(serializers.ModelSerializer):
         ),
     }
 
+    status = serializers.ChoiceField(
```
reupen: I would use `Meta.extra_kwargs` so that only the attributes that need to be overridden are overridden (for example, overriding a field in this way caused us problems in the past with `max_length`s being removed from CharFields).

datahub/interaction/serializers.py
```diff
@@ -150,6 +157,7 @@ class InteractionSerializer(serializers.ModelSerializer):
         ),
     )
     created_by = NestedAdviserField(read_only=True)
+    location = serializers.CharField(default='')
```
reupen: Same comment as above about using `Meta.extra_kwargs` – this is stripping out e.g. the max length validation.

datahub/interaction/validators.py
```diff
@@ -36,3 +37,37 @@ def __call__(self, data):
                 'The interaction contacts must belong to the specified company.',
                 code='inconsistent_contacts_and_company',
             )
+
+
+class StatusChangeValidator:
+    """
+    Validates that an interaction's status cannot change from complete back to
```
reupen: It would be good to explain the reason for this restriction.

datahub/interaction/test/views/test_interaction.py
```diff
@@ -202,6 +211,108 @@ def test_add(self, extra_data, permissions):
             },
             'created_on': '2017-04-18T13:25:30.986208Z',
             'modified_on': '2017-04-18T13:25:30.986208Z',
+            'location': request_data.get('location', ''),
+        }
+
+    @freeze_time('2017-04-18 13:25:30.986208')
+    @pytest.mark.parametrize('permissions', NON_RESTRICTED_ADD_PERMISSIONS)
+    def test_add_draft(self, permissions):
```
reupen: What about adding this as another case to test_add? We would probably need to specify service and communication_channel explicitly in each set of arguments.

reupen:
The added tests are all in test_interaction.py but it looks like the changes will affect service deliveries too.

So perhaps a few of the tests could be moved to test_common.py, and maybe a few tests could be added to test_service_delivery.py too (such as adding a draft service delivery).

datahub/interaction/validators.py
```diff
@@ -36,3 +37,37 @@ def __call__(self, data):
...
+        self.instance = serializer.instance
+
+    def __call__(self, data):
+        """
+        Performs validation.
+        """
+        if not self.instance:
+            return
+        existing_interaction_complete = self.instance.status == Interaction.STATUSES.complete
+        new_status = data.get('status', Interaction.STATUSES.complete)
```
marcofucci: This is incorrect as if status is not passed in, this line assumes that it's complete whilst it might not be.
Safer to use `DataCombiner`

datahub/interaction/serializers.py
currycoder: Thanks - that is neat!

datahub/interaction/validators.py
currycoder: Nice spot.  `DataCombiner` is neat!

datahub/interaction/test/views/test_interaction.py
currycoder: You're right.  I had already done this actually..  No idea why I also did an extra test method..

PR #1611, opened 2019-04-09
Add management command to upload company match candidates from CSV file

elcct:
### Description of change

This adds a management command that enables upload of CSV file with company match candidates.

It it possible to upload different sets of match candidates for a company by using different `batch_number`.

This code is temporary and the consumed CSV file is machine generated so basic validation should be enough.

### Checklist

### DIFF
--- a/changelog/company/update_company_match_candidates.feature
+++ b/changelog/company/update_company_match_candidates.feature
@@ -1,0 +1,1 @@
+Company match candidates can now be updated with a management command using data from CSV file
--- a/datahub/dbmaintenance/management/commands/update_company_match_candidates.py
+++ b/datahub/dbmaintenance/management/commands/update_company_match_candidates.py
@@ -1,0 +1,103 @@
+import json
+from base64 import b64decode
+from functools import lru_cache
+from logging import getLogger
+
+from datahub.company.models import Company
+from datahub.dbmaintenance.management.base import CSVBaseCommand
+from datahub.dbmaintenance.utils import parse_uuid
+from datahub.dnb_match.constants import DNB_COUNTRY_CODE_MAPPING
+from datahub.dnb_match.models import DnBMatchingCSVRecord
+from datahub.metadata.models import Country
+
+logger = getLogger(__name__)
+
+DNB_COUNTRY_MAPPING = {
+    entry['name']: entry['iso_alpha2_code']
+    for entry in DNB_COUNTRY_CODE_MAPPING.values()
+}
+
+
+class Command(CSVBaseCommand):
+    """Command to update DnBMatchingCSVRecord."""
+
+    EXPECTED_FIELDS = {
+        'duns_number',
+        'name',
+        'global_ultimate_duns_number',
+        'global_ultimate_name',
+        'global_ultimate_country',
+        'address_1',
+        'address_2',
+        'address_town',
+        'address_postcode',
+        'address_country',
+        'confidence',
+        'source',
+    }
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--batch_number',
+            type=int,
+            help='Batch number - version of the match candidates.',
+        )
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        company = Company.objects.get(pk=parse_uuid(row['id']))
+        batch_number = options['batch_number']
+
+        raw_data = b64decode(row['data'])
+        data = json.loads(raw_data)
+
+        if not self._validate_data(data):
+            logger.warning(
+                'Required fields are missing for given company: %s, %s',
+                company.pk,
+                data,
+            )
+            return
+
+        enriched_data = self._resolve_dnb_address_country(data)
+        if not enriched_data:
+            logger.warning(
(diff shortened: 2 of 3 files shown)


changelog/company/update_company_match_candidates.feature
```diff
@@ -0,0 +1 @@
+Added a management command to update company match candidates data using a CSV file
```
reupen: Should be something like 'Company match candidates can now be...' or 'A management command was added to...'.

(See https://github.com/uktrade/data-hub-leeloo/tree/develop/changelog.)

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
```diff
@@ -0,0 +1,97 @@
...
+from datahub.dbmaintenance.utils import parse_uuid
+from datahub.dnb_match.constants import DNB_COUNTRY_CODE_MAPPING
+from datahub.dnb_match.models import DnBMatchingCSVRecord
+from datahub.metadata.models import Country
+
+logger = getLogger(__name__)
+
+DNB_COUNTRY_MAPPING = {
+    entry['name']: entry['iso_alpha2_code']
+    for _, entry in DNB_COUNTRY_CODE_MAPPING.items()
```
reupen:
```suggestion
    for entry in DNB_COUNTRY_CODE_MAPPING.values()
```

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
```diff
@@ -0,0 +1,97 @@
...
+            DnBMatchingCSVRecord.objects.update_or_create(
+                company_id=company.id,
+                defaults={
+                    'batch_number': batch_number,
+                    'data': enriched_data,
+                },
+            )
+
+    def _validate_data(self, data):
+        return all(row.keys() == self.REQUIRED_FIELDS for row in data)
```
reupen: If they are are required fields, I think would expect >= rather than ==.

With == `EXPECTED_FIELDS` might be a better name than `REQUIRED_FIELDS`.

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _resolve_address_country(self, data):
+        """Resolve address_country with Data Hub country."""
+        for row in data:
+            country = self._map_country(row['address_country'])
+            if country:
```
marcofucci: Wouldn't this create problems with some records potentially having different formats if country is None and address_country doesn't get resolved? We could also skip the record with a warning if that happens.

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
marcofucci: Also, I did not intend to use that `name` in actual logic so please double-check that the those values match the ones in your file. I got the list from the D&B website but you never know, if could be out-of-date.

datahub/dbmaintenance/test/commands/test_update_company_match_candidates.py
```diff
@@ -0,0 +1,245 @@
...
+    """Test that the command updates the specified records (ignoring ones with errors)."""
+    caplog.set_level('WARNING')
+
+    companies = CompanyFactory.create_batch(5)
+    data = []
+
+    for _ in companies:
+        record = [_get_data_for_company(), _get_data_for_company()]
+        raw_record = json.dumps(record)
+        base64_record = b64encode(raw_record.encode('utf-8')).decode('utf-8')
```
marcofucci: I would have a small util function to dump and encode the record as it's repeated a few times and it's distracting from the main logic.

marcofucci: Do you actually need the loop? You could construct only one and reuse it in the actual test logic.

changelog/company/update_company_match_candidates.feature
elcct: Out of all news fragments about commands I used the only wrong one as base for my comment. Now, that's luck :) I updated it now. Thanks!

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
elcct: I have checked the values in the file and they all match the constants.

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
elcct: Great idea. Fixed!

datahub/dbmaintenance/management/commands/update_company_match_candidates.py
elcct: Good point. I have added a warning in case the country is not recognised and additional check in the tests. Also clarified that it looks for DnB country.

datahub/dbmaintenance/test/commands/test_update_company_match_candidates.py
elcct: Very good points. I should not leave this way. I have moved this to a new function and dropped the loop etc. Thanks!

PR #1612, opened 2019-04-09
Add command to data cleanse companies using D&B matches

marcofucci:
### Description of change

The new Django command `dnb_match.cleanse_companies_using_worldbase_match` loops over
all Data Hub companies with `duns_number == NULL` and D&B match and updates company fields from the matched D&B Worldbase record.

Duplicated matches are ignored as they cannot be automatically cleansed without potential error.

Data Hub companies are also archived if the Worldbase record indicates that they are out of business and if they have related investment projects or OMIS orders, their URL is logged so that can be manually checked.

Transactions are not used and if a record fails, it is logged and the execution of the command resumed.

The command disables search signal receivers to avoid overloading the queue. 
After it completes, ES needs to be made up-to-date with the command `sync_es`.

~### Not in this PR~

~The command should probably also print out the archived companies with Investment Projects and OMIS orders so that these can be manually checked (only a handful of them at the moment).~

~### Questions~

~Is this okay or it would trigger too many `sync_object_task` tasks?~

### How to test

Create a Worldbase match
```python
from django.utils.crypto import get_random_string
from datahub.dnb_match.utils import (
    EmployeesIndicator,
    NATIONAL_ID_SYSTEM_CODE_UK,
    OutOfBusinessIndicator,
    TurnoverIndicator,
)

company = Company.objects.filter(
    duns_number__isnull=True, 
    dnbmatchingresult__isnull=True, 
    archived=False,
).first()

duns_number = get_random_string(9, allowed_chars='0123456789')
DnBMatchingResult.objects.create(
    company=company,
    data={
        'dnb_match': {
            'name': 'WB Corp',
            'country': 'ENGLAND',
            'duns_number': duns_number,
            'global_ultimate_name': '',
            'global_ultimate_country': '',
            'global_ultimate_duns_number': ''
        },
        'wb_record': {
            'DUNS Number': duns_number,
            'Business Name': 'WB Corp',
            'Secondary Name': 'Known as...',
            'Employees Total': '11',
            'Employees Total Indicator': EmployeesIndicator.ESTIMATED.value,
            'Employees Here': '0',
            'Employees Here Indicator': EmployeesIndicator.ESTIMATED.value,
            'Annual Sales in US dollars': '100',
            'Annual Sales Indicator': TurnoverIndicator.ESTIMATED.value,
            'Street Address': '1',
            'Street Address 2': 'Main Street',
            'City Name': 'London',
            'State/Province Name': 'Camden',
            'Country Code': '785',
            'Postal Code for Street Address': 'SW1A 1AA',
            'National Identification Number': '12345678',
            'National Identification System Code': str(NATIONAL_ID_SYSTEM_CODE_UK),
            'Out of Business indicator': OutOfBusinessIndicator.NOT_OUT_OF_BUSINESS.value,
        },
        'matched_by': 'data-science'
    }
)
```

Simulate the command
```
./manage.py cleanse_companies_using_worldbase_match --simulate
```

Run the command
```
./manage.py cleanse_companies_using_worldbase_match
```

You should see
```
Started
Finished - total cleansed: 1, archived: 0
```

### Checklist

### DIFF
--- a/datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
+++ b/datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
@@ -1,0 +1,209 @@
+from logging import getLogger
+
+import reversion
+from django.contrib.postgres.fields import JSONField
+from django.contrib.postgres.fields.jsonb import KeyTextTransform, KeyTransform
+from django.core.management.base import BaseCommand
+from django.db.models import Count, Exists, OuterRef, Q
+from django.db.models.functions import Cast
+
+from datahub.company.models import Company
+from datahub.dnb_match.models import DnBMatchingResult
+from datahub.dnb_match.utils import update_company_from_wb_record
+from datahub.search.signals import disable_search_signal_receivers
+
+
+logger = getLogger(__name__)
+
+# list of (related field, query annotation name)
+# where the 'query annotation name' can be used to access a bool value indicating
+# if the company has any related field objects)
+RELATED_FIELDS_MAPPING = [
+    (
+        'intermediate_investment_projects',
+        'intermediate_investment_projects__exists',
+    ),
+    (
+        'investee_projects',
+        'investee_projects__exists',
+    ),
+    (
+        'investor_investment_projects',
+        'investor_investment_projects__exists',
+    ),
+    (
+        'orders',
+        'orders__exists',
+    ),
+]
+
+
+def _filter_by_existing_related_objects(companies):
+    """
+    Can be used to check if archived companies need to be manually checked
+    because some related objects exist which might need updating.
+
+    :returns: a QuerySet with companies in the `companies` list that have any
+        of the existing related objects defined in RELATED_FIELDS_MAPPING
+    """
+    qs = Company.objects.filter(
+        id__in=[company.id for company in companies],
+    )
+
+    filters = Q()
+    for field_name, annotation_name in RELATED_FIELDS_MAPPING:
+        related_field = Company._meta.get_field(field_name).field
+
+        subquery = related_field.model.objects.filter(
+            **{related_field.attname: OuterRef('pk')},
+        ).only('pk')
+
+        qs = qs.annotate(
+            **{annotation_name: Exists(subquery)},
+        )
+
+        filters = filters | Q(**{annotation_name: True})
+
+    return qs.filter(filters)
+
+
+class Command(BaseCommand):
+    """
+    Loops over all Data Hub companies with duns_number == NULL and D&B match and updates company
+    fields from the matched D&B Worldbase record.
+    Duplicated matches are ignored as they cannot be automatically cleansed without potential
+    error.
+
+    If the Worldbase record indicates that the company is out of business, the Data Hub
+    company is archived.
+
+    In case of errors, the failures are captured without stopping the command.
+
+    NOTE: The command disables search signals so avoid overloading the queue.
+    After it completes, you need to run `sync_es` manually to make sure
+    all documents are up-to-date.
+    """
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            help='Simulates the command by performing the cleansing and rolling things back.',
+        )
+
+    def _get_companies_queryset(self):
+        # subquery used to only get the matches without duplicates
+        subquery = DnBMatchingResult.objects.filter(
+            company__archived=False,
+        ).annotate(
+            dnb_match=KeyTransform('dnb_match', 'data'),
+            matched_duns_number=KeyTextTransform(
+                'duns_number',
+                Cast('dnb_match', JSONField()),
+            ),
+        ).values(
+            'matched_duns_number',
+        ).annotate(
+            group_count=Count('matched_duns_number'),
+        ).filter(
+            matched_duns_number__isnull=False,
+            group_count=1,
+        )
+
+        return Company.objects.annotate(
+            wb_record=KeyTransform('wb_record', 'dnbmatchingresult__data'),
+            matched_duns_number=KeyTextTransform(
+                'DUNS Number',
+                Cast('wb_record', JSONField()),
+            ),
+        ).filter(
+            duns_number__isnull=True,
+            archived=False,
+            matched_duns_number__in=subquery.values('matched_duns_number'),
+        )
+
+    @disable_search_signal_receivers(Company)
+    def handle(self, *args, **options):
+        """Handles the command."""
+        logger.info('Started')
+
+        result = {True: 0, False: 0}
+        archived_companies = []
+
+        for company in self._get_companies_queryset().iterator():
+            succeeded_or_not, archived = self.process_company(
+                company,
+                simulate=options['simulate'],
+            )
+
+            result[succeeded_or_not] += 1
+            if archived:
+                archived_companies.append(company)
+
+        self._print_results(result, archived_companies)
+
+    def _process_company(self, company, simulate=False):
+        """
+        Updates company fields from matched Worldbase record.
+        :returns: True if the company was/would have been archived
+        """
+        wb_record = company.dnbmatchingresult.data['wb_record']
+
+        with reversion.create_revision():
+            updated_fields = update_company_from_wb_record(
+                company,
+                wb_record,
+                commit=not simulate,
+            )
+
+            if not simulate:
+                reversion.set_comment('Updated from Dun & Bradstreet data.')
+
+            return 'archived' in updated_fields
+
+    def process_company(self, company, simulate=False):
+        """
+        Wrapper around _process_company to catch potential problems.
+
+        :param simulate: if True, the changes will not be saved
+        :returns: tuple of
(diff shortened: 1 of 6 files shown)


marcuspp:
Massive :+1: for the notes on testing the PR.

When I run the simulate mode locally I get.
``django.db.utils.IntegrityError: duplicate key value violates unique constraint "company_company_duns_number_key"
DETAIL:  Key (duns_number)=(000000001) already exists.``

I understand this is because I already have a company in my DB with a duns_number of '0000001'

So my question is can this ever happen? Or is this just because of test data and a DnBMatchingResult would only every be associated with a company with same duns_number.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+                    if archived:
+                        archived_companies.append(company)
+
+                logger.info(
+                    f'Finished - total cleansed: {total_cleansed}, '
+                    f'archived: {len(archived_companies)}',
+                )
+
+                if is_simulation:
+                    logger.info(f'Rolling back changes...')
```
marcuspp: No need to be f string.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+
+    If the Worldbase record indicates that the company is out of business, the Data Hub
+    company is archived.
+
+    In case of exception, all changes are rolled back.
+    """
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        simulation_group = parser.add_mutually_exclusive_group()
```
reupen: If the add_mutually_exclusive_group() stuff meant to be here?

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+            duns_number__isnull=True,
+            archived=False,
+            matched_duns_number__in=subquery.values('matched_duns_number'),
+        )
+
+    def handle(self, *args, **options):
+        """Handles the command."""
+        is_simulation = options['simulate']
+
+        logger.info(f'Started')
```
reupen: This one doesn't need to be an f-string as well.

marcofucci:
@marcuspp 

> So my question is can this ever happen? Or is this just because of test data and a DnBMatchingResult would only every be associated with a company with same duns_number.

Yeah sorry, I should have used a random number. That is just for setting up the environment for the manual test, it's never going to happen in real life.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
marcofucci: No, you are right. I wanted to keep the `only-print-queries` as well but it wasn't worth it so I got rid of it and forgot to remove this.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _get_companies_queryset(self):
+        # subquery used to only get the matches without duplicates
+        subquery = DnBMatchingResult.objects.filter(
+            company__archived=False,
+        ).annotate(
+            dnb_match=KeyTransform('dnb_match', 'data'),
+            matched_duns_number=KeyTextTransform(
+                'duns_number',
+                Cast('dnb_match', JSONField()),
```
reupen: Just FYI, the error without the cast looks like a known bug: https://code.djangoproject.com/ticket/29139

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+            ),
+        ).values(
+            'matched_duns_number',
+        ).annotate(
+            group_rank=Count('matched_duns_number'),
```
reupen: I find rank a bit confusing as RANK is a different SQL function.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+        is_simulation = options['simulate']
+
+        logger.info(f'Started')
+
+        total_cleansed = 0
+        archived_companies = []
+
+        try:
+            with transaction.atomic():
+                for company in self._get_companies_queryset().iterator():
```
reupen: Just FYI, whether .iterator() is used or not, the data will be cached from when the query is first run (whether locally or on the server). I can't see that mattering much here, anyway.

Also I assume that this either runs quick enough or we're only going to run it during a quiet period to make locking a non-concern.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,109 @@
...
+                    logger.info(f'Rolling back changes...')
+                    raise SimulationRollback()
+        except SimulationRollback:
+            logger.info(f'Changes rolled back')
+
+    def process_company(self, company):
+        """Updates company fields from matched Worldbase record."""
+        wb_record = company.dnbmatchingresult.data['wb_record']
+
+        with reversion.create_revision():
```
reupen: Wouldn't it make more sense creating a single revision for the entire update? Or did you run into problems that way?

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
marcofucci: 😮

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
marcofucci: It was mainly:
- to align with some other logic (e.g. the `dbmaintenance` one)
- to avoid any problems when reverting a single company from the admin (not sure if there are side effects)

Do you think it's worth using one revision for all versions?

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
marcofucci: Very true, thanks for the pointer. I was thinking to run it in the evening or over the weekend because of the sync_es and the several updates but I should probably check how quickly this runs (in local for example).

marcofucci:
As waiting for the transaction to commit locks the records, I've changed the strategy and I'm now not rolling back all changes if an error occurs but instead catching the error and continuing.

For this reason, I had to get rid of the atomic context manager and the logic in simulation mode is now skipping the save call instead. This is less accurate but I couldn't find any better way.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
reupen: Not now, since the the transaction has been removed.

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,207 @@
...
+            bool indicating if the company was processed successfully
+            bool indicating if the company was archived
+        """
+        succeeded = True
+        archived = False
+        try:
+            archived = self._process_company(company, simulate=simulate)
+        except Exception as exc:
+            succeeded = False
+            logger.info(f'Company {company.name} - {company.id} failed: {repr(exc)}')
```
reupen: logger.exception? (We used logger.exception in dbmaintenance too.)

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,207 @@
...
+            'Finished - '
+            f'succeeded: {result[True]}, '
+            f'failed: {result[False]}, '
+            f'archived: {len(archived_companies)}',
+        )
+
+        archived_companies_with_related_objects = _filter_by_existing_related_objects(
+            archived_companies,
+        )
+        if archived_companies_with_related_objects.exists():
```
reupen: Minor thing, but using .exists() and then iterating over the query set will execute two queries. You could instead use bool() or similar here to force the query set to be evaluated (and hence only execute one query).

datahub/dnb_match/management/commands/cleanse_companies_using_worldbase_match.py
```diff
@@ -0,0 +1,207 @@
...
+            bool indicating if the company was archived
+        """
+        succeeded = True
+        archived = False
+        try:
+            archived = self._process_company(company, simulate=simulate)
+        except Exception as exc:
+            succeeded = False
+            logger.info(f'Company {company.name} - {company.id} failed: {repr(exc)}')
+
```
reupen: The other thing is if the command gets killed part way through (for whatever reason), is there a way to work out what it did? It might be useful to log successful updates as well in that respect (similar to dbmaintenance).

PR #1618, opened 2019-04-11
Financial year for GVA

marcuspp:
### Description of change
This adds the logic to GVA for selecting the correct financial year for the GVA Multiplier.

The logic uses fallbacks if the Multipliers have not been updated then the most recent years data is used. 

It's important to remember the GVA prior to actually landing is just an estimate and that when a project is updated the GVA calculation is always checked.

### How to test
- Update a project so that it has a GVA value (set FDI, sector, foreign equity investment)
- View http://localhost:8000/v3/investment/ and check gross value added is set
- Go to http://localhost:8000/admin/investment/gvamultiplier/
- Add a new multiplier for the same sector as your project for a year in the future 2030
- Update the projects actual land date to be that year.
- Make sure that multiplier is used
- Create a new project with an actual land date of 2035
- Make sure the multiplier being used is that of the year 2030

### Checklist

### DIFF
--- a/changelog/investment/gva-logic-update.internal.rst
+++ b/changelog/investment/gva-logic-update.internal.rst
@@ -1,0 +1,1 @@
+The logic has been updated for selecting which financial year's data is used to calculate the GVA for an investment project.
--- a/datahub/investment/project/gva_utils.py
+++ b/datahub/investment/project/gva_utils.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from logging import getLogger
 
 from django.utils.functional import cached_property
@@ -6,6 +7,7 @@ from datahub.core.constants import (
     InvestmentBusinessActivity as InvestmentBusinessActivityConstant,
     InvestmentType as InvestmentTypeConstant,
 )
+from datahub.core.utils import get_financial_year
 from datahub.investment.project.constants import (
     FDISICGrouping as FDI_SICGroupingConstant,
 )
@@ -94,14 +96,40 @@ class GrossValueAddedCalculator:
         return investment_sector.fdi_sic_grouping
 
     def _get_gva_multiplier(self, fdi_sic_grouping_id):
-        """:returns a GVA Sector."""
+        """
+        :returns a GVA Multiplier or None.
+
+        Firstly retrieves all the GVA Multipliers for a FDI SIC Grouping. Then checks
+        if there is a multiplier for the given financial year. If one is not found
+        then returns the GVA Multiplier for the last year there is an entry for.
+
+        GVA multipliers will be reviewed each year to ensure they continue to be accurate with
+        a changing economy. If the GVA values have not been updated, the latest year's information
+        should continue to be used as it represents the most accurate data available
+
+        If there are no entries returns None.
+
+        """
+        gva_multipliers_for_grouping = GVAMultiplier.objects.filter(
+            fdi_sic_grouping_id=fdi_sic_grouping_id,
+        ).order_by(
+            '-financial_year',
+        )
+
+        financial_year = self._get_gva_multiplier_financial_year()
         try:
-            return GVAMultiplier.objects.get(
-                fdi_sic_grouping_id=fdi_sic_grouping_id,
-                financial_year=self._get_gva_multiplier_financial_year(),
+            return gva_multipliers_for_grouping.get(
+                financial_year=financial_year,
             )
         except GVAMultiplier.DoesNotExist:
-            return None
+            if gva_multipliers_for_grouping.filter(
+                financial_year__gt=financial_year,
+            ).exists():
+                logger.exception(
+                    f'Unable to find a GVA Multiplier for financial year {financial_year} '
+                    f'fdi sic grouping id {fdi_sic_grouping_id}',
+                )
+            return gva_multipliers_for_grouping.first()
 
     def _get_investment_sector(self, root_sector):
         """:returns the investment sector for a root DIT sector if one found else returns None."""
@@ -115,13 +143,24 @@ class GrossValueAddedCalculator:
 
     def _get_gva_multiplier_financial_year(self):
         """
-        TODO: Check the investment project actual land date and
-        to future proof when new GVA multiplier data is added and when no
-        GVA Multiplier data exists for a given year.
-
         :returns the financial year that should be used for the investment project.
+
+        If the investment project does not have an actual land date return the current
+        financial year. If the actual land date is less than or equal to 2019 return 2019
+        this is because there is no multiplier data prior to the 2019 financial year.
+
+        Due to the multiplier data being for a financial year for all investment projects
+        with actual land date greater than 2019 return the financial year that the project landed.
         """
-        return 2019
+        if not self.investment_project.actual_land_date:
+            return get_financial_year(
+                datetime.today(),
+            )
+
+        return max(
+            get_financial_year(self.investment_project.actual_land_date),
+            2019,
+        )
 
 
 def set_gross_value_added_for_investment_project(investment_project):
--- a/datahub/investment/project/test/test_gva_utils.py
+++ b/datahub/investment/project/test/test_gva_utils.py
@@ -1,13 +1,17 @@
+from datetime import date
 from decimal import Decimal
 from unittest import mock
 
 import pytest
+from freezegun import freeze_time
 
 from datahub.core.constants import (
     InvestmentBusinessActivity as InvestmentBusinessActivityConstant,
     InvestmentType as InvestmentTypeConstant,
     Sector as SectorConstant,
 )
+from datahub.investment.project.constants import FDISICGrouping as FDISICGroupingConstant
+from datahub.investment.project.gva_utils import GrossValueAddedCalculator
 from datahub.investment.project.test.factories import (
     GVAMultiplierFactory,
     InvestmentProjectFactory,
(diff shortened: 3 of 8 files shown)


changelog/investment/gva-logic-update.internal.rst
```diff
@@ -0,0 +1 @@
+Updated the logic for selecting which financial years data to use for the GVA calculation for an investment project.
```
elcct: Could be `year's` ?

datahub/investment/project/test/test_gva_utils.py
```diff
@@ -186,3 +193,22 @@ def test_calculate_gva(
...
+
+    @pytest.mark.parametrize(
+        'actual_land_date,expected_financial_year',
+        (
+            (None, 2030),
+            (date(1980, 1, 1), 2019),
+            (date(2018, 1, 1), 2019),
+            (date(2019, 3, 1), 2019),
+            (date(2019, 8, 1), 2019),
+            (date(2025, 3, 1), 2024),
```
elcct: That's probably not necessary, but I would add one more with `(date(2025, 3, 31), 2024),`.

datahub/investment/project/gva_utils.py
```diff
@@ -94,14 +95,26 @@ def _get_fdi_sic_grouping_for_sector(self):
...
+
         try:
-            return GVAMultiplier.objects.get(
-                fdi_sic_grouping_id=fdi_sic_grouping_id,
+            return gva_multipliers_for_grouping.get(
                 financial_year=self._get_gva_multiplier_financial_year(),
             )
         except GVAMultiplier.DoesNotExist:
-            return None
+            return gva_multipliers_for_grouping.first()
```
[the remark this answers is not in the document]
marcuspp: There should not be a gap the data. The data is analysed for a year before it is required to be loaded. If it is not loaded in the system then the team that uses the GVA will notice (as they will be responsible for loading the multipliers). There is also a script that can be run to refresh all GVA values for all projects if they become out of date (This has been cleared with Jim). And if a multiplier is edited all associated projects will automatically get updated.

marcuspp:
> Looks cool - just a couple of optional thoughts inline.
> 
> Also, something around the way GVA Multipliers works in general... It seems like you've now added a way for GVA multipliers to be picked up more dynamically for a particular investment project - e.g. "get me the most up to date multiplier for the expected land date" However, it seems like the attributed GVA multiplier for an investment project is only updated when the investment project is saved - not when a new/more relevant multiplier is added. Is a more dynamic update mechanism in scope for future work?

@currycoder It uses the actual land date not estimated which should be in the past. There could be edge cases where data isn't loaded etc but the refresh script can be run if required. If the process that has been put in place is followed then the data should always be up-to-date.

datahub/investment/project/gva_utils.py
```diff
@@ -115,13 +129,23 @@ def _get_investment_sector(self, root_sector):
...
+
+        Due to the multiplier data being for a financial year for all investment projects
+        with actual land date greater than 2019 we have to check the month it landed
+        and if it is before April minus 1 for the year of landing.
         """
-        return 2019
+        if not self.investment_project.actual_land_date:
+            return datetime.today().year
+
+        if self.investment_project.actual_land_date.year <= 2019:
```
reupen: Should this be checking the financial year?

reupen:
Similar questions as @currycoder – it looks like there is now scope for GVAs to change when a new GVA multiplier is added, and also when the current date changes.

Did we say that the multipliers for a given financial year are calculated using the previous year's data? In which case, it seems inevitable that they will be added after the previous financial year has ended.

There may be a workaround of running `./manage.py refresh_gross_value_added_values` but it doesn't seem obvious. Perhaps it could be run on a schedule?

marcuspp:
> Did we say that the multipliers for a given financial year are calculated using the previous year's data? In which case, it seems inevitable that they will be added after the previous financial year has ended.

No there is a year in between because they need to analyse the data. So the data loaded now (financial year 2018-2019) is derived from all the projects landed before April 2018. So at the moment they will be analysing all projects landed before April 2019 which then becomes financial year (2019-2020) data. So there should be plenty of time to load the data.

datahub/investment/project/gva_utils.py
marcuspp: Yes? If its 2019 or earlier we don't need to check if it's before April as we don't have any data for 2018. So it should just use 2019.

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
marcofucci: I think it would still be good to log this in sentry if Brendan's case happens so that people are warned.

datahub/investment/project/gva_utils.py
```diff
@@ -94,14 +96,26 @@ def _get_fdi_sic_grouping_for_sector(self):
         return investment_sector.fdi_sic_grouping
 
     def _get_gva_multiplier(self, fdi_sic_grouping_id):
-        """:returns a GVA Sector."""
+        """
+        :returns a GVA Multiplier or None.
+
+        Firstly retrieves all the GVA Multipliers for a FDI SIC Grouping. Then tries to see
+        if there is a multiplier for the given financial year. If one is not found
+        then returns the GVA Multiplier for the last year there is an entry for.
```
marcofucci: > If one is not found then returns the GVA Multiplier for the last year there is an entry for.
It would be good to explain why you are using the latest one as it might not be obvious.

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
marcuspp: Sure but this is very unlikely to happen and would alert on bad data. ie. Someone puts an actual land date in 2050?

datahub/investment/project/gva_utils.py
```diff
@@ -115,13 +129,23 @@ def _get_investment_sector(self, root_sector):
...
+        If the actual land date is less than or equal to 2019 return 2019 this is because there is
+        no multiplier data prior to the 2019 financial year.
+
+        Due to the multiplier data being for a financial year for all investment projects
+        with actual land date greater than 2019 we have to check the month it landed
+        and if it is before April minus 1 for the year of landing.
         """
-        return 2019
+        if not self.investment_project.actual_land_date:
+            return datetime.today().year
```
reupen: Similar to the below, why calendar year rather than financial year?

marcofucci: Shouldn't this be the financial year instead of the year?

datahub/investment/project/gva_utils.py
reupen: I find it confusing because it switches between calendar year and financial year. Even if the net effect is the same, it takes more effort to understand what is going on.

Also, if it's just to put a lower bound on the year returned, I'm not sure why it can't just be something like:

```python
financial_year = get_financial_year(self.investment_project.actual_land_date)
return max(financial_year, 2019)
```

datahub/investment/project/gva_utils.py
elcct: I missed this when reviewing. It would be great to address it.

marcuspp: Updated!

datahub/investment/project/gva_utils.py
marcuspp: Updated

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
marcuspp: Updated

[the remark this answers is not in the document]
reupen: The current config is that errors go to Sentry, but not warnings. So if we want this to go to Sentry, it would either have to be an error/`logger.exception()`, or [we would have to explicitly send it to Sentry](https://github.com/uktrade/data-hub-leeloo/blob/d9e6a84b7e6f3cd3a101901909894e582b3ca89b/datahub/search/execute_query.py#L35-L41).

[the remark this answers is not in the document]
marcuspp: @reupen thanks. Updated to logger.exception.

changelog/investment/gva-logic-update.internal.rst
```diff
@@ -0,0 +1 @@
+Updated the logic for selecting which financial year's data to use for the GVA calculation for an investment project.
```
marcofucci: [Full sentences](https://github.com/uktrade/data-hub-leeloo/blob/develop/changelog/README.rst) ;-)

datahub/investment/project/gva_utils.py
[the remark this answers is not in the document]
marcofucci: There are many investment projects with `actual_land_date >= 04/2019` so this would trigger many exceptions unless I'm missing something. 
Is there a problem with the data and a plan to fix that?
I think it would make more sense for this to raise an exception only if there's a gap in the data (meaning there's a multiplier for a financial year after this one but not this one).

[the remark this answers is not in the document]
marcuspp: `_get_gva_multiplier_financial_year` won't return a year less than 2019 as it does max(value, 2019)

changelog/investment/gva-logic-update.internal.rst
marcuspp: Updated

PR #1622, opened 2019-04-12
Add CSV download for large capital profiles

marcuspp:
### Description of change
This adds the option to download large capital profiles as a csv file.

### To test
- ./manage.py shell_plus
```python
from datahub.investment.investor_profile.test.factories import (
    LargeInvestorProfileFactory, 
    CompleteLargeInvestorProfileFactory,
)

CompleteLargeInvestorProfileFactory()
LargeInvestorProfileFactory()

```
- Go to http://localhost:8000/v4/search/large-investor-profile/export
- Click POST
- Check the CSV file

### Checklist


datahub/investment/investor_profile/models.py
```diff
@@ -156,6 +156,10 @@ class InvestorProfile(BaseModel):
 
     class Meta:
         unique_together = ('investor_company', 'profile_type')
+        verbose_name_plural = 'Large Capital Profiles'
```
reupen: Model verbose names should generally be in lower case.

datahub/search/large_investor_profile/test/test_views.py
```diff
@@ -561,3 +576,182 @@ def _get_data_item(self, result, check_item_key):
...
+                'Data Hub link': (
+                    f'{settings.DATAHUB_FRONTEND_URL_PREFIXES["company"]}'
+                    f'/{profile.investor_company.pk}/investments/large-capital-profile'
+                ),
+            }
+            for profile in sorted_profiles
+        ]
+
+        expected_rows = format_csv_data(expected_row_data)
+        actual_rows = [dict(item) for item in reader]
```
reupen: I assume the OrderedDict to dict conversion is to make pytest give better diffs, could you add a comment to explain the reason for it?

datahub/search/large_investor_profile/views.py
```diff
@@ -88,11 +94,93 @@ class SearchInvestorProfileAPIViewMixin:
...
+        ),
+        uk_region_locations_names=get_string_agg_subquery(
+            DBInvestorProfile,
+            'uk_region_locations__name',
+        ),
+        other_countries_being_considered_names=get_string_agg_subquery(
+            DBInvestorProfile,
+            'other_countries_being_considered__name',
+        ),
+        link=get_front_end_url_expression(
```
marcofucci: I would add another item in `DATAHUB_FRONTEND_URL_PREFIXES` instead.

marcuspp: I considered this but because this url needs the company pk in the middle of it I thought it was cleaner this way. 

Or an alternative could be
```

DATAHUB_FRONTEND_URL_PREFIXES =
{ 
    'large-capital-profile': '{DATAHUB_FRONTEND_BASE_URL}/companies/(company_pk}/investments/large-capital-profile'
}
# Not an f-string as the company_pk is not known in the settings file.

# When using it.
url = DATAHUB_FRONTEND_URL_PREFIXES.format(DATAHUB_FRONTEND_BASE_URL='...', company_pk=1234)

```

What do you think?

datahub/search/large_investor_profile/test/test_views.py
marcuspp: Sure. This was just copied from the investment csv test.

datahub/search/large_investor_profile/views.py
marcofucci: I think you can use:
```
DATAHUB_FRONTEND_URL_PREFIXES = {
    'large-capital-profile': f'{DATAHUB_FRONTEND_BASE_URL}/companies/{{company_pk}}/investments/large-capital-profile'}
}

# and when using it:
url = DATAHUB_FRONTEND_URL_PREFIXES['large-capital-profile'].format(company_pk=1234)
```
?

marcuspp: Initial thoughts are I don't like it as it would need to do something special for this url in `get_front_end_url_expression` anyway as it doesn't need to append a pk on the end of the string. The only advantage of this is so that all the urls will be in within the PREFIXES list. 
I will have a think and see if I can come up with an elegant solution and look more into how I could use a different function to Concat to populate the value.

marcuspp: I don't think this will work (not unless I am missing something). Concat is DB function I would need to split the string up anyway as company_id is part of the queryset (so I don't actually have the id). The result leaves me with what I have.

marcofucci: oh yeah I forgot that this is used in the `get_front_end_url_expression` as well. Yeah it looks like the whole `get_front_end_url_expression` and related pks logic might need to be refactored to support this case as well.
It would be nice to add the pk part to the `DATAHUB_FRONTEND_URL_PREFIXES` as it's used in many places.

marcuspp: Would you mind if I do the refactor in a separate PR?

datahub/search/large_investor_profile/test/test_views.py
reupen: I probably wrote it then – if I'm confused by my own code, I think that's even more reason for a comment :slightly_smiling_face:

datahub/search/large_investor_profile/test/test_views.py
```diff
@@ -732,6 +732,9 @@ def test_export(self, setup_es, request_sortby, orm_ordering):
         ]
 
         expected_rows = format_csv_data(expected_row_data)
+
+        # item is a list of tuples so is cast to a dict to make the comparison easier to
```
reupen: Isn't `item` an OrderedDict?

datahub/search/large_investor_profile/views.py
marcofucci: 👍 with a TODO

marcuspp: Done. Added the todo in `common.py`

PR #1623, opened 2019-04-12
Add select file page of import interactions admin tool

reupen:
### Description of change

This adds the first page of the import interactions admin tool (with preliminary content). As the feature is incomplete, the feature is behind a feature flag.

A link to the new page is displayed on the interaction change list in the admin site if the feature flag is active.

For now, submitting a valid file simply redirects back to the change list page.

When complete, the feature will also:

- validate the rows in the CSV file, displaying errors back to the user
- display a preview of what will be imported if the file validates
- match interactions to contacts using email addresses
- return unmatched interactions to the user (either displayed or a download)

The reason for this feature is so that we can load interactions to Data Hub from partners that do not have direct access to Data Hub. Each file will be fairly small (a few hundred interactions).

I would recommend looking at each commit separately (note: I reordered the commits, but GitHub is still displaying them in the original order).

#### Import interactions link

image

#### Select file page

image

#### Form error

image

#### Importing a file

2. Create a CSV file with the following columns: `kind`, `date`, `service`, `contact_email`, `adviser_1` (for a valid file, use different columns for an invalid file)
3. Navigate to the interaction change list page in the admin site
4. Click on the 'Import interaction' link
5. Select your CSV file
6. Submit the form
7. If the file is valid, you should be redirected to the change list page. Otherwise, an error should be displayed above the form field.

#### Feature hidden if feature flag not active

1. Delete or deactivate the `admin-interaction-csv-importer` feature flag if it exists
2. Navigate to the interaction change list page in the admin site
3. The 'Import interactions' link should not be present

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import.py
+++ b/datahub/interaction/admin_csv_import.py
@@ -1,0 +1,79 @@
+from django.contrib.admin.templatetags.admin_urls import admin_urlname
+from django.core.exceptions import PermissionDenied
+from django.http import HttpResponseRedirect
+from django.template.response import TemplateResponse
+from django.urls import path, reverse
+
+from datahub.core.admin_csv_import import BaseCSVImportForm
+from datahub.feature_flag.utils import feature_flagged_view
+
+
+INTERACTION_IMPORTER_FEATURE_FLAG_NAME = 'admin-interaction-csv-importer'
+
+
+class InteractionCSVForm(BaseCSVImportForm):
+    """Form used for loading a CSV file to import interactions."""
+
+    csv_file_field_label = 'Interaction list (CSV file)'
+    required_columns = {
+        'kind',
+        'date',
+        'service',
+        'contact_email',
+        'adviser_1',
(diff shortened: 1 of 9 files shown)


datahub/interaction/admin_csv_import.py
```diff
@@ -0,0 +1,79 @@
...
+class InteractionCSVForm(BaseCSVImportForm):
+    """Form used for loading a CSV file to import interactions."""
+
+    csv_file_field_label = 'Interaction list (CSV file)'
+    required_columns = {
+        'kind',
+        'date',
+        'service',
+        'contact_email',
+        'adviser_1',
```
elcct: Why is it called with `_1`? What is the meaning of `_1` in this context?

reupen: An interaction can have multiple advisers, and the tool will allow adding up to two advisers to an interaction (the first one is mandatory).

PR #1627, opened 2019-04-15
Make interactions archivable

currycoder:
### Description of change

This change makes Interactions archivable - using the `ArchivableModel` mixin.  This adds the following fields to Interaction:
- ``archived`` - boolean representing whether the interaction has been archived or not
- ``archived_by`` - foreign key to adviser - who archived this interaction
- ``archived_reason`` - free text justification for archiving this interaction
- ``archived_on`` - datetime for when the interaction was archived

Additionally, add `archive` and `unarchive` views to the Interactions API using the `ArchivableViewSetMixin` class.

### Context

As part of the email/meetings interactions project, we need to be able to record that an upcoming meeting has been cancelled.  To do this, we are reusing the existing archive pattern which is in use across other models in datahub.  An interaction whose `status='draft'` and `archived=True` will be considered as cancelled - the frontend will display a badge conveying this.  
In the future, we may want to allow interactions to be arbitrarily archived outside of this meeting feature - this PR provides the foundations to allow this.

### Data Migration

The `archived` field will be required in the DB in future. We will run through the following process to ensure this:

- After releasing/deploying this code, we will run the `datahub.dbmaintenance.tasks.replace_null_with_default` task to set all existing Interactions to have `archived=False`.
- A follow-up PR will be created to make this field required at the model level as well as adding appropriate defaults to ensure that existing working copies are migrated successfully.

### Checklist


datahub/interaction/serializers.py
```diff
@@ -150,6 +150,10 @@ class InteractionSerializer(serializers.ModelSerializer):
         ),
     )
     created_by = NestedAdviserField(read_only=True)
+    # TODO: make this read-only once we have migrated all existing interactions
+    # to have archived=False
+    archived = serializers.BooleanField(default=False)
```
marcofucci: Not sure why you are making this editable, to add the default when new records are created? I'd instead override the `validate` method to set that field instead, see https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/interaction/serializers.py#L220

datahub/interaction/test/views/test_common.py
```diff
@@ -495,15 +499,22 @@ def test_cannot_update_read_only_fields(self):
...
+                'archived_on': date.today(),
+                'archived_reason': 'test',
             },
         )
 
         assert response.status_code == status.HTTP_200_OK
         assert response.data['archived_documents_url_path'] == 'old_path'
+        assert response.data['archived_by'] is None
+        assert response.data['archived_on'] is None
+        assert response.data['archived_reason'] is None
```
marcofucci: I'd explicitly set those archive_* attributes in the factory creator so that we don't rely on defaults and we make it clear here that the initial values were None.

datahub/interaction/serializers.py
currycoder: That's exactly why.  Good idea - will do

datahub/interaction/serializers.py
```diff
@@ -230,6 +231,14 @@ def validate(self, data):
             data['dit_adviser'] = first_participant['adviser']
             data['dit_team'] = first_participant['adviser'].dit_team
 
+        # Ensure that archived=False is set for creations/updates, when the
+        # existing instance does not have a value for it
+        # TODO: remove this once we give archived a model-level default
+        is_create = not self.instance
+        archived_not_set = not is_create and self.instance.archived is None
+        if is_create or archived_not_set:
```
reupen: This is a bit confusing as `is_create` is checked twice. Isn't the condition just `not self.instance or self.instance.archived is None`?

currycoder: Point.  My quest to make it self-documenting went a bit OTT, will swap out the logic as you suggest.

PR #1628, opened 2019-04-16
Add a source JSONField to Interaction

currycoder:
### Description of change

Adds a field `Interaction.source' - to represent data about external sources that we may create interactions from.  This is a JSONField to allow developers to be unconstrained in how they use this field for different data sources.

### Context

This will allow us to record/look up meeting ids for interactions that are automatically created for meeting invite emails.  This will assist us in making sure that we don't create duplicate interactions for the same meeting.

### Points of note

I haven't proposed that we set a default for this field - as this might be a little complex with json fields.  It might be preferable to perform a manual migration across the DB to set every interaction to have `source={}` - this could be coupled with an API change to ensure that newly created interactions have a default source of `{}`.  Any thoughts?

### Checklist


reupen:
> I haven't proposed that we set a default for this field - as this might be a little complex with json fields. It might be preferable to perform a manual migration across the DB to set every interaction to have source={} - this could be coupled with an API change to ensure that newly created interactions have a default source of {}. Any thoughts?

It it would help with querying, but it seems pretty forgiving to me as both of these work (don't return an error):

```
select null::jsonb @> '{"b": 2}'::jsonb;
select null::jsonb -> 'b';
```

Also, even if the field were non-nullable it wouldn't preclude strings and numbers as values.

So, null as the empty value seems OK to me (and nothing would stop us changing our minds later on). Are there any plans for the front end to use `source`?

changelog/interaction/source.db.rst
```diff
@@ -0,0 +1,4 @@
+A supporting field was added to ``interaction_interaction`` for the 
+purpose of allowing interactions to be archived:
```
reupen: Is this the right text for this news fragment?

currycoder:
> At the moment the field is editable in the admin site, should we make it read-only and format the same way we did for the user event log?

Good idea - I think this makes a lot of sense.  I can't see a use-case for admins to adjust the source on an interaction through the admin panel.

changelog/interaction/source.api.rst
```diff
@@ -0,0 +1,6 @@
+``GET /v3/interaction`` and ``GET /v3/interaction/<uid>``: The following field was added:
+
+* ``source`` - object, nullable - object representing the external source of
+  an interaction, e.g. a calendar invite
+
+This field is read only.
```
marcofucci: Why does the frontend need this field?

currycoder: Good point.  It doesn't for any of the work that is currently in scope, I suppose I didn't know what our approach should be in this case.  Should what we expose be on a "need to know" basis in general?

changelog/interaction/source.db.rst
currycoder: Good spot.  copypasta.

currycoder:
> So, null as the empty value seems OK to me (and nothing would stop us changing our minds later on). Are there any plans for the front end to use `source`?

Thanks for thinking about this.  I'll leave it as null for now.  There are no plans for frontend to use source, and as per the comment from @marcofucci I might just remove it from the API entirely.

PR #1629, opened 2019-04-16
Convert None values for address fields to empty strings in company API v3

marcofucci:
### Description of change

**Before**: Using None values in POST/PATCH company v3 API would set the model fields to None.

**After**: None values are now converted to empty strings so that we can make them NOT NULL in the future.

### How to test

`POST /v3/company` using
```json
{
    "business_type": "98d14e94-5d95-e211-a939-e4115bead28a",
    "name": "Acme",
    "registered_address_1": "75 Stramford Road",
    "registered_address_2": null,
    "registered_address_country": "80756b9a-5d95-e211-a939-e4115bead28a",
    "registered_address_county": null,
    "registered_address_postcode": null,
    "registered_address_town": "London",
    "sector": "73535406-6095-e211-a939-e4115bead28a",
    "trading_address_1": null,
    "trading_address_2": null,
    "trading_address_country": null,
    "trading_address_county": null,
    "trading_address_postcode": null,
    "trading_address_town": null,
    "uk_region": "8a4cd12a-6095-e211-a939-e4115bead28a"
}
```

The following response fields should get converted to `''`
```
    'registered_address_2',
    'registered_address_county',
    'registered_address_postcode',
    'trading_address_1',
    'trading_address_2',
    'trading_address_town',
    'trading_address_county',
     'trading_address_postcode',
```

### Checklist

### DIFF
--- a/changelog/company/convert_None_to_blank.api.rst
+++ b/changelog/company/convert_None_to_blank.api.rst
@@ -1,0 +1,1 @@
+``POST /v3/company`` and ``PATCH /v3/company/<uuid:pk>``: None values for address CharFields are now internally converted to empty strings as Django recommends: https://docs.djangoproject.com/en/2.1/ref/models/fields/#null
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -586,6 +586,15 @@ class CompanySerializerV3(BaseCompanySerializer):
 
         return data
 
+    def _get_address_field_names_in_data(self, data):
+        all_address_field_names = {
+            field
+            for mapping in self.ADDRESS_FIELDS_MAPPING.values()
+            for field in mapping.values()
+        }
+
+        return all_address_field_names & data.keys()
+
     def _populate_address_fields(self, combiner, data):
         """
         Populates the address_* fields with the values from trading address or
@@ -596,14 +605,8 @@ class CompanySerializerV3(BaseCompanySerializer):
         Doing this will allow us to implement a variant that updates the addresses
         in a different way.
         """
-        all_address_field_names = {
-            field
-            for mapping in self.ADDRESS_FIELDS_MAPPING.values()
-            for field in mapping.values()
-        }
-
         # was any address field specified?
-        if not all_address_field_names & data.keys():
+        if not self._get_address_field_names_in_data(data):
             return
 
         trading_fields_mapping = Company.TRADING_ADDRESS_VALIDATION_MAPPING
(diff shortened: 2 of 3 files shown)


datahub/company/serializers.py
```diff
@@ -621,7 +624,41 @@ def _populate_address_fields(self, combiner, data):
             target_value = combiner.get_value(source_field_name)
             data[target_field_name] = target_value
 
+    def run_validation(self, data=serializers.empty):
+        """
+        Converts all None address values into their default (if default != None).
```
currycoder: To double check the intention here...

This method converts API calls coming in with address fields set as `null` to the default value set on the field?  I had to re-read the logic of this method a few times to understand what it was doing, perhaps the docstring could be a little more detailed that this is converting incoming API requests to use default values?

changelog/company/convert_None_to_blank.api.rst
```diff
@@ -0,0 +1 @@
+``POST /v3/company`` and ``PATCH /v3/company/<uuid:pk>``: None values for address charfields are now internally converted to empty strings.
```
currycoder: Could you perhaps give some business context on why this is needed?  An alternative here would be to make the frontend explicitly set these defaults instead of adding some dynamic rewrites to the API layer.

datahub/company/serializers.py
```diff
@@ -621,7 +624,41 @@ def _populate_address_fields(self, combiner, data):
             target_value = combiner.get_value(source_field_name)
             data[target_field_name] = target_value
 
+    def run_validation(self, data=serializers.empty):
```
reupen: Should we handle the `data=serializers.empty` case since it's the default value?

reupen:
I wonder if we're better off using our own CharField subclass to make the logic more generic (the standard one could be swapped out with an overridden ModelSerializer field type mapping).

datahub/company/serializers.py
marcofucci: Yes that's the intention; that's because it's a django recommends to _avoid using null on string-based fields such as CharField and TextField_  and we have null=True everywhere. https://docs.djangoproject.com/en/2.1/ref/models/fields/#null

changelog/company/convert_None_to_blank.api.rst
marcofucci: 👍 I don't think we should trust input coming from the frontend (it can be any frontend) but I'll try expand on this in the fragment.

marcofucci:
> I wonder if we're better off using our own CharField subclass to make the logic more generic (the standard one could be swapped out with an overridden ModelSerializer field type mapping).

@reupen I was thinking about this as well. It's not really needed in this case as v3 will be gone soon and the v4 already doesn't allow None values.
It would be better to wait for next time we need it otherwise I'll create dead code. It's also a bit delicate and probably bigger so it needs a bit more time.

datahub/company/serializers.py
marcofucci: Well spotted! I've added some code to handle that but I'm not sure how to trigger that in practice

PR #1632, opened 2019-04-16
Add file size limit to interaction admin CSV import

reupen:
### Description of change

This adds a file size limit to the interaction CSV import tool. This is because we only need to process small files and aren't set up to handle large files.

The check is done in a file upload handler to stop large temp files from being written (which could cause us to run out of disk space – if it is done in the form it is too late).

If the limit is hit, an error message is displayed at the top of the page using the messages framework.

I haven't added a news fragment for this as the feature is still behind a feature flag.

#### Limit displayed as help text

image

#### Error message

image

### To test

1. Navigate to http://localhost:8000/admin/interaction/interaction/import
2. Try to upload a file bigger than 2 MB. You should see an error message
3. Try to upload a file smaller than 2 MB. There should be no error message

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the 
 | `GUNICORN_WORKER_CLASS`  | No | [Type of Gunicorn worker.](http://docs.gunicorn.org/en/stable/settings.html#worker-class) Uses async workers via gevent by default. |
 | `GUNICORN_WORKER_CONNECTIONS`  | No | Maximum no. of connections for async workers (default=10). |
 | `HAWK_RECEIVER_IP_WHITELIST` | No | IP addresses (comma-separated) that can access the Hawk-authenticated endpoints. |
+| `INTERACTION_ADMIN_CSV_IMPORT_MAX_SIZE` | No | Maximum file size in bytes for interaction admin CSV uploads (default=2MB). |
 | `INVESTMENT_DOCUMENT_AWS_ACCESS_KEY_ID` | No | Same use as AWS_ACCESS_KEY_ID, but for investment project documents. |
 | `INVESTMENT_DOCUMENT_AWS_SECRET_ACCESS_KEY` | No | Same use as AWS_SECRET_ACCESS_KEY, but for investment project documents. |
 | `INVESTMENT_DOCUMENT_AWS_REGION` | No | Same use as AWS_DEFAULT_REGION, but for investment project documents. |
--- a/datahub/core/admin_csv_import.py
+++ b/datahub/core/admin_csv_import.py
@@ -17,6 +17,7 @@ class BaseCSVImportForm(forms.Form):
     )
 
     csv_file_field_label = 'CSV file'
+    csv_file_field_help_text = None
     required_columns = set()
 
     csv_file = forms.FileField(
@@ -27,6 +28,7 @@ class BaseCSVImportForm(forms.Form):
         """Initialises the form, dynamically setting the label of the csv_file field."""
         super().__init__(*args, **kwargs)
         self.fields['csv_file'].label = self.csv_file_field_label
+        self.fields['csv_file'].help_text = self.csv_file_field_help_text
 
     def clean_csv_file(self):
         """Validates the uploaded CSV file and creates a CSV DictReader from it."""
--- a/datahub/interaction/test/views/test_admin_csv_import.py
+++ b/datahub/interaction/test/views/test_admin_csv_import.py
@@ -1,6 +1,8 @@
 import io
 
 import pytest
+from django.conf import settings
+from django.contrib import messages as django_messages
 from django.contrib.admin.templatetags.admin_urls import admin_urlname
 from django.test import Client
 from django.urls import reverse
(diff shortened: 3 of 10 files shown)


datahub/interaction/test/views/test_admin_csv_import.py
```diff
@@ -146,6 +147,30 @@ def test_does_not_allow_file_without_correct_columns(
             'adviser_1, contact_email, date, kind, service.',
         ]
 
+    def test_rejects_large_files(self, interaction_importer_feature_flag):
+        """
+        Test that large files are rejected.
+
+        Note: INTERACTION_ADMIN_CSV_IMPORT_MAX_SIZE is set to 1024 in config.settings.test
+        """
+        file = io.BytesIO(b'-' * 1025)
```
marcofucci: Why not `INTERACTION_ADMIN_CSV_IMPORT_MAX_SIZE + 1`

datahub/core/test/test_admin_csv_import.py
```diff
@@ -90,3 +94,45 @@ def test_does_not_allow_files_with_invalid_utf8(self, file_contents):
 
         assert 'csv_file' in form.errors
         assert form.errors['csv_file'] == ['There was an error decoding the file contents.']
+
+
+class TestMaxUploadSize:
+    """Tests BaseCSVImportForm."""
```
marcofucci: docstring needs updating I suppose

datahub/core/test/test_admin_csv_import.py
```diff
@@ -90,3 +94,45 @@ def test_does_not_allow_files_with_invalid_utf8(self, file_contents):
...
+
+    @pytest.mark.parametrize(
+        'file_size,error_expected',
+        (
+            (50, False),
```
marcofucci: It's not very obvious why 50, it could be clearer to use a constant in `datahub/core/test/support/views.py` and import that here.

datahub/core/admin_csv_import.py
```diff
@@ -1,11 +1,71 @@
...
+def max_upload_size(max_size):
+    """View decorator to enforce a maximum size on uploads."""
+    def decorator(view_func):
+        @csrf_exempt
+        @wraps(view_func)
+        def wrapped_view(request, *args, **kwargs):
+            # Note: request.upload_handlers must be manipulated before CSRF-protection
+            # routines run
+            request.upload_handlers.insert(0, MaxSizeFileUploadHandler(request, max_size))
+            return csrf_protect(view_func)(request, *args, **kwargs)
```
marcofucci: This has the side effect that if a developer uses `@csrf_exempt` in the view and this decorator as well, this decorator would ignore the `csrf_exempt` in a non-obvious way. I guess it could be okay if you document it.

Another option would be to always add MaxSizeFileUploadHandler to `settings.FILE_UPLOAD_HANDLERS` and have a flag set by the decorator to enable it (maybe something like `request.enabled_max_upload_size`). I don't know if that works well so I'll leave it to you.

reupen: `csrf_protect` respects `csrf_exempt` so it should be OK if the decorators are applied in the correct order, right? I have updated the docstring.

I think we need to think about the global behaviour as well, we should have some protection on all views I think. Perhaps something to look at separately from this?

README.md
```diff
@@ -290,6 +290,7 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the
 | `GUNICORN_WORKER_CLASS`  | No | [Type of Gunicorn worker.](http://docs.gunicorn.org/en/stable/settings.html#worker-class) Uses async workers via gevent by default. |
 | `GUNICORN_WORKER_CONNECTIONS`  | No | Maximum no. of connections for async workers (default=10). |
 | `HAWK_RECEIVER_IP_WHITELIST` | No | IP addresses (comma-separated) that can access the Hawk-authenticated endpoints. |
+| `INTERACTION_ADMIN_CSV_IMPORT_MAX_SIZE` | No | Maximum file size in bytes for interaction admin CSV uploads (default=2MB). |
```
currycoder: Could a sensible default for this be added to `sample.env`?

currycoder: Oh I see that it's optional...  Disregard!

datahub/core/admin_csv_import.py
```diff
@@ -1,11 +1,81 @@
...
+            django_messages.error(self.request, error_msg)
+
+            raise SkipFile()
+        return raw_data
+
+    def file_complete(self, file_size):
+        """Does nothing, so that the next handler processes the upload."""
+
+
+def max_upload_size(max_size):
```
currycoder: I feel like this is general purpose enough to be in a different module TBH.  It's not really specific to CSV imports and may be useful for other file uploads.  I was surprised that django didn't offer this functionality as a built-in file handler.

reupen: I was seeing it as specific to the admin site – there's a fair bit in `datahub.core.admin` now, we could turn it into a sub-package.

reupen: @currycoder I moved it to `datahub.core.admin`. I'll leave splitting `datahub.core.admin` up as a separate task.

PR #1633, opened 2019-04-16
Make NestedRelatedField accept model instances

reupen:
### Description of change

This makes NestedRelatedField accept model instances. In general, DRF serialiser fields accept internal values, so this helps maintain consistency with other fields.

The reason for the change is that I may use the serialiser to create interactions as part of the interaction import tool (to avoid duplicating any logic e.g. DIT participant creation and defaults).

### Checklist

### DIFF
--- a/datahub/core/serializers.py
+++ b/datahub/core/serializers.py
@@ -67,6 +67,7 @@ class NestedRelatedField(serializers.RelatedField):
         'does_not_exist': 'Invalid pk "{pk_value}" - object does not exist.',
         'incorrect_type': 'Incorrect type. Expected object, received {'
                           'data_type}.',
+        'unsaved_object': 'Model instances must be saved.',
     }
 
     def __init__(self, model, extra_fields=('name',), **kwargs):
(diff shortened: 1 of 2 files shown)


datahub/core/serializers.py
```diff
@@ -98,6 +99,11 @@ def get_queryset(self):
 
     def to_internal_value(self, data):
         """Converts a user-provided value to a model instance."""
+        if isinstance(data, self._model):
+            if data._state.adding:
```
currycoder: Why did you choose this way instead of `if not data.pk`?

I'm a little wary about accessing non-public attributes.

reupen: Having a value for `pk` doesn't mean that it's been saved (if you instantiate a Company it will already have a value for `pk` as we give it a default).

`_state.adding` is partially documented here: https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.from_db

Although it is `_`-prefixed, that seems to be more of a Django convention to avoid conflicts with field names (for example, there is `_meta` as well).

reupen:
I'm not sure that I'm going to be able to use the serialiser for the import functionality anyway, so I'm going to close this for now.

PR #1634, opened 2019-04-17
Simplify docker working copy startup to one command

currycoder:
### Description of change

Running through a stack of commands from the README each time my docker-compose project was brought up was getting tedious.

I've made the `Dockerfile` run these commands instead (and silenced any failures if the DB is already in a loaded state).  The Dockerfile also creates a superuser for you if you have `SUPERUSER_USERNAME` and `SUPERUSER_PASSWORD` environment variables set.  This reduces the docker readme from 8 steps to 4 and gets us to a point where there is only one meaningful command needed to bring up a working docker-compose project for the API.

### Points of note

- Restarts of a running development container are slower - as there are more things to run through to ensure that the working copy has a workable data set.
- The automatic superuser creation is a bit hacky, but I thought that some hardcoded python was preferable to making/adding a dependency for a whole extra django command.  Especially since this is something we only want to run for local docker copies.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Leeloo uses Docker compose to setup and run all the necessary components. The do
 3.  Build and run the necessary containers for the required environment:
 
     ```shell
-    docker-compose up -d
+    docker-compose up
     ```
 
     * It will take time for the leeloo API container to come up - it will run
--- a/README.md
+++ b/README.md
@@ -29,22 +29,16 @@ Leeloo uses Docker compose to setup and run all the necessary components. The do
 3.  Build and run the necessary containers for the required environment:
 
     ```shell
-    docker-compose build
+    docker-compose up -d
     ```
 
-4.  Populate the database and initialise Elasticsearch:
-
-    ```shell
-    docker-compose run leeloo ./manage.py migrate
-    docker-compose run leeloo ./manage.py init_es
-    docker-compose run leeloo ./manage.py loadinitialmetadata
-    docker-compose run leeloo ./manage.py createinitialrevisions
-    ```
-
+    * It will take time for the leeloo API container to come up - it will run
+      migrations on both DBs, load initial data, sync elasticsearch etc. Watch
+      along in the api container's logs.
     * **NOTE:** 
-      If you are using a linux system, these commands may hang on 
-      waiting for the elasticsearch container to come up (`data-hub-leeloo_es_1`) - 
-      it might be perpetually restarting.
+      If you are using a linux system, the  elasticsearch container may not 
+      come up successfully (`data-hub-leeloo_es_1`) - it might be perpetually 
+      restarting.
       If the logs for that container mention something like 
       `max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]`,
       you will need to run the following on your host machine:
@@ -62,26 +56,7 @@ Leeloo uses Docker compose to setup and run all the necessary components. The do
 
       For more information, [see the elasticsearch docs on vm.max_map_count](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/vm-max-map-count.html).
 
-5. Optionally, you can load some test data and update elasticsearch:
-
-    ```shell
-    docker-compose run leeloo ./manage.py loaddata /app/fixtures/test_data.yaml
-    docker-compose run leeloo ./manage.py sync_es
-    ```
-
-6.  Create a superuser:
-
-    ```shell
-    docker-compose run leeloo ./manage.py createsuperuser
-    ```
-
-7.  Run the services:
-
-    ```shell
-    docker-compose up
-    ```
-
-8.  Optionally, you may want to run a local copy of the data hub frontend. 
+4.  Optionally, you may want to run a local copy of the data hub frontend. 
     By default, you can run both leeloo and the frontend under one docker-compose
     project.  [See the instructions in the frontend readme to set it up](https://github.com/uktrade/data-hub-frontend/#setting-up-with-docker-compose).
 
--- a/datahub/core/management/commands/createsuperuserwithpassword.py
+++ b/datahub/core/management/commands/createsuperuserwithpassword.py
@@ -22,12 +22,12 @@ class Command(BaseCommand):
         parser.add_argument(
             'username',
             type=str,
-            help='The superuser\'s username',
+            help="The superuser's username",
         )
         parser.add_argument(
             'password',
             type=str,
-            help='The superuser\'s password',
+            help="The superuser's password",
         )
 
     def handle(self, *args, **options):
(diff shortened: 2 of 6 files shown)


start.sh
```diff
@@ -2,7 +2,24 @@
 python /app/manage.py migrate --noinput
 python /app/manage.py migrate --database mi --noinput
 python /app/manage.py init_es
+# Load initial metadata - ignore errors as we may have already loaded it in to
+# this DB
+python /app/manage.py loadinitialmetadata || true
+# Load initial revisions - ignore errors as we may have already loaded it in to
+# this DB
+python /app/manage.py createinitialrevisions || true
+python /app/manage.py loaddata /app/fixtures/test_data.yaml
```
reupen: Probably want to load `test_ch_data.yaml` as well (as `setup-uat.sh`)

start.sh
```diff
@@ -2,7 +2,24 @@
...
+python /app/manage.py sync_es
 python /app/manage.py collectstatic --noinput
+# Create superuser - ignore errors as we may have already loaded it in to
+# this DB.
+if [[ -n "${SUPERUSER_USERNAME}" && -n "${SUPERUSER_PASSWORD}" ]]; then 
+    # Unfortunately, we have to be a bit hacky here as the standard 
+    # django createsuperuser command does not allow password to be specified
+    # programmatically
+    echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('$SUPERUSER_USERNAME', '$SUPERUSER_PASSWORD')" | python manage.py shell || true
```
reupen: Can the Python code be spread over multiple lines? (There's an example in `setup-uat.sh`.)

README.md
```diff
@@ -29,22 +29,16 @@ Leeloo uses Docker compose to setup and run all the necessary components. The do
...
-    ```
-
-4.  Populate the database and initialise Elasticsearch:
-
-    ```shell
-    docker-compose run leeloo ./manage.py migrate
-    docker-compose run leeloo ./manage.py init_es
-    docker-compose run leeloo ./manage.py loadinitialmetadata
-    docker-compose run leeloo ./manage.py createinitialrevisions
+    docker-compose up -d
```
reupen: I noticed the addition of `-d` – not sure everyone will want to run it in detached mode.

start.sh
```diff
@@ -2,7 +2,24 @@
...
 python /app/manage.py migrate --database mi --noinput
 python /app/manage.py init_es
+# Load initial metadata - ignore errors as we may have already loaded it in to
+# this DB
+python /app/manage.py loadinitialmetadata || true
+# Load initial revisions - ignore errors as we may have already loaded it in to
+# this DB
+python /app/manage.py createinitialrevisions || true
+python /app/manage.py loaddata /app/fixtures/test_data.yaml
+python /app/manage.py sync_es
```
reupen: Not sure if `sync_es` is really necessary – I thought the ES syncing of the objects created by loading fixtures would happen via Celery anyway.

reupen:
Is there any scope to share logic with `setup-uat.sh`?

start.sh
```diff
@@ -2,7 +2,24 @@
...
 python /app/manage.py collectstatic --noinput
+# Create superuser - ignore errors as we may have already loaded it in to
+# this DB.
+if [[ -n "${SUPERUSER_USERNAME}" && -n "${SUPERUSER_PASSWORD}" ]]; then 
+    # Unfortunately, we have to be a bit hacky here as the standard 
+    # django createsuperuser command does not allow password to be specified
```
marcuspp: I'm not a fan of python code like this in shell scripts as you can't reuse it or re-run it without running everything else in the script. Could you not create a command that wraps createsuperuser and call that instead?

README.md
marcuspp: I always run with `-d` then `docker-compose logs -f` the result is the same but if you kill that tail then the containers are still running.

currycoder:
> Is there any scope to share logic with `setup-uat.sh`?

I've had a look at how we would split out the common logic in to a shared script and honestly it gets a bit messy.  The main reason is that certain errors need to be suppressed for running on local docker copies (where data already exists etc) - but we want errors to be flagged up when running the test suites.

So you end up having to write a script wrapper function and things get especially problematic when wrapping the big chunk of embedded python in `setup-uat.sh`.  I think we should probably look at cleaning up both of these scripts a bit in a wider body of work.  Where should I log that?

start.sh
currycoder: Nice - you're right!

start.sh
currycoder: I agree with you, but TBH I didn't want to include something that makes it easier to create a superuser directly in the project's python code.  Especially since this is something we only want to run for local docker copies.

marcuspp: I can't see the harm in creating a wrapped version of `createsuperuser` it doesn't make it easier because if I wanted a super user I just would use `createsuperuser` directly. It is just a way to get python code out of the shell script.

marcuspp:
Sorting `setup-uat.sh` out would be great. I have just tried rebuilding and helping a new starter get up and running and not being able to load the test advisers makes things difficult and getting the python code out of the script would be a bonus. Happy if this is out of scope for this PR though.

> > Is there any scope to share logic with `setup-uat.sh`?
> 
> I've had a look at how we would split out the common logic in to a shared script and honestly it gets a bit messy. The main reason is that certain errors need to be suppressed for running on local docker copies (where data already exists etc) - but we want errors to be flagged up when running the test suites.
> 
> So you end up having to write a script wrapper function and things get especially problematic when wrapping the big chunk of embedded python in `setup-uat.sh`. I think we should probably look at cleaning up both of these scripts a bit in a wider body of work. Where should I log that?

currycoder:
This is ready for re-review.  I've created a new `createsuperuserwithpassword` command as per your suggestions @marcuspp 
For now, I'd prefer to avoid reworking the start/start-uat scripts as discussed.  I think it would mean a lot of scope creep for what I'm trying to achieve by iterating the usability of docker copies.

datahub/core/management/commands/createsuperuserwithpassword.py
```diff
@@ -0,0 +1,39 @@
...
+            help='The superuser\'s password',
+        )
+
+    def handle(self, *args, **options):
+        """
+        Handle invocation of the command.
+        """
+        username = options['username']
+        password = options['password']
+        user = User.objects.create_superuser(username, password)
```
marcuspp: Could you add a catch for an integrity error? Just incase the user already exists.

start.sh
```diff
@@ -2,7 +2,21 @@
 python /app/manage.py migrate --noinput
 python /app/manage.py migrate --database mi --noinput
 python /app/manage.py init_es
+# Load initial metadata - ignore errors as we may have already loaded it in to
+# this DB
+python /app/manage.py loadinitialmetadata || true
+# Load initial revisions - ignore errors as we may have already loaded it in to
+# this DB
+python /app/manage.py createinitialrevisions || true
```
reupen: Don't think there would be much to create initial revisions for at this point (for a new database). So it would probably be better to run this after `test_data.yaml` has been loaded (and the super user created actually).

Also it should be safe to run multiple times, so we probably wouldn't want `|| true` here.

currycoder: :+1:

PR #1635, opened 2019-04-17
Add columns to DnBMatchingCSVRecord to store selected match candidate

elcct:
### Description of change

This adds new columns to the `DnBMatchingCSVRecord` so that we can store which company match candidate has been or has not been selected, why, when and by whom.

If `selected_duns_number` is null, it means the adviser cannot find a candidate from the given list and should specify a reason for that (`no_match_reason` column). The front end should provide with a selection of typical reasons. 
The `no_match_description` would be used if adviser has other reason than typical or when she/he thinks that multiple companies could match - in that case, that field should contain comma separated duns numbers of possible candidates - there is no validation for that yet.

I tried to keep is as simple as possible and most logic for selecting candidates will be in the front end.

There is basic validation on save, so that user cannot select company that doesn't exist in the list of candidates etc.
We don't expect for the list of matches to change. In case we need a new set of matches, it will be uploaded with new `batch_number`.

There will be a PR with API endpoints coming.

### Checklist


datahub/dnb_match/models.py
```diff
@@ -30,11 +32,102 @@ class DnBMatchingCSVRecord(models.Model):
...
+        null=True, blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+    )
+    selected_on = models.DateTimeField(null=True, blank=True)
+
+    no_match_reason = models.CharField(
+        blank=True,
+        null=True,
+        help_text='Reason why does not match.',
```
currycoder: More readable: "The reason that a match could not be found."

datahub/dnb_match/models.py
```diff
@@ -30,11 +32,102 @@ class DnBMatchingCSVRecord(models.Model):
...
+
+    selected_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        null=True, blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+    )
+    selected_on = models.DateTimeField(null=True, blank=True)
+
+    no_match_reason = models.CharField(
```
currycoder: Could it be better to make this a choice field..?  What possible reasons are we expecting?  By leaving it as free-text, I don't expect the data quality to be great/unified.

datahub/dnb_match/models.py
```diff
@@ -30,11 +32,102 @@ class DnBMatchingCSVRecord(models.Model):
...
+        blank=True,
+        null=True,
+        help_text='Reason why does not match.',
+        max_length=settings.CHAR_FIELD_MAX_LENGTH,
+    )
+    no_match_description = models.TextField(blank=True, null=True)
+
+    def select_match_candidate(
```
currycoder: I feel like it could be more readable to make `select_match_candidate` only handle the case where a match is found and split out the "no match" in to a separate method: `record_no_match()` or something.

datahub/dnb_match/models.py
elcct: Indeed! Thank you

marcofucci:
@elcct I'm not really sure we need this.
The `DnBMatchingResult` has a `data` field with currently the following output:
```
{
    'dnb_match': {
        'name': 'lorem',
        'country': 'ENGLAND',
        'duns_number': '123456789',
        'global_ultimate_name': '',
        'global_ultimate_country': '',
        'global_ultimate_duns_number': ''
    },
    'matched_by': 'data-science'
}
```

When a user selects a candidate we could just add a `DnBMatchingResult` with the `dnb_match` details obtained from `DnBMatchingCSVResult` and with a `matched_by: adviser` or something like that. In case of no match we could have another format similar to the one you suggested in this PR.
It's slightly more fragile as the format is not as rigid as the one imposed by columns but it reuses what we already have and makes it obvious where things are.

What do you think?

elcct:
@marcofucci 
I see your point, and I was thinking about creating a record in `DnBMatchingResult` table once user makes a selection in the following API PR. 

The purpose of `DnBMatchingCSVRecord` is now to support the UI in case you need to edit the match etc. and maybe the name of the model doesn't reflect that.
I can see that we could achieve this using `DnBMatchingResult` too, but that seems to me a little bit messier, so I wanted to keep it separate.

So when user selects a match and it gets "commited" to the `DnBMatchingResult` table, we still get to reuse what we have.

datahub/dnb_match/models.py
elcct: I have added a choice of the reasons we currently have. I agree it will be better.

datahub/dnb_match/models.py
elcct: Fair point. The no match part was a bit awkward there so I split it. Thanks!

elcct:
I gave it more thought and I think now that @marcofucci idea is better, so I am going to close this one. I have reused some bits of that one in another PR that will come.

I wanted to self-contain the match within the given batch of candidates, but that doesn't really give any benefits, apart from slightly more defined data model. 
If I use `DnBMatchingResult` to present the match information in the API response, it will enable user to see matches done by machine and also edit them and this solution doesn't allow that.

PR #1644, opened 2019-04-23
Add company matching API.

elcct:
### Description of change

This adds the API for manual company matching.

There are three endpoints available:
`GET /v4/dnb-match/<company_pk>`: gets information about the result of matching and available candidates. The result is stored in `DnBMatchingResult` table and the listed candidates come from `DnBMatchingCSVRecord` record of highest `batch_number`.

`POST /v4/dnb-match/<company_pk>/select-match`: updates the `DnBMatchingResult` with selected candidate data under `dnb_match` key. It accepts `duns_number` in the POST body.

`POST /v4/dnb-match/<company_pk>/select-no-match`: updates `DnBMatchingResult` with information about not being able to find a match under `no_match` key. 

Both POST endpoints also store the adviser information under `adviser` key and mark the information source as `adviser`.

### Checklist

### DIFF
--- a/datahub/dnb_match/serializers.py
+++ b/datahub/dnb_match/serializers.py
@@ -1,0 +1,168 @@
+from django.utils.translation import gettext_lazy
+from rest_framework import serializers
+from rest_framework.settings import api_settings
+
+from datahub.dnb_match.constants import NoMatchReason
+from datahub.dnb_match.models import DnBMatchingResult
+from datahub.dnb_match.utils import _get_list_of_latest_match_candidates
+
+
+class SelectMatchingCandidateSerializer(serializers.Serializer):
+    """DRF serializer for company match candidate selection."""
+
+    duns_number = serializers.ChoiceField(choices=[])
+
+    def __init__(self, *args, **kwargs):
+        """Initialise the serializer with candidates data."""
+        super().__init__(*args, **kwargs)
+
+        company = self.context['company']
+
+        self.context['candidates'] = _get_list_of_latest_match_candidates(company.pk)
+        self.fields['duns_number'].choices = self._candidates_to_choices()
+
+    def save(self):
+        """Save matched candidate data into DnBMatchingResult."""
+        for candidate in self.context['candidates']:
+            if self.validated_data['duns_number'] == str(candidate.get('duns_number')):
+                data = self._candidate_to_result(
+                    candidate,
+                    self.context['request'].user,
+                )
+
+                DnBMatchingResult.objects.update_or_create(
+                    company=self.context['company'],
+                    defaults={
+                        'data': data,
+                    },
+                )
+
+    def _candidates_to_choices(self):
+        return [
+            (str(candidate['duns_number']), candidate['name'])
+            for candidate in self.context['candidates']
+        ]
+
+    @staticmethod
+    def _candidate_to_result(candidate, adviser):
+        """Transform matching candidate to matching result."""
+        fields = (
+            'duns_number',
+            'name',
+            'address_country',
+            'global_ultimate_duns_number',
+            'global_ultimate_name',
+            'global_ultimate_country',
+        )
+        remap_key = {
+            'address_country': 'country',
+        }
+        dnb_match = {
+            remap_key.get(key, key): value for key, value in candidate.items() if key in fields
+        }
+
+        return {
+            'dnb_match': dnb_match,
+            'matched_by': 'adviser',
+            'adviser': {
+                'id': adviser.id,
+                'first_name': adviser.first_name,
+                'last_name': adviser.last_name,
+                'name': adviser.name,
+            },
+        }
+
+
+class SelectNoMatchSerializer(serializers.Serializer):
+    """DRF serializer for no match candidate selection."""
+
+    default_error_messages = {
+        'list_of_candidates_required': gettext_lazy(
+            'List of candidates is required if the reason is "more_than_one".',
+        ),
+        'description_required': gettext_lazy(
+            'The "description" is required if the reason is "other".',
+        ),
+        'too_many_fields': gettext_lazy(
+            'If the reason is "not_listed" or "not_confident", other fields '
+            'should not need to be filled.',
+        ),
+    }
+
+    reason = serializers.ChoiceField(choices=NoMatchReason)
+
+    description = serializers.CharField(required=False)
+    candidates = serializers.ListField(child=serializers.CharField(), required=False)
+
+    def save(self):
+        """Save no match information to DnBMatchingResult."""
+        data = self._get_request_data_to_result(self.context['request'].user)
+
+        DnBMatchingResult.objects.update_or_create(
+            company=self.context['company'],
+            defaults={
+                'data': data,
+            },
+        )
+
+    def validate(self, data):
+        """Check if correct fields are filled in."""
+        self._check_reason_more_than_one(data)
+        self._check_reason_other(data)
+        self._check_reason_not_listed_or_not_confident(data)
+        return data
+
+    def _check_reason_more_than_one(self, data):
+        """Check if the reason "more_than_one" has a list of candidates."""
+        if data['reason'] == NoMatchReason.more_than_one and not data.get('candidates'):
+            error = {
+                'candidates': [
+                    self.error_messages['list_of_candidates_required'],
+                ],
+            }
+            raise serializers.ValidationError(error, code='list_of_candidates_required')
+
+    def _check_reason_other(self, data):
+        """Check if the reason "other" has a description."""
+        if data['reason'] == NoMatchReason.other and not data.get('description'):
+            error = {
+                'description': [
+                    self.error_messages['description_required'],
+                ],
+            }
+            raise serializers.ValidationError(error, code='description_required')
+
+    def _check_reason_not_listed_or_not_confident(self, data):
+        """Check if the reason "not_listed" or "not_confident" does not have other fields."""
+        reason_not_listed_or_not_confident = data['reason'] in (
+            NoMatchReason.not_listed, NoMatchReason.not_confident,
+        )
+        other_fields_than_reason = data.keys() != {'reason'}
+
+        if reason_not_listed_or_not_confident and other_fields_than_reason:
+            error = {
+                api_settings.NON_FIELD_ERRORS_KEY: [
+                    self.error_messages['too_many_fields'],
+                ],
+            }
+            raise serializers.ValidationError(error, code='too_many_fields')
+
+    def _get_request_data_to_result(self, adviser):
+        no_match = {
+            'reason': self.validated_data['reason'],
+        }
+        if self.validated_data['reason'] == NoMatchReason.more_than_one:
+            no_match['candidates'] = self.validated_data['candidates']
+        if self.validated_data['reason'] == NoMatchReason.other:
+            no_match['description'] = self.validated_data['description']
+
+        return {
+            'no_match': no_match,
+            'matched_by': 'adviser',
+            'adviser': {
+                'id': adviser.id,
+                'first_name': adviser.first_name,
+                'last_name': adviser.last_name,
+                'name': adviser.name,
+            },
+        }
--- a/datahub/dnb_match/test/test_views.py
+++ b/datahub/dnb_match/test/test_views.py
@@ -1,0 +1,458 @@
+"""Tests for matching information views."""
+
+from uuid import uuid4
+
+import pytest
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.company.test.factories import CompanyFactory
+from datahub.core.test_utils import APITestMixin
+from datahub.core.test_utils import create_test_user
+from datahub.dnb_match.test.factories import DnBMatchingCSVRecord, DnBMatchingResultFactory
+from datahub.dnb_match.views import _replace_dnb_country_fields
+
+
+def _create_match_result_selected():
+    """Create match result test dictionary."""
+    return {
+        'dnb_match': {
+            'duns_number': '111',
+            'name': 'NAME OF A COMPANY',
+            'country': 'United Kingdom',
+            'global_ultimate_duns_number': '112',
+            'global_ultimate_name': 'NAME OF A GLOBAL COMPANY',
+            'global_ultimate_country': 'United Kingdom',
+        },
+        'matched_by': 'data-science',
+    }
+
+
+def _create_match_result_no_match(reason, description=None, candidates=None):
+    """Create no match result test dictionary."""
+    no_match = {
+        'reason': reason,
+    }
+    if description:
+        no_match['description'] = description
+    if candidates:
+        no_match['candidates'] = candidates
+
+    return {
+        'no_match': no_match,
+        'matched_by': 'adviser',
+    }
+
+
+class TestMatchingInformationView(APITestMixin):
+    """Tests for the matching information view."""
+
+    def test_access_is_denied_if_without_permissions(self):
+        """Test that a 403 is returned if the user has no permissions."""
+        company = CompanyFactory()
+        user = create_test_user(permission_codenames=[])
+        api_client = self.create_api_client(user=user)
+
+        url = reverse(
+            'api-v4:dnb-match:item',
+            kwargs={
+                'company_pk': company.pk,
+            },
+        )
+
+        response = api_client.get(url)
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_cannot_get_match_information_of_company_that_doesnt_exist(self):
+        """Test that a 404 is returned if company doesn't exist."""
+        url = reverse(
+            'api-v4:dnb-match:item',
+            kwargs={
+                'company_pk': uuid4(),
+            },
+        )
+
+        response = self.api_client.get(url)
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    @pytest.mark.parametrize('result', (
+        _create_match_result_selected(),
+        _create_match_result_no_match('not_listed'),
+        _create_match_result_no_match('more_than_one', candidates=['111', '222', '333']),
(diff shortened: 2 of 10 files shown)


changelog/dnb-matchin-api.api
```diff
@@ -0,0 +1,20 @@
+New API endpoints were added to aid matching Data Hub companies with D&B companies:
```
reupen: Could you add .rst to the file name so that we can preview the file on GitHub?

datahub/dnb_match/serializers.py
```diff
@@ -0,0 +1,38 @@
...
+
+    reason = serializers.ChoiceField(choices=NoMatchReason)
+
+    description = serializers.CharField(required=False)
+    candidates = serializers.ListField(child=serializers.CharField(), required=False)
+
+    def validate(self, data):
+        """Check if correct fields are filled in."""
+        if data['reason'] == NoMatchReason.more_than_one and not data.get('candidates'):
+            raise serializers.ValidationError(
```
reupen: We should set the error messages on the `default_error_messages` like other serialisers e.g. https://github.com/uktrade/data-hub-leeloo/blob/c099465f235432549cd54a295040f58dd2afac4a/datahub/event/serializers.py#L14 (except use `gettext_lazy` instead of `ugettext_lazy`) and then access them using `self.error_messages`.

datahub/dnb_match/serializers.py
```diff
@@ -0,0 +1,38 @@
...
+                'List of candidates is required if the reason is "more_than_one".',
```
reupen: It looks like we could assign this error to the field that becomes required.

datahub/dnb_match/serializers.py
```diff
@@ -0,0 +1,38 @@
...
+            )
+        if data['reason'] == NoMatchReason.other and not data.get('description'):
+            raise serializers.ValidationError(
+                'The "description" is required if the reason is "other".',
+            )
+        if all([
```
reupen: The if condition looks a bit complicated, are we able to split it up? https://docs.python-guide.org/writing/style/#one-statement-per-line

datahub/dnb_match/serializers.py
```diff
@@ -0,0 +1,38 @@
...
+            data['reason'] in (NoMatchReason.not_listed, NoMatchReason.not_confident),
+            set(data.keys()) != {'reason'},
```
reupen: The set conversion can be skipped here.

datahub/dnb_match/views.py
```diff
@@ -0,0 +1,166 @@
...
+        try:
+            return Company.objects.get(pk=company_pk)
+        except Company.DoesNotExist:
+            raise Http404
+
+    def _get_matching_information(self, request, company, **kwargs):
+        """Get matching candidates and current selection."""
+        try:
+            record = DnBMatchingResult.objects.get(company=company)
+            matching_result = record.data
```
marcofucci: This can be `company.dnbmatchingresult.data` but your version is okay as well.

datahub/dnb_match/views.py
```diff
@@ -0,0 +1,166 @@
...
+from rest_framework.views import APIView
+
+from datahub.company.models import Company
+from datahub.dnb_match.models import DnBMatchingCSVRecord, DnBMatchingResult
+from datahub.dnb_match.serializers import (
+    SelectMatchingCandidateSerializer,
+    SelectNoMatchSerializer,
+)
+from datahub.oauth.scopes import Scope
+from datahub.search.dict_utils import company_dict, contact_or_adviser_dict
```
marcofucci: I know this is just temporary so I think this is minor but it looks weird that this is using a search package. It could potentially be replaced by the django `django.forms.models.model_to_dict`? Not sure if it's even better 🤔

datahub/dnb_match/views.py
```diff
@@ -0,0 +1,166 @@
...
+            if str(duns_number) == str(candidate.get('duns_number')):
+                data = self._candidate_to_result(candidate, request.user)
+
+                DnBMatchingResult.objects.update_or_create(
+                    company=company,
+                    defaults={
+                        'data': data,
+                    },
+                )
+                break
```
marcofucci: I think in general it would be better if all this validation and save logic was on the serializer.
The duns_number could be a ChoiceField with duns_number as key so that `serializer.is_valid` also validates that the duns_number is in the selection.

datahub/dnb_match/test/test_views.py
```diff
@@ -0,0 +1,413 @@
...
+        url = reverse(
+            'api-v4:dnb-match:item',
+            kwargs={
+                'company_pk': company.pk,
+            },
+        )
+        response = self.api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.data
```
marcofucci: It should be `response.json()` I think ;-)

datahub/dnb_match/test/test_views.py
```diff
@@ -0,0 +1,413 @@
...
+
+        response = self.api_client.get(url)
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    @pytest.mark.parametrize('result', (
+        _create_match_result_selected(),
+        _create_match_result_no_match('not_listed'),
+        _create_match_result_no_match('more_than_one', candidates=['111', '222', '333']),
+        _create_match_result_no_match('not_confident'),
+        _create_match_result_no_match('other', 'These candidates are out of place.'),
```
marcofucci: Maybe `_create_match_result_no_match('other', description='These candidates are out of place.'),` is better

changelog/dnb-matching-api.api.rst
```diff
@@ -0,0 +1,20 @@
...
+
+``reason`` and depending on reason - ``description`` or ``candidates``
+
+Key ``candidates`` contains a list of potential candidates, each having following fields:
+
+``duns_number``, ``name``, ``global_ultimate_duns_number``, ``global_ultimate_name``, ``global_ultimate_country``, ``address_1``, ``address_2``, ``address_town``, ``address_postcode``, ``address_country``, ``confidence``, ``source``
+
+Key ``matched_by`` contains information who / what performed the match (e.g. adviser or data-science)
+
+Finally ``adviser`` key contains ``id``, ``name`` information about adviser performing the match
```
marcofucci: I would rewrite this, the endpoints return `result` which is a dict with keys = `'dnb_match', 'no_match', 'matched_by', 'adviser'` plus `candidates`.

datahub/dnb_match/test/test_views.py
```diff
@@ -0,0 +1,413 @@
...
+        response_data = response.data
+        assert 'result' in response_data
+        assert response_data['result'] == {
+            'dnb_match': {
+                'name': 'test name',
+                'duns_number': 12345,
+                'country': {
+                    'id': '81756b9a-5d95-e211-a939-e4115bead28a',
+                    'name': 'United States',
+                },
```
marcofucci: It's a bit weird that this country is a Data Hub country as a dict whilst the `global_ultimate_country` is the raw D&B string, maybe we can keep this as a D&B string as well instead of converting? I don't think we lose anything.

elcct: The problem is that the upload command that stores the match candidates resolves D&B string to Data Hub country. I would have to resolve it back to D&B string or just support both formats? If it is a string convert to DH and if it is a dict leave it. I could also convert `global_ultimate_country` to DH dict?

datahub/dnb_match/serializers.py
elcct: I have changed the messages now. Thanks!

datahub/dnb_match/serializers.py
elcct: That is true. Updated it. Thanks!

datahub/dnb_match/serializers.py
elcct: I agree. It is split now. Thanks!

datahub/dnb_match/serializers.py
elcct: True. Updated. Thanks!

datahub/dnb_match/test/test_views.py
elcct: True. Fixed. Thanks!

datahub/dnb_match/views.py
elcct: You're right. I have moved all this logic to serialiser now. Thanks!

datahub/dnb_match/test/test_views.py
elcct: Yes! Fixed. Thanks!

datahub/dnb_match/views.py
elcct: I realised that mode_to_dict wouldn't copy computed fields, so I just inlined the dictionary for adviser. It seems small enough to be okay. I used `model_to_dict` for company dict though. So now there is no dependency on search. Thanks!

datahub/dnb_match/views.py
elcct: Updated. Thanks!

elcct:
@currycoder I think that is something we need to live with for now, as we didn't design what to do in such situation. I see the problem if one adviser gives correct answer and then another in the meantime fills in that a match cannot be found erasing the correct answer. I will include that in the documentation for future development. Hopefully we can communicate that this form should be used only if you are certain about the answer to minimise this.

changelog/dnb-matching-api.api.rst
elcct: I have rewritten the news fragment using your suggestion. Not sure if it is too vague now :)

changelog/dnb-matching-api.api.rst
```diff
@@ -0,0 +1,9 @@
+New API endpoints were added to aid matching Data Hub companies with D&B companies:
+
+All endpoints return a result which is a dict with keys = ``dnb_match``, ``no_match``, ``matched_by``, ``adviser`` plus ``candidates``.
+
+``GET /v4/dnb-match/<company_pk>`` returns a result described above.
+
+``POST /v4/dnb-match/<company_pk>/select-match`` accepts ``duns_number`` of a candidate to be selected as a match.
+
+``POST /v4/dnb-match/<company_pk>/select-no-match`` accepts ``reason`` that can be one of ``not_listed``, ``more_than_one``, ``not_confident``, ``other`` and ``description`` or ``candidates`` (as a list of DUNS numbers) where applicable.
```
marcofucci: Yeah I think you are right, it's still a bit vague; I would be better to be very explicit. What do you think about this? (change as you want)

New API endpoints were added to aid matching Data Hub companies with D&B companies:

All endpoints return a response body with the following format:
```
{
    'result': {
        ...
    },
    'candidates': [
        { ... },
        { ... }
    ],
    company: {
        'id': '81756b9a-5d95-e211-a939-e4115bead28a',
        'name': 'My Corp',
        'trading_names': ['trading name']
    }
}
```

The value of ``result`` depends on the type of match.

If a match was found and recorded:

```
{
    'dnb_match': {
        'duns_number': '111',
        'name': 'NAME OF A COMPANY',
        'country': 'United Kingdom',
        'global_ultimate_duns_number': '112',
        'global_ultimate_name': 'NAME OF A GLOBAL COMPANY',
        'global_ultimate_country': 'United Kingdom',
    },
    'matched_by': 'data-science'
},
```

If a match wasn't found because the company isn't listed or the adviser is not confident to make the match:
```
{
    'no_match': {
        'reason': 'not_listed',  # or 'not_confident'
    },
    'matched_by': 'adviser'
},
```

If a match wasn't found because there were multiple potential matches:
```
{
    'no_match': {
        'reason': 'more_than_one',
        'candidates': [  # list of duns numbers
            '123456789',
            '987654321'
        ]
    },
    'matched_by': 'adviser'
},
```

If a match wasn't found because of other reasons:
```
{
    'no_match': {
        'reason': 'other',
        'description': 'explaination...',
    },
    'matched_by': 'adviser'
},
```

The top level ``candidates`` is a list of objects with this format:

```
{
    'duns_number': 12345,
    'name': 'test name',
    'global_ultimate_duns_number': 12345,
    'global_ultimate_name': 'test name global',
    'global_ultimate_country': 'USA',
    'address_1': '1st LTD street',
    'address_2': '',
    'address_town': 'London',
    'address_postcode': 'SW1A 1AA',
    'address_country': {
        'id': '81756b9a-5d95-e211-a939-e4115bead28a',
        'name': 'United States',
    },
    'confidence': 10,
    'source': 'cats',
}
```

Endpoints:

``GET /v4/dnb-match/<company_pk>`` returns the response above

``POST /v4/dnb-match/<company_pk>/select-match`` accepts the ``duns_number`` of the candidate to be selected as a match from the list of candidates

``POST /v4/dnb-match/<company_pk>/select-no-match`` accepts ``reason`` with value:

- ``not_listed``: if none of the candidates is a good match
- ``not_confident``: if the adviser is not confident to make the match
- ``more_than_one``: if there are multiple potential matches. In this case an extra ``candidates`` field is required with the list of valid duns numbers.
- ``other``: for other reasons. In this case an extra freetext ``description`` field is required

datahub/dnb_match/views.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            matching_result = company.dnbmatchingresult.data
+        except ObjectDoesNotExist:
+            matching_result = {}
+
+        try:
+            candidates = DnBMatchingCSVRecord.objects.get(company_id=company.pk)
+            candidates_data = candidates.data or []
+        except DnBMatchingCSVRecord.DoesNotExist:
+            candidates_data = []
```
marcofucci: Shouldn't you use something similar to the logic defined by `_get_latest_match_candidates` in the serializer?

elcct: That code obviously is wrong! I have moved the method that get the candidates to `utils` and made it return just a list of candidates as the `DnBMatchingCSVRecord` itself is not being used anyway. Thanks!

changelog/dnb-matching-api.api.rst
elcct: Wow this is amazing! I have added it as a newsfragment and edited a bit. Also added information about the `country` issue (that is it can be either a string or a dictionary). 
Not sure about the best way to solve it. What about always returning the dictionary? Maybe I wouldn't touch the data in the database but added a simple filter in the response to change DnB country to a dict on the fly?

marcofucci: Yeah I think that making them consistent would be better if it doesn't add too much work. See if you can convert the global one to a dict on the fly but if it's too much work we can document it as leave it as it is as this API is just temporary anyway.

changelog/dnb-matching-api.api.rst
```diff
@@ -0,0 +1,140 @@
+New API endpoints were added to aid matching Data Hub companies with D&B companies:
+
+All endpoints return a response body with the following format:
+```
```
reupen: I'm not sure this is valid reStructuredText – I think I'd have expected something more like this (for the code block):

```
New API endpoints were added to aid matching Data Hub companies with D&B companies:

All endpoints return a response body with the following format::

    {
        "result": {
            ...
        },
        "candidates": [
            { ... },
            { ... }
        ],
        "company": {
            "id": "81756b9a-5d95-e211-a939-e4115bead28a",
            "name": 'My Corp',
            "trading_names": ["trading name"]
        }
    }
```

changelog/dnb-matching-api.api.rst
```diff
@@ -0,0 +1,140 @@
+New API endpoints were added to aid matching Data Hub companies with D&B companies:
```
reupen: Should the news fragment be under company?

datahub/dnb_match/serializers.py
```diff
@@ -0,0 +1,168 @@
...
+            raise serializers.ValidationError(error, code='description_required')
+
+    def _check_reason_not_listed_or_not_confident(self, data):
+        """Check if the reason "not_listed" or "not_confident" does not have other fields."""
+        reason_not_listed_or_not_confident = data['reason'] in (
+            NoMatchReason.not_listed, NoMatchReason.not_confident,
+        )
+        other_fields_than_reason = data.keys() != {'reason'}
+
+        if all((reason_not_listed_or_not_confident, other_fields_than_reason)):
```
reupen: Can this be a simple `if reason_not_listed_or_not_confident and other_fields_than_reason:`?

datahub/dnb_match/serializers.py
```diff
@@ -0,0 +1,168 @@
...
+                    self.error_messages['too_many_fields'],
+                ],
+            }
+            raise serializers.ValidationError(error, code='too_many_fields')
+
+    def _get_request_data_to_result(self, adviser):
+        no_match = {
+            'reason': self.validated_data['reason'],
+        }
+        if self.validated_data['reason'] == 'more_than_one':
```
reupen: Are we able to use `NoMatchReason.more_than_one` here and in similar lines?

changelog/dnb-matching-api.api.rst
elcct: I added a function to the view that looks up the fields with countries in the response and replace them if applicable.

changelog/dnb-matching-api.api.rst
elcct: Nice one! I updated the news fragment and checked with `.rst` viewer. Thanks.

changelog/dnb-matching-api.api.rst
elcct: I think it could as well be under company. I was a bit split where to put it. Moved to company. Thanks!

datahub/dnb_match/serializers.py
elcct: hahaha yes! Thanks!

datahub/dnb_match/serializers.py
elcct: Yes, that should use `NoMatchReason`. Updated. Thanks!

PR #1649, opened 2019-04-23
Filter company autocomplete by country

marcuspp:
### Description of change
For large capital profiles we would like to make it easier for the user to find the company they would like to set a profile up for using typeahead search To do this it was decided to filter the company typeahead by country (if provided).

### To Test
-  Rebuild your search index ( `./manage.py migrate_es` you may need to set `CELERY_TASK_ALWAYS_EAGER=True` in your local.py settings file.
- In your browser http://localhost:8000/v4/search/company/autocomplete?term=one&format=json
- You should see 2 companies returned
- Then add france to the search http://localhost:8000/v4/search/company/autocomplete?term=one&country=82756b9a-5d95-e211-a939-e4115bead28a&format=json
- You should see 1 company that is based in france

NB. The DRF Browseable view for company autocomplete seems to be broken (no longer has a queryset defined) on develop so format=json is required to test.

### Checklist

### DIFF
--- a/datahub/search/company/models.py
+++ b/datahub/search/company/models.py
@@ -6,6 +6,7 @@ from elasticsearch_dsl import Boolean, Completion, Date, Keyword, Text
 from datahub.search import dict_utils
 from datahub.search import fields
 from datahub.search.models import BaseESModel
+from datahub.search.utils import get_unique_values_and_exclude_nulls_from_list
 
 
 DOC_TYPE = 'company'
@@ -13,7 +14,8 @@ DOC_TYPE = 'company'
 
 def get_suggestions(db_company):
     """
-    A list of fields used by the completion suggester to
+    Returns a dictionary with the keys input and context.
+    input contains a list of fields used by the completion suggester to
     find a record when using an autocomplete search.
 
     https://www.elastic.co/guide/en/elasticsearch/
@@ -28,9 +30,15 @@ def get_suggestions(db_company):
 
     Optional weighting could be added here to boost particular suggestions.
     See above link.
+
+    contexts contains a dictionary with any supported filters.
+
+    https://www.elastic.co/guide/en/elasticsearch/reference/current/suggester-context.html
+
+    context - country a list of UUIDs of the countries where the company is based.
     """
     if db_company.archived:
-        return []
+        return {}
 
     names = [
         db_company.name,
@@ -44,7 +52,17 @@ def get_suggestions(db_company):
         *names,
     ]
 
-    return list(filter(None, set(data)))
+    countries = [
+        db_company.registered_address_country_id,
+        db_company.address_country_id,
+    ]
+
+    return {
+        'input': get_unique_values_and_exclude_nulls_from_list(data),
+        'contexts': {
+            'country': get_unique_values_and_exclude_nulls_from_list(countries),
+        },
+    }
 
 
 class Company(BaseESModel):
@@ -110,7 +128,14 @@ class Company(BaseESModel):
     vat_number = Keyword(index=False)
     duns_number = Keyword()
     website = Text()
-    suggest = Completion()
+    suggest = Completion(
+        contexts=[
+            {
+                'name': 'country',
+                'type': 'category',
+            },
+        ],
+    )
 
     COMPUTED_MAPPINGS = {
         'suggest': get_suggestions,
--- a/datahub/search/company/test/test_models.py
+++ b/datahub/search/company/test/test_models.py
@@ -3,6 +3,7 @@ from collections import Counter
 import pytest
 
 from datahub.company.test.factories import CompanyFactory
+from datahub.core.constants import Country as CountryConstant
 from datahub.search.company.models import Company as ESCompany, get_suggestions
 
 pytestmark = pytest.mark.django_db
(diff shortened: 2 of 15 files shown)


datahub/search/views.py
```diff
@@ -294,17 +295,30 @@ def list(self, request, *args, **kwargs):
...
 
+    def get_search_context(self, query_params):
+        """
+        Add context (filters) to the autocomplete search if provided.
+        """
+        if not self.autocomplete_context_serializer_class:
+            return {}
+        serializer = self.autocomplete_context_serializer_class(data=query_params)
+        serializer.is_valid()
+        inner_context = serializer.validated_data
```
currycoder: My understanding here is that this will use all validated data from a serializer.  Say that in the future we wanted to offer a few different fields to provide context...  e.g. for Company...
`business_type`, `one_list_tier`, `sector`

These would all need to be fields on the autocomplete context serializer, right?

A few follow up questions if this is the case:
1) If the user only selects the country context and the other contexts have null values, would that be handled by the query correctly here?
2) Should we be able to combine these contexts?  e.g. OR/AND

It feels to me like there might need to be an extra layer of logic between the context serializer and building the final context, with these considerations.

datahub/search/company/models.py
```diff
@@ -11,9 +11,20 @@
 DOC_TYPE = 'company'
 
 
+def get_unique_values_and_exclude_nulls_from_list(data):
```
elcct: Should these functions be moved to something like `utils.py`?

datahub/search/company/models.py
```diff
@@ -11,9 +11,20 @@
...
+    """
+    :param data: a list of values
+    :return: a list of none empty unique values
```
marcofucci: non-empty instead of none empty

datahub/search/views.py
marcuspp: Good spot. I've now added `required=False` to the `country` field in the serializer so now all cases should be supported. I tested adding a second context field and both cases should now be supported. The `inner_context` will be just what has been passed in and validated.

datahub/search/company/test/test_models.py
```diff
@@ -87,28 +88,31 @@ def test_company_dbmodels_to_es_documents(self, setup_es):
                     'Good', 'uk', 'Hello Hello uk',
                     'us', 'Good Hello us', 'fr', 'Hello',
                 ],
+                {'country': [CountryConstant.united_kingdom.value.id]},
```
elcct: I think it would be great to pass at least two countries when testing a list.

marcuspp:
> @marcuspp trading_address is going. Can you please use `address` fields and `registered_address` fields instead?

@marcofucci  I have updated the code so it longer uses `trading_address`

datahub/search/views.py
```diff
@@ -294,17 +295,30 @@ def list(self, request, *args, **kwargs):
...
+        return {
+            'context': inner_context,
+        } if inner_context else {}
```
marcofucci: Small thing but I would find:
```python
if not inner_context:
    return {}

return {
    'context': inner_context,
}
```

slightly easier to read.

datahub/search/test/test_query_builder.py
```diff
@@ -326,11 +329,55 @@ def test_build_entity_permission_query_no_conditions(filters, expected):
...
+            ['id', 'name'],
+            {'country': 'hello'},
+            {
+                '_source': {'includes': ['id', 'name']},
+                'suggest': {
+                    'autocomplete': {
+                        'completion': {
+                            'field': 'suggest',
+                            'size': 1,
+                            'country': 'hello',
```
marcofucci: Shouldn't this be `"contexts"=...`?

datahub/search/execute_query.py
```diff
@@ -9,13 +9,14 @@
 logger = getLogger(__name__)
 
 
-def execute_autocomplete_query(es_model, keyword_search, limit, fields_to_include):
+def execute_autocomplete_query(es_model, keyword_search, limit, fields_to_include, context):
```
marcofucci: Not sure we have appropriate unit tests for this.

datahub/search/query_builder.py
```diff
@@ -123,7 +123,11 @@ def build_autocomplete_query(es_model, keyword_search, limit, fields_to_include)
     return autocomplete_search.suggest(
         'autocomplete',
         keyword_search,
-        completion={'field': 'suggest', 'size': limit},
+        completion={
+            'field': 'suggest',
+            'size': limit,
+            **context,
```
marcofucci: I think it would be better to have ```'contexts': context``` here instead of constructing the dict with context field in `get_search_context`?

datahub/search/company/test/test_entity_search_views_v3.py
```diff
@@ -627,45 +627,96 @@ def test_no_permissions_returns_403(self):
...
+            'registered_address_1': company.registered_address_1,
+            'registered_address_2': company.registered_address_2,
+            'registered_address_town': company.registered_address_town,
+            'registered_address_county': company.registered_address_county,
+            'registered_address_postcode': company.registered_address_postcode,
+            'registered_address_country': {
+                'id': str(company.registered_address_country_id),
+                'name': company.registered_address_country.name,
+            },
+            'trading_names': company.trading_names,
```
marcofucci: Maybe keep this file as it was so that we don't get many merge conflicts as it's going to be deleted soon.

datahub/search/query_builder.py
marcuspp: Unfortunately I can't do this as this causes an error for any suggestion field does not support contexts. Supplying a contexts key with an empty dictionary/None causes `suggester [suggest] doesn't expect any context` to be raised. I will add a comment to the code that details this.

PR #1654, opened 2019-04-24
Initial work for email ingestion

currycoder:
### Description of change
This adds a `Mailbox` and `EmailProcessor` class under a new `email_ingestion` package.

`Mailbox` acts as a gateway for an IMAP email inbox and allows convenience methods for getting new "unread" mail and providing hooks for processing that mail.
`EmailProcessor` is an abstract superclass which provides a pattern for any email processing business logic that we need.

The overall aspiration is to tie these components together, such that the datahub backend can monitor multiple IMAP mailboxes and a mailbox can have one or more concrete email processor classes for processing incoming mail.  Right now, this PR is concerned with providing the building blocks in a way that is well tested.

### Points of note

- The tests for `Mailbox` are fairly mock-heavy.  I've used a mixture of testing the end-to-end logic of public methods (with IMAP/mail processing mocked out) and some unit tests.  I did debate making wrapper classes for IMAP/mail processing in a way that we could use dependency injection to swap out the concrete classes, but I decided against this: it adds more layers and I can't see us needing to swap out these libraries in future.  It probably would have made the tests cleaner, however.

### simplejson
The failing test case for investment/GVA searches  has now been fixed.

Since we now have simplejson as a dependency, py-elasticsearch will use that when serializing/deserializing JSON for the elasticsearch API.  This means that any floats that we are indexing to elasticsearch have the potential to be more accurate.  

To use the data indexed for the GVA test case as an example, here is some sample code for the differences between py-elasticsearch using `json` and `simplejson` for serialization:
```
import json
import simplejson
from decimal import Decimal

from elasticsearch.serializer import JSONSerializer

gva_multiplier = Decimal('9.999999')
foreign_equity_investment = 9999999999999999999
# A decimal value
gva = foreign_equity_investment * gva_multiplier

# Simulate py-elasticsearch serialization using json
json_result = json.dumps(gva, default=JSONSerializer().default)
# Simulate py-elasticsearch serialization using simplejson
simplejson_result = simplejson.dumps(gva, default=JSONSerializer().default)

print("json result:")
print(json_result)
print("simplejson result:")
print(simplejson_result)
```
Yields the following output:
```
json result:
9.999998999999999e+19
simplejson result:
99999989999999999990.000001
```

**The implication here is that when we deploy this, floats across the codebase which are saved to elasticsearch will become more accurate.  We should discuss whether we need to fully rebuild the ES indexes because of this.**

### Checklist

### DIFF
--- a/datahub/email_ingestion/email_processor.py
+++ b/datahub/email_ingestion/email_processor.py
@@ -24,9 +24,4 @@ class EmailProcessor(ABC):
 
             (False, "The email was not sent by a known DIT address")
         """
-        raise NotImplementedError(
-            (
-                'EmailProcessor subclass %s does not implement `process_email` '
-                'method'
-            ) % self.__class__.__name__,
-        )
+        pass
--- a/datahub/email_ingestion/email_processor.py
+++ b/datahub/email_ingestion/email_processor.py
@@ -1,0 +1,32 @@
+from abc import ABC, abstractmethod
+
+
+class EmailProcessor(ABC):
+    """
+    EmailProcessor subclasses should take a message and perform some actions
+    based on it's metadata/content in order to process it.  EmailProcessor is
+    an abstract superclass.
+    """
+
+    @abstractmethod
+    def process_email(self, message):
+        """
+        Perform business logic to process a `mailparser.Message` instance - this
+        may involve interacting with the DB or other services as necessary.
+
+        :param message: mailparser.Message object - the email message to process
+
+        :returns: Tuple of boolean (whether the message was successfully processed
+            by this processor) and string (a meaningful message explaining why
+            this was not processed or explaining that this was processed
+            successfully).
+            e.g.
+
+            (False, "The email was not sent by a known DIT address")
+        """
+        raise NotImplementedError(
+            (
+                'EmailProcessor subclass %s does not implement `process_email` '
+                'method'
+            ) % self.__class__.__name__,
+        )
--- a/datahub/email_ingestion/email_processor.py
+++ b/datahub/email_ingestion/email_processor.py
@@ -24,4 +24,3 @@ class EmailProcessor(ABC):
 
             (False, "The email was not sent by a known DIT address")
         """
-        pass
--- a/datahub/email_ingestion/email_processor.py
+++ b/datahub/email_ingestion/email_processor.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 class EmailProcessor(ABC):
     """
     EmailProcessor subclasses should take a message and perform some actions
-    based on it's metadata/content in order to process it.  EmailProcessor is
+    based on its metadata/content in order to process it.  EmailProcessor is
     an abstract superclass.
     """
 
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -125,6 +125,19 @@ class Mailbox:
     def _parse_message(self, message_bytes):
         return mailparser.parse_from_bytes(message_bytes)
 
+    def _get_message(self, uid, server):
+        typ, msg_contents = server.uid('fetch', uid, '(RFC822)')
+        if not msg_contents:
+            return None
+        try:
+            message = self._parse_message(msg_contents[0][1])
+        except TypeError:
+            # This may happen if another thread/process deletes the
+            # message between our generating the ID list and our
+            # processing it here.
+            raise
+        return message
+
     def get_new_mail(self):
         """
         Generator method which gets new messages from the email inbox. After a
@@ -141,16 +154,9 @@ class Mailbox:
 
             for uid in message_ids:
                 try:
-                    typ, msg_contents = server.uid('fetch', uid, '(RFC822)')
-                    if not msg_contents:
+                    message = self._get_message(uid, server)
+                    if not message:
                         continue
-                    try:
-                        message = self._parse_message(msg_contents[0][1])
-                    except TypeError:
-                        # This may happen if another thread/process deletes the
-                        # message between our generating the ID list and our
-                        # processing it here.
-                        raise
                     yield message
                 except (MessageParseError, MailParserError):
                     # If we have some problem parsing the email, it's likely
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -40,7 +40,6 @@ class Mailbox:
         imap_domain,
         mail_processor_classes,
         imap_port=None,
-        use_ssl=True,
     ):
         """
         Initialise a Mailbox object.
@@ -51,8 +50,6 @@ class Mailbox:
         :param mail_processor_classes: iterable - EmailProcessor classes which
             should be used to process incoming mail to this mailbox
         :param imap_port: optional int - the port to use when connecting with imap
-        :param use_ssl: optional boolean - whether or not to use SSL with imap, defaults
-            to True
         """
         self.email = email
         self.password = password
@@ -60,11 +57,7 @@ class Mailbox:
         if imap_port:
             self.imap_port = imap_port
         else:
-            if use_ssl:
-                self.imap_port = 993
-            else:
-                self.imap_port = 465
-        self.use_ssl = use_ssl
+            self.imap_port = 993
         # Make a copy of the processor class iterable
         self.processor_classes = [processor_class for processor_class in mail_processor_classes]
 
@@ -76,11 +69,7 @@ class Mailbox:
 
         :yields: An active imaplib server connection object.
         """
-        if self.use_ssl:
-            transport = imaplib.IMAP4_SSL
-        else:
-            transport = imaplib.IMAP4
-        connection = transport(self.imap_domain, self.imap_port)
+        connection = imaplib.IMAP4_SSL(self.imap_domain, self.imap_port)
         connection.login(self.email, self.password)
         connection.select()
         try:
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -1,0 +1,173 @@
+import imaplib
+from contextlib import contextmanager
+from email.errors import MessageParseError
+from logging import getLogger
+
+import mailparser
+from mailparser.exceptions import MailParserError
+
+logger = getLogger(__name__)
+
+
+class Mailbox:
+    """
+    The Mailbox class acts as a gateway to an email inbox - accessed via IMAP.
+    It offers methods for getting new messages from the inbox and processing
+    them using associated processing classes.
+
+    When an email is retrieved from the inbox successfully, it is marked as
+    deleted on the upstream IMAP inbox.  This is the mechanism we use to ensure
+    that emails are only ever processed once.
+
+    The first processor to successfully consume and process the message will stop
+    the chain.  Processors will be called in order - so ordering of
+    `mail_processor_classes` is important.
+    """
+
+    def __init__(
+        self,
+        email,
+        password,
+        imap_domain,
+        mail_processor_classes,
+        imap_port=None,
+        use_ssl=True,
+    ):
+        """
+        Initialise a Mailbox object.
+
+        :param email: string - the email address of the inbox
+        :param password: string - the password for the inbox
+        :param imap_domain: string - the domain of the imap server to connect to
+        :param mail_processor_classes: iterable - EmailProcessor classes which
+            should be used to process incoming mail to this mailbox
+        :param imap_port: optional int - the port to use when connecting with imap
+        :param use_ssl: optional boolean - whether or not to use SSL with imap, defaults
+            to True
+        """
+        self.email = email
+        self.password = password
+        self.imap_domain = imap_domain
+        if imap_port:
+            self.imap_port = imap_port
+        else:
+            if use_ssl:
+                self.imap_port = 993
+            else:
+                self.imap_port = 465
+        self.use_ssl = use_ssl
+        self.processor_classes = []
+        for processor_class in mail_processor_classes:
+            self.processor_classes.append(processor_class)
+
+    @contextmanager
+    def _connect(self):
+        """
+        Connect to the email inbox over IMAP.  The connection is closed down
+        automatically when execution leaves the scope of the context.
+
+        :yields: An active imaplib server connection object.
+        """
+        transport = imaplib.IMAP4_SSL
+        if not self.use_ssl:
+            transport = imaplib.IMAP4
+        server = transport(self.imap_domain, self.imap_port)
+        server.login(self.email, self.password)
+        server.select()
+        try:
+            yield server
+        finally:
+            server.close()
+            server.logout()
+
+    def _get_all_message_ids(self, server):
+        """
+        Fetch all of the message ids that are present in the inbox.
+
+        :param server: imaplib connection object - the active connection to
+            use to retrieve message ids from
+
+        :returns: A list of message id strings.
+        """
+        # Fetch all the message uids
+        response, message_ids = server.uid('search', None, 'ALL')
+        message_id_string = message_ids[0].strip()
+        # Usually `message_id_string` will be a list of space-separated
+        # ids; we must make sure that it isn't an empty string before
+        # splitting into individual UIDs.
+        if message_id_string:
+            return message_id_string.decode().split(' ')
+        return []
+
+    def _process_email(self, message):
+        """
+        Run through the EmailProcessor objects associated with this Mailbox and
+        attempt to process the message.
+
+        The message will be processed by at most one processor; EmailProcessor
+        objects will be called in the order that they were passed in on
+        Mailbox instantiation. The method returns as soon as the message has been
+        successfully processed by a processor.
+
+        :param message: mailparser.Mailparser object - the message to process
+
+        :returns: True if the message was processed, False otherwise.
+        """
+        for processor_class in self.processor_classes:
+            processor = processor_class()
+            processed, message = processor.process_email(message)
+            if processed:
+                processor_name = processor_class.__name__
+                message = f'Email {message} was processed by processor: {processor_name}'
+                logger.info(message)
+                return True
+        return False
+
+    def _parse_message(self, message_bytes):
+        return mailparser.parse_from_bytes(message_bytes)
+
+    def get_new_mail(self):
+        """
+        Generator method which gets new messages from the email inbox. After a
+        message has been yielded, it is flagged as deleted on the inbox so that
+        it will not be ingested again later.
+
+        :yields: A mailparser.Message object for each parsed message.
+        """
+        with self._connect() as server:
+            message_ids = self._get_all_message_ids(server)
+            if not message_ids:
+                # No new messages to ingest
+                return
+
+            for uid in message_ids:
+                try:
+                    typ, msg_contents = server.uid('fetch', uid, '(RFC822)')
+                    if not msg_contents:
+                        continue
+                    try:
+                        message = self._parse_message(msg_contents[0][1])
+                    except TypeError:
+                        # This may happen if another thread/process deletes the
+                        # message between our generating the ID list and our
+                        # processing it here.
+                        raise
+                    yield message
+                except (MessageParseError, MailParserError):
+                    # If we have some problem parsing the email, it's likely
+                    # to be spam/malicious so skip it
+                    continue
+
+                # Mark the email for deletion in the inbox
+                server.uid('store', uid, '+FLAGS', '(\\Deleted)')
+            # Carry out deletion for the emails marked for deletion
+            server.expunge()
+
+    def process_new_mail(self):
+        """
+        Gets all of the new mail in the inbox and goes through the associated
+        EmailProcessor classes to process each message.
+        """
+        messages = self.get_new_mail()
+        for message in messages:
+            self._process_email(message)
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -131,9 +131,9 @@ class Mailbox:
             processor_name = processor_class.__name__
             try:
                 processed, processing_output = processor.process_email(message)
-            except Exception as exc:
+            except Exception:
                 error_message = (
-                    f'Error "{exc.__class__.__name__}" processing email "{message.message_id}" '
+                    f'Error processing email "{message.message_id}" '
                     f'which was processed by processor "{processor_name}"'
                 )
                 logger.exception(error_message)
@@ -145,11 +145,10 @@ class Mailbox:
                 )
                 logger.info(success_message)
                 return True
-            else:
-                logger.debug(
-                    f'Email {message.message_id} could not be processed by {processor_name}. '
-                    f'Reason: "{processing_output}"',
-                )
+            logger.debug(
+                f'Email {message.message_id} could not be processed by {processor_name}. '
+                f'Reason: "{processing_output}"',
+            )
         return False
 
     def _parse_message(self, message_bytes):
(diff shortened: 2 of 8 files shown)


datahub/email_ingestion/email_processor.py
```diff
@@ -0,0 +1,32 @@
...
+
+        :returns: Tuple of boolean (whether the message was successfully processed
+            by this processor) and string (a meaningful message explaining why
+            this was not processed or explaining that this was processed
+            successfully).
+            e.g.
+
+            (False, "The email was not sent by a known DIT address")
+        """
+        raise NotImplementedError(
```
marcofucci: I don't think you need `NotImplementedError` as abstractmethods cannot be called so I think this could never be executed.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+        if imap_port:
+            self.imap_port = imap_port
+        else:
+            if use_ssl:
+                self.imap_port = 993
+            else:
+                self.imap_port = 465
+        self.use_ssl = use_ssl
+        self.processor_classes = []
+        for processor_class in mail_processor_classes:
```
marcofucci: Is this to keep a copy of the list? You could use `list(mail_processor_classes)` or maybe `[processor_class for processor_class in mail_processor_classes]`?

datahub/email_ingestion/test/test_mailbox.py
```diff
@@ -0,0 +1,258 @@
...
+        Helper contextmanager to patch IMAP4_SSL which is used in the Mailbox
+        class. We can specify email details returned by our mocked instance by
+        providing sample email messages.
+
+        :param email_messages: iterable of email message dictionaries with keys
+            uid (the email uid) and message_content (a string of message payload)
+        :yields: A mock.Mock object which has been patched in place of the IMAP4_SSL
+            object in the datahub.email_ingestion.mailbox module
+        """
+        with mock.patch('datahub.email_ingestion.mailbox.imaplib.IMAP4_SSL') as mocked_imap_class:
```
marcofucci: We tend to use the pytest [https://docs.pytest.org/en/latest/monkeypatch.html](monkeypatch) and pytest fixtures for these kind of setups but 👍

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+        :yields: A mailparser.Message object for each parsed message.
+        """
+        with self._connect() as server:
+            message_ids = self._get_all_message_ids(server)
+            if not message_ids:
+                # No new messages to ingest
+                return
+
+            for uid in message_ids:
+                try:
```
reupen: There is a lot of indentation here, so it would be good to move the loop body to another method (or similar).

datahub/email_ingestion/mailbox.py
currycoder: Good spot!  This is some left over cruft from a refactor

datahub/email_ingestion/email_processor.py
currycoder: Neat.

datahub/email_ingestion/test/test_mailbox.py
currycoder: Ok - will bear these in mind in future.

currycoder:
> A couple of questions on the design:
> 
> 1. what happens if there's an exception in an EmailProcessor.process_email? it would be nice to add a test for that

Good point - I'll introduce that change.

> 2. Have you thought about race conditions? If there are 2 `Mailbox` instances parsing the same message at the same time, it could be processed twice.

On the prototype (which I am incrementally cleaning up/feeding in as PRs), this race condition is handled by the celery task which orchestrates the mailboxes.  The task initially acquires a lock (this will now be a postgres advisory lock thanks to Reupen's comments) before manipulating the mailboxes to ingest new mail.  

I could instead move the lock acquisition in to the `Mailbox` class and have a lock per mailbox.  This does guarantee that race conditions are handled by any code that accesses mailboxes, but it makes the error handling and overall logic a little bit heavier than just acquiring a lock for the whole task.  My preference would be to keep things simple and keep the lock acquisition on the task only, but I'm happy to discuss it further @marcofucci - what do you think?

datahub/email_ingestion/mailbox.py
currycoder: Good idea, will do.

datahub/email_ingestion/test/test_mailbox.py
reupen: It would be good to cut down on the indentation in the module in any case. The main thing is the use of this context manager. I can think of a couple of alternatives:

- pytest fixture and monkeypatch as @marcofucci mentioned
- use `patch_imap` as a decorator instead of as a context manager

marcofucci:
@currycoder ok with the lightweight version but worth documenting that a lock must be acquired when using the class to avoid race conditions

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+    @contextmanager
+    def _connect(self):
+        """
+        Connect to the email inbox over IMAP.  The connection is closed down
+        automatically when execution leaves the scope of the context.
+
+        :yields: An active imaplib server connection object.
+        """
+        transport = imaplib.IMAP4_SSL
+        if not self.use_ssl:
```
reupen: I would suggest either a one-liner `transport = imaplib.IMAP4_SSL if self.use_ssl else imaplib.IMAP4` or an if/else i.e.:

```python
if self.use_ssl:
    transport = imaplib.IMAP4_SSL
else:
    transport = imaplib.IMAP4
```

(rather than assigning one value to `transport` and then overwriting it).

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+            transport = imaplib.IMAP4
+        server = transport(self.imap_domain, self.imap_port)
```
reupen: `server` might be a bit confusing as I'd have thought this was a client. Could go with something simple like `connection` and `connection_cls` for `server` and `transport`?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+                    try:
+                        message = self._parse_message(msg_contents[0][1])
+                    except TypeError:
+                        # This may happen if another thread/process deletes the
+                        # message between our generating the ID list and our
+                        # processing it here.
+                        raise
+                    yield message
+                except (MessageParseError, MailParserError):
+                    # If we have some problem parsing the email, it's likely
```
reupen: Should we log these? Seems like we should have some monitoring for it.

currycoder: You're right we should.  I'm considering how to log this in a way that is useful for debugging without accidentally logging any private details

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+                        continue
+                    try:
+                        message = self._parse_message(msg_contents[0][1])
+                    except TypeError:
+                        # This may happen if another thread/process deletes the
+                        # message between our generating the ID list and our
+                        # processing it here.
+                        raise
+                    yield message
+                except (MessageParseError, MailParserError):
```
reupen: Are we expecting these to be raised by `server.uid()` or during the `yield`?

datahub/email_ingestion/mailbox.py
elcct: I agree, `server` seems confusing, `connection` would be more appropriate.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+        Mailbox instantiation. The method returns as soon as the message has been
+        successfully processed by a processor.
+
+        :param message: mailparser.Mailparser object - the message to process
+
+        :returns: True if the message was processed, False otherwise.
+        """
+        for processor_class in self.processor_classes:
+            processor = processor_class()
+            processed, message = processor.process_email(message)
```
elcct: I am not sure if `message` is a good name, perhaps `outcome_detail` or something similar could be better. I though that `message` would be an actual content of the email.

datahub/email_ingestion/mailbox.py
elcct: Could we just log the message id and reason why it is logged if possible?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,173 @@
...
+                    yield message
+                except (MessageParseError, MailParserError):
+                    # If we have some problem parsing the email, it's likely
+                    # to be spam/malicious so skip it
+                    continue
+
+                # Mark the email for deletion in the inbox
+                server.uid('store', uid, '+FLAGS', '(\\Deleted)')
+            # Carry out deletion for the emails marked for deletion
+            server.expunge()
```
elcct: Do we need to delete messages?

datahub/email_ingestion/mailbox.py
currycoder: Quick question, what level is sentry showing log messages for in production?

datahub/email_ingestion/mailbox.py
currycoder: Good spot!  It's overwriting `message` which is bad!

datahub/email_ingestion/mailbox.py
currycoder: They are flagged for deletion (it's up to the IMAP server how this is handled - often the actual messages are just archived).

We depend on this mechanism for marking emails as processed for future ingestion runs.  
This saves us having to do more complicated things like "get me all the emails since the date of the last run" - instead, when we ingest an email successfully we mark it as deleted immediately.  In this way the ingester is atomic.

Though if there is a problem in the processing step (which happens after ingestion) the email will be lost - this is an acceptable shortcoming.

datahub/email_ingestion/mailbox.py
reupen: It's ERROR at the moment (across all loggers – it's configured in `config.settings.common_sentry`).

(Of course, directly sending messages to Sentry is also an option, e.g.: https://github.com/uktrade/data-hub-leeloo/blob/d9e6a84b7e6f3cd3a101901909894e582b3ca89b/datahub/search/execute_query.py#L34-L41)

datahub/email_ingestion/mailbox.py
elcct: So that is also assuming only one client will be using the email?
It is weird because that is a main IMAP's "selling points", the ability to keep messages on the server so that multiple devices can use the email without the need of downloading it.
It seems like the pattern here would fit how POP3 works - that is messages are being deleted once downloaded.

datahub/email_ingestion/mailbox.py
currycoder: I've reworked the scope of the try/except here so it is as narrow as possible

datahub/email_ingestion/mailbox.py
currycoder: Good point well made :)  I'm going to modify things so that it is based on "SEEN"/"UNSEEN" flags which is much cleaner and less destructive.

currycoder:
Going to improve test coverage BTW

currycoder:
This is now ready for re-review.  Coverage as follows:
```
datahub/email_ingestion/email_processor.py                                                                      4      1      0      0    75%
datahub/email_ingestion/mailbox.py                                                                             82      1     22      0    99%
```

`email_processor.py` is only at 75% because coverage is counting the body of an abstract method as a miss - which as far as I understand can't be executed.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,200 @@
...
+            processor_name = processor_class.__name__
+            try:
+                processed, processing_output = processor.process_email(message)
+            except Exception as e:
+                error_message = (
+                    f'Error "{e.__class__.__name__}" processing email "{message.message_id}" '
+                    f'which was processed by processor "{processor_name}"'
+                )
+                logger.error(error_message)
+                raise e
```
reupen: Just `raise` would be better for re-raising (maintains the original context). However, where is it re-raising it to? Will the error be logged twice?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,200 @@
...
+logger = getLogger(__name__)
+
+
+class EmailRetrievalError(Exception):
+    """
+    Custom exception for when an email was not successfully retrieved from the
+    IMAP server.
+    """
+
+    pass
```
reupen: Just a small thing, but `pass` can omitted if something has a docstring.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,200 @@
...
+        return mailparser.parse_from_bytes(message_bytes)
+
+    def _get_message(self, uid, connection):
+        try:
+            typ, msg_contents = connection.uid('fetch', uid, '(RFC822)')
+        except TypeError:
+            # This may happen if something deletes the
+            # message between our generating the ID list and our
+            # processing it here.
+            raise EmailRetrievalError()
```
reupen: Would suggest using the [`raise ... from ...` syntax](https://docs.python.org/3/reference/simple_stmts.html#raise) if we are logging this later on. (So `except TypeError as exc:` and `raise EmailRetrievalError() from exc`).

datahub/email_ingestion/email_processor.py
```diff
@@ -0,0 +1,27 @@
...
+
+        :returns: Tuple of boolean (whether the message was successfully processed
+            by this processor) and string (a meaningful message explaining why
+            this was not processed or explaining that this was processed
+            successfully).
+            e.g.
+
+            (False, "The email was not sent by a known DIT address")
+        """
+        pass
```
marcofucci: Actually, I think you can get rid of this pass (and get more coverage 😄 )

datahub/email_ingestion/test/test_mailbox.py
```diff
@@ -73,7 +73,8 @@ class TestMailbox:
 
     def mock_mailbox_parse_message(self, mailbox):
         """
-        Helper to ensure that the _parse_message method of a supplied mailbox
+        Helper to ensure that the _a
+        EmailRetrievalError as expected.parse_message method of a supplied mailbox
```
marcofucci: This needs updating?

datahub/email_ingestion/mailbox.py
marcofucci: Yeah shouldn't it just return False?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,200 @@
...
+    It offers methods for getting new messages from the inbox and processing
+    them using associated processing classes.
+
+    When an email is retrieved from the inbox successfully, it is marked as
+    SEEN on the upstream IMAP inbox.  This is the mechanism we use to ensure
+    that emails are only ever processed once.
+
+    The first processor to successfully consume and process the message will stop
+    the chain.  Processors will be called in order - so ordering of
+    `mail_processor_classes` is important.
```
marcofucci: It would be good to say that you need to acquire a look when using this, maybe with an example?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,200 @@
...
+                try:
+                    message = self._get_message(uid, connection)
+                except (MessageParseError, MailParserError) as e:
+                    # If we have some problem parsing the email, it's likely
+                    # to be spam/malicious so skip it
+                    error_message = (
+                        f'Mailbox "{self.email}" failed to parse message '
+                        f'{uid} with error {e}'
+                    )
+                    logger.error(error_message)
```
marcofucci: This is generally okay but have you thought that the message is not marked as SEEN so the error will be constantly logged every time `process_new_mail` is called until the problem is solved manually?

datahub/email_ingestion/test/test_mailbox.py
```diff
@@ -0,0 +1,389 @@
...
+            except IndexError:
+                assert not processor.process_email.called
+
+    def test_process_email_processing_error(self, caplog):
+        """
+        Test that _process_email can handle processing errors from a bad email
+        processor class.
+        """
+        processor_classes = []
+        for i in range(2):
```
marcofucci: Why are you using `range(2)` if you are not using the second element?

datahub/email_ingestion/mailbox.py
currycoder: Will do. :+1:

datahub/email_ingestion/test/test_mailbox.py
currycoder: Looks like an accidental paste! Good spot

datahub/email_ingestion/mailbox.py
currycoder: Very good point.  I'm going to ensure that the message is flagged as seen regardless of these errors.
A manual resolution can then be done/deployed if necessary, and we can mark the email as unread to reprocess it if necessary.

datahub/email_ingestion/test/test_mailbox.py
currycoder: Good point.  I've added an assertion to ensure that the second processor is not called.

currycoder:
This is good to go again - thanks for the comments :+1:

datahub/email_ingestion/email_processor.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABC, abstractmethod
+
+
+class EmailProcessor(ABC):
+    """
+    EmailProcessor subclasses should take a message and perform some actions
+    based on it's metadata/content in order to process it.  EmailProcessor is
```
elcct: That should be `its`?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,217 @@
...
+                    # Just set the message to None so that we still mark it as
+                    # read later
+                    message = None
+                except EmailRetrievalError as e:
+                    # We should fail and exit immediately in this case, as it's
+                    # probable that another process is processing the inbox
+                    error_message = (
+                        f'Mailbox "{self.email}" could not retrieve message {uid} successfully '
+                        f'with error {e}'
```
elcct: I that way it would sound better: `could not retrieve message {uid} successfully. Error {e}.`.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,217 @@
...
+            processor = processor_class()
+            processor_name = processor_class.__name__
+            try:
+                processed, processing_output = processor.process_email(message)
+            except Exception as e:
+                error_message = (
+                    f'Error "{e.__class__.__name__}" processing email "{message.message_id}" '
+                    f'which was processed by processor "{processor_name}"'
+                )
+                logger.error(error_message)
```
reupen: Should we use logger.exception() here?

datahub/email_ingestion/mailbox.py
reupen: Was this one missed?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,217 @@
...
+        with self._connect() as connection:
+            message_ids = self._get_all_message_ids(connection)
+            if not message_ids:
+                # No new messages to ingest
+                return
+
+            for uid in message_ids:
+                try:
+                    message = self._get_message(uid, connection)
+                except (MessageParseError, MailParserError) as e:
```
reupen: We usually use `exc` for exception variables – it would be good to be consistent.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,217 @@
...
+
+        :param message: mailparser.Mailparser object - the message to process
+
+        :returns: True if the message was processed, False otherwise.
+        """
+        for processor_class in self.processor_classes:
+            processor = processor_class()
+            processor_name = processor_class.__name__
+            try:
+                processed, processing_output = processor.process_email(message)
```
reupen: `processing_output` is not used, is that intentional?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,217 @@
...
+                try:
+                    message = self._get_message(uid, connection)
+                except (MessageParseError, MailParserError) as e:
+                    # If we have some problem parsing the email, it's likely
+                    # to be spam/malicious so skip it
+                    error_message = (
+                        f'Mailbox "{self.email}" failed to parse message '
+                        f'{uid} with error {e}'
+                    )
+                    logger.error(error_message)
```
reupen: As above – should we use `logger.exception()` here?

datahub/email_ingestion/mailbox.py
currycoder: I think there's a better approach with Reupen's suggestion below - `logger.exception()` gives a nice exception trace.

datahub/email_ingestion/mailbox.py
currycoder: Good to know!

datahub/email_ingestion/mailbox.py
currycoder: Good spot. I wanted to use this for debug info, so I have now.

datahub/email_ingestion/mailbox.py
currycoder: Done.  How did I not know about `logger.exception()`!?

datahub/email_ingestion/test/test_mailbox.py
```diff
@@ -0,0 +1,402 @@
...
+            mocked_processor_class = mock.Mock(spec=EmailProcessor)
+            mocked_processor_class.__name__ = f'Processor {i}'
+            mocked_processor = mocked_processor_class.return_value
+            mocked_processor.process_email.return_value = (False, 'Bad email')
+            processor_classes.append(mocked_processor_class)
+        bad_processor = processor_classes[0].return_value
+
+        def processing_error(*args, **kwargs):
+            raise TypeError('Ooops!')
+        bad_processor.process_email.side_effect = processing_error
```
reupen: Would `bad_processor.process_email.side_effect = TypeError('Ooops!')` work here?

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,220 @@
...
+        :returns: True if the message was processed, False otherwise.
+        """
+        for processor_class in self.processor_classes:
+            processor = processor_class()
+            processor_name = processor_class.__name__
+            try:
+                processed, processing_output = processor.process_email(message)
+            except Exception as exc:
+                error_message = (
+                    f'Error "{exc.__class__.__name__}" processing email "{message.message_id}" '
```
reupen: Should be able to drop the exception name from the message now.

datahub/email_ingestion/mailbox.py
```diff
@@ -0,0 +1,220 @@
...
+                logger.exception(error_message)
+                return False
+            if processed:
+                success_message = (
+                    f'Email {message.message_id} was processed by processor: '
+                    f'{processor_name}'
+                )
+                logger.info(success_message)
+                return True
+            else:
```
reupen: Could drop this else.

datahub/email_ingestion/test/test_mailbox.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        Unit test of _process_email method to ensure that the chain of email
+        processing works as expected.
+        """
+        processor_classes = []
+        # Set up our mocked processor classes, ensuring that they return the
+        # results we expect
+        for i in range(processor_count):
+            mocked_processor_class = mock.Mock(spec=EmailProcessor)
+            mocked_processor_class.__name__ = ''
```
reupen: Will `mock.Mock(spec=EmailProcessor, __name__='')` work?

datahub/email_ingestion/test/test_mailbox.py
```diff
@@ -0,0 +1,402 @@
...
+                (
+                    (False, 'Could not be processed'),
+                    (True, 'Processed successfully'),
+                ),
+                True,
+            ),
+            # Ensure that when there are no processors, the email is not processed
+            (
+                0,
+                tuple(),
```
reupen: Just `()` can be used for an empty tuple.

PR #1656, opened 2019-04-25
Reset registered address fields and copy over from companies house table

alixedi:
### Description of change

Add a Django command: `updated_company_registered_address` that copies the `registered_address` from `CompaniesHouseCompany` to `Company`. 

If a `Company` doesn't have a corresponding `CompaniesHouseCompany`, the `registered_address` fields are set to blank.

The command has a `simulate` flag that does everything except `.save()`.

The command creates revisions for the updates to the `registered_address` fields and tried to ensure that the revisions are only created if there is an actual change in the state. 

### Checklist

### DIFF
--- a/datahub/dbmaintenance/management/commands/update_company_registered_address.py
+++ b/datahub/dbmaintenance/management/commands/update_company_registered_address.py
@@ -1,0 +1,119 @@
+from logging import getLogger
+from types import SimpleNamespace
+
+import reversion
+from django.core.management.base import BaseCommand
+
+from datahub.company.models import Company
+from datahub.search.signals import disable_search_signal_receivers
+
+logger = getLogger(__name__)
+
+
+def copy_registered_address(destination_company, source_company):
+    """
+    Copy registered_address field values that are different from source
+    Company to destination Company.
+    :param destination_company: The registered_address fields for this
+    object will be updated.
+    :param source_company: The registered_address fields for this object will
+    be copied to destination object.
+    :returns copy_count: Number of fields that were copied
+    """
+    fields = {
+        'registered_address_1',
+        'registered_address_2',
+        'registered_address_town',
+        'registered_address_county',
+        'registered_address_country',
+        'registered_address_postcode',
+    }
+
+    copy_count = 0
+
+    for field in fields:
+        destination = getattr(destination_company, field)
+        source = getattr(source_company, field)
+        if destination != source:
+            setattr(destination_company, field, source)
+            copy_count += 1
+
+    return copy_count > 0
+
+
+class Command(BaseCommand):
+    """
+    Command to update registered_address for Company with
+    the corresponding ComapiesHouseCompany registered_address.
+    """
+
+    def add_arguments(self, parser):
+        """Define extra arguments."""
+        parser.add_argument(
+            '--simulate',
+            action='store_true',
+            help='Simulates the command by running the code without saving the changes',
+        )
+
+    @staticmethod
+    def _get_companies_queryset():
+        return Company.objects.all()
+
+    @disable_search_signal_receivers(Company)
+    def handle(self, *args, **options):
+        """Handles the command."""
+        logger.info('Started')
+
+        result = {True: 0, False: 0}
+
+        for company in self._get_companies_queryset().iterator():
+            succeeded = self.process_company(
+                company,
+                simulate=options['simulate'],
+            )
+            result[succeeded] += 1
+
+        logger.info(f'Finished - succeeded: {result[True]}, failed: {result[False]}')
+
+    @staticmethod
+    def _process_company(company, simulate=False):
+        """
+        Update the address of the Company to the
+        address of the CompanyHouseCompany.
+        """
+        ch_company = company.companies_house_data
+
+        if ch_company is None:
+            # There is no corresponding CH record so we reset
+            # the address by assigning empty values to fields
+            ch_company = SimpleNamespace(**{
+                'registered_address_1': '',
+                'registered_address_2': '',
+                'registered_address_town': '',
+                'registered_address_county': '',
+                'registered_address_country': None,
+                'registered_address_postcode': '',
+            })
+
+        copied = copy_registered_address(company, ch_company)
+
+        if not simulate and copied:
+            with reversion.create_revision():
+                company.save()
+                reversion.set_comment('Updated registered address using CompaniesHouse data.')
+
+    def process_company(self, company, simulate=False):
+        """
+        Wrapper around _process_company to catch potential problems.
+
+        :param company: Company object
+        :param simulate: if True, the changes will not be saved
+        :returns: bool indicating if the company was processed successfully
+        """
+        try:
+            self._process_company(company, simulate=simulate)
+            logger.info(f'Company {company.name} - OK')
+            return True
+        except Exception as exc:
+            logger.exception(f'Company {company.name} - {company.id} failed: {repr(exc)}')
+            return False
--- a/datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
+++ b/datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
@@ -1,51 +1,265 @@
+from unittest.mock import Mock, patch
+
 import pytest
 from django.core.management import call_command
+from reversion.models import Version
 
 from datahub.company.models import Company
 from datahub.company.test.factories import (
     CompaniesHouseCompanyFactory,
     CompanyFactory,
 )
+from datahub.core.constants import Country
+from datahub.dbmaintenance.management.commands.update_company_registered_address import (
+    copy_registered_address,
+)
 
 
 @pytest.fixture()
-def company_company_house_data(db):
+def company(db):
     """
-    Create a CompaniesHouseCompany that has a company number that matches
-    Company and one that doesn't.
+    Create a Company, CompaniesHouseCompany pair with the same address.
     """
-    CompanyFactory(company_number='1')
-    CompaniesHouseCompanyFactory(company_number='1')
-    CompanyFactory(company_number='2')
-    CompaniesHouseCompanyFactory(company_number='3')
+    address = {
+        'registered_address_1': '12',
+        'registered_address_2': 'Foo St.',
+        'registered_address_town': 'London',
+        'registered_address_county': 'Westminster',
+        'registered_address_country_id': Country.united_kingdom.value.id,
+        'registered_address_postcode': 'SW1 T2E',
+    }
+    company = CompanyFactory(company_number='1', **address)
+    CompaniesHouseCompanyFactory(company_number='1', **address)
+
+    return company
 
 
-def _is_address_same(company1, company2):
+def is_address_same(company_number):
     """
-    Check if the given companies have same
-    address.
+    Check if address is the same for the Company & CompaniesHouseCompany
+    with the given company_number.
     """
-    fields = [
-        'registered_address_town',
-        'registered_address_town',
+    fields = {
         'registered_address_1',
         'registered_address_2',
+        'registered_address_town',
         'registered_address_county',
         'registered_address_country',
         'registered_address_postcode',
-    ]
+    }
+
+    company = Company.objects.get(company_number=company_number)
+    ch_company = company.companies_house_data
+
     field_matches = [
-        getattr(company1, field) == getattr(company2, field)
+        getattr(company, field) == getattr(ch_company, field)
         for field in fields
     ]
+
     return all(field_matches)
 
 
-@pytest.mark.usefixtures('company_company_house_data')
-def test_run():
-    """."""
+@pytest.mark.parametrize(
+    'field, value',
+    (
+        ('registered_address_1', '10'),
+        ('registered_address_2', 'Bar St.'),
+        ('registered_address_town', 'Brighton'),
+        ('registered_address_county', 'Surrey'),
+        ('registered_address_country_id', Country.italy.value.id),
+        ('registered_address_postcode', 'TW3 2YT'),
+    ),
+)
+def test_different_address(field, value, company):
+    """
+    The command should copy the registered_address from CompaniesHouseCompany
+    to Company and this should create a revision.
+    """
+    setattr(company, field, value)
+    company.save()
+
     call_command('update_company_registered_address')
-    for company in Company.objects.all():
-        ch_data = company.companies_house_data
-        if ch_data is not None:
-            assert _is_address_same(company, ch_data)
+
+    assert is_address_same(company.company_number)
+
+    versions = Version.objects.get_for_object(company)
+    assert versions.count() == 1
+
+
+def test_same_address(company):
+    """
+    When the address is same, we do not create a new revision.
+    """
+    call_command('update_company_registered_address')
+
+    versions = Version.objects.get_for_object(company)
+    assert versions.count() == 0
+
+
+def test_no_ch_company(company):
+    """
+    When there is no matching CompaniesHouseCompany for a
+    Company, we reset the address to blank and create revision.
+    """
+    company.company_number = 'NOT 1'
+    company.save()
+    call_command('update_company_registered_address')
+
+    company.refresh_from_db()
+    assert company.registered_address_1 == ''
+    assert company.registered_address_2 == ''
+    assert company.registered_address_town == ''
+    assert company.registered_address_county == ''
+    assert company.registered_address_country is None
+    assert company.registered_address_postcode == ''
+
+    versions = Version.objects.get_for_object(company)
+    assert versions.count() == 1
+
+
+def test_no_ch_company_address_blank(company):
+    """
+    When there is no matching CompaniesHouseCompany for a
+    Company and the address is already blank, we do not
+    create a revision.
+    """
+    company.company_number = 'NOT 1'
+    company.registered_address_1 = ''
+    company.registered_address_2 = ''
+    company.registered_address_town = ''
+    company.registered_address_county = ''
+    company.registered_address_country = None
+    company.registered_address_postcode = ''
+    company.save()
+
+    call_command('update_company_registered_address')
+
+    versions = Version.objects.get_for_object(company)
+    assert versions.count() == 0
+
+
+def test_simulate(company):
+    """
+    When the command is run in simulate mode, we do not save
+    the changes or create a revision.
+    """
+    company.registered_address_postcode = 'TW3 2TY'
+    company.save()
+    call_command('update_company_registered_address', simulate=True)
+
+    assert not is_address_same(company.company_number)
+
+    versions = Version.objects.get_for_object(company)
+    assert versions.count() == 0
+
+
+@pytest.mark.usefixtures('company')
+@pytest.mark.parametrize('simulate', [True, False])
+def test_logs(simulate, caplog):
+    """
+    The normal as well as simulated run of the command should produce logs
+    when there are no exceptions.
+    """
+    with caplog.at_level('INFO'):
+        call_command('update_company_registered_address', simulate=simulate)
+        assert caplog.records[0].message == 'Started'
+        assert caplog.records[-1].message == 'Finished - succeeded: 1, failed: 0'
+
+
+@pytest.mark.usefixtures('company')
+@patch(
+    'datahub.dbmaintenance.management.commands'
+    '.update_company_registered_address.Command'
+    '._process_company',
+    side_effect=Exception('Testing'),
+)
+@pytest.mark.parametrize('simulate', [True, False])
+def test_logs_fail(_, simulate, caplog):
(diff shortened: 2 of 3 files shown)


datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,94 @@
...
+from django.core.management.base import BaseCommand
+
+from datahub.company.models import Company
+from datahub.search.signals import disable_search_signal_receivers
+
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """."""
```
reupen: The docstring needs updating here :slightly_smiling_face:

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+    @staticmethod
+    def _get_companies_queryset():
+        return Company.objects.all()
+
+    @disable_search_signal_receivers(Company)
+    def handle(self, *args, **options):
+        """Handles the command."""
+        logger.info(f'Started')
```
reupen: This one doesn't need to be an f-string (copied from somewhere else that was wrong I guess...)

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        succeeded = True
+        try:
+            self._process_company(company, simulate=simulate)
+            logger.info(f'Company {company.name} - OK')
+        except Exception as exc:
+            succeeded = False
+            logger.exception(f'Company {company.name} - {company.id} failed: {repr(exc)}')
+
+        return succeeded
```
reupen: Not sure `succeeded` is really needed. You could just return True here and False above in the exception handler.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @staticmethod
+    def _copy_address(destination, source):
+        """
+        Copy address fields from source to
+        destination record.
+        """
+        fields = [
+            'registered_address_town',
+            'registered_address_town',
```
reupen: Looks like you have `registered_address_town` twice here.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _is_address_same(company1, company2):
+    """
+    Check if the given companies have same
+    address.
+    """
+    fields = [
+        'registered_address_town',
+        'registered_address_town',
```
reupen: As above – looks like you have `registered_address_town` twice here.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,106 @@
...
+
+from datahub.company.models import Company
+from datahub.company.test.factories import (
+    CompaniesHouseCompanyFactory,
+    CompanyFactory,
+)
+
+
+@pytest.fixture()
+def company_company_house_data(db):
```
reupen: I assume the middle company should be companies? As Companies House is the name of an organisation.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if ch_company is None:
+            return
+
+        self._copy_address(company, ch_company)
+
+        if not simulate:
+            with reversion.create_revision():
+                company.save()
+                reversion.set_comment('Registered address correction.')
```
reupen: It would be good to make 'Registered address correction' more specific (something like 'Updated registered address using Companies House data' perhaps).

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Update the address of the Company to the
+        address of the CompanyHouseCompany.
+        """
+        ch_company = company.companies_house_data
+
+        if ch_company is None:
+            return
+
+        self._copy_address(company, ch_company)
```
reupen: In other commands we've checked if the values of the fields being updated are actually different from the new values and not doing anything if they are the same. Perhaps we should be doing that here as well?

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,106 @@
...
+    field_matches = [
+        getattr(company1, field) == getattr(company2, field)
+        for field in fields
+    ]
+    return all(field_matches)
+
+
+@pytest.mark.usefixtures('company_company_house_data')
+def test_run():
+    """."""
```
reupen: Docstrings to update (here and below).

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,131 @@
...
+        Update the address of the Company to the
+        address of the CompanyHouseCompany.
+        """
+        ch_company = company.companies_house_data
+
+        if ch_company is None:
+            return
+
+        if is_address_same(company, ch_company):
+            return
```
reupen: This line isn't covered according to the coverage report, it would be good to update the test data so that it gets covered.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    with caplog.at_level('INFO'):
+        call_command('update_company_registered_address', simulate=simulate)
+        assert caplog.records[0].message == 'Started'
+        assert caplog.records[-1].message == 'Finished - succeeded: 2, failed: 0'
+
+
+@pytest.mark.usefixtures('company_companies_house_data')
+@patch(
+    (
```
reupen: Small thing, but I don't think this pair of brackets is needed.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,131 @@
...
+    Check if the given companies have same address.
+    :param company1: Company object
+    :param company2: Company object
+    :param address_fields: List of address fields
+    :returns: bool indicating if the company1 and company2
+    have the same registered_address.
+    """
+    fields = address_fields or REGISTERED_ADDRESS_FIELDS
+
+    assert all([hasattr(company1, field) for field in fields])
```
reupen: Do we need these assertions? Not sure what they are for. (`getattr()` will raise an exception if the attributes don't exist.)

alixedi: Removed.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
alixedi: Removed.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
alixedi: I am going to add a test that checks for the absence of version (from reversion) in case the addresses match.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,177 @@
...
+@pytest.fixture()
+def company_companies_house_data(db):
+    """
+    Create a CompaniesHouseCompany that has a company number that matches
+    Company and one that doesn't.
+    """
+    CompanyFactory(company_number='1')
+    CompaniesHouseCompanyFactory(company_number='1')
+    CompanyFactory(company_number='2')
+    CompaniesHouseCompanyFactory(company_number='3')
```
marcofucci: I would set different registered addresses explicitly instead of relying on the factory as this could change in the future.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,177 @@
...
+    destination = type('Foo', (object,), {'a': 'Foo'})
+    source = type('Bar', (object,), {'a': 'Bar'})
+
+    copy_address(destination, source, ['a'])
+
+    assert is_address_same(destination, source, ['a'])
+
+
+@pytest.mark.usefixtures('company_companies_house_data')
+def test_same_address():
```
marcofucci: It would be better to test this via calling the command instead we don't know the internal logic of the _process_company here unless we also test the other case here which seems less clean.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,177 @@
...
+import pytest
+from django.core.management import call_command
+from reversion.models import Version
+
+from datahub.company.models import Company
+from datahub.company.test.factories import (
+    CompaniesHouseCompanyFactory,
+    CompanyFactory,
+)
+from datahub.dbmaintenance.management.commands \
```
reupen: Just a small thing, but

```python
from datahub.dbmaintenance.management.commands.update_company_registered_address import (
    copy_address,
    is_address_same,
    Command,
)
```
would fit here (with a 99 character line-length limit).

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,131 @@
...
+    @staticmethod
+    def _process_company(company, simulate=False):
+        """
+        Update the address of the Company to the
+        address of the CompanyHouseCompany.
+        """
+        ch_company = company.companies_house_data
+
+        if ch_company is None:
+            return
```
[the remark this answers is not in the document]
alixedi: I didn't realise this was part of the spec.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
alixedi: I think I didn't this b/c the CI complained but I will double check.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
alixedi: Good point.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
alixedi: NP

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,125 @@
...
+    'registered_address_1',
+    'registered_address_2',
+    'registered_address_town',
+    'registered_address_county',
+    'registered_address_country',
+    'registered_address_postcode',
+]
+
+
+def is_address_same(company1, company2, address_fields=None):
```
elcct: It doesn't seem to be doing anything address specific, just the list of fields to compare defaults to `REGISTERED_ADDRESS_FIELDS` if nothing is provided. I would make it generic and just pass the list of address fields as an argument. If you passed fields that are not address fields the function wouldn't "know".
Probably I would optimize this and `copy_address` into one function.  You could take a list of attributes that differ and then copy only those. Then return the number of attributes copied over. If it was zero then save operation would be skipped.

Or to make it address specific I would validate that the `address_fields` are indeed address fields.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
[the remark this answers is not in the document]
reupen: @marcofucci Is that also for non-UK companies?

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
alixedi: Actually, I have designed the tests so that they are not relying on a pair of `Company` and `CompaniesHouseCompany` having the same address: https://github.com/uktrade/data-hub-leeloo/pull/1656/files#diff-c11876a6290236297465c3cda715734fR175

datahub/dbmaintenance/management/commands/update_company_registered_address.py
alixedi: 👌 ideas, thanks.

One Q: How would you validate if `address_fields` are indeed `address_fields`?

elcct: Probably the easiest way is to remove `address_fields` param and only check `REGISTERED_ADDRESS_FIELDS` fields.

alixedi: Having `address_fields` as parameter makes for better tests: https://github.com/uktrade/data-hub-leeloo/blob/65c7e6c854c77402036f018c180427d0fe75e904/datahub/dbmaintenance/test/commands/test_update_company_registered_address.py#L152-L153

elcct: That test doesn't check if you can actually copy the address. You could modify the test so that it reflects the intended use of the function - that is to add address fields to the test objects - this way you won't need `address_fields` and test functionality that actually will be used.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,140 @@
...
+    fields = address_fields or REGISTERED_ADDRESS_FIELD_DEFAULTS
+
+    for field in fields:
+        setattr(destination, field, getattr(source, field))
+
+
+def reset_address(company, address_fields=None):
+    """
+    Resets the address fields to empty string. Country is a ForegnKey
+    and so it is set yo None.
```
marcofucci: `yo` typo :)

datahub/dbmaintenance/management/commands/update_company_registered_address.py
```diff
@@ -0,0 +1,140 @@
...
+    @staticmethod
+    def _process_company(company, simulate=False):
+        """
+        Update the address of the Company to the
+        address of the CompanyHouseCompany.
+        """
+        ch_company = company.companies_house_data
+
+        if ch_company is None:
+            reset_address(company)
```
marcofucci: If would be good if we could reset and save only when needed so that we avoid multiple SQLs and empty revisions.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
alixedi: We have other tests that check if we copied the address - as in all the fields in the `REGISTERED_ADDRESS_FIELD_DEFAULTS`.

Parametrising the `address_fields` in this function means that we can test for the copy functionality *only*. I don't think this is necessarily a bad thing.

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,172 @@
...
+def test_run():
+    """
+    The normal run of the command should update the registered_address
+    fields for the Company object.
+    """
+    call_command('update_company_registered_address')
+    for company in Company.objects.all():
+        ch_data = company.companies_house_data
+        if ch_data is not None:
+            assert is_address_same(company, ch_data)
```
marcofucci: This doesn't test the reset case when the CH data doesn't exist or is invalid

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    foo = type('Foo', (object,), {'a': 'A property'})
+    bar = type('Bar', (object,), {})
+
+    with pytest.raises(AttributeError):
+        copy_address(foo, bar, ['a'])
+        copy_address(bar, foo, ['a'])
+
+
+def test_is_address_same():
```
marcofucci: I would move this up so that it's closer to the `test_is_address_same_missing_field` one

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    is_address_same function should return True if all the
+    address_fields for the given objects are the same.
+    """
+    foo = type('Foo', (object,), {'a': 'A property'})
+    bar = type('Bar', (object,), {'a': 'A property'})
+    baz = type('Baz', (object,), {'a': 'Another property'})
+
+    assert is_address_same(foo, bar, address_fields=['a'])
+    assert not is_address_same(foo, baz, address_fields=['a'])
```
marcofucci: I would use `pytest.mark.parametrize` instead of testing both so that it's easier to scan mentally (especially as `bar` and `baz` are similar names :) ) but it's not a major thing.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
marcofucci: I'm also not convinced by having  the `address_fields` argument just for the tests. Maybe you could remove it and use a Company instance in the tests instead?

elcct: But that extra functionality is created solely for the tests and it allows user of this function to copy fields that are not address fields. I am just not sure how that is adding value. You are not testing something that is being used in the product and that is confusing.
The tests assume that if those functions work with provided fields then those functions should also work with the `REGISTERED_ADDRESS_FIELD_DEFAULTS`.

alixedi: P.S. Had a chat IRL and going to change this now. TY @marcofucci and @elcct

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,172 @@
...
+        call_command('update_company_registered_address', simulate=simulate)
+        assert caplog.records[-1].message == 'Finished - succeeded: 0, failed: 2'
+
+
+def test_is_address_same_missing_field():
+    """
+    is_address_same function should fail if one of the objects
+    does not have any of the specified address_fields.
+    """
+    foo = type('Foo', (object,), {'a': 'A property'})
```
reupen: Any reason for not going for [https://docs.python.org/3/library/types.html#types.SimpleNamespace](types.SimpleNamespace) (or just a `Mock`)?

datahub/dbmaintenance/management/commands/update_company_registered_address.py
alixedi: Great point. I had to refactor a bit to take care of this generally and cleanly. TY @elcct for the ideas.

datahub/dbmaintenance/management/commands/update_company_registered_address.py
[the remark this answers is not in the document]
marcofucci: @reupen yes, that field is called primary address in the frontend so it's technically wrong to start considering that address the "registered address" and there's no way we can get the real registered address for now (we'll eventually get it from D&B anyway).

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def test_no_ch_company(company):
+    """
+    When there is no matching CompaniesHouseCompany for a
+    Company, we reset the address to blank and create revision.
+    """
+    company.company_number = 'NOT 1'
+    company.save()
+    call_command('update_company_registered_address')
```
marcofucci: I think you missed the lines that check that the registered address fields are actually reset :-)

datahub/dbmaintenance/test/commands/test_update_company_registered_address.py
```diff
@@ -0,0 +1,210 @@
...
+)
+@pytest.mark.parametrize('simulate', [True, False])
+def test_logs_fail(_, simulate, caplog):
+    """
+    The normal as well as simulated run of the command should produce logs
+    when it encounters exceptions.
+    """
+    with caplog.at_level('INFO'):
+        call_command('update_company_registered_address', simulate=simulate)
+        assert caplog.records[-1].message == 'Finished - succeeded: 0, failed: 1'
```
marcofucci: Would it be okay to change this slightly so that it processes 2 companies, the first one fails and the second one succeeds so that we know that the first error doesn't stop the whole command?

alixedi: I added a new test for this: https://github.com/uktrade/data-hub-leeloo/blob/373d6cb20c2cd100a9870558441bc8410a7966a6/datahub/dbmaintenance/test/commands/test_update_company_registered_address.py#L221

PR #1619, opened 2019-04-12
Make admin site CSV validation logic reusable

reupen:
### Description of change

This moves the CSV validation logic that was being used in the admin site for loading marketing opt-outs into a separate class so that it can be reused for similar functionality elsewhere in the admin site.

This is being done in advance of another tool being added into the admin site for loading interactions from a CSV file.

This is just a refactor, so I haven't added a news fragment.

I would suggest looking at each commit individually.

### Checklist

* [ ] Has a new newsfragment been created? Check [changelog/README.rst](https://github.com/uktrade/data-hub-leeloo/blob/master/changelog/README.rst) for instructions
* [ ] Have any relevant search models been updated?
* [ ] Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?
* [ ] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?
* [ ] Has the admin site been updated (for new models, fields etc.)?
* [ ] Has the README been updated (if needed)?

PR #1659, opened 2019-04-26
Add form for validating rows for import interactions tool

reupen:
### Description of change

This adds a form that will be used to validate CSV rows when importing interactions via the admin site. The form will only be used for validation; it will not be displayed anywhere.

For context, please see previous related PRs: #1619, #1623, #1632 and #1650.

The spec the rows are being validated against is covered in on the select file page:

image

### To test

The form is not used yet, so there is nothing to manually test. However, you can look at #1660 if you want to see how it will be used.

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/row_form.py
+++ b/datahub/interaction/admin_csv_import/row_form.py
@@ -1,0 +1,173 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy
+
+from datahub.company.models import Advisor
+from datahub.core.query_utils import get_full_name_expression
+from datahub.event.models import Event
+from datahub.interaction.models import CommunicationChannel, Interaction
+from datahub.metadata.models import Service, Team
+
+
+OBJECT_DISABLED_MESSAGE = gettext_lazy('This option is disabled.')
+ADVISER_NOT_FOUND_MESSAGE = gettext_lazy(
+    'An active adviser could not be found with the specified name.',
+)
+ADVISER_WITH_TEAM_NOT_FOUND_MESSAGE = gettext_lazy(
+    'An active adviser could not be found with the specified name and team.',
+)
+MULTIPLE_ADVISERS_FOUND_MESSAGE = gettext_lazy(
+    'Multiple matching advisers were found.',
+)
+INTERACTION_CANNOT_HAVE_AN_EVENT_MESSAGE = gettext_lazy(
+    'An interaction cannot have an event.',
+)
+
+
+def _validate_not_disabled(obj):
+    if obj.disabled_on:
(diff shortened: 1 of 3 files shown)


datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -0,0 +1,168 @@
...
+        CommunicationChannel.objects.filter(),
+        to_field_name='name__iexact',
+        required=False,
+        validators=[_validate_not_disabled],
+    )
+    event_id = forms.ModelChoiceField(
+        Event.objects.filter(),
+        required=False,
+        validators=[_validate_not_disabled],
+    )
```
marcofucci: Wouldn't it be better to exclude all the disabled ones and remove the validator? In some cases it would also be more correct as 2 objects having the same name with the old one being disabled might imply that you want the non-disabled one?

reupen: I did that initially, but it means that the error message (if the value entered is for a disabled item) is a bit confusing. We should not be having metadata items with the same name anyway (they will show up as duplicates in the filters on the collection pages, for example).

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -0,0 +1,168 @@
...
+    communication_channel = NoDuplicatesModelChoiceField(
+        CommunicationChannel.objects.filter(),
+        to_field_name='name__iexact',
+        required=False,
+        validators=[_validate_not_disabled],
+    )
+    event_id = forms.ModelChoiceField(
+        Event.objects.filter(),
```
marcofucci: Why these 2 have `.filter()` instead of `.all()`

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+
+
+class InteractionCSVRowForm(forms.Form):
+    """Form used for validating a single row in a CSV of interactions."""
+
+    kind = forms.ChoiceField(choices=Interaction.KINDS)
+    date = forms.DateField(input_formats=['%d/%m/%Y', '%Y-%m-%d'])
+    contact_email = forms.EmailField()
+    adviser_1 = forms.CharField()
```
marcofucci: A bit of help text or a comment to document how this adviser will be obtained would be good.

datahub/interaction/admin_csv_import/row_form.py
reupen: Have updated these.

datahub/interaction/admin_csv_import/row_form.py
reupen: Added some additional comments.

PR #1661, opened 2019-04-26
Make company.Company.registered_address_country blank=True

marcofucci:
### Description of change

This makes the field `company.Company.registered_address_country` `blank=True` so that it becomes optional in the Django admin.

It was already `null=True` so no other logic should change.

### Generated SQL
This shouldn't create any SQL but Django does this:

```sql
SET CONSTRAINTS "c_registered_address_country_id_4a27b881_fk_metadata_country_id" IMMEDIATE; ALTER TABLE "company_company" DROP CONSTRAINT "c_registered_address_country_id_4a27b881_fk_metadata_country_id"
ALTER TABLE "company_company" ADD CONSTRAINT "company_company_registered_address_c_4a27b881_fk_metadata_" FOREIGN KEY ("registered_address_country_id") REFERENCES "metadata_country" ("id") DEFERRABLE INITIALLY DEFERRED
```

It seems okay though. Another option would be to edit an older migration file to add `blank=True` and avoid an extra migration but seems messier.

### Checklist


reupen:
> This shouldn't create any SQL but Django does this:
> 
> ```sql
> SET CONSTRAINTS "c_registered_address_country_id_4a27b881_fk_metadata_country_id" IMMEDIATE; ALTER TABLE "company_company" DROP CONSTRAINT "c_registered_address_country_id_4a27b881_fk_metadata_country_id"
> ALTER TABLE "company_company" ADD CONSTRAINT "company_company_registered_address_c_4a27b881_fk_metadata_" FOREIGN KEY ("registered_address_country_id") REFERENCES "metadata_country" ("id") DEFERRABLE INITIALLY DEFERRED
> ```
> 
> It seems okay though. Another option would be to edit an older migration file to add `blank=True` and avoid an extra migration but seems messier.

It would be good to check how long that takes. We have edited old migrations to avoid that before.

marcofucci:
@reupen good call, I've just run the sql in staging and it completed in 0.4 seconds

PR #1671, opened 2019-04-30
Add changelog for 11.11.0

marcofucci:
Changelog for 11.11.0


marcofucci:
I moved the section related to the company data change to the deprecation section so that it's more prominent. It's actually a deprecation so it makes more sense for it to be there.

PR #1673, opened 2019-05-01
Add contact matching logic to import interaction tool

reupen:
### Description of change

This adds the contact matching logic for the import interactions admin site tool.

Interactions are matched with contacts by email address.

The logic works as follows:

- First, `Contact.email` is checked. If a unique match is found, this is used.
- If multiple matches were found using `Contact.email`, matching aborts.
- Otherwise, `Contact.email_alternative` is checked. If there is a unique match, this is used.

An index on `UPPER(email_alternative)` for the `company_contact` table was also added in the same way as was done for the `email` field in https://github.com/uktrade/data-hub-leeloo/pull/1181. The migration should take about half a second, so is more than quick enough.

For additional context, see previous PRs related to the import interactions tool:
https://github.com/uktrade/data-hub-leeloo/issues?utf8=%E2%9C%93&q=label%3A%22import+interactions%22

### To test

Matched contacts are not yet used so there is nothing to manually test.

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/row_form.py
+++ b/datahub/interaction/admin_csv_import/row_form.py
@@ -2,6 +2,7 @@ from django import forms
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy
 
+from datahub.company.contact_matching import find_active_contact_by_email_address
 from datahub.company.models import Advisor
 from datahub.core.query_utils import get_full_name_expression
 from datahub.event.models import Event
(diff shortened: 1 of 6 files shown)


datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -133,41 +146,87 @@ def clean(self):
...
                 data.get(team_field),
             )
         except ValidationError as exc:
             self.add_error(adviser_field, exc)
 
-    @staticmethod
-    def _look_up_adviser(adviser_name, team):
-        if not adviser_name:
-            return None
+    def _populate_contact(self, data):
```
currycoder: This looks really very handy - especially considering that I'm about to do very similar things for the email calendar invite ingestion (looking up a number of contacts based on recipient emails).  Would you consider moving this matching logic over to a more reusable util somewhere appropriate?

reupen: Yep – something like `datahub.company.contact_matching`?

currycoder: Sounds good to me :+1:

PR #1653, opened 2019-04-24
Make company fields NOT NULL

marcofucci:
### Description of change

This makes the following columns `NOT NULL` to align with what [Django recommends](https://docs.djangoproject.com/en/2.1/ref/models/fields/#null):

 - ``company_company.registered_address_2``
- ``company_company.registered_address_county``
- ``company_company.registered_address_postcode``
- ``company_company.address_1``
- ``company_company.address_2``
- ``company_company.address_town``
- ``company_company.address_county``
- ``company_company.address_postcode``
- ``company_company.trading_address_1``
- ``company_company.trading_address_2``
- ``company_company.trading_address_town``
- ``company_company.trading_address_county``
- ``company_company.trading_address_postcode``

The migration is split into separate files because of [this Django bug](https://code.djangoproject.com/ticket/25105).

There's no point in refactoring the company V3 endpoints in any way as they will be removed soon.

### Generated SQL

```sql
BEGIN
ALTER TABLE "company_company" ALTER COLUMN "registered_address_2" SET DEFAULT ''
UPDATE "company_company" SET "registered_address_2" = '' WHERE "registered_address_2" IS NULL
ALTER TABLE "company_company" ALTER COLUMN "registered_address_2" SET NOT NULL
ALTER TABLE "company_company" ALTER COLUMN "registered_address_2" DROP DEFAULT
COMMIT

BEGIN
ALTER TABLE "company_company" ALTER COLUMN "registered_address_county" SET DEFAULT ''
UPDATE "company_company" SET "registered_address_county" = '' WHERE "registered_address_county" IS NULL
ALTER TABLE "company_company" ALTER COLUMN "registered_address_county" SET NOT NULL
ALTER TABLE "company_company" ALTER COLUMN "registered_address_county" DROP DEFAULT
COMMIT

BEGIN
ALTER TABLE "company_company" ALTER COLUMN "registered_address_postcode" SET DEFAULT ''
UPDATE "company_company" SET "registered_address_postcode" = '' WHERE "registered_address_postcode" IS NULL
ALTER TABLE "company_company" ALTER COLUMN "registered_address_postcode" SET NOT NULL
ALTER TABLE "company_company" ALTER COLUMN "registered_address_postcode" DROP DEFAULT
COMMIT

BEGIN
ALTER TABLE "company_company" ALTER COLUMN "address_1" SET DEFAULT ''
UPDATE "company_company" SET "address_1" = '' WHERE "address_1" IS NULL
[...]

PR #1679, opened 2019-05-02
Make status, location and archived fields on Interaction not nullable

currycoder:
### Description of change

This follows on from the work to add `status`, `location` and archive fields to the Interaction model.  To allow backwards compatibility with the existing data (without a very long/blocking data migration) a number of fields were set to be nullable in the DB.

The work has been carried out to set a default value for these fields on production.  This change makes the fields not nullable and ensures that any new Interaction objects created by the application have a default value.  

The resultant migration will also serve to update any development environments with default values.

### Checklist


reupen:
@currycoder We discovered relatively recently that we need to split these kind of migrations due to https://code.djangoproject.com/ticket/25105 and #1653.

(If it actually updates values for one of the first two columns you'll get an error like `django.db.utils.OperationalError: cannot ALTER TABLE "interaction_interaction" because it has pending trigger events`.)

(`atomic = False` should work as well, but I don't think it's a good solution as we may run the risk of things being in an inconsistent state if the app gets killed while migrations are running.)

currycoder:
> @currycoder We discovered relatively recently that we need to split these kind of migrations due to https://code.djangoproject.com/ticket/25105 and #1653.

Onit. :+1:

datahub/interaction/models.py
```diff
@@ -199,11 +189,7 @@ class Interaction(ArchivableModel, BaseModel):
         'metadata.Service', blank=True, null=True, on_delete=models.SET_NULL,
     )
     subject = models.TextField()
-    # TODO: Ensure that this is required (with default='')
-    # once we have ensured that a default value '' is set on
-    # all existing Interactions and we have ensured that new interactions are
-    # being created with a '' location
-    location = models.CharField(max_length=MAX_LENGTH, blank=True, null=True)
+    location = models.CharField(max_length=MAX_LENGTH, default='')
```
marcofucci: This is interesting, I was expecting this to be `location = models.CharField(max_length=MAX_LENGTH, blank=True)` which disallow null values and uses `''` behind the scenes.
Any reasons why you choose `location = models.CharField(max_length=MAX_LENGTH, default='')?

currycoder:
> can't we remove the `'location': {'default': ''},` from the `InteractionSerializer.Meta`?

That's possible yes - will do that now.  

As a side note: I thought it would also be possible to remove the overrides for `status` in the `Meta` too - but this gave me some unexpected failures which I believe are to do with how `RulesBasedValidator`  builds the state of the model before validating.  I think that it doesn't take model field defaults in to account before running a validation step?  Do you know if this is a limitation of `DataCombiner` or expected behaviour..?

datahub/interaction/models.py
currycoder: No particular reason, I think that's just the way that my mental map works when thinking about blankable charfields.
I'll switch it over to `blank=True` for consistency.

marcofucci:
@currycoder yeah I found out about that limitation when dealing with addresses. In my case it was the DataCombiner.get_value returning None instead of the field default in case of new instances.
I didn't work on it as my logic was temporary and I didn't know if it was a huge issue but it seems it would be nice to fix it eventually.

changelog/interaction/meeting-fields-not-nullable.db.rst
```diff
@@ -0,0 +1,6 @@
+The `interaction_interaction` table has been modified such that the following
+columns are no longer nullable:
+
+- `status` - this has an application-enforced default of 'complete'
+- `location` - this has an application-enforced default of ''
+- `archived` - this has an application-enforced default of false
```
marcofucci: double `` required in this fragment

datahub/interaction/migrations/0054_location_not_null.py
```diff
@@ -0,0 +1,16 @@
...
+
+    dependencies = [
+        ('interaction', '0053_status_not_null'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='interaction',
+            name='location',
+            field=models.CharField(default='', max_length=255),
```
marcofucci: needs updating (check `makemigrations`)

currycoder:
> @currycoder yeah I found out about that limitation when dealing with addresses. In my case it was the DataCombiner.get_value returning None instead of the field default in case of new instances.
> I didn't work on it as my logic was temporary and I didn't know if it was a huge issue but it seems it would be nice to fix it eventually.

Is this something that is logged somewhere?  If not, where shall I log it?  I'd be more than happy to take a look once calendar interaction ingestion is done.

PR #1683, opened 2019-05-03
Use serialiser validation rules in admin site import interaction tools 

reupen:
### Description of change

This makes the admin site import interactions tool uses the interaction serialiser validation rules as a second phase of validation.

Errors are mapped to CSV fields where possible.

This adds a bit of complexity, but we agreed that it was better to reuse the existing validation logic if possible.

This allowed some of the validation logic in the form for the `event_id` field to be removed. It also revealed that some validation logic for `communication_channel` has been previously missed.

### Checklist


datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -177,6 +217,54 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+            'service': data['service'],
+            'status': Interaction.STATUSES.complete,
+            'subject': subject,
+            'was_policy_feedback_provided': False,
+        }
+
+        if kind == Interaction.KINDS.service_delivery:
+            creation_data['is_event'] = bool(event)
+
+        return creation_data
```
marcofucci: I find this method slightly more complicated than it should be. Could you just do something like this?
```
    def _as_serializer_dict(self):
        """
        Transforms cleaned data into a dict suitable for use with validation rules from
        InteractionSerializer.
        """
        data = self.cleaned_data

        dit_participants = [
            {
                'adviser': adviser,
                'team': adviser.dit_team,
            }
            for adviser in [data.get('adviser_1'), data.get('adviser_2')]
            if adviser
        ]

        creation_data = {
            **data,

            'contacts': [data['contact']],
            'company': data['contact'].company,
            'date': datetime.combine(data['date'], time(), tzinfo=utc),
            'dit_participants': dit_participants,
            'event': data.get('event_id'),
            'status': Interaction.STATUSES.complete,
            'subject': data.get('subject') or data['service'].name,
            'was_policy_feedback_provided': False,
        }

        if creation_data['kind'] == Interaction.KINDS.service_delivery:
            creation_data['is_event'] = bool(creation_data['event'])

        return creation_data
```

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -177,6 +217,54 @@ def _populate_adviser(self, data, adviser_field, team_field):
         except ValidationError as exc:
             self.add_error(adviser_field, exc)
 
+    def _as_serializer_dict(self):
```
marcofucci: I would call this something like `_cleaned_data_as_serializer_data`

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -160,13 +174,39 @@ def clean(self):
...
+
+        Errors are mapped to CSV fields where possible. If not possible, they are
+        added to NON_FIELD_ERRORS (but this should not happen).
+        """
+        if not self.is_valid() or not self.cleaned_data['contact']:
+            return
+
+        transformed_data = self._as_serializer_dict()
+        serializer = InteractionSerializer()
+        validator = RulesBasedValidator(*INTERACTION_SERIALIZER_VALIDATION_RULES)
```
currycoder: What prevents us from using the `validators` attached to the InteractionSerializer?  e.g. the `ContactsBelongToCompanyValidator` still looks applicable to a CSV...

datahub/interaction/serializers.py
```diff
@@ -394,101 +504,5 @@ class Meta:
...
-                    OperatorRule('contacts', lambda value: len(value) <= 1),
-                    when=OperatorRule('is_event', bool),
-                ),
-                ValidationRule(
-                    'invalid_for_non_event',
-                    OperatorRule('event', not_),
-                    when=OperatorRule('is_event', not_),
-                ),
-            ),
+            RulesBasedValidator(*INTERACTION_SERIALIZER_VALIDATION_RULES),
```
marcofucci: I think that this is a bit confusing as a new dev wouldn't know where to put extra logic and what validator to use in other places.
To easily fix this you can override `RulesBasedValidator`, give it a meaningful name and write it so that it initialises itself with those INTERACTION_SERIALIZER_VALIDATION_RULES items so that you can use the class in here and in the form.

An even better solution would be to somehow have a constant or a custom Validator with a list of sub validators so that all validators except `HasAssociatedInvestmentProjectValidator` can be in one object and could be called here and in the form.

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -160,13 +174,39 @@ def clean(self):
...
         return data
 
-    def get_flat_error_list_iterator(self):
-        """Get a generator of CSVRowError instances representing validation errors."""
-        return (
-            CSVRowError(self.row_index, field, self.data.get(field, ''), error)
-            for field, errors in self.errors.items()
-            for error in errors
-        )
+    def _post_clean(self):
```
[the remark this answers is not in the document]
reupen: Done :+1:

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -160,13 +174,39 @@ def clean(self):
...
+        """
+        Performs validation using the rules from InteractionSerializer
+        if we have a full record.
+
+        Errors are mapped to CSV fields where possible. If not possible, they are
+        added to NON_FIELD_ERRORS (but this should not happen).
+        """
+        if not self.is_valid() or not self.cleaned_data['contact']:
```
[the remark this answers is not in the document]
reupen: Not sure how that would work as `ValidationError`s would normally be handled by the internal Django cleaning logic in the form. I have added a `is_valid_and_matched` method to make things a bit clearer. Let me know if you have a better idea.

datahub/interaction/admin_csv_import/row_form.py
reupen: The problem was with `HasAssociatedInvestmentProjectValidator` (as, unfortunately, it's tied to DRF views). I have reworked the logic so it uses all of the validators except that one, so it should be better and clearer now.

datahub/interaction/admin_csv_import/row_form.py
reupen: Done.

datahub/interaction/admin_csv_import/row_form.py
reupen: Done, except that I have still kept subject as a variable, as I think the logic gets lost when it's put directly into a dict literal like that.

datahub/interaction/serializers.py
reupen: I have reworked it so that it uses all the validators except for `HasAssociatedInvestmentProjectValidator`. That one is now excluded using the serialiser context. I used `is_bulk_import` for the context key but could rename it if anyone has a better suggestion.

datahub/interaction/admin_csv_import/row_form.py
currycoder: Cool, this looks much neater now!  Feel free to re-request me for review when you are ready - I'm keen to see this pattern merge for validating imported interactions as there's probably a lot that I can lift from it for the calendar interactions project. :)

PR #1684, opened 2019-05-03
Email ingestion task and sender validation

currycoder:
### Description of change

This PR adds the following functionality in aid of the calendar interaction ingestion project:
- A new MailboxManager class which uses a new django setting `MAILBOXES` to create and manage Mailbox instances.
- A new `ingest_emails` celery task which uses `MailboxManager` as a gateway to orchestrate imap inboxes and process new emails.
- An `email_sent_by_dit` validation function for verifying that an email (returned by the Mailbox class) was sent by a valid DIT email.

This does not switch on the email ingestion functionality (that will be in a future PR).  Nor does it add any concrete processor classes for dealing with new emails (another future PR).  This should conclude the overarching logic to allow datahub to ingest and process emails from IMAP inboxes - i.e. 'the framework'.

### Points of note
- I couldn't find anywhere that recorded valid DIT email domains for the `email_sent_by_dit` validator.  It would be good to have feedback on whether this is present in the codebase or where this should sit to keep things DRY.
- I'm not entirely happy with the `_verify_authentication` helper which accompanies `email_sent_by_dit`.  I've held off improving it too much at this stage as I still do not have a final IT-provided IMAP inbox to test against (I've been testing against a personal GSuite inbox instead).  My suspicion is that the authentication header could be a little different from what gmail is providing so I'd prefer to avoid improving this too early.
- It would be good to have a little discussion around my use of `django_pglocks.advisory_lock`.  My understanding is that my current usage of it will mean that the lock will be relinquished when the context manager is exited.  It would be good to verify that there is no opportunity for a celery worker process to hold on to the lock indefinitely.

### Checklist


datahub/email_ingestion/mailbox.py
```diff
@@ -217,3 +220,72 @@ def process_new_mail(self):
...
+        """
+        self.mailboxes = {}
+        self.initialise_mailboxes()
+
+    def initialise_mailboxes(self):
+        """
+        Initialise all of the mailboxes detailed in the MAILBOXES django setting.
+        """
+        for mailbox_name, config in settings.MAILBOXES.items():
+            properly_configured = (
```
reupen: Would simply this to something like `properly_configured = config.keys() >= {'email', 'password', 'imap_domain'}`

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,50 @@
...
+                assert line.startswith('spf=pass')
+                assert line.endswith('smtp.mailfrom=%s;' % from_email)
+                auth_results['spf'] = True
+            if line.startswith('dmarc'):
+                assert line.startswith('dmarc=pass')
+                auth_results['dmarc'] = True
+        except AssertionError:
+            return False
+    all_auth_pass = all(auth_results.values())
+    if not all_auth_pass:
```
reupen: Could be simplified to `return all_auth_pass`.

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,50 @@
...
+    ]
+    auth_results = {
+        'dkim': False,
+        'spf': False,
+        'dmarc': False,
+    }
+    for line in authentication_lines:
+        try:
+            if line.startswith('dkim'):
+                assert line.startswith('dkim=pass')
```
reupen: Would not suggest using assertions for checks that are important (one reason is that they are stripped out if you run python with `-O`). 

Here they are critical to the logic of the function. This one could be changed to just `auth_results['dkim'] = line.startswith('dkim=pass')` (for example).

Perhaps you could

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,50 @@
...
+                auth_results['dkim'] = True
+            if line.startswith('spf'):
+                assert line.startswith('spf=pass')
+                assert line.endswith('smtp.mailfrom=%s;' % from_email)
```
reupen: It would be good to add a comment about the purpose of this line (since it's additional to the `spf=pass` check).

Also would be good to make it an f-string (instead of using the %-formatting).

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,50 @@
...
+    Checks whether an email message was sent by a valid DIT address.
+
+    :param message: mailparse.MailParser object - the message to check
+
+    :returns: True if the email was sent by DIT, False otherwise.
+    """
+    from_email = message.from_[0][1]
+    # TODO: See if valid domains are recorded elsewhere in the codebase
+    # and use this to keep things DRY
+    dit_domains = ['@trade.gov.uk', '@digital.trade.gov.uk']
```
reupen: It would be good to define these in the settings.

datahub/email_ingestion/mailbox.py
```diff
@@ -217,3 +220,72 @@ def process_new_mail(self):
...
+    ```
+    from django_pglocks import advisory_lock
+
+    with advisory_lock('ingest_emails', wait=False) as acquired:
+        if not acquired:
+            logger.info('Emails are already being ingested by something else')
+            return
+        mailbox.process_new_mail()
+    ```
+    """
```
marcofucci: I personally find that using a logic similar to what django does with databases for example is a cleaner solution to managers (I used manager many times).
See [ConnectionHandler](https://github.com/django/django/blob/master/django/db/utils.py#L135) (also see its dict-like implementation) and [how it's used / initialised](https://github.com/django/django/blob/master/django/db/__init__.py#L16).

I think it's cleaner because 1. if checks at startup if there are problems, 2. devs don't need to instantiate managers when needed so it's clear that it's a singleton 3. doesn't introduce another concept (manager).

currycoder: I like the rationale here.  In practical terms I'm thinking of renaming this class to `MailboxHandler` and instantiating it as a singleton in `email_ingestion/__init__.py`.  

I was originally intending on this being a singleton but ended up moving the class instantiation in to the calling task since I could only envision it being needed in the task for the near future.

datahub/email_ingestion/mailbox.py
currycoder: :candy:

datahub/email_ingestion/validation.py
currycoder: Thanks, didn't know there was a way to prevent assertion checks in python.

currycoder:
This is ready for re-review following syntactic fixups from Reupen's suggestions and reworking the MailboxManager to be a MailboxHandler singleton as per Marco's

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,49 @@
...
+        'spf': False,
+        'dmarc': False,
+    }
+    for line in authentication_lines:
+        try:
+            if line.startswith('dkim'):
+                auth_results['dkim'] = line.startswith('dkim=pass')
+            if line.startswith('spf'):
+                spf_valid = (
+                    line.startswith('spf=pass') and line.endswith('smtp.mailfrom=%s;' % from_email)
```
marcofucci: did you forget to use the f string here?

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,49 @@
...
+                )
+                auth_results['spf'] = spf_valid
+            if line.startswith('dmarc'):
+                auth_results['dmarc'] = line.startswith('dmarc=pass')
+        except AssertionError:
```
marcofucci: Do you need `AssertionError`?

currycoder: Ooops - missed this in my assertion refactor

datahub/email_ingestion/test/test_validation.py
```diff
@@ -0,0 +1,106 @@
...
+                'dkim=pass header.i=@trade.gov.uk header.s=selector1 header.b=foobar;',
+                'spf=pass (google.com: domain of bill.adama@trade.gov.uk designates ',
+                'XX.XXX.XX.XX as permitted sender) smtp.mailfrom=foobar@trade.gov.uk;',
+                'dmarc=pass (p=QUARANTINE sp=QUARANTINE dis=NONE) header.from=trade.gov.uk',
+            ]),
+            False,
+        ),
+    ),
+)
+def test_email_sent_by_dit(email, authentication_results, is_valid):
```
marcofucci: Small thing but we tend to prefix the expected param with `expected_` so your `is_valid` would become `expected_value` or similar

datahub/email_ingestion/validation.py
```diff
@@ -0,0 +1,46 @@
...
+                line.startswith('spf=pass') and line.endswith(f'smtp.mailfrom={from_email};')
+            )
+            auth_results['spf'] = spf_valid
+        if line.startswith('dmarc'):
+            auth_results['dmarc'] = line.startswith('dmarc=pass')
+    all_auth_pass = all(auth_results.values())
+    return all_auth_pass
+
+
+def email_sent_by_dit(message):
```
marcofucci: Minor thing but I'd call this `was_email_sent_by_dit` to make it clear that the return value is a bool?

reupen:
Just a bit concerned that the amount of work that the Celery task has to do is not really bounded. That's not a huge problem in itself, but if it was running for a while it would block the worker and other tasks could queue up behind it.

I can think of a couple of things that would help:

1. putting an upper limit on how many messages can be processed in one go
2. using sub-tasks to process individual message

I know we're fairly far down this path, but did you have a look at https://imapclient.readthedocs.io/? (It also supports IMAP IDLE.)

currycoder:
> I can think of a couple of things that would help:
> 
> 1. putting an upper limit on how many messages can be processed in one go
> 2. using sub-tasks to process individual messages

I think this is a good point that you raise @reupen - it would be good to discuss in further detail.  How many celery workers are there on each of the environments?

The original plan for this work separated email ingestion from processing by using separate tasks, but along the way it became a little more complex to achieve.  We could definitely look at this as a further iteration on this work.

One thing that I'm also considering is whether we could put the email ingestion task on a different queue?  I see in `Procfile` that there are workers for the default celery queue and a "long-running" queue.  I could either assign this to the long-running queue or to it's own email ingestion queue.  This should avoid congestion of the worker(s) servicing the default queue.  What do you think?

config/settings/common.py
```diff
@@ -520,3 +520,18 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
...
+    #    "email": env("MAILBOX_MEETINGS_EMAIL", default=''),
+    #    "password": env("MAILBOX_MEETINGS_PASSWORD", default=''),
+    #    "imap_domain": env("MAILBOX_MEETINGS_IMAP_DOMAIN", default=''),
+    #    "processor_classes": [
+    #        "datahub.interaction.email_processors.CalendarInteractionEmailProcessor",
+    #    ],
+    #},
+}
+
+DIT_EMAIL_DOMAINS = ['trade.gov.uk', 'digital.trade.gov.uk']
```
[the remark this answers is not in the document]
currycoder: Done!

reupen:
> I think this is a good point that you raise @reupen - it would be good to discuss in further detail. How many celery workers are there on each of the environments?

Should be one node, four processes for the default queue.

> The original plan for this work separated email ingestion from processing by using separate tasks, but along the way it became a little more complex to achieve. We could definitely look at this as a further iteration on this work.

:+1: 

> One thing that I'm also considering is whether we could put the email ingestion task on a different queue? I see in Procfile that there are workers for the default celery queue and a "long-running" queue. I could either assign this to the long-running queue or to it's own email ingestion queue. This should avoid congestion of the worker(s) servicing the default queue. What do you think?

I think it should be OK in the default queue (at least for now). The long-running queue was because the ES tasks can take a long time (10s of minutes plus) so if a few of those tasks were running it would completely block all worker processes. We should be OK here I think (and we also have the number of worker processes and worker instances as other knobs to fiddle with).

currycoder:
Thanks for talking it over @reupen.  I'm going to go ahead and merge once the other approvals are in.

We can always pick up this conversation again in the future PRs to roll out this work - namely the one which switches it on with a celery beat setting :)

PR #1691, opened 2019-05-07
Large Capital profile fix required checks

marcuspp:
###  Issue raised
This is a problem when using the UI as a user can select Cleared add a date and save, the user then goes back in and selects Issues identified and immediately saves, at this point the API should error but doesn't as the date has previously been saved via Cleared. In other words Issues identified now has the date from Cleared.

In short, you shouldn't be able to save either Cleared or Issues identified without adding an accompanying date.

### Description of change
 - The required checks conducted fields/validation have been split from the `LargeCapitalInvestorProfileSerializer` as this functionality will be reused on UK Opportunities.
- Validation was then reworked to only use the validators rather than field validation function to hopefully ensure all errors are returned at once. 

### To test
- Ensure you have the latest Frontend code running (corresponding PR has only just been merged)
- Go to http://localhost:3000/companies/375094ac-f79a-43e5-9c88-059a7caa17f0/investments/large-capital-profile
- Create a profile
- Save an empty profile
- Select  Required checks conducted `Cleared` and try and save (should see an error)
- Add an adviser and date and save
- Change to `Issue identified` and click save, You should see an error

### Checklist


changelog/investment/required_checks_conducted.api.rst
```diff
@@ -0,0 +1,3 @@
+The validation for the endpoint `PATCH /v4/investor-profile/` has been updated.
```
reupen: Backticks here :slightly_smiling_face:  (Should be double rather than single...)

datahub/core/test/test_validate_utils.py
```diff
@@ -40,6 +49,135 @@ def test_is_not_blank(value, blank):
     assert is_not_blank(value) == blank
 
 
+@pytest.mark.parametrize(
```
reupen: I'd suggest using `@freeze_time` rather than relying on `date.today()` (as a general principle) to make sure that the tests are deterministic (and don't depend on what the current date is).

Also, the ones in the params will be evaluated at import time, so e.g. if you start the tests at 23:55 and you might get failures from tests that run after midnight (although it would be an uncommon occurence).

An example of using `@freeze_time` is in https://github.com/uktrade/data-hub-leeloo/blob/31f58d040838c5828ac8fd0a72053668d165ac99/datahub/cleanup/test/test_cleanup_config.py

datahub/core/test/test_validate_utils.py
```diff
@@ -40,6 +49,135 @@ def test_is_not_blank(value, blank):
...
+    'data_date,expected_result',
+    (
+        (
+            date.today() + timedelta(days=1),
```
reupen: I would cover the case when the date is the current date as well (as that should be one of the boundaries for True or False).

datahub/core/validate_utils.py
```diff
@@ -94,3 +96,53 @@ def is_not_blank(value):
...
+
+
+def _is_date_less_than_a_year_ago(value):
+    """Returns True if the date value is within the last year."""
+    if not value:
+        return False
+    return _is_date_in_the_past(value) and (value > date.today() - relativedelta(years=1))
+
+
+def is_field_present_in_post_data(combiner, field_name):
```
reupen: Will it necessarily be a POST request?

datahub/core/validate_utils.py
```diff
@@ -94,3 +96,53 @@ def is_not_blank(value):
...
+    """Returns True of the field is in the post data."""
+    if field_name in combiner.data:
```
reupen: Could be simplified to just `return field_name in combiner.data`.

datahub/core/validators/rules_based.py
```diff
@@ -61,6 +61,14 @@ def __call__(self, combiner) -> bool:
         return self._operator(value)
 
 
+class InstanceAwareOperatorRule(OperatorRule):
```
reupen: What does 'instance-aware' mean? It would be good to explain a bit (or maybe another name would help?)

Also, not sure about inheriting from `OperatorRule` since the required signature of `self.operator_` is different so the docstring of `OperatorRule.__init__()` doesn't apply.

datahub/core/validate_utils.py
```diff
@@ -94,3 +96,53 @@ def is_not_blank(value):
...
+def _is_date_in_the_past(value):
+    """Returns True if the date value is today or in the past."""
+    if not value:
+        return False
+    return date.today() >= value
+
+
+def _is_date_less_than_a_year_ago(value):
+    """Returns True if the date value is within the last year."""
+    if not value:
```
reupen: Is this checking two things (that the value is not None and it's in the last year)?

Could have missed something (and maybe there is some overlap with how it currently works), but I think I would think of it as two separate rules (with different error messages):

1. the field value can't be None under condition X
2. the field value must be in the last twelve months under condition X and if the value is not None

datahub/core/validate_utils.py
```diff
@@ -94,3 +96,53 @@ def is_not_blank(value):
...
+        return False
+    return _is_date_in_the_past(value) and (value > date.today() - relativedelta(years=1))
```
reupen: I think we could get rid of `_is_date_in_the_past()` and go for something like:

```python
today = date.today()
return today - relativedelta(years=1) < value <= today
```

marcuspp: I have separated this as I would like to use date in the past elsewhere. `actual_land_date`

datahub/core/validate_utils.py
marcuspp: Yes but is that a problem? If you phrase the question like "is the value provided a date less than a year ago?" (which is the intention) then checking and returning False for a None is a valid.

There is a separate check in validators in the serializer for part 1.

datahub/core/validate_utils.py
marcuspp: Updated. Removed post from function names and updated comments.

datahub/investment/investor_profile/serializers.py
```diff
@@ -104,8 +111,119 @@ class LargeCapitalInvestorProfileSerializer(serializers.ModelSerializer):
...
+                    ),
+                ),
+                ValidationRule(
+                    'invalid_required_checks_conducted_by',
+                    InstanceAwareOperatorRule(
+                        'required_checks_conducted_by',
+                        is_field_not_blank_in_data,
+                    ),
+                    when=AndRule(
+                        InstanceAwareOperatorRule(
```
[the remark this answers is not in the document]
marcuspp: I like this idea. Do you think there is a need to reference in the name to the fact the rule only applies to the data dict and not what is currently saved. I'm considering do the same for `FieldIsNotBlank`?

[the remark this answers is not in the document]
marcuspp: I suppose `IsBeing` suggests that.

datahub/core/validators/rules_based.py
marcuspp: I've updated and gone is a `IsFieldRule` I wanted to try and follow one convention the `IsField...` that we established below. If you'd like more details add in the comments let me know. I'm happy to change this if you think a different name would make things clearer.

reupen:
Just forget to mention something to be aware of – not sure if it will be a problem, but some users will be in a time zone ahead of the server (in UTC) so today for them could potentially be `date.today()` + 1 day.

datahub/investment/investor_profile/test/test_views.py
```diff
@@ -583,3 +620,62 @@ def test_patch_large_capital_required_checks_conducted_by_error(self):
...
+        ),
+    )
+    @freeze_time('2011-01-01')
+    def test_patch_large_capital_required_checks_conducted_by_error_update(
+        self,
+        request_data,
+        expected_error_response,
+    ):
+        """
+        Test updating required checks conducted by cannot be set when required checks
```
elcct: I feel like something is missing in this doc string.

datahub/core/validators/rules_based.py
```diff
@@ -36,6 +36,62 @@ def field(self):
...
+        """
+        Returns True if the field is being updated and the value is not blank.
+        """
+        if self.field not in combiner.data:
+            return False
+        return is_not_blank(combiner.data[self.field])
+
+
+class IsFieldRule(BaseRule):
+    """Rule to check if a field meets a provided condition."""
```
elcct: I think it would be great to expand a little what provided condition is, e.g. a function that is passed as a parameter. It took me a little time to understand what it is for as I couldn't get it from the doc string.

datahub/core/validators/rules_based.py
```diff
@@ -36,6 +36,62 @@ def field(self):
...
+
+        Checks the post data to see if the field has been supplied then if
+        so checks the current value to see if it is being updated.
+        """
+        if self.field not in combiner.data:
+            return False
+        value = combiner.data[self.field]
+        if combiner.instance and getattr(combiner.instance, self.field) == value:
+            return False
+        return True
```
elcct: This could be simplified to something like:

```
if self.field not in combiner.data or not combiner.instance:
  return False
return getattr(combiner.instance, self.field) == combiner.data[self.field]
``

File: datahub/core/validate_utils.py [hunk 13]
```diff
@@ -94,3 +96,28 @@ def is_not_blank(value):
 @lru_cache()
 def _get_model_field_info(model):
     return model_meta.get_field_info(model)
+
+
+def _is_provided_and_is_date_in_the_past(value):
```
reupen: Main last comment from me is that I'm not sure how directly application these three functions will be to other apps, so it might be better to keep them in `datahub.investment` for now (and reconsider later if needed).

datahub/core/validators/rules_based.py
marcuspp: Updated.

datahub/investment/investor_profile/test/test_views.py
marcuspp: Updated

datahub/core/validate_utils.py
marcuspp: Sure I can move them if thats preferred.

marcuspp:
> Just forget to mention something to be aware of – not sure if it will be a problem, but some users will be in a time zone ahead of the server (in UTC) so today for them could potentially be `date.today()` + 1 day.

@reupen 🤔 True. The majority of users will be UK based but you are correct this could happen but would be an extreme edge case. I will add a comment to warn other developers of this condition.

datahub/core/validators/rules_based.py
```diff
@@ -36,6 +36,62 @@ def field(self):
...
+                          value is valid). Will be called with the combiner and the
+                          field name as arguments.
+        """
+        super().__init__(field)
+        self._function = function_
+
+    def __call__(self, combiner) -> bool:
+        """Test whether the rule passes or fails."""
+        if self.field not in combiner.data:
+            return False
```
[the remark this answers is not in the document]
marcuspp: I could make it configurable if you'd prefer. But I would've thought if there was a requirement to return value if the value is not in combiner.data then this code could be amended accordingly then?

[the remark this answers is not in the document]
marcofucci: I was thinking that you should be able to delete these 2 lines without any problem. The tests related to the views pass without as `self._function` checks for that as well.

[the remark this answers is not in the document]
marcofucci: Well, it was in the previous version :-/

PR #1693, opened 2019-05-09
Remove trading address fields from django

alixedi:
### Description of change

Removes the following fields from the codebase:

```
trading_address_1
trading_address_2
trading_address_town
trading_address_county
trading_address_country
trading_address_postcode
```

Includes the migration that sets these fields as `NULLABLE` in the DB and removes them from Django state.

### Checklist


datahub/company/test/test_company_views_archiving.py
```diff
@@ -36,12 +36,10 @@ def test_archive_company_reason(self):
 
     def test_archive_company_invalid_address(self):
         """
-        Test archiving a company when the company has an invalid trading address and missing
```
reupen: The test name is still `test_archive_company_invalid_address` – it might be better to change the test to set `address_town` to `''` so it still tests the invalid address case (the original problem was that companies with an invalid address could not be archived).

datahub/company/test/test_company_views_archiving.py
```diff
@@ -36,12 +36,10 @@ def test_archive_company_reason(self):
...
-        UK region.
+        Test archiving a company when the company has a missing UK region.
         """
         company = CompanyFactory(
             registered_address_country_id=Country.united_kingdom.value.id,
```
reupen: Not directly related to the changes in this PR but this should be `address_country_id` rather than `registered_address_country_id` (as `address` is now the one used to determine if `uk_region` is required: https://github.com/uktrade/data-hub-leeloo/blob/47fde0ebccf1272d397a94a971f0a771a316685b/datahub/company/serializers.py#L558-L565).

datahub/company/test/test_company_views_archiving.py
```diff
@@ -58,12 +56,10 @@ class TestUnarchiveCompany(APITestMixin):
 
     def test_unarchive_company_invalid_address(self):
         """
-        Test unarchiving a company when the company has an invalid trading address and missing
-        UK region.
+        Test unarchiving a company when the company has a missing UK region.
         """
         company = CompanyFactory(
             address_country_id=Country.united_kingdom.value.id,
-            trading_address_town='',
```
reupen: Same comment as for `test_archive_company_invalid_address` – probably would be better to set `address_town` instead,

PR #1696, opened 2019-05-09
Add preview page to import interactions tool

reupen:
### Description of change

This adds a preview page to the admin site import interactions tool. This displays:

* counts of the loaded rows by contact matching status
* the first 100 rows that will be imported

If no interactions were matched to contacts, a slightly different template is used.

(The actual saving will be added in an upcoming PR.)

[(Previous related PRs.)](https://github.com/uktrade/data-hub-leeloo/pulls?q=is%3Apr+is%3Aclosed+label%3A%22import+interactions%22)

Any suggestions for how to simplify some of the `blocktrans` blocks in the templates?

### To test

1. Create a valid or invalid CSV file according to the spec on the page
1. Select your CSV file
1. Submit the form

```
Digital Data Hub - Live Service,<an adviser name>,,fooimport@bar.com,Email/Website,Account managemenT,interaction,22/02/2019,,some notes
```

#### Ready to import interactions

image

#### No matched contacts

image

### Checklist


datahub/interaction/templates/admin/interaction/interaction/import_preview.html
```diff
@@ -0,0 +1,106 @@
...
+
+  <ul>
+    <li>
+      {% blocktrans with num_matched_human=num_matched|intcomma count num_matched=num_matched %}
+        {{ num_matched_human }} interaction is ready to import
+      {% plural %}
+        {{ num_matched_human }} interactions are ready to import
+      {% endblocktrans %}
+    </li>
+    <li>
```
currycoder: IMO, both of these "x records skipped" statements end up as noise if nothing has been skipped.  Would be good to only show them if they provide some information.

currycoder: Also, the only real difference between these blocks and the blocks in `import_no_matches.html` is the tense...  Could they be combined in to include template snippets to avoid repitition?  e.g. `{{ num_unmatched_human }} record skipped as it could not be matched with an active contact`

reupen: I'll change it to omit the messages if the counts are zero :+1: 

Regarding the second point, I don't like the repetition either but would prefer to maintain the tense distinction for clarity. Haven't been able to come up with something that does that and doesn't add complexity at the same time...

currycoder: Sounds good. :+1:  It's a shame we can't just refactor the english language.

reupen: Came up with something actually – just pushed it, it should be better :slightly_smiling_face:

datahub/interaction/templates/admin/interaction/interaction/import_preview.html
```diff
@@ -0,0 +1,95 @@
...
+  {{ block.super }}
+  <link rel="stylesheet" type="text/css" href="{% static 'interaction/admin/css/import-preview.css' %}">
+{% endblock %}
+
+{% block content %}
+  <h2>{% trans 'Ready to import interactions' %}</h2>
+  <p>{% trans 'The file was successfully validated and is now ready to import.' %}</p>
+
+  <ul>
+    {% trans 'interaction is ready to import' as matched_singluar_text %}
```
marcuspp: typo `singular`

datahub/interaction/templates/admin/interaction/interaction/import_no_matches.html
```diff
@@ -0,0 +1,30 @@
...
+  <link rel="stylesheet" type="text/css" href="{% static 'interaction/admin/css/import-preview.css' %}">
+{% endblock %}
+
+{% block content %}
+  <h2>{% trans 'No matching contacts found' %}</h2>
+
+  <p>{% trans 'No interactions have been imported as none of the loaded interactions could be matched to active contacts.' %}</p>
+
+  <ul>
+    {% trans 'record was skipped as it could not be matched with an active contact' as unmatched_singluar_text %}
```
marcuspp: typo `singular`

datahub/interaction/templates/admin/interaction/interaction/import_preview.html
reupen: Thanks – caught some of those before I pushed but missed most of them :disappointed:

datahub/interaction/admin_csv_import/views.py
```diff
@@ -90,6 +91,23 @@ def _error_list_response(self, request, errors):
...
+        matching_counts, matched_rows = form.get_matching_summary(MAX_PREVIEW_ROWS_TO_DISPLAY)
+        num_matched = matching_counts[ContactMatchingStatus.matched]
+
+        template_filename = 'import_preview.html' if num_matched else 'import_no_matches.html'
+
+        return self._template_response(
+            request,
+            f'admin/interaction/interaction/{template_filename}',
+            PAGE_TITLE,
+            num_matched=num_matched,
```
alixedi: This can be tighter with a bit of work:

```
return self._template_response(
   ...
    **matching_counts,
   ...
)
```

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -25,3 +27,33 @@ def get_row_error_iterator(self):
...
+                if not form.is_valid():
+                    # This should not happen. Just raise an exception to alert us if it does.
+                    raise DataHubException('CSV row unexpectedly failed revalidation')
+
+                contact_matching_status = form.cleaned_data['contact_matching_status']
+                matching_counts[contact_matching_status] += 1
+
+                is_row_matched = contact_matching_status == ContactMatchingStatus.matched
+
+                if is_row_matched and len(matched_rows) < max_rows:
```
alixedi: This seems like we are bringing presentation logic too deep down the call stack. I would reckon this function shouldn't care about `max_rows`. We can truncate `matched_rows` in the caller or even better - inside the template.

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -25,3 +27,33 @@ def get_row_error_iterator(self):
...
+    def get_matching_summary(self, max_rows):
+        """
+        Get a summary of the contact matching results of the rows.
+
+        This is used as part of the preview page displayed once a file has been uploaded.
+
+        :returns: dict of counts by `ContactMatchingStatus` and list of matched rows as
+            serializer dicts
+        """
+        matching_counts = {status: 0 for status in ContactMatchingStatus}
```
alixedi: Opportunity to use https://docs.python.org/2/library/collections.html#collections.Counter here.

datahub/interaction/test/admin_csv_import/utils.py
```diff
@@ -15,3 +23,99 @@ def random_service(disabled=False):
     return random_obj_for_queryset(
         Service.objects.filter(disabled_on__isnull=not disabled),
     )
+
+
+def make_csv_file(fieldnames, *rows, encoding='utf-8-sig', filename='test.csv'):
```
alixedi: Why is the `*` here?

datahub/interaction/test/admin_csv_import/utils.py
```diff
@@ -15,3 +23,99 @@ def random_service(disabled=False):
...
+        *row_dicts,
+        fieldnames=fieldnames,
+        encoding=encoding,
+        filename=filename,
+    )
+
+
+def make_csv_file_from_dicts(*rows, fieldnames=None, encoding='utf-8-sig', filename='test.csv'):
+    """Make a CSV file from a number of dicts."""
+    if fieldnames is None:
```
alixedi: I think we can enforce an API here that forces us to pass in the `fieldnames`. I can't think of any use case where the caller of this function doesn't have `fieldnames`.

datahub/interaction/test/admin_csv_import/utils.py
alixedi: Also, don't really understand why there seems to be 2 functions both `make_csv_files...` in test utilities. Why can't we have a single function that makes CSV files from a list of `dict`?

datahub/interaction/admin_csv_import/file_form.py
reupen: Do you have a more specific example in mind? I looked at it already but I came to the conclusion that it can't be easily used here:

* we need all members of ContactMatchingStatus to always be in the `dict` (even if the count is zero)
* we don't have an iterable to feed into it so would end up being updated in a similar way anyway

datahub/interaction/admin_csv_import/file_form.py
reupen: One of the reasons was to avoid unbounded memory usage. We have limited memory in the environment the app runs in production.

datahub/interaction/admin_csv_import/views.py
reupen: Nice idea, but won't work as the keys of matching_counts` aren't strings.

datahub/interaction/test/admin_csv_import/utils.py
reupen: > Why is the * here?

Before, `fieldnames` was in rows, I just brought it out of there. Is there a specific problem you are seeing?

> Also, don't really understand why there seems to be 2 functions both `make_csv_files...` in test utilities. Why can't we have a single function that makes CSV files from a list of `dict`?

It can probably go, I just didn't want to update all the existing tests in this PR (when there are more on the way...)

datahub/interaction/test/admin_csv_import/utils.py
reupen: I think what matters is what makes the tests clearer... Not overly bothered either way, but right now only `make_csv_file()` is making use of that.

datahub/interaction/admin_csv_import/file_form.py
alixedi: The reason I mentioned Counter was b/c you seem to be initialising counters here (to 0) while:

```
>>> c = Counter()
>>> c['lol']
0
```

datahub/interaction/admin_csv_import/file_form.py
alixedi: Fair enough.

datahub/interaction/test/admin_csv_import/utils.py
alixedi: The reason I am asking about `*` is because we do not seem to want to unpack rows. Do we?

datahub/interaction/test/admin_csv_import/utils.py
alixedi: Ok

datahub/interaction/admin_csv_import/views.py
alixedi: `**{status.name for status in matching_counts}`?

datahub/interaction/admin_csv_import/file_form.py
reupen: We had a short discussion offline – from my point of view, it would be better not to rely on that implicit logic once the `dict` has been returned from the function.

PR #1698, opened 2019-05-10
 Add ability to save InteractionCSVForm data to the cache 

reupen:
### Description of change

This is part of the import interactions tool.

This allows the uploaded file to be saved to the cache while the preview page is being displayed to the user. The cache is used for simplicity as we are only dealing with small files.

This functionality will be used in an upcoming PR that implements the saving of loaded rows.

### To test

The functionality is not used yet so there is nothing to manually test.

### Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -123,7 +123,12 @@ def local_memory_cache(monkeypatch):
         'default',
         {'BACKEND': 'django.core.cache.backends.locmem.LocMemCache'},
     )
-    monkeypatch.setattr('django.core.cache.caches', CacheHandler())
+    cache_handler = CacheHandler()
+    monkeypatch.setattr('django.core.cache.caches', cache_handler)
+
+    yield
+
+    cache_handler['default'].clear()
 
 
 @pytest.fixture
--- a/datahub/interaction/admin_csv_import/file_form.py
+++ b/datahub/interaction/admin_csv_import/file_form.py
@@ -1,10 +1,19 @@
+import gzip
+from datetime import timedelta
+from secrets import token_urlsafe
+
 from django.conf import settings
+from django.core.cache import cache
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.template.defaultfilters import filesizeformat
 
 from datahub.company.contact_matching import ContactMatchingStatus
 from datahub.core.admin_csv_import import BaseCSVImportForm
 from datahub.core.exceptions import DataHubException
 from datahub.interaction.admin_csv_import.row_form import InteractionCSVRowForm
+
+
+FILE_CACHE_TIMEOUT_SECS = int(timedelta(minutes=30).total_seconds())
 
 
 class InteractionCSVForm(BaseCSVImportForm):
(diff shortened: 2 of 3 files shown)


conftest.py
```diff
@@ -123,7 +123,12 @@ def local_memory_cache(monkeypatch):
         'default',
         {'BACKEND': 'django.core.cache.backends.locmem.LocMemCache'},
     )
-    monkeypatch.setattr('django.core.cache.caches', CacheHandler())
+    cache_handler = CacheHandler()
+    monkeypatch.setattr('django.core.cache.caches', cache_handler)
+
+    yield
+
+    cache_handler['default'].clear()
```
currycoder: Neat pattern

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -57,3 +67,55 @@ def get_matching_summary(self, max_rows):
                     matched_rows.append(form.cleaned_data_as_serializer_dict())
 
         return matching_counts, matched_rows
+
+    def save_to_cache(self):
+        """
+        Generate a token and stores the file in the configured cache with a timeout.
+
+        Can only be called on a validated form.
```
currycoder: You say this..  Is it worth making this explicit through raising an exception with a meaningful error message?  Otherwise you are depending on inferring the problem from an `AttributeError` when an attempt to access `self.cleaned_data` happens

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -57,3 +67,55 @@ def get_matching_summary(self, max_rows):
...
+    def from_token(cls, token):
+        """
+        Create a InteractionCSVForm instance using a token.
+
+        Returns None if serialised data for the token can't be found in the cache.
+        """
+        data_key, name_key = _cache_keys_for_token(token)
+        cache_keys_and_values = cache.get_many((data_key, name_key))
+
+        if {data_key, name_key} - cache_keys_and_values.keys():
```
currycoder: I feel like this could be more self documenting.  I realise that it is succinct right now but just splitting out the line in to a separate variable could help IMO, e.g.
```
cache_value_missing = {data_key, name_key} - cache_keys_and_values.keys()
if cache_value_missing:
...
```

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -57,3 +67,55 @@ def get_matching_summary(self, max_rows):
...
+        compressed_data = gzip.compress(data)
+
+        token = _make_token()
+        data_key, name_key = _cache_keys_for_token(token)
+        cache_keys_and_values = {
+            data_key: compressed_data,
+            name_key: csv_file.name,
+        }
+        cache.set_many(cache_keys_and_values, timeout=FILE_CACHE_TIMEOUT_SECS)
+        return token
```
currycoder: I'm finding it a little hard to understand how the token will be used by the surrounding views.  Presumably it will be incorporated in to the URL for the preview page.  Could you shed some extra light on this?

reupen: Yes – I was planning on using a URL path of `/admin/interaction/interaction/import/<token>/save` for the form submission that does the saving. It may be used for downloading unmatched contacts as well.

datahub/interaction/admin_csv_import/file_form.py
reupen: My view was that it's what happens generally if you try to access `cleaned_data` before calling `Form.is_valid()` or accessing `Form.errors` so it wasn't worth the extra noise of a check. But can add one if we think it's important.

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -57,3 +67,56 @@ def get_matching_summary(self, max_rows):
...
+
+        any_cache_keys_missing = {data_key, name_key} - cache_keys_and_values.keys()
+        if any_cache_keys_missing:
+            return None
+
+        decompressed_data = gzip.decompress(cache_keys_and_values[data_key])
+        name = cache_keys_and_values[name_key]
+        csv_file = SimpleUploadedFile(name=name, content=decompressed_data)
+
+        return InteractionCSVForm(
```
marcofucci: Non critical: you could use cls here instead of InteractionCSVForm

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -7,6 +13,10 @@
 from datahub.interaction.admin_csv_import.row_form import InteractionCSVRowForm
 
 
+FILE_CACHE_KEY_LENGTH = 32
+FILE_CACHE_TIMEOUT_SECS = int(timedelta(minutes=30).total_seconds())
```
marcofucci: Non critical: this is okay but I find 30 * 60 easier to read and what many other packages tend to use.

datahub/interaction/admin_csv_import/file_form.py
currycoder: Yep I think this is fair!

datahub/interaction/admin_csv_import/file_form.py
reupen: Good spot!

datahub/interaction/admin_csv_import/file_form.py
currycoder: Ok makes sense.  And to double check, there is no way that we will re-use the cached version if the user uploads a new file?  My only observation is that the cache key does not vary on the contents of the file, so this needs to be considered in usage.

datahub/interaction/admin_csv_import/file_form.py
reupen: Not too fussed, I only prefer this because it works better for more complex cases.

datahub/interaction/admin_csv_import/file_form.py
reupen: The chance of the same token being re-used should be virtually nil (it has just over double the bits of randomness than a v4 UUID). [It's considered safe against brute-force attacks](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe) so that happening by accident should be a non-concern.

The intention was to identify the upload operation (or session) rather than the file.

datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -57,3 +66,56 @@ def get_matching_summary(self, max_rows):
                     matched_rows.append(form.cleaned_data_as_serializer_dict())
 
         return matching_counts, matched_rows
+
+    def save_to_cache(self):
+        """
+        Generate a token and stores the file in the configured cache with a timeout.
```
elcct: I think there is a typo there - `stores`.

reupen: :+1: 
Will fix before merging.

PR #1710, opened 2019-05-14
Mechanism to parse interaction data from calendar emails

currycoder:
### Description of change

Adds a mechanism for parsing interaction data from calendar invite emails.  This builds on the email ingestion framework established in previous PRs to lay the groundwork for introducing the first email processor to datahub.

The parsing mechanism will extract the details necessary to build a draft interaction from the invite email.  This PR does not include the code for actually creating the draft interaction - this will follow in the next PR.

### Points of note

- There is actually only ~450 lines of python code for this PR.  The remainder of the additions are from email test files.
- There's a lot of sample email files for test purposes.  I did consider templating these files to reduce the raw number of lines added, but I decided to use files based on actual samples instead as this highlighted some oddities for specific email clients.  We need to be exhaustive here to make sure that we have as many edge cases covered as possible - the email clients that advisers can use are varied and we should accommodate all of the formats that they produce.

### Checklist


reupen:
@currycoder I'm getting some test failures locally with `django.core.exceptions.ValidationError: ['Email not sent by DIT']`. Is it relying on `DIT_EMAIL_DOMAINS` being set in the environment?

There are (at least) a couple of options to avoid that:

* update `config.settings.test`
* override the settings for particular tests

currycoder:
> @currycoder I'm getting some test failures locally with `django.core.exceptions.ValidationError: ['Email not sent by DIT']`. Is it relying on `DIT_EMAIL_DOMAINS` being set in the environment?
> 
> There are (at least) a couple of options to avoid that:
> 
> * update `config.settings.test`
> * override the settings for particular tests

Ok will do - thanks for flagging this.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        Initialise the CalendarInteractionEmailParser with a MailParser email
+        object.
+        """
+        self.message = message
+
+    def _verify_email_sent_by_dit(self):
+        if not was_email_sent_by_dit(self.message):
+            raise ValidationError('Email not sent by DIT')
+
+    def _get_adviser_from_email(self, email):
```
reupen: I often see the same email address typed with different capitalisation – do we need to make the look-up case insensitive?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        for field in ['contact_email', 'email']:
+            criteria = {field: email}
+            try:
+                return Advisor.objects.get(**criteria)
+            except Advisor.DoesNotExist:
```
reupen: Doesn't look like we have `unique=True` on `contact_email` so we may need to think about that as well.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _extract_sender_adviser(self):
+        sender_email = self.message.from_[0][1]
+        sender_adviser = self._get_adviser_from_email(sender_email)
+        if not sender_adviser:
+            raise ValidationError('Email not sent by recognised DIT Adviser')
+        return sender_adviser
+
+    def _get_all_recipients(self):
+        all_recipients = set()
```
reupen: Looks like a good candidate for a comprehension e.g. (not tested):

```
from itertools import chain

return {
   recipient[1].strip() for recipient in chain(self.message.to, self.message.cc)
}
```

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        contacts = []
+        for recipient_email in all_recipients:
+            contact, matching_status = find_active_contact_by_email_address(recipient_email)
+            if contact:
+                contacts.append(contact)
+        if not contacts:
+            raise ValidationError('No email recipients were recognised as Contacts')
+        return contacts
+
+    def _extract_secondary_advisers(self, all_recipients, sender_adviser):
```
reupen: What does secondary mean in this context? Might be good to add an explanation.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        for field in ['contact_email', 'email']:
+            criteria = {field: email}
+            try:
+                return Advisor.objects.get(**criteria)
+            except Advisor.DoesNotExist:
+                continue
+        return None
+
+    def _extract_sender_adviser(self):
+        sender_email = self.message.from_[0][1]
```
alixedi: This seem a bit fragile. Any chance of an `IndexError` here?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        interaction_data['company'] = self._get_company_from_contacts(interaction_data['contacts'])
+        calendar_event = self._get_calendar_event_metadata()
+        interaction_data['date'] = calendar_event['start']
+        interaction_data['location'] = calendar_event['location']
+        interaction_data['meeting_details'] = calendar_event
+        interaction_data['subject'] = calendar_event['subject']
+        meeting_confirmed = calendar_event['status'] == 'CONFIRMED'
+        if not meeting_confirmed:
+            raise ValidationError('Calendar event was not status: CONFIRMED')
+        return interaction_data
```
marcofucci: I think it would be good to introduce some kind of convention. There are quite a few methods, some starting with `_extract` some with `_get` and some with `_grab`. Also, some only extract data, some extract and do some partial validation and some complete some other validation. Finally, some use `self.message` and some only use the given arguments.
I would:
- have some type of convention on names
- either have methods that only extract data and other methods and use the first group and add validation *or* have methods that do extraction and validation at the same time
- maybe general purpose methods can become functions of the module if that helps?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+                secondary_advisers.append(adviser)
+        return secondary_advisers
+
+    def _get_company_from_contacts(self, contacts):
+        """
+        Get the company from the given contacts.
+        If these contacts are related to different companies, return the company
+        with the highest number of contacts in our iterable.
+        """
+        company_counts = Counter([contact.company for contact in contacts])
```
reupen: Just a small thing but `Counter(contact.company for contact in contacts)` will also work.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _grab_calendar_string_from_text(self):
+        for text_part in self.message.text_plain:
+            if text_part.startswith('BEGIN:VCALENDAR'):
+                return text_part
+        return False
+
+    def _grab_calendar_string_from_attachments(self):
+        for attachment in self.message.attachments:
+            if attachment['mail_content_type'] == 'application/ics':
```
reupen: Would be good to defined a named constant for `'application/ics'`.

datahub/interaction/email_processors.py
reupen: `sender_adviser` is unused here.

datahub/interaction/email_processors.py
alixedi: Please consider:

`set([email.strip() for _, email in (*self.message.to, *self.message.cc)])`

alixedi: Also, it would be good to find out what `_` is here and is it possible for the tuple to only have the email?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        for recipients in [self.message.to, self.message.cc]:
+            all_recipients = all_recipients.union(
+                {recipient[1].strip() for recipient in recipients},
+            )
+        return all_recipients
+
+    def _extract_contacts(self, all_recipients):
+        contacts = []
+        for recipient_email in all_recipients:
+            contact, matching_status = find_active_contact_by_email_address(recipient_email)
```
alixedi: Please consider using `in` https://stackoverflow.com/a/5482050 to reduce the number of queries here to 1.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+    def _get_calendar_event_component(self):
+        error = ValidationError('No calendar event could be extracted')
+        calendar_string = self._grab_calendar_string_from_text()
+        if not calendar_string:
+            calendar_string = self._grab_calendar_string_from_attachments()
+        if not calendar_string:
+            raise error
+        try:
+            calendar = icalendar.Calendar.from_ical(calendar_string)
+        except Exception as exc:
```
alixedi: Can we catch a more specific exception here?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+            raise error from exc
+        calendar_event_components = []
```
alixedi: Please consider:

`calendar_event_components = [comp for comp in calendar.walk() if comp.name =='VEVENT']`

or

`calendar_event_components = filter(lambda comp: comp.name == 'VEVENT', calendar.walk())`

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        if len(calendar_event_components) > 1:
+            raise ValidationError(
+                f'There were {len(calendar_event_components)} events in the calendar '
+                '- expected 1 event',
+            )
+        return calendar_event_components[0]
+
+    def _get_calendar_event_metadata(self):
+        event_component = self._get_calendar_event_component()
+        location = event_component.get('location')
```
alixedi: `location = str(event_component.get('location') or '')`

datahub/interaction/email_processors.py
reupen: Is there a reason for `set([...])` rather than a set comprehension?

datahub/interaction/email_processors.py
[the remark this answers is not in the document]
currycoder: After discussing with Reupen, I've moved this logic in to a combined module for contact and adviser matching - refactored from Reupen's `datahub.company.contact_matching` utility module.  The problem was extremely similar conceptually so I think this made the best sense.

currycoder: Done as part of reworking `contact_matching.py` in to `contact_adviser_matching.py`

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _verify_email_sent_by_dit(self):
+        if not was_email_sent_by_dit(self.message):
+            raise ValidationError('Email not sent by DIT')
+
+    def _get_adviser_from_email(self, email):
+        for field in ['contact_email', 'email']:
+            criteria = {field: email}
+            try:
+                return Advisor.objects.get(**criteria)
```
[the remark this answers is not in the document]
currycoder: This was resolved as part of moving the logic to `contact_adviser_matching.py`

currycoder:
Just running through the feedback on this.  Will comment/resolve as I go but please don't re-review until I request!  I'll hopefully do this in a batch later today.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _extract_contacts(self, all_recipients):
+        contacts = []
+        for recipient_email in all_recipients:
+            contact, matching_status = find_active_contact_by_email_address(recipient_email)
+            if contact:
+                contacts.append(contact)
+        if not contacts:
+            raise ValidationError('No email recipients were recognised as Contacts')
+        return contacts
```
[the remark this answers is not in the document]
currycoder: I thought about where this limitation should be represented.  
Although there is still a fair bit of business logic contained within this parser, I thought that it would be better for the parser to responsible for validation/extraction only - and that the caller would need to be responsible for ensuring that these entities were safe to save on an interaction.

So for this case, this limitation is covered by the next PR which has code which instantiates and uses this parser class.  Happy to discuss this further of course.

datahub/interaction/email_processors.py
currycoder: Thanks for pointing this out.  The suggestions were helpful to work out how to tidy the class up a bit.

datahub/interaction/email_processors.py
currycoder: This is fixed up from copying your contact matching pattern for advisors.

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+@pytest.mark.django_db
+class TestCalendarInteractionEmailParser:
+    """
+    Test the CalendarInteractionEmailParser class.
+    """
+
+    def _get_parser_for_email_file(self, relative_email_file_path):
+        email_file_path = os.path.join(
```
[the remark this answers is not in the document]
currycoder: Nice - didn't know about `__file__` - with retrospect my path handling was very python2 :)

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -0,0 +1,269 @@
...
+    fixture = {
+        'contacts': [],
+        'advisers': [],
+    }
+    adviser_emails = [
+        'brendan.smith@trade.gov.uk',
+        'marco.fucci@digital.trade.gov.uk',
+        'ali.zaidi@digital.trade.gov.uk',
+    ]
+    for adviser_email in adviser_emails:
```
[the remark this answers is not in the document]
currycoder: Neat - good tip

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -0,0 +1,269 @@
...
+        return CalendarInteractionEmailParser(message)
+
+    @pytest.mark.parametrize(
+        'email_file,expected_event_details',
+        (
+            (
+                'email_samples/valid/outlook_online/sample.eml',
+                {
+                    'subject': 'test meet',
+                    'start': datetime(2019, 3, 29, 12, 00, tzinfo=pytz.utc),
```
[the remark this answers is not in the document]
currycoder: Thanks - will use this instead of the direct `pytz` import

datahub/interaction/email_processors.py
currycoder: > Also, it would be good to find out what _ is here and is it possible for the tuple to only have the email?

If the email message supplies the item in the following format: 
`"Bill Adama <bill.adama@example.net>"`, the resulting tuple will be `("Bill Adama", "bill.adama@example.net")`

I'm not quite sure (without investigation) about what determines whether the raw email string provides name/email pairs as described or just email values.  From the raw samples that I sourced from live mailboxes/emails, I did not find any messages which supplied the pair format.

currycoder: Thanks for the comprehension feedback here, I've folded that in.

datahub/interaction/email_processors.py
currycoder: I've added an explainer docstring for this helper method:
```
Extract the secondary (non-sender) advisers for the calendar invite - that is,
any advisers that received the invite who did not send it.
```

currycoder: It is used now - I've ensured that the sender's adviser entry is not returned in the result of this call.

datahub/interaction/email_processors.py
currycoder: I don't think this will happen in practice, but there is some chance that an email will be malformed so we might as well catch it.  I've wrapped this in a try/except block.

datahub/interaction/email_processors.py
currycoder: I did think about optimising this lookup to a single query.  I ultimately decided against it for the following reason:
The utility function `find_active_contact_by_email_address` wraps up the approach for "firstly try to find using the `email` field, then try to find using `email_alternative`".  I could enrich that module a little bit with some additional methods for handling bulk lookups, but I'm not anticipating that this will be a massive performance issue as the average meeting will likely only have a handful of recipients.

IMO, this is one of those times where the desire for optimisation collides with the need for simplicity and I couldn't really justify the extra work.  Happy to discuss further if you see this as critical!

datahub/interaction/email_processors.py
currycoder: I believe so, I've changed it to `ValueError` - though I'm a little off-put by icalendar's lack of documentation on exceptions that the logic raises.

currycoder:
Thanks for the feedback!  I believe that I have responded to each comment.  Please consider this as ready for re-review.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,186 @@
...
+
+def _get_utc_datetime(localised_datetime):
+    try:
+        return localised_datetime.astimezone(pytz.utc)
+    # When the calendar event is just a date
+    except AttributeError:
+        return localised_datetime
+
+
+class CalendarInteractionEmailParser:
```
reupen: Is there any more logic planned for this class? Just wondering if it would be simpler to make `extract_interaction_data_from_email()` a module-level function that takes a message.

datahub/company/contact_adviser_matching.py
```diff
@@ -0,0 +1,90 @@
...
+from typing import Optional, Tuple
+
+from datahub.company.models import Advisor, Contact
+
+
+class MatchingStatus(IntEnum):
+    """Matching status (for when searching for a model by a lookup field)."""
+
+    matched = 1
+    unmatched = 2
```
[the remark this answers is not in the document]
currycoder: I think this sounds like a good idea, but I'm wary that it will touch a lot of files and supporting functions and I would rather that this PR didn't balloon too much further now.  Could be a nice small follow up PR

datahub/company/test/test_contact_adviser_matching.py
```diff
@@ -0,0 +1,159 @@
...
+        assert not contact
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    'email,expected_matching_status,match_on_alternative',
+    (
+        # same case, match on email
+        ('unique1@primary.com', MatchingStatus.matched, False),
+        # same case, match on contact_email
```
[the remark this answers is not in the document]
currycoder: Could you be more specific, I'm not seeing a problem with these comments.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,186 @@
...
+        Initialise the CalendarInteractionEmailParser with a MailParser email
+        object.
+        """
+        self.message = message
+
+    def _extract_and_validate_sender_adviser(self):
+        try:
+            sender_email = self.message.from_[0][1]
+        except IndexError:
+            raise('Email was malformed - missing "from" header')
```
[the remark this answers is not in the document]
currycoder: Good spot.  Added a test case for this too

datahub/interaction/email_processors.py
currycoder: It might do.  I suppose on paper we would lose a level of indentation and make all of the `_extract_` helpers need to pass `message` around.  On the other hand, I do like the class wrapper as a sort of stated theme for the logic within.  How strongly do you feel about it?

datahub/company/test/test_contact_adviser_matching.py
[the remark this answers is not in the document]
reupen: Sorry, you're right, I was confused by the word 'contact'. However, it would be good to rename `match_on_alternative` as that was referring to `email_alternative` on the Contact model.

datahub/interaction/email_processors.py
reupen: It's not critical, partly an observation that some of the methods don't use `self` and some of the existing module-level functions take `message`.

My other observation was that `extract_interaction_data_from_email()` is a little bit buried at the bottom of the file. I'd (personally) have tried to keep it at the top of the file as everything else is essentially internal to it. But we might be in subjective territory there.

datahub/company/contact_adviser_matching.py
```diff
@@ -0,0 +1,100 @@
...
+    The following the logic is used:
+    - if a unique match is found using Advisor.email, this is used
+    - if there are multiple matches found using Advisor.contact_email, matching aborts
+    - otherwise, if there is a unique match on Contact.contact_email, this is used
+    """
+    adviser, matching_status = _find_active_adviser_using_field(email, 'email')
+    if matching_status == MatchingStatus.unmatched:
+        adviser, matching_status = _find_active_adviser_using_field(email, 'contact_email')
+
+    return adviser, matching_status
```
marcofucci: My thoughts:

1. in theory you should return `multiple_matches` if an email is used in `Adviser.email` and some other `Adviser.contact_email` to be 100% accurate
2. I would have preferred to stick with the Django: return 1 object, None or raise MultipleObjectsReturned instead of introducing another concept (`MatchingStatus`)
3. I still think that it would have been nice to have this on the Manager as it looks like a general purpose method.

datahub/company/test/test_contact_adviser_matching.py
```diff
@@ -0,0 +1,159 @@
...
+    },
+    {
+        'email': 'archived1@primary.com',
+        'contact_email': 'archived1@alternative.com',
+        'is_active': False,
+    },
+]
+
+
+@pytest.mark.django_db
```
marcofucci: You can probably add `pytestmark = pytest.mark.django_db` at the top of the module as they all depend on db?

datahub/interaction/email_processors.py
currycoder: Sure.  My understanding was that "private"/helper code generally sits above public stuff.  Could you elaborate a little more on the general pattern followed at DIT for this sort of code ordering?  I think it would add a lot of noise to the PR to rejig this now, but I'm happy to stick with an agreed style for other changes going forward.

datahub/company/contact_adviser_matching.py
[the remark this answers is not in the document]
reupen: Yes, we could do that later on.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,186 @@
...
+        calendar_string = _extract_calendar_string_from_text(self.message)
+        if not calendar_string:
+            calendar_string = _extract_calendar_string_from_attachments(self.message)
+        if not calendar_string:
+            raise error
+        try:
+            calendar = icalendar.Calendar.from_ical(calendar_string)
+        except ValueError as exc:
+            raise error from exc
+        calendar_event_components = [comp for comp in calendar.walk() if comp.name == 'VEVENT']
```
marcofucci: `VEVENT` could be a constant

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,186 @@
...
+            all_recipients,
+            interaction_data['sender'],
+        )
+        interaction_data['company'] = _get_company_from_contacts(interaction_data['contacts'])
+        calendar_event = self._extract_and_validate_calendar_event_metadata()
+        interaction_data['date'] = calendar_event['start']
+        interaction_data['location'] = calendar_event['location']
+        interaction_data['meeting_details'] = calendar_event
+        interaction_data['subject'] = calendar_event['subject']
+        meeting_confirmed = calendar_event['status'] == 'CONFIRMED'
```
marcofucci: `CONFIRMED` could be a constant

marcofucci: Can't this check be in `_extract_and_validate_calendar_event_metadata`?

datahub/company/contact_adviser_matching.py
reupen: > 1. in theory you should return multiple_matches if an email is used in Adviser.email and some other Adviser.contact_email to be 100% accurate

I agree that we may need different logic for advisers than contacts. Just to note that it was intentional how it works for contacts (on the basis that the alternative email address could be a generic email address or someone else's email address).

datahub/interaction/email_processors.py
reupen: I don't think we have one, so that might be one for further discussion.

datahub/interaction/email_processors.py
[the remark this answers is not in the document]
marcofucci: What I meant was that `extract_interaction_data_from_email` now returns `contacts` probably from different companies and a `company` related to just one of them. This doesn't seem obvious. If you are planning to have some business logic outside you could maybe leave out the `company` bit or return `companies` instead to have a more accurate representation of the parsed data. Another option could be to rename the key to `top_company` or something.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,186 @@
...
+
+def _extract_calendar_string_from_text(message):
+    """
+    Extract an icalendar string from the plain text of a message.
+
+    :param message: MailParser message object
+    :returns: A string of the icalendar body or None if it could not be found in plain text
+    """
+    for text_part in message.text_plain:
+        if text_part.startswith('BEGIN:VCALENDAR'):
```
marcofucci: `BEGIN:VCALENDAR` could be a constant

datahub/company/test/test_contact_adviser_matching.py
reupen: Only thing with applying it to the module is that it's easy to end up applying it to things that don't need it...

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -0,0 +1,283 @@
...
+    attributed to some DB entities.
+    """
+    fixture = {
+        'contacts': [],
+        'advisers': [],
+    }
+    adviser_emails = [
+        'brendan.smith@trade.gov.uk',
+        'marco.fucci@digital.trade.gov.uk',
+        'ali.zaidi@digital.trade.gov.uk',
```
marcofucci: There's no point in using these exact email addresses here and in the sample files. I'd use example1 2 3 to minimise reconnaissance.

datahub/company/test/test_contact_adviser_matching.py
currycoder: I'm going to leave this as-is for now, unless this is critical?

marcofucci: 👍 Non-critical

datahub/company/contact_adviser_matching.py
currycoder: I see what you mean about 2/3 and will keep this in mind for future changes.  I would prefer not to rework this here as my objective was to keep code that was solving similar problems DRY rather than embark on a full refactor.

> 1. in theory you should return multiple_matches if an email is used in Adviser.email and some other Adviser.contact_email to be 100% accurate

I think I need a better understanding of what these two fields are storing and what will be most accurate for my use case.  My objective is to make the "best guess" about the correct adviser for a given email address (as supplied through user input).  In this case, should I actually check `contact_email` first and then `email`?  I'm happy to rename the helper function to get across that we are making a best guess that prioritises making a match over accuracy if this helps too?

datahub/interaction/email_processors.py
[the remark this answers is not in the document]
currycoder: Thanks for the clarification here.  I realise that this broke exactly what I was trying to say about having too much business logic in the parser!
I've named the key `top_company` as suggested to make things explicit.

datahub/company/contact_adviser_matching.py
marcofucci: @reupen Yes we could leave contacts as it is and only rework advisers. 

@currycoder Adviser.email is the login username (can’t be changed) and contact_email is the actual accurate email address. No matter what, I suspect there will be cases when you want to prioritise email over contact_email and viceversa. In theory you shouldn’t have many Adviser records with contact_email same as another record’s email. You could check manually on staging.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,214 @@
...
+    email address - the context here is that we are dealing with the email
+    accounts that advisers use for setting up meetings/emailing companies.
+
+    :param email: string email address
+    :returns: an Advisor object or None, if a match could not be found
+    """
+    for field in ['contact_email', 'email']:
+        criteria = {field: email, 'is_active': True}
+        try:
+            return Advisor.objects.filter(**criteria).order_by('date_joined')[0]
```
reupen: Looks like a good use case for using `.first()` on the query set.

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,214 @@
...
+            raise ValidationError('Email not sent by DIT')
+        sender_adviser = _get_best_match_adviser_by_email(sender_email)
+        if not sender_adviser:
+            raise ValidationError('Email not sent by recognised DIT Adviser')
+        return sender_adviser
+
+    def _extract_and_validate_contacts(self, all_recipients):
+        contacts = []
+        for recipient_email in all_recipients:
+            contact, matching_status = find_active_contact_by_email_address(recipient_email)
```
marcofucci: Non-critial: `contact, _ = find_active_contact_by_email_address(recipient_email)` would be better as you are not using `matching_status`

datahub/interaction/email_processors.py
jim68000: We should probably also rename 'email' to avoid confusion - not for you Brendan, just another thing to remember

datahub/interaction/test/email_processors/email_samples/invalid/multiple_calendar_events.eml
```diff
@@ -0,0 +1,190 @@
...
+STATUS:CONFIRMED
+SUMMARY:initial
+TRANSP:OPAQUE
+END:VEVENT
+BEGIN:VEVENT
+DTSTART:20190329T163000Z
+DTEND:20190329T173000Z
+DTSTAMP:20190329T113633Z
+ORGANIZER;CN=3Dadviser1@trade.gov.uk:mailto:adviser1@digi=
+ta
```
reupen: Any idea how the line breaks ended up like this?

datahub/interaction/email_processors.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def _get_top_company_from_contacts(contacts):
+    """
+    Get the company from the given contacts.
+    If these contacts are related to different companies, return the company
+    with the highest number of contacts in our iterable.
+    """
+    company_counts = Counter(contact.company for contact in contacts)
+    top_company = company_counts.most_common()[0][0]
```
alixedi: Please consider: `top_company, _ = company_counts.most_common(1)`

datahub/interaction/email_processors.py
alixedi: Please consider using [earliest](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#earliest) and catch the `DoesNotExist` exception.

datahub/interaction/email_processors.py
alixedi: I wonder how this cleared the linter 😕

reupen: `flake8-bugbear` checks that (we don't have it at the moment).

datahub/interaction/test/email_processors/email_samples/invalid/multiple_calendar_events.eml
currycoder: This is the email as I received it (on another branch where email ingestion is active for one of our test IMAP inboxes).  So I'm not exactly sure what determines the line breaks - it could be anything in the chain; outlook client, email server, IMAP retrieval.  My approach has been to capture the format in the way that is most representative of the final format.

reupen: That's fine. Just wanted to double-check that that's how they actually were.

PR #1711, opened 2019-05-16
Delete trading address fields from database

alixedi:
### Description of change

Remove `trading_address` columns from the database.

### Checklist


datahub/company/migrations/0085_remove_trading_address_from_db.py
```diff
@@ -0,0 +1,78 @@
...
+MAX_LENGTH = 255
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0084_remove_trading_address_from_django'),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
```
currycoder: I commented this on the other review:
Out of interest, why add the state back and then remove the whole field instead of just using `database_operations`?

currycoder: One thing I'm not sure about the answer on though, is it worth generally only marking a follow-up PR as ready for review once it's predecessor has merged?  I can see it being a source of confusion for reviewers seeing the same code twice - though it's not too problematic in this instance as the changes here are relatively small.

alixedi: > Out of interest, why add the state back and then remove the whole field instead of just using database_operations?

The key to understand this is `migrations.` before `AddField` and `RemoveField`. We cannot `migrations.RemoveField` if the field doesn't exist in Django's state.

> One thing I'm not sure about the answer on though, is it worth generally only marking a follow-up PR as ready for review once it's predecessor has merged?

My bad, I mistakenly removed the WIP label from the other PR in my pre-coffee morning rampage (IIRC) which in hindsight was quite disastrous.

alixedi: Ref: https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/company/migrations/0058_remove_classification_from_database.py

currycoder: :+1:

PR #1714, opened 2019-05-17
Add logic for saving a single imported interaction

reupen:
### Description of change

This adds the logic for saving a single imported interaction in the import interactions tool.

This will be used when the full logic for completing the import is added.

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/row_form.py
+++ b/datahub/interaction/admin_csv_import/row_form.py
@@ -3,6 +3,7 @@ from typing import NamedTuple
 
 from django import forms
 from django.core.exceptions import NON_FIELD_ERRORS, ValidationError
+from django.db.transaction import atomic
 from django.utils.timezone import utc
 from django.utils.translation import gettext_lazy
 from rest_framework import serializers
@@ -12,10 +13,11 @@ from datahub.company.contact_matching import (
     find_active_contact_by_email_address,
 )
 from datahub.company.models import Advisor
+from datahub.core.exceptions import DataHubException
 from datahub.core.query_utils import get_full_name_expression
 from datahub.core.utils import join_truthy_strings
 from datahub.event.models import Event
-from datahub.interaction.models import CommunicationChannel, Interaction
+from datahub.interaction.models import CommunicationChannel, Interaction, InteractionDITParticipant
 from datahub.interaction.serializers import InteractionSerializer
 from datahub.metadata.models import Service, Team
 
@@ -147,10 +149,15 @@ class InteractionCSVRowForm(forms.Form):
 
     def is_valid_and_matched(self):
         """Return if the form is valid and the interaction has been matched to a contact."""
-        return (
-            self.is_valid()
-            and self.cleaned_data['contact_matching_status'] == ContactMatchingStatus.matched
-        )
+        return self.is_valid() and self.is_matched()
+
+    def is_matched(self):
+        """
+        Returns whether the interaction was matched to a contact.
+
+        Can only be called post-cleaning.
+        """
+        return self.cleaned_data['contact_matching_status'] == ContactMatchingStatus.matched
 
     def clean(self):
         """Validate and clean the data for this row."""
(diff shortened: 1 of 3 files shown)


datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -205,6 +206,35 @@ def full_clean(self):
             for field, errors in normalised_errors.items():
                 self._add_serializer_error(field, errors)
 
+    @atomic
+    def save(self, user, source):
+        """Creates an interaction from the cleaned data."""
+        serializer_data = self.cleaned_data_as_serializer_dict()
```
currycoder: Question:  Why can't you use the `serializer.save()` method with pertinent overrides instead?

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -776,6 +777,132 @@ def test_cleaned_data_as_serializer_dict_for_service_delivery(self):
...
+                (adviser_1, adviser_1.dit_team),
+                (adviser_2, adviser_2.dit_team),
+            ),
+        )
+
+        assert actual_advisers_and_teams == expected_advisers_and_teams
+
+    def test_save_rolls_back_on_error(self, monkeypatch):
+        """Test that save() rolls back if there's an error."""
+        monkeypatch.setattr(
```
currycoder: Nice test!

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -776,6 +777,132 @@ def test_cleaned_data_as_serializer_dict_for_service_delivery(self):
...
+            'date': '02/03/2018',
+            'adviser_1': adviser.name,
+            'contact_email': contact.email,
+            'service': service.name,
+            'communication_channel': communication_channel.name,
+        }
+        form = InteractionCSVRowForm(data=data)
+
+        assert form.is_valid()
+        interaction = form.save(user, source=source)
```
currycoder: Rather than trusting the return value, could you look up the interaction independently to ensure that it was persisted before making assertions?

reupen: I'll add an `interaction.refresh_from_db()` :+1:

currycoder: :+1:

datahub/interaction/admin_csv_import/row_form.py
reupen: Hmm, I could call `InteractionSerializer.create()` directly. Could be a bit fragile though...

currycoder: Sure!  I was mainly hoping to open a discussion about it, my next PR for calendar processing makes use of `serializer.save()` and things are working as I expect.  I suppose you are concerned about future changes to serializer field validation etc for the API which could prove to be problematic?

reupen: Sort of. To use `InteractionSerializer.save()`, I thought I'd have to reinstate #1633 (or put the data in the format used for an API request). And, run full serialiser validation and pass read-only etc. fields separately as keyword arguments to the `InteractionSerializer.save()` call.

(Alternatively, `InteractionSerializer.create()` could be used which would make things much simpler, but seems a bit hacky.)

So, in summary, not against using `InteractionSerializer.save()` but we would need to make more changes as this would be the first time calling `is_valid()` on the serialiser (so we would need to change the earlier validation stage of the import to do that as well).

currycoder: Yep makes sense.  I'm easy either way really - I think the code that you have here is neat and understandable.  I think perhaps we just need to document on the serialiser that there is some usage for the validation (or full usage) outside of the API layer, as this may not be obvious for somebody making modifications to it that is unaware.

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -776,6 +777,134 @@ def test_cleaned_data_as_serializer_dict_for_service_delivery(self):
...
+        }
+        form = InteractionCSVRowForm(data=data)
+
+        assert form.is_valid()
+
+        with pytest.raises(ValueError):
+            form.save(user, source=source)
+
+        assert not Interaction.objects.exists()
+
```
marcofucci: It would be good to have a test for when a contact is not matched. My understanding is that `is_valid` would silently return but the save would fail?

reupen: I've added a test and made it a little bit more robust for that situation.

datahub/interaction/admin_csv_import/row_form.py
reupen: Yes, it would be good to make things as clear as possible.

There was a small note above the validators on the serialiser, but I've added a more detailed one in the main docstring for `InteractionSerializer`.

PR #1721, opened 2019-05-20
Feature/interaction activity stream

alixedi:
### Description of change

This introduces a new endpoint: `/v3/activity-stream/interactions` that exposes all the interactions to be hoovered up by the [activity-stream](https://github.com/uktrade/activity-stream) service.

Following pieces were needed for this to work in addition to the usual ones:

1. InteractionCursorPagination: This brings in `CursorPagination` which was needed because the interactions table is ~3M rows and using the default limit-order pagination got slower as the user progressed through the pages.

2. InteractionActivitySerializer: Needed for conversion to the activity-stream format as specified in the [RFC](https://www.w3.org/TR/activitystreams-core/#introduction) and specifically the [spec](https://github.com/uktrade/activity-stream-datahub-spec/tree/master/interactions) written by @marcofucci for interactions.

P.S. I am not completely happy with the tests but wanted to put up the PR so that we can discuss suggestions for improvement.

### Checklist


datahub/interaction/views.py
```diff
@@ -44,4 +53,25 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
...
     )
-    ordering = ('-date', '-created_on')
+
+
+class InteractionCursorPagination(CursorPagination):
+    """
+    Cursor pagination for interaction.
+    """
+
+    page_size = 5
```
alixedi: This is for testing.

datahub/interaction/views.py
```diff
@@ -44,4 +53,25 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
         'last_name_of_first_contact',
         'subject',
     )
-    ordering = ('-date', '-created_on')
+
+
+class InteractionCursorPagination(CursorPagination):
```
alixedi: Cursor pagination solves the performance problem when paging a large-ish model such as `Interaction`.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +133,19 @@ def test_succeeds_with_valid_redentials(api_client):
...
+    Get a list of interactions.
+    """
+    sender = _auth_sender()
+    response = api_client.get(
+        f'http://testserver{reverse("api-v3:activity-stream:interactions")}',
+        content_type='',
+        HTTP_AUTHORIZATION=sender.request_header,
+        HTTP_X_FORWARDED_FOR='1.2.3.4, 123.123.123.123',
+    )
+    assert response.status_code == status.HTTP_200_OK
```
alixedi: This gives me a 401 at the moment. Any ideas?

alixedi:
This is where I am at right now with interactions, activity stream and the like.

I have a few questions that I have put inline with the code.

datahub/interaction/views.py
```diff
@@ -44,4 +53,25 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
...
+class InteractionActivityStreamViewSet(HawkResponseSigningMixin, ListModelMixin, ViewSet):
+    """
+    Interaction ViewSet for the activity stream
+    """
+
+    authentication_classes = (HawkAuthentication,)
+    permission_classes = (HawkScopePermission,)
+    required_hawk_scope = HawkScope.activity_stream
+    pagination_class = InteractionCursorPagination
+    serializer_class = InteractionSerializer
```
alixedi: I want to serialise `Interaction`s to [activity streams format](https://www.w3.org/TR/activitystreams-core/#examples). 

Current idea is to write a special serialiser for this - `InteractionActivityStreamSerializer`.  Any thoughts?

datahub/activity_stream/test/test_activity_stream_views.py
marcofucci: Check `_auth_sender` it uses `_url`

datahub/interaction/views.py
```diff
@@ -44,4 +55,46 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
...
+
+    page_size = 5
+    ordering = ('created_on', )
+
+    def get_paginated_response(self, data):
+        """
+        Override this function to re-format the response accourding to
+        activity stream spec.
+        """
+        return Response(OrderedDict([
```
alixedi: Reference: https://www.w3.org/TR/activitystreams-core/#example-3 and https://www.w3.org/TR/activitystreams-vocabulary/#dfn-orderedcollectionpage

datahub/interaction/serializers.py
```diff
@@ -503,3 +503,71 @@ class Meta:
...
+            'archived_by',
+            'archived_on',
+            'archived_reason',
+        )
+
+    def to_representation(self, instance):
+        """
+        S
+        """
+        return {
```
alixedi: Reference: [Collection/CollectionPage](https://www.w3.org/TR/activitystreams-core/#example-3) and [OrderedCollection](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-orderedcollectionpage)

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -3,15 +3,22 @@
...
+    CompanyInteractionFactory,
+    EventServiceDeliveryFactory,
+    InvestmentProjectInteractionFactory,
+    ServiceDeliveryFactory,
+)
+
 
-def _url():
-    return 'http://testserver' + reverse('api-v3:activity-stream:index')
+def _url(name):
```
alixedi: Refactoring this to remove the hard-coded reverse call.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +144,346 @@ def test_succeeds_with_valid_redentials(api_client):
...
+        'partOf': 'http://testserver/v3/activity-stream/interactions',
+        'orderedItems': [
+            {
+                'id': f'dit:DataHubInteraction:{interaction.id}:Announce',
+                'type': 'Announce',
+                'published': iso_fix(interaction.created_on),
+                'generator': {'name': 'dit:dataHub', 'type': 'Application'},
+                'object': {
+                    'id': f'dit:DataHubInteraction:{interaction.id}',
+                    'type': ['Event', 'dit:Interaction'],
```
michalc: At the moment, I think the public-user facing Great search queries on all things of type `Event` https://github.com/uktrade/great-domestic-ui/blob/develop/activitystream/helpers.py#L129 . Ideally, it would query on something more specific like `'dit:aventiEvent'`, and from memory I think this was the plan, but I don't think this has happened yet. This may need discussion with @WillTaylor22

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +144,346 @@ def test_succeeds_with_valid_redentials(api_client):
...
+                            },
+                        },
+                        *[
+                            {
+                                'id': f'dit:DataHubContact:{contact.pk}',
+                                'type': ['Person', 'dit:Contact'],
+                                'dit:emailAddress': contact.email,
+                                'name': contact.name,
+                            }
+                            for contact in interaction.contacts.all()
```
michalc: Apologies if this is already covered by `.all()`, but would be good for them to be some order in the contacts. Doesn't particularly matter what it is I suspect, just something that is consistent between requests.

datahub/activity_stream/test/test_activity_stream_views.py
marcofucci: Thanks, @michalc . In that case, should we change this to `'type': 'dit:Interaction'` for the time being? @WillTaylor22 when are you planning to refactor your data/query?

datahub/interaction/views.py
```diff
@@ -44,4 +55,41 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
...
+    """
+    Interaction ViewSet for the activity stream
+    """
+
+    authentication_classes = (HawkAuthentication,)
+    permission_classes = (HawkScopePermission,)
+    required_hawk_scope = HawkScope.activity_stream
+    pagination_class = InteractionCursorPagination
+    serializer_class = InteractionActivitySerializer
+    queryset = Interaction.objects.all()
```
reupen: Presumably we could make use of `select_related` and `prefetch_related` to help with performance and avoid extra queries for relationships? There's an example in `datahub.interaction.queryset` (unfortunately it has an annotation, but maybe it could be reused if that was separated out).

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+            'id': f'dit:DataHubInvestmentProject:{project.pk}',
+            'type': 'dit:InvestmentProject',
+            'name': project.name,
+            'url': reverse('api-v3:investment:investment-item', kwargs={'pk': project.pk}),
+        } if project is not None else {}
+
+    def _get_event_context(self, event):
+        return {
+            'id': f'DataHubEvent:{event.pk}',
```
marcofucci: Sorry forgot `dit` here, this should have been 'dit:DataHubEvent:...`

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+                'type': ['Event', f'dit:{instance.kind.title().replace("_", "")}'],
+                'startTime': instance.date,
+                'dit:status': instance.status,
+                'dit:archived': instance.archived,
+                'dit:communicationChannel': {
+                    'name': instance.communication_channel.name
+                    if instance.communication_channel else None,
+                },
+                'dit:subject': instance.subject,
+                'dit:service': {'name': instance.service.name},
```
reupen: `service` is nullable – do we have any protection against it being null? (It shouldn't be in general, but some of the old data from legacy systems isn't in perfect shape...)

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+            'dit:team': {
+                'id': f'dit:DataHubTeam:{event.lead_team.pk}',
+                'type': ['Group', 'dit:Team'],
+                'name': event.lead_team.name,
+            },
+            'url': reverse('api-v3:event:item', kwargs={'pk': event.pk}),
+        } if event is not None else {}
+
+    def _get_context(self, instance):
+        context = []
```
reupen: Looks like this function could be simplified a bit since kind can only be one of the two values (so no need to build up a list).

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+        if instance.kind == 'interaction':
```
reupen: Would be good to use `Interaction.KINDS.interaction` and `Interaction.KINDS.service_delivery` (instead of the strings).

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+            context.append(self._get_project_context(
+                instance.investment_project,
+            ))
```
marcofucci: The agreed style for this is:
```
context.append(
    self._get_project_context(
        instance.investment_project,
    )
)
```

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+                'type': ['Group', 'dit:Team'],
+                'name': event.lead_team.name,
+            },
+            'url': reverse('api-v3:event:item', kwargs={'pk': event.pk}),
+        } if event is not None else {}
+
+    def _get_context(self, instance):
+        context = []
+        if instance.kind == 'interaction':
+            context.append(self._get_project_context(
```
reupen: Not sure if you've seen them, but there are some notes on indentation style here: https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Coding%20style.md#indentation

(Following the examples, there should be a line break before the first argument here.)

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+                },
+                'dit:subject': instance.subject,
+                'dit:service': {'name': instance.service.name},
+                'attributedTo': [
+                    self._get_company(instance.company),
+                    self._get_advisor(instance.dit_adviser),
+                    *self._get_contacts(instance.contacts),
+                ],
+                'url': reverse('api-v3:interaction:item', kwargs={'pk': instance.pk}),
+                'context': self._get_context(instance),
```
marcofucci: It would be good to only add context if it's != {}

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +144,346 @@ def test_succeeds_with_valid_redentials(api_client):
...
+                'type': 'Announce',
+                'published': interaction.created_on.isoformat().replace('+00:00', 'Z'),
+                'generator': {'name': 'dit:dataHub', 'type': 'Application'},
+                'object': {
+                    'id': f'dit:DataHubInteraction:{interaction.id}',
+                    'type': ['Event', 'dit:ServiceDelivery'],
+                    'startTime': interaction.date.isoformat().replace('+00:00', 'Z'),
+                    'dit:status': interaction.status,
+                    'dit:archived': interaction.archived,
+                    'dit:communicationChannel': {'name': None},
```
reupen: Is `'dit:communicationChannel': {'name': None}` expected for an event service delivery? Couldn't see it [in the spec](https://github.com/uktrade/activity-stream-datahub-spec/blob/master/interactions/service_delivery_event.json).

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+                'dit:archived': instance.archived,
+                'dit:communicationChannel': {
+                    'name': instance.communication_channel.name
+                    if instance.communication_channel else None,
+                },
+                'dit:subject': instance.subject,
+                'dit:service': {'name': instance.service.name},
+                'attributedTo': [
+                    self._get_company(instance.company),
+                    self._get_advisor(instance.dit_adviser),
```
marcofucci: Shouldn't we use `InteractionDITParticipant` instead of the deprecated `instance.dit_adviser`?

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+            for contact in contacts.all()
+        ]
+
+    def to_representation(self, instance):
+        """
+        Serialize the interaction as per Activity Stream interaction spec:
+        https://github.com/uktrade/activity-stream-datahub-spec/tree/master/interactions
+        """
+        return {
+            'id': f'dit:DataHubInteraction:{instance.pk}:Announce',
```
marcofucci: This is `object.id` + ':Announce' so you could save `f'dit:DataHubInteraction:{instance.pk}'` in a var?

datahub/interaction/serializers.py
marcofucci: `service` is not mandatory so it could be None unfortunately :-/

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+    def _get_event_context(self, event):
+        return {
+            'id': f'DataHubEvent:{event.pk}',
+            'type': 'Event',
+            'dit:eventType': {'name': event.event_type.name},
+            'name': event.name,
+            'startTime': event.start_date,
+            'endTime': event.end_date,
+            'dit:team': {
+                'id': f'dit:DataHubTeam:{event.lead_team.pk}',
```
marcofucci: Shouldn't this use `_get_team`?

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+            'id': f'dit:DataHubTeam:{team.pk}',
+            'type': ['Group', 'dit:Team'],
+            'name': team.name,
+        } if team is not None else {}
+
+    def _get_advisor(self, adviser):
+        return {
+            'id': f'dit:DataHubAdviser:{adviser.pk}',
+            'type': ['Person', 'dit:Adviser'],
+            'dit:emailAddress': adviser.contact_email,
```
marcofucci: This should probably be "contact_email or email"

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +144,346 @@ def test_succeeds_with_valid_redentials(api_client):
...
             content_type='incorrect',
         )
+
+
+def iso_fix(dt):
+    """
+    Change given datetime to isoformat with +00:00
+    replaced by Z.
+    """
+    return dt.isoformat().replace('+00:00', 'Z')
```
marcofucci: There's a `format_date_or_datetime` in `datahub.core.test_utils`.

datahub/activity_stream/test/test_activity_stream_views.py
michalc: >  In that case, should we change this to 'type': 'dit:Interaction' for the time being? 

Yes, sounds good.

datahub/activity_stream/urls.py
```diff
@@ -1,11 +1,17 @@
...
+from datahub.interaction.views import InteractionActivityViewSet
 
 activity_stream_urls = [
     path(
         'activity-stream/',
         ActivityStreamViewSet.as_view({'get': 'list'}),
         name='index',
     ),
+    path(
+        'activity-stream/interactions',
```
michalc: Ah... from memory the `'activity-stream/` endpoint is a dummy one, essentially created for the pen-test. Could use that? (Or, if _not_ using it, remove it?)

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +144,346 @@ def test_succeeds_with_valid_redentials(api_client):
...
+        'partOf': 'http://testserver/v3/activity-stream/interactions',
+        'orderedItems': [
+            {
+                'id': f'dit:DataHubInteraction:{interaction.id}:Announce',
+                'type': 'Announce',
+                'published': interaction.created_on.isoformat().replace('+00:00', 'Z'),
+                'generator': {'name': 'dit:dataHub', 'type': 'Application'},
+                'object': {
+                    'id': f'dit:DataHubInteraction:{interaction.id}',
+                    'type': ['Event', 'dit:ServiceDelivery'],
```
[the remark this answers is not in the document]
marcofucci: @reupen: 
- types are PascalCase except namespaces, hence `dit:ServiceDelivery`. 
- field names and values are CamelCase, hence `dit:archived` and `dit:dataHub`

It looks a bit rubbish, it's not 100% clear from the W3C doc but [seems that way](https://www.w3.org/TR/activitystreams-core/#extensibility).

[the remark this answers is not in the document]
marcofucci: @reupen 
> is it strictly necessary for us to convert service_delivery (our internal value) to ServiceDelivery

It's not strictly but they seem different objects and they will probably be displayed differently in the frontend so they should in theory be different types. We don't have to convert service_delivery to ServiceDelivery this way we could have a MAPPING dict or something like that if that's what you are worried about. It could also be more stable that way in case new kinds are introduced.

datahub/activity_stream/test/test_activity_stream_views.py
marcofucci: yeah `dit:communicationChannel` shouldn't be in the response in that case

datahub/interaction/serializers.py
```diff
@@ -503,3 +504,116 @@ class Meta:
...
+                    if instance.communication_channel else None,
+                },
+                'dit:subject': instance.subject,
+                'dit:service': {'name': instance.service.name},
+                'attributedTo': [
+                    self._get_company(instance.company),
+                    self._get_advisor(instance.dit_adviser),
+                    *self._get_contacts(instance.contacts),
+                ],
+                'url': reverse('api-v3:interaction:item', kwargs={'pk': instance.pk}),
```
marcofucci: I think we should use frontend urls instead of API one. This way another non-DataHub service can just add a link instead of constructing the url.

datahub/activity_stream/test/test_activity_stream_views.py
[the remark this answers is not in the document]
michalc: I would probably lean to consistency with W3C docs, which do appear to be as @marcofucci suggests. I suspect it's not the end of the world to _not_ be consistent it's it's too tricky.

I also would probably be more pro an explicit mapping somewhere, rather than the automatic conversion elsewhere in this PR `instance.kind.title().replace("_", ""`... etc. So to have a hard-coded dictionary a bit like:
```python
{
...
   'service_delivery': 'dit:ServiceDelivery',
...
}
```
So there is a bit of a review process before adding types to the AS, it's a bit more grep-able, so it fails if an attempt is made to add something that "shouldn't" be there, and I guess a bit more flexible in case there needs to be deviation from the internal name.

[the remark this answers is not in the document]
michalc: Ah I didn't see @marcofucci 's reply!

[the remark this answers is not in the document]
reupen: Cool. Yes, a mapping sounds good to me. (Could use a test to make sure all KINDS are covered in the mapping.)

datahub/activity_stream/urls.py
alixedi: Removing it. Thanks.

datahub/activity_stream/test/test_activity_stream_views.py
alixedi: I had a chat with @michalc and @marcofucci on this one and we have decided to go with `s/Event/dit:Event`.

datahub/interaction/serializers.py
alixedi: Discussed IRL. We have decided to drop the team for now.

datahub/interaction/serializers.py
```diff
@@ -583,37 +580,47 @@ def _get_contacts(self, contacts):
                 'dit:emailAddress': contact.email,
                 'name': contact.name,
             }
-            for contact in contacts.all()
+            for contact in contacts.order_by('pk').all()
```
reupen: Small thing to consider: calling `order_by()` will usually defeat the prefetching. 

Only alternative I can think of is doing the sorting in Python... (e.g. `sorted(contacts.all(), key=attrgetter('pk')`)

alixedi: I wonder if that is the case because this is `contacts` here are `interaction.contracts` and not the entire table.

Do you know of any way to find out if the prefetching is indeed being defeated?

reupen: Yes, there are various options. 

You can use `./manage.py runserver_plus --print-sql` to monitor the SQL queries being executed (also `./manage.py shell_plus --print-sql`).

If you can get it working in a browser you can use Django Debug Toolbar.

The documentation for `QuerySet.prefetch_related` covers how it works: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.prefetch_related

datahub/interaction/serializers.py
```diff
@@ -503,3 +503,124 @@ class Meta:
...
+                    *self._get_dit_participants(instance.dit_participants),
+                    *self._get_contacts(instance.contacts),
+                ],
+                'url': instance.get_absolute_url(),
+            },
+        }
+
+        context = self._get_context(instance)
+        if context:
+            interaction['object']['context'] = [context]
```
michalc: A very minor style point / feel free to ignore, but I think here (and similarly for `dit:communicationChannel`) you can have something like...

```python
context = self._get_context(instance)
...
interaction = {
  ...
  **(
    {'context': context} if context else {}
  )
  ...
}
```
to only have `context` if it's not the empty dict and avoid the mutation

alixedi: I really like this 😍 

TY @michalc !

datahub/interaction/views.py
```diff
@@ -44,4 +52,41 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
...
     )
-    ordering = ('-date', '-created_on')
+
+
+class InteractionCursorPagination(CursorPagination):
+    """
+    Cursor pagination for interaction.
+    """
+
+    ordering = ('created_on', )
```
michalc: Can there ever be items created at the same moment? Should it be something like (`created_on`, `id`) to deal with this case so items are not missed?

Also, can an interaction ever change, and if so, is there an `updated_on`/similar available, and is it set to `created_on` on the instance creation? If so, then this might be a better choice: it would means the last page of the feed, which is polled for updates, will contain both newly created interactions _and_ just updated interactions.

alixedi: This is a great point. I should probably write a couple of tests to ensure that the right interactions get on the last page. Thanks!

datahub/interaction/views.py
```diff
@@ -44,4 +52,41 @@ class InteractionViewSet(ArchivableViewSetMixin, CoreViewSet):
...
+
+    def _get_url(self):
+        return self.encode_cursor(self.cursor) if self.cursor else self.base_url
+
+    def get_paginated_response(self, data):
+        """
+        Override this function to re-format the response accourding to
```
reupen: Typo in 'accourding' here.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +143,312 @@ def test_succeeds_with_valid_redentials(api_client):
...
+                            'name': interaction.dit_adviser.name,
+                        },
+                        *[
+                            {
+                                'id': f'dit:DataHubContact:{contact.pk}',
+                                'type': ['Person', 'dit:Contact'],
+                                'dit:emailAddress': contact.email,
+                                'name': contact.name,
+                            }
+                            for contact in interaction.contacts.order_by('pk').all()
```
reupen: Minor point: `.all()` should be redundant here.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +143,312 @@ def test_succeeds_with_valid_redentials(api_client):
...
+                        ],
+                    ],
+                    'url': interaction.get_absolute_url(),
+                },
+            },
+        ],
+    }
+
+
+def test_interaction_investment_project_activity(api_client, db):
```
reupen: `test_interaction_activity` was updated to use `@pytest.mark.django_db` but the other tests are still using the `db` fixture.

datahub/interaction/serializers.py
```diff
@@ -503,3 +503,124 @@ class Meta:
                 ),
             ),
         ]
+
+
+class InteractionActivitySerializer(serializers.ModelSerializer):
+    """Interaction serialiser."""
+
+    KINDS_JSON = {
```
reupen: It would be good to use `Interaction.KINDS.interaction` and `Interaction.KINDS.service_delivery` for the keys.

datahub/interaction/serializers.py
```diff
@@ -503,3 +503,124 @@ class Meta:
...
+            'published': instance.created_on,
+            'generator': {'name': 'dit:dataHub', 'type': 'Application'},
+            'object': {
+                'id': interaction_id,
+                'type': ['dit:Event', f'dit:{self.KINDS_JSON[instance.kind]}'],
+                'startTime': instance.date,
+                'dit:status': instance.status,
+                'dit:archived': instance.archived,
+                'dit:subject': instance.subject,
+                'dit:service': {'name': self._get_service_name(instance)},
```
marcofucci: Service can be None, it would be good not to add it to the response if so

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +143,312 @@ def test_succeeds_with_valid_redentials(api_client):
...
+                            'id': f'dit:DataHubInvestmentProject:{project.pk}',
+                            'name': project.name,
+                            'type': 'dit:InvestmentProject',
+                            'url': project.get_absolute_url(),
+                        },
+                    ],
+                },
+            },
+        ],
+    }
```
marcofucci: It would be good to have some tests for the pagination as well, just to check that the next in the response body is there as it’s a new pagination style and it’s a good way of documenting things

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -133,3 +143,312 @@ def test_succeeds_with_valid_redentials(api_client):
...
+        '@context': 'https://www.w3.org/ns/activitystreams',
+        'summary': 'Interaction Activities',
+        'type': 'OrderedCollectionPage',
+        'id': 'http://testserver/v3/activity-stream/interactions',
+        'partOf': 'http://testserver/v3/activity-stream/interactions',
+        'orderedItems': [
+            {
+                'id': f'dit:DataHubInteraction:{interaction.id}:Announce',
+                'type': 'Announce',
+                'published': interaction.created_on.isoformat().replace('+00:00', 'Z'),
```
marcofucci: Did you forget to update this line? :-) there are other places as well that need to be updated

reupen:
Apart from the pre-fetching point, the main other last point from me is about where things sit in the source tree. It seems a bit confusing that the view is under `datahub.interaction` but the tests under `datahub.activity_stream`.

I suppose perhaps the first question is are we happy with the URL path `/v3/activity-stream/interactions` or would `/v3/interaction/activity-stream` suit us better?

(Also it in the first case it should be `/v3/activity-stream/interaction` to be consistent with the other endpoints.)

datahub/activity_stream/test/test_activity_stream_views.py
alixedi: Oops, I have been having trouble in keeping track of the change requests :(

alixedi:
> Apart from the pre-fetching point, the main other last point from me is about where things sit in the source tree. It seems a bit confusing that the view is under `datahub.interaction` but the tests under `datahub.activity_stream`.
> 
> I suppose perhaps the first question is are we happy with the URL path `/v3/activity-stream/interactions` or would `/v3/interaction/activity-stream` suit us better?
> 
> (Also it in the first case it should be `/v3/activity-stream/interaction` to be consistent with the other endpoints.)

@reupen Do you have a specific suggestion?

I like `/activity-stream/interaction` primarily because I read it as "activity stream of type interaction" which is what it is in my opinion instead of "interaction's representation as an activity stream". The reason for this opinion is that the difference between this and the regular endpoint is more than just formatting. 

In terms of the placement in source tree,  I am guessing that `InteractionActivityViewSet` will have more in common with `datahub.interaction` than `datahub.activity_stream` but as we add a couple of more activity stream types, it might turn out that I am wrong.

changelog/activity_stream/interaction-activity-stream.feature.rst
```diff
@@ -0,0 +1,6 @@
+It is now possible to get a list of interactions in activity-stream `format. 
+<https://github.com/uktrade/activity-stream-datahub-spec/tree/master/interactions>`_
```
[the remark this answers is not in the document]
alixedi: Ok, I will point the reference to [W3's doc](https://www.w3.org/TR/activitystreams-core/#introduction)

reupen:
Another option would be to make it like search so we have e.g. `datahub.activity_stream.interaction`.

datahub/activity_stream/test/test_activity_stream_views.py
alixedi: I have added some tests for ordering.

datahub/interaction/serializers.py
```diff
@@ -503,3 +503,125 @@ class Meta:
...
+            },
+        }
+
+        context = self._get_context(instance)
+        if context:
+            interaction['object']['context'] = [context]
+
+        if instance.kind == Interaction.KINDS.interaction:
+            interaction['object']['dit:communicationChannel'] = {
+                'name': instance.communication_channel.name,
```
marcofucci: Unfortunately bad data means that some interactions have `communication_channel == None` so you'd have to add a check here.

datahub/interaction/serializers.py
```diff
@@ -503,3 +503,125 @@ class Meta:
...
+            }
+
+        if instance.service is not None:
+            interaction['object']['dit:service'] = {
+                'name': instance.service.name,
+            }
+
+        return interaction
```
marcofucci: As I suggestion, what do we think about what @michalc suggested? I find it better self-documented this way:

```
        interaction_id = f'dit:DataHubInteraction:{instance.pk}'

        context = self._get_context(instance)

        if instance.kind == Interaction.KINDS.interaction and instance.communication_channel:
            communication_channel = {
                'name': instance.communication_channel.name,
            }
        else:
            communication_channel = {}

        if instance.service:
            service = {
                'name': instance.service.name,
            }
        else:
            service = {}

        interaction = {
            'id': f'{interaction_id}:Announce',
            'type': 'Announce',
            'published': instance.created_on,
            'generator': {'name': 'dit:dataHub', 'type': 'Application'},
            'object': {
                'id': interaction_id,
                'type': ['dit:Event', f'dit:{self.KINDS_JSON[instance.kind]}'],
                'startTime': instance.date,
                'dit:status': instance.status,
                'dit:archived': instance.archived,
                'dit:subject': instance.subject,
                **({'context': [context]} if context else {}),
                **(
                    {'dit:communicationChannel': communication_channel}
                    if communication_channel else {}
                ),
                **({'dit:service': service} if service else {}),
                'attributedTo': [
                    self._get_company(instance.company),
                    *self._get_dit_participants(instance.dit_participants),
                    *self._get_contacts(instance.contacts),
                ],
                'url': instance.get_absolute_url(),
            },
        }

        return interaction
```

alixedi: I <3 it initially and tried it but I felt it didn't read very well. In the present form, there is a clear distinction on keys that are optional as opposed to ^

datahub/interaction/serializers.py
alixedi: I can fix it like the others but I do feel we are going overboard with not having keys that have a `None` value in the serialised JSON.

alixedi: I would rather have serialised JSON with a fixed schema that has some keys set to `None` instead of changing the schema b/c we have bad data.

I admit this is more of a second thought so happy to listen to arguments against.

datahub/activity_stream/test/test_activity_stream_views.py
```diff
@@ -95,17 +110,17 @@ def test_succeeds_with_valid_redentials(api_client):
     """If the Authorization and X-Forwarded-For headers are correct, then
     the correct, and authentic, data is returned
     """
-    sender = _auth_sender()
+    sender = _auth_sender(
```
marcofucci: Small thing but can you please fix the name of this test? There's a `c` missing

datahub/interaction/serializers.py
alixedi: We had a chat on Slack and decided to go with what @marcofucci is suggesting for consistency across mandatory and optional fields.

datahub/activity_stream/interaction/tests/test_views.py
```diff
@@ -0,0 +1,635 @@
...
+    """
+    interactions = []
+
+    # We create 2 interactions with the same modified_on time
+    with freeze_time():
+        interactions += CompanyInteractionFactory.create_batch(2)
+
+    interactions += CompanyInteractionFactory.create_batch(8)
+
+    sender = _auth_sender(_url('api-v3:activity-stream:interactions'))
```
reupen: This construct is repeated a fair few times (
```python
    sender = _auth_sender(_url('api-v3:activity-stream:interactions'))
    response = api_client.get(
        _url('api-v3:activity-stream:interactions'),
        content_type='',
        HTTP_AUTHORIZATION=sender.request_header,
        HTTP_X_FORWARDED_FOR='1.2.3.4, 123.123.123.123',
    )
```
)

Would be good to try and make it a bit less repetitive – maybe a helper function or adding a `hawk_api_client` fixture?

datahub/activity_stream/interaction/tests/test_views.py
```diff
@@ -0,0 +1,635 @@
...
+        content_type='',
+        HTTP_AUTHORIZATION=sender.request_header,
+        HTTP_X_FORWARDED_FOR='1.2.3.4, 123.123.123.123',
+    )
+
+    assert response.status_code == status.HTTP_200_OK
+
+    sorted_interaction_ids = [
+        f'dit:DataHubInteraction:{i.pk}'
+        for i in sorted(interactions, key=lambda i: (i.modified_on, i.pk))
```
reupen: IMO `i` sounds too much like an index or other integer – would suggest something like `obj`.

datahub/activity_stream/interaction/tests/test_views.py
```diff
@@ -0,0 +1,635 @@
...
+    ]
+
+    response_interaction_ids = [
+        i['object']['id']
```
reupen: As last comment (about `i`) – maybe `item` could be used here instead of `i`.

datahub/interaction/models.py
```diff
@@ -273,6 +273,8 @@ def __str__(self):
     class Meta:
         indexes = [
             models.Index(fields=['-date', '-created_on']),
+            # For activity-stream
+            models.Index(fields=['modified_on', 'id']),
```
marcofucci: I don't see a migration for this?

changelog/activity_stream/interaction-activity-stream.feature.rst
```diff
@@ -0,0 +1,6 @@
+It is now possible to get a list of interactions in activity-stream `format. 
+<https://www.w3.org/TR/activitystreams-core/>`_ 
+
+The URL for this is:
+
+| /v3/activity-stream/interactions
```
marcofucci: The folder `changelog/activity_stream` is not mapped so this won't work. I think it would be better to have the file in `changelog/interaction-activity-stream.api.rst`

datahub/interaction/models.py
alixedi: I brought this up few days back.

@reupen suggested that we can let Django take care of this b/c similar migrations (create index) in the past didn't take long to complete.

Happy to add a migration if you think otherwise?

datahub/activity_stream/interaction/views.py
```diff
@@ -0,0 +1,54 @@
...
+    HawkScopePermission,
+)
+from datahub.core.viewsets import CoreViewSet
+from datahub.interaction.queryset import get_base_interaction_queryset
+
+
+class InteractionCursorPagination(CursorPagination):
+    """
+    Cursor pagination for interaction.
+    """
```
reupen: It would be good to add a bit of an explanation about `modified_on` being chosen despite it not being non-changing for a particular object.

alixedi: Good idea.

PR #1725, opened 2019-05-20
Enable whitelist of exempt domains for email ingestion verification

currycoder:
### Description of change
This PR enables us to set a whitelist of domains which are exempt from email ingestion `Authentication-Results` verification.  This is achieved through the new `DIT_EMAIL_DOMAINS_AUTHENTICATION_EXEMPT` setting.

Our `was_email_sent_by_dit()` validation function checks the email's `Authentication-Results` header to ensure that the email was sent by a verified email address.  However, for some domains, verification is handled by our mailbox at the SMTP server level.  Emails for these domains will not have an `Authentication-Results` header at all, so we need to use this whitelist for such domains.

### Checklist


datahub/email_ingestion/validation.py
```diff
@@ -33,14 +32,22 @@ def was_email_sent_by_dit(message):
 
     :returns: True if the email was sent by DIT, False otherwise.
     """
-    from_email = message.from_[0][1]
-    # TODO: See if valid domains are recorded elsewhere in the codebase
-    # and use this to keep things DRY
+    try:
+        from_email = message.from_[0][1]
+        from_domain = from_email.split('@')[1]
```
reupen: Looks like an email address can technically have multiple @ symbols in it. So would suggest using either `str.rpartition()` or `str.rsplit()` (with `maxsplit=1`).

PR #1726, opened 2019-05-21
Add logic for saving imported interactions CSV file

reupen:
### Description of change

This adds the remainder of the logic for saving imported interactions.

This will be used in an upcoming PR that will plug the logic into the view.

This sets the source field of imported interactions to:

```json
{
    "file": {
        "name": "<file name>",
        "size": <file size>,
        "sha256": "<SHA-256 hash>",
    },
}
```

(Potentially we could add the row number as well, but I've kept it at that for now.)

I picked SHA-256 for the hash as it's fairly ubiquitous (although Ubuntu comes with `b2sum` these days...)

I would suggest reviewing the two commits separately.

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/file_form.py
+++ b/datahub/interaction/admin_csv_import/file_form.py
@@ -1,7 +1,9 @@
 import gzip
+import hashlib
 from datetime import timedelta
 from secrets import token_urlsafe
 
+import reversion
 from django.conf import settings
 from django.core.cache import cache
 from django.core.files.uploadedfile import SimpleUploadedFile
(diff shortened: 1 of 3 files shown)


datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -87,6 +83,32 @@ def save_to_cache(self):
...
+    @reversion.create_revision()
+    def save(self, user):
+        """Saves all loaded rows matched with contacts."""
+        reversion.set_comment(REVISION_COMMENT)
+
+        csv_file = self.cleaned_data['csv_file']
+        sha256 = _sha256_for_file(csv_file)
+
+        source = {
+            'file': {
```
currycoder: I realise that `source` is deliberately schema-less as a JSON field, but is it worth documenting this format as a comment on the model field?  I think this format is better than what I had in mind for meetings - signposting the source type as the key for the source-specific object.

reupen: Yes, that's a good point. I'll add something.

We could also add a `type` field if we think it will be useful (but not sure I have a use for it, if I were getting the rows for a particular hash I would probably just do something like `source @> '{"file": {"sha256": "<some hash>"}}`) (in SQL).

reupen: I've updated the existing comment.

PR #1732, opened 2019-05-22
Calendar invite email processing - saving an interaction

currycoder:
### Description of change

This introduces the `CalendarInteractionEmailProcessor` class which is a concrete implementation of the `datahub.email_ingestion.email_processor.EmailProcessor` abstract class.  
This class will use the `CalendarInteractionEmailParser` class to extract and validate meeting invite data from an ingested email message, perform further validation and then save the data as a draft interaction.

This class stitches everything together to provide the first concrete implementation of email ingestion in datahub.  This functionality is not actually switched on yet - we need to wire up this processing class with the MAILBOXES setting in `config/settings/common.py` in a subsequent PR.

### Points of note
- I've used the `InteractionSerializer` as a final mechanism of validation for the incoming data.  I thought this was best to keep things DRY - the parser class firstly does some basic validation to ensure that we have everything that we need and the serializer follows this up with our full validation for interaction data.

### Checklist


datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +68,207 @@ def calendar_data_fixture():
...
+    return {
+        'sender': 'adviser1@trade.gov.uk',
+        'contacts': ['bill.adama@example.net', 'saul.tigh@example.net'],
+        'secondary_advisers': [],
+        'date': datetime(2019, 5, 1, 13, 00, tzinfo=utc),
+        'top_company': 'Company 1',
+        'location': 'Windsor House',
+        'meeting_details': {'uid': '12345'},
+        'subject': 'A meeting',
+    }
```
marcofucci: I think the way of setting field values and then overriding it with objects is a bit magic. 
Maybe call these 'sender_email' etc. and pop them when creating the actual data? Or we could add some comments to explain that the item value be replaced by an instance of the object

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,114 @@ def _get_utc_datetime(localised_datetime):
...
+        interaction_data['contacts'] = sanitised_contacts
+        try:
+            serializer = self.validate_with_serializer(interaction_data)
+        except RestValidationError as exc:
+            return (False, _flatten_serializer_errors_to_string(exc.detail))
+        # For our initial iteration of this feature, we are ignoring meeting updates
+        matching_interactions = Interaction.objects.filter(
+            source__meeting__id=interaction_data['meeting_details']['uid'],
+        )
+        if matching_interactions:
```
reupen: It would be good to use `matching_interactions.exists()` so that the query stops when as soon as it finds a match.

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,114 @@ def _get_utc_datetime(localised_datetime):
...
+            interaction_data['contacts'],
+            interaction_data['top_company'],
+        )
+        interaction_data['contacts'] = sanitised_contacts
+        try:
+            serializer = self.validate_with_serializer(interaction_data)
+        except RestValidationError as exc:
+            return (False, _flatten_serializer_errors_to_string(exc.detail))
+        # For our initial iteration of this feature, we are ignoring meeting updates
+        matching_interactions = Interaction.objects.filter(
```
reupen: I think we'll want to consider performance here as if it doesn't use an index it will do a sequential scan on the entire table.

I think the most generic approach is to create a GIN index (could be done in a separate PR) and use the contains lookup (e.g. `source__contains={'meeting': {'id': 123}}`).

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +68,207 @@ def calendar_data_fixture():
...
+        interaction_data_overrides,
+        calendar_data_fixture,
+        base_interaction_data_fixture,
+        monkeypatch,
+    ):
+        """
+        Test that process_email saves a draft interaction can be saved successfully
+        when the calendar parser yields good data.
+        """
+        interaction_data = {**base_interaction_data_fixture, **interaction_data_overrides}
```
reupen: Quite a long bit of consecutive lines of code here – would be nice to add some vertical white space.

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,114 @@ def _get_utc_datetime(localised_datetime):
...
+            source__meeting__id=interaction_data['meeting_details']['uid'],
+        )
+        if matching_interactions:
+            return (False, 'Meeting already exists as an interaction')
+        # We've validated and marshalled everything we need to build a draft interaction
```
elcct: Is this comment necessary?

datahub/interaction/test/email_processors/test_email_processors.py
currycoder: Agreed.  I like this approach better

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,114 @@ def _get_utc_datetime(localised_datetime):
         return localised_datetime
 
 
+def _flatten_serializer_errors_to_string(serializer_errors):
+    """
+    Flatten DRF Serializer validation errors to a string with one field per line.
+    """
+    error_parts = []
```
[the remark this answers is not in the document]
currycoder: Eventually, I would like to surface validation errors as a notification to users - so that they get some feedback about why their meeting email did not create a draft interaction on Data Hub.  There will certainly be some more work to do to make the validation errors meaningful enough to an end user (and wrap them in translation calls etc), but I would like to do this in a subsequent PR.  

Mainly I was considering here about how to transpose DRF-format validation errors in to something that was user-consumable.  It is a start but there is more to do later.

datahub/interaction/email_processors.py
currycoder: Great spot - thanks.  I'll change this as you specify and will queue up a subsequent PR to add the GIN index.

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,114 @@ def _get_utc_datetime(localised_datetime):
         return localised_datetime
 
 
+def _flatten_serializer_errors_to_string(serializer_errors):
```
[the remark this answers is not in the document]
currycoder: I don't quite agree on this.  I might be missing some context about DIT house-style, but right now the module is 330 lines of quite tightly related business logic.  I would certainly agree if we introduced another email processor to this file for a different business function.

Happy to discuss further.

datahub/interaction/email_processors.py
currycoder: It used to be back when the method was a bit of a mess!  You are right though, it is self-documenting now.

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +68,207 @@ def calendar_data_fixture():
...
+        email_parser_mock = self._get_email_parser_mock(interaction_data, monkeypatch)
+        processor = CalendarInteractionEmailProcessor()
+        result, message = processor.process_email(mock.Mock())
+        assert result is True
```
alixedi: It seems result is False when something goes wrong?

Can we use exceptions if this is the case?

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,114 @@ def _get_utc_datetime(localised_datetime):
...
+    Flatten DRF Serializer validation errors to a string with one field per line.
+    """
+    error_parts = []
+    for field_name, details in serializer_errors.items():
+        details_string = ','.join([str(detail) for detail in details])
+        error_parts.append(f'{field_name}: {details_string}')
+    return '\n'.join(error_parts)
+
+
+def _reduce_contacts_to_single_company(contacts, company):
```
alixedi: The `reduce` in the function's name is a bit confusing b/c it means something different in functional programming. Can we use `filter` instead? E.g. `_filter_contacts_by company`?

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,116 @@ def _get_utc_datetime(localised_datetime):
...
+
+
+class CalendarInteractionEmailProcessor(EmailProcessor):
+    """
+    An EmailProcessor which checks whether incoming email is a valid DIT/company
+    meeting, parses meeting information and creates a draft Interaction model
+    instance for it if the information is valid.
+    """
+
+    def _transform_to_serializer_format(self, data):
```
alixedi: Rename to `_to_serializer_format`?

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,116 @@ def _get_utc_datetime(localised_datetime):
...
+        return serializer
+
+    @transaction.atomic
+    def save_serializer_as_interaction(self, serializer, interaction_data):
+        """
+        Create the interaction model instance from the validated serializer.
+        """
+        interaction = serializer.save(
+            source={
+                'meeting': {'id': interaction_data['meeting_details']['uid']},
```
alixedi: This seems a bit confusing. The function is called `save_serializer_as_interaction` but we seem to be saving the `meeting_details` only?

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,116 @@ def _get_utc_datetime(localised_datetime):
...
+            return (False, exc.message)
+
+        # Make the same-company check easy to remove later if we allow Interactions
+        # to have contacts from more than one company
+        sanitised_contacts = _reduce_contacts_to_single_company(
+            interaction_data['contacts'],
+            interaction_data['top_company'],
+        )
+        interaction_data['contacts'] = sanitised_contacts
+        try:
```
alixedi: Can we not just try to save and catch the validation errors? Why are we having to validate as a separate step?

datahub/interaction/email_processors.py
currycoder: It does what it says it does!  Up until now, the serializer has not been saved as an interaction object yet, we save it with an extra override for the source JSON field. I'll add a comment to help clarify.

datahub/interaction/test/email_processors/test_email_processors.py
currycoder: This is the contract established in a previous PR for email processors to follow.  
They return a pair; first value is whether the email was processed successfully, second value is a meaningful success/error message.  
It could be argued that raising an exception would be more pythonic but I think it works either way really.

datahub/interaction/email_processors.py
currycoder: It's just the way DRF works:
```
In [1]: from datahub.interaction.serializers import InteractionSerializer                                                      

In [2]: s = InteractionSerializer({})                                                                                          

In [3]: s.save()                                                                                                               
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-d2c7335da61e> in <module>
----> 1 s.save()

/usr/local/lib/python3.7/site-packages/rest_framework/serializers.py in save(self, **kwargs)
    178 
    179         assert hasattr(self, '_errors'), (
--> 180             'You must call `.is_valid()` before calling `.save()`.'
    181         )
    182 

AssertionError: You must call `.is_valid()` before calling `.save()`.
```

datahub/interaction/test/email_processors/test_email_processors.py
alixedi: I think we should consider changing the contract. Exceptions are really nice and we are using them elsewhere in the codebase so this sticks out. If you are on the fence about Exceptions, do it for consistency please :)

currycoder: I think your arguments are valid here.  I would rather not vastly extend the scope of this PR though as I believe that will have a number of touch points in the `email_ingestion` package/tests.  I'll follow up on this change in a subsequent PR.

datahub/interaction/email_processors.py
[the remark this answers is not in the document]
reupen: I'm not sure I understand why it can't be converted into a sub-package with two modules and why that would be a problem.

That would also make it possible to keep the internal functions with the corresponding class they are used in.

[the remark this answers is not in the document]
currycoder: Something like `email_processors/processors.py` and `email_processors/calendar_invite_parser.py`?

I'm not saying that it is problematic - I want to fully understand and assimilate the reasoning.  My assertion here is that it is not a very large module at the moment and my understanding of general python practice is that multiple related classes in one module is fine.  Though I do see smatterings of separation in the codebase - e.g. the `company.models` sub-package.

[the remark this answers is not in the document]
marcofucci: yeah splitting would help a bit

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +66,212 @@ def calendar_data_fixture():
...
+    """
+    Basic interaction data spec which can be used to build a return value
+    which a mock of CalendarInteractionEmailParser can return.
+    """
+    return {
+        'sender_email': 'adviser1@trade.gov.uk',
+        'contact_emails': ['bill.adama@example.net', 'saul.tigh@example.net'],
+        'secondary_adviser_emails': [],
+        'date': datetime(2019, 5, 1, 13, 00, tzinfo=utc),
+        'top_company': 'Company 1',
```
marcofucci: `top_company_name`?

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +66,212 @@ def calendar_data_fixture():
...
+        interaction_data_overrides,
+        calendar_data_fixture,
+        base_interaction_data_fixture,
+        monkeypatch,
+    ):
+        """
+        Test that process_email saves a draft interaction when the calendar
+        parser yields good data.
+        """
+        interaction_data = {**base_interaction_data_fixture, **interaction_data_overrides}
```
marcofucci: I think 
```
        interaction_data = {
            **base_interaction_data_fixture, 
            **interaction_data_overrides,
        }
```

would be easier to read.

datahub/interaction/email_processors.py
```diff
@@ -98,6 +103,118 @@ def _get_utc_datetime(localised_datetime):
...
+            return (False, _flatten_serializer_errors_to_string(exc.detail))
+
+        # For our initial iteration of this feature, we are ignoring meeting updates
+        matching_interactions = Interaction.objects.filter(
+            source__contains={'meeting': {'id': interaction_data['meeting_details']['uid']}},
+        )
+        if matching_interactions.exists():
+            return (False, 'Meeting already exists as an interaction')
+
+        return self.save_serializer_as_interaction(serializer, interaction_data)
```
marcofucci: it's not obvious that `self.save_serializer_as_interaction(serializer, interaction_data)` is expected to return a tuple `(True, message)`. I would just return the interaction in `self.save_serializer_as_interaction(serializer, interaction_data)` and return `return (True, f'Successfully created interaction #{interaction.id}')` here instead

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +66,212 @@ def calendar_data_fixture():
...
+        email_parser_mock = self._get_email_parser_mock(interaction_data, monkeypatch)
+        assert not Interaction.objects.count()
+
+        # Process the message and save a draft interaction
+        processor = CalendarInteractionEmailProcessor()
+        result, message = processor.process_email(mock.Mock())
```
marcofucci: Test the `message` as well?

datahub/interaction/test/email_processors/test_email_processors.py
```diff
@@ -56,6 +66,212 @@ def calendar_data_fixture():
...
+        assert result is True
+        interaction = Interaction.objects.first()
+
+        # Verify dit_participants holds all of the advisers for the interaction
+        expected_adviser_emails = [interaction_data['sender_email']]
+        expected_adviser_emails.extend(interaction_data['secondary_adviser_emails'])
```
marcofucci:
```
        expected_adviser_emails = [
            interaction_data['sender_email'],
            **interaction_data['secondary_adviser_emails']
        ]
```
?

datahub/interaction/email_processors.py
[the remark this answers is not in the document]
currycoder: I've split this down now.

datahub/interaction/email_processors/processors.py
```diff
@@ -0,0 +1,126 @@
...
+from django.db import transaction
+from rest_framework import serializers
+
+from datahub.email_ingestion.email_processor import EmailProcessor
+from datahub.interaction.email_processors.parsers import CalendarInteractionEmailParser
+from datahub.interaction.models import Interaction
+from datahub.interaction.serializers import InteractionSerializer
+
+
+logger = getLogger(__name__)
```
marcofucci: I don't think you are using this?

datahub/interaction/test/email_processors/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+        email_prefix = name.lower().replace(' ', '.')
+        email = f'{email_prefix}@example.net'
+        contact = ContactFactory(
+            first_name=first_name,
+            last_name=last_name,
+            email=email,
+            company=company,
+        )
+        fixture['contacts'].append(contact)
+    yield fixture
```
marcofucci: It looks like you are not using `fixture` in your tests so maybe you can get rid of the dict and just have `yield` here?

datahub/interaction/email_processors/processors.py
```diff
@@ -0,0 +1,121 @@
...
+            serializer = self.validate_with_serializer(interaction_data)
+        except serializers.ValidationError as exc:
+            return (False, _flatten_serializer_errors_to_string(exc.detail))
+
+        # For our initial iteration of this feature, we are ignoring meeting updates
+        matching_interactions = Interaction.objects.filter(
+            source__contains={'meeting': {'id': interaction_data['meeting_details']['uid']}},
+        )
+        if matching_interactions.exists():
+            return (False, 'Meeting already exists as an interaction')
```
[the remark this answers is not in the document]
currycoder: Good point.  I'll have a proper think about this WRT exceptions in the follow-up PR, but I think there are 3 cases:
- This was processed, stop the chain.
- This was not processed, this processor is not interested in it.
- This was not processed, this processor was interested in it but the message was not valid.

It might be difficult to get across the distinction from the latter two cases, but I'll be thinking about it.

PR #1734, opened 2019-05-24
Activity feed endpoint

marcofucci:
### Description of change

This adds a new django app called `activity_feed` which acts as a proxy between the frontend and Activity Stream so that activities can be read from it.

I suggest you review the commits individually.

## Still to come

The endpoint should return an empty list if the logged-in user doesn't have permissions on all interactions.

### Checklist


datahub/activity_feed/views.py
```diff
@@ -0,0 +1,46 @@
...
+            settings.ACTIVITY_STREAM_OUTGOING_URL,
+            hawk_auth,
+            raise_for_status=False,
+        )
+        response = api_client.request(
+            request.method,
+            '',
+            data=request.body,
+            headers={
+                'Content-Type': request.content_type,
```
reupen: How does this behave if the request doesn't have a Content-Type header? Would be good to add a test for that case as well.

marcofucci: Thanks for the heads up! It sometimes defaults to `'application/x-www-form-urlencoded'` which makes things quite complicated. I'll add a check to make sure content type is always application/json

datahub/activity_feed/test/test_views.py
```diff
@@ -49,3 +50,28 @@ def test_get(self, requests_mock, request_data, response_status_code, response_c
...
         assert requests_mock.last_request.body == request_data
+
+    @pytest.mark.parametrize(
+        'content_type,expected_status_code',
+        (
+            (None, status.HTTP_400_BAD_REQUEST),
+            ('text/html', status.HTTP_400_BAD_REQUEST),
+            ('text/javascript', status.HTTP_400_BAD_REQUEST),
+            ('application/json', status.HTTP_200_OK),
+            ('application/ld+json', status.HTTP_200_OK),
```
reupen: Is `application/ld+json` accepted by the upstream server? Just asking as my local Elasticsearch seems to reject it:

`{"error":"Content-Type header [application/ld+json] is not supported","status":406}`

Also would be good to add one with an encoding e.g. `application/json; charset=utf-8`

datahub/activity_feed/views.py
```diff
@@ -20,6 +25,12 @@ class ActivityFeedView(APIView):
 
     def get(self, request):
         """Proxy for GET requests."""
+        if not JSON_CONTENT_TYPE_RE.match(request.content_type):
```
reupen: Not particularly important, but we could use one of the `parse_header` functions we are using elsewhere to make the check a bit tighter.

datahub/activity_feed/test/test_views.py
marcofucci: I'm trying to add as less custom validation as possible so the idea and delegate as much as possible but I think your suggestion go use cgi.parse_header and check for exactly application/json is a nice one. I'll update

README.md
```diff
@@ -35,19 +35,19 @@ Leeloo uses Docker compose to setup and run all the necessary components. The do
     * It will take time for the leeloo API container to come up - it will run
       migrations on both DBs, load initial data, sync elasticsearch etc. Watch
       along in the api container's logs.
-    * **NOTE:** 
-      If you are using a linux system, the  elasticsearch container may not 
-      come up successfully (`data-hub-leeloo_es_1`) - it might be perpetually 
+    * **NOTE:**
+      If you are using a linux system, the  elasticsearch container may not
```
elcct: `the  elas` <- too many spaces?

datahub/activity_feed/views.py
```diff
@@ -0,0 +1,56 @@
...
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
+
+    def get(self, request):
+        """Proxy for GET requests."""
+        content_type = request.content_type or ''
+        base_media_type, _ = parse_header(content_type.encode(HTTP_HEADER_ENCODING))
+        if base_media_type != 'application/json':
+            return HttpResponse(
+                'Please set Content-Type header value to application/json',
```
elcct: I think it will be more appropriate to ask for the request to have `JSON` content. If someone sends some random data, just changing the header won't make it valid.

datahub/activity_feed/views.py
```diff
@@ -0,0 +1,56 @@
...
+
+        api_client = APIClient(
+            settings.ACTIVITY_STREAM_OUTGOING_URL,
+            hawk_auth,
+            raise_for_status=False,
+        )
+        response = api_client.request(
+            request.method,
+            '',
+            data=request.body,
```
elcct: It seems like we are sending some body in the request, then why we don't use `POST` instead?

marcofucci: yeah, ES seems to prefer GET with the query in the body
https://www.elastic.co/guide/en/elasticsearch/reference/6.7/search-request-body.html

> Both HTTP GET and HTTP POST can be used to execute search with body. Since not all clients support GET with body, POST is allowed as well.

datahub/activity_feed/views.py
marcofucci: Yeah, I'd like to keep this as transparent as I can, I didn't even want to add the json check but we have some custom logic that does some magic if mime type is 'application/x-www-form-urlencoded' hence this check.


If the content is not JSON, Activity Stream will return some error and this is going to return it untouched.

datahub/activity_feed/views.py
elcct: That's not great but what to do.

PR #1735, opened 2019-05-24
Add duplicate checking to import interactions tool

reupen:
### Description of change

This adds additional validation to the import interactions tool that checks that:

- rows are not duplicates of existing objects
- rows are not duplicates of other rows in the file

Duplicate checking is done based on date, service and contact.

These fields were chosen for simplicity, and because we are primarily dealing with overseas delivery partner interactions which have specific services and are only created by importing CSV files.

I would suggest reviewing the commits individually.

### Screenshots

image

### To test

1. Create a valid CSV file according to the spec on the page, but containing rows where service, date and contact are the same as another row or an existing interaction
1. Select your CSV file
1. Submit the form to bring up the preview page
1. Click the import button on the preview page

```
Digital Data Hub - Live Service,<an adviser name>,,fooimport@bar.com,Email/Website,Account managemenT,interaction,22/02/2019,,some notes
```

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/duplicate_checking.py
+++ b/datahub/interaction/admin_csv_import/duplicate_checking.py
@@ -1,0 +1,63 @@
+"""
+Logic for detecting probable duplicates when importing interactions.
+
+This includes:
+
+- duplicates of existing interactions in the database
+
+Duplicate checking is done based on date, service and contact. These fields were chosen for
+simplicity, and because we are primarily dealing with overseas delivery partner interactions
+which have specific services and are only created by importing CSV files.
+"""
+
+from datahub.interaction.models import Interaction
+
+
+# They keys of this mapping correspond the cleaned data of an InteractionCSVRowForm and are used
+# to determine if an interaction is a duplicate.
+#
+# For example, if two contacts have the same date, contact and service they are considered a
+# duplicate.
+#
+# The values are field lookups for the Interaction model.
+#
+# (Note that the date field on the model is a DateTime field – only the date part is considered
+# when checking if a row is a duplicate of existing interactions.)
+DUPLICATE_FIELD_MAPPING = {
+    'date': 'date__date',
+    'contact': 'contacts',
+    'service': 'service',
+}
+
+
+def is_duplicate_of_existing_interaction(cleaned_data):
+    """
+    Check if a cleaned InteractionCSVRowForm is a duplicate of an existing interaction in the
+    database.
+
+    The query appears to be fairly fast, and makes use of existing indexes.
+    """
+    if not _cleaned_data_to_key(cleaned_data):
+        return False
+
+    filter_kwargs = {
+        queryset_field: cleaned_data[data_field]
+        for data_field, queryset_field in DUPLICATE_FIELD_MAPPING.items()
+    }
+
+    return Interaction.objects.filter(**filter_kwargs).exists()
+
+
+def _cleaned_data_to_key(cleaned_data):
+    """
+    Return a tuple representing a unique key for the cleaned data of an InteractionCSVRowForm.
+    """
+    # As an optimisation we could just track the pk for model instances,
+    # but that is omitted for simplicity
+    key = tuple(cleaned_data.get(field) for field in DUPLICATE_FIELD_MAPPING)
+
+    if all(key):
+        return key
+
+    # Some of the fields did not pass validation
+    return None
--- a/datahub/interaction/admin_csv_import/duplicate_checking.py
+++ b/datahub/interaction/admin_csv_import/duplicate_checking.py
@@ -4,6 +4,7 @@ Logic for detecting probable duplicates when importing interactions.
 This includes:
 
 - duplicates of existing interactions in the database
+- duplicates of other interactions in the CSV file being imported
 
 Duplicate checking is done based on date, service and contact. These fields were chosen for
 simplicity, and because we are primarily dealing with overseas delivery partner interactions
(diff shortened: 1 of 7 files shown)


datahub/interaction/admin_csv_import/duplicate_checking.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    if not _cleaned_data_to_key(cleaned_data):
+        return False
+
+    filter_kwargs = {
+        queryset_field: cleaned_data[data_field]
+        for data_field, queryset_field in DUPLICATE_FIELD_MAPPING.items()
+    }
+
+    return Interaction.objects.filter(**filter_kwargs).exists()
```
currycoder: How large do we expect the average (and largest) CSV files to be?  As I understand it, this change equates to `n` extra individual queries where `n` is the valid lines in the file.

datahub/interaction/admin_csv_import/duplicate_checking.py
```diff
@@ -0,0 +1,90 @@
...
+# to determine if an interaction is a duplicate.
+#
+# For example, if two contacts have the same date, contact and service they are considered a
+# duplicate.
+#
+# The values are field lookups for the Interaction model.
+#
+# (Note that the date field on the model is a DateTime field – only the date part is considered
+# when checking if a row is a duplicate of existing interactions.)
+DUPLICATE_FIELD_MAPPING = {
```
currycoder: I realise that you've detailed your reasons for looking up duplicates with these fields, but is there any value on matching by `source` as well?  I'm just thinking through a future case where we might re-use this mechanism and ignore as duplicate some interactions that occurred which were similar to interactions manually logged in the system.

datahub/interaction/admin_csv_import/duplicate_checking.py
reupen: Roughly 200 or so. I don't expect it to add much to the running time from some quick tests (we had c*N queries already from all the look-ups for contacts, advisers etc.)

datahub/interaction/admin_csv_import/duplicate_checking.py
reupen: Was there a particular scenario you had in mind? I think if a practical reason comes up why these particular criteria aren't working for us (I didn't actually select them) we can review it later on (and it should be easy to change them).

In terms of re-use, it's really only for `datahub.interaction.admin_csv_import` at the moment, so it would need some refactoring to make it more generic.

datahub/interaction/admin_csv_import/duplicate_checking.py
currycoder: Good point.

datahub/interaction/admin_csv_import/duplicate_checking.py
currycoder: What I meant by "re-use this mechanism", is in the sense that we are creating a new mechanism for ingesting CSVs of interactions.  This may have a particular company/use case in mind right now, but we may reuse the mechanism for a different use case in future - one where there is not a clear separation between CSV-ingested and user-created interactions.

For me, the practical criteria for a new use case is as follows:
An adviser logs an interaction for 05/09/2019 using the data hub frontend - it has a certain set of  contacts and is for a particular service.
There is subsequently a CSV import for another interaction which matches this criteria, but is distinct from the adviser-submitted interaction.  In this case, we erroneously mark it as a duplicate.

What I am suggesting is that the `source` field is just checked for a partial match.  e.g. "Only consider for duplicates interactions which were also imported from CSV files"

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -133,10 +140,11 @@ class InteractionCSVRowForm(forms.Form):
     subject = forms.CharField(required=False)
     notes = forms.CharField(required=False)
 
-    def __init__(self, *args, row_index=None, **kwargs):
+    def __init__(self, *args, duplicate_tracker=None, row_index=None, **kwargs):
         """Initialise the form with an optional zero-based row index."""
         super().__init__(*args, **kwargs)
         self.row_index = row_index
+        self.duplicate_tracker = duplicate_tracker
```
marcofucci: Non-critical: it would have been nice to add `duplicate_tracker` in the commit where it was used as well - it's not used in the commit I'm reviewing.

reupen: Whoops! Will fix when rebasing.

datahub/interaction/admin_csv_import/duplicate_checking.py
reupen: >  There is subsequently a CSV import for another interaction which matches this criteria, but is distinct from the adviser-submitted interaction. In this case, we erroneously mark it as a duplicate.

I don't have any evidence that that is likely to be a false positive – that seems to be an assumption, if I haven't missed anything?

Ideally, we wouldn't block the load completely if duplicates are detected and let the user decide what to do with each one. However, we're blocking for now as it's a much simpler solution.

currycoder: I think perhaps my assumption is that there is no other way that these CSV interactions can be entered in to the system; other than the CSV mechanism.  If that is the case, I would see a good case to additionally use the `source` field as the duplicate criteria.  I think you have the context here that I don't though!

Sounds good as a simple solution though - this can always be iterated on later if it becomes a problem.

elcct:
Looks good to me

PR #1742, opened 2019-05-28
Add service question and answers to metadata endpoint

elcct:
### Description of change

This adds `interaction_questions` field to `/metadata/service/` response. This field contains a representation of service questions and answer options from `ServiceQuestion`, `ServiceAnswerOption` ~and `ServiceAdditionalQuestion`~ models. The response has following format:

```
    [ # Array of ServiceQuestion
        {
            'id': <uuid>,
            'name: <str>,
            'disabled_on': <datetime>,
            'answer_options': [ # Array of ServiceAnswerOption
                {
                    'id': <uuid>,
                    'name': <str>,
                    'disabled_on': <datetime>,
                },
                ...
            ]
        },
        ...
    ]
```

The endpoint is supposed to return all questions and answers (including the disabled ones) and it will be up to the client to display appropriate ones.

This also adds the initial set of questions and answer options in a database migration.

Since the service metadata is going to be managed from the migrations, the ability to edit services in Django admin has been disabled.

Before merging I'll check if required services exist in the database and if not I will add them manually, but I expect them to exist.

### Checklist


datahub/interaction/migrations/0057_add_service_questions_and_answers.yaml
```diff
@@ -0,0 +1,734 @@
...
+    disabled_on:
+    name: Signed Export Growth Plan Received from company
+    question_id: 7360689f-3a9b-4139-b888-265ae1149bff
+    order: 400.0
+- model: interaction.servicequestion
+  pk: 13bdafaf-5a1f-4a8f-b622-46a29b413510
+  fields:
+    disabled_on:
+    name: Status
+    service_id: 380bba2b-3499-e211-a939-e4115bead28a
```
marcofucci: This is a bit dangerous as it's expecting an existing record which could have been deleted manually.

It also needs more work as existing TAP fields need to be deprecated and the logic replaced by these fields before they can go live.

Maybe exclude the TAP bit from this MVP and try to think about how to release that bit at a later stage?

datahub/interaction/migrations/0057_add_service_questions_and_answers.yaml
```diff
@@ -0,0 +1,734 @@
...
+  fields:
+    disabled_on:
+    name: Who was the company introduced to?
+    service_id: 1477622e-9adb-4017-8d8d-fe3221f1d2fc
+    order: 100.0
+- model: interaction.servicequestion
+  pk: 335f31fd-8d43-4061-9648-a68d00834a0e
+  fields:
+    disabled_on:
+    name: Was this of significant assistance?
```
marcofucci: It seems like it would be useful to have this on the actual interaction as a field as it will need to be queryable? Also what's the difference between this and a _service delivery_?

elcct: This only applies to some services. At the moment we only want to capture the additional information and querying is not in scope for MVP.

datahub/interaction/migrations/0057_add_service_questions_and_answers.yaml
elcct: I think we need the service that the question is attached to. I could add a step in the migration to add those services if found missing. I think we should not be allowed to delete them (use disabled_on field instead) so I could also remove that option from the admin.

Re TAP, I agree but let's talk about it offline.

datahub/metadata/test/test_views.py
```diff
@@ -244,6 +246,65 @@ def test_list_filter_by_has_any(self, api_client, contexts):
...
+        db_service = service_additional_question.answer_option.question.service
+
+        for service in services:
+            if service['id'] == str(db_service.id):
+                response_service = service
+                break
+
+        disabled_on = format_date_or_datetime(
+            db_service.disabled_on,
+        ) if db_service.disabled_on else None
```
marcofucci: You are using this in a couple of places so you should use a small help function.

datahub/interaction/migrations/0057_add_service_questions_and_answers.yaml
marcofucci: As discussed, I leave this to you To decide

datahub/interaction/migrations/0057_add_service_questions_and_answers.yaml
marcofucci: As discussed, I leave this to you to decide

datahub/metadata/test/test_views.py
```diff
@@ -244,6 +246,63 @@ def test_list_filter_by_has_any(self, api_client, contexts):
...
+        url = reverse(viewname='service')
+        response = api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+        services = response.json()
+
+        service_additional_question = ServiceAdditionalQuestion.objects.first()
+        db_service = service_additional_question.answer_option.question.service
+
+        for service in services:
```
reupen: I think you could do something like:

```python
response_service = next(service for service in services if service['id'] == str(db_service.id))
```

datahub/interaction/migrations/0057_add_service_questions_and_answers.yaml
```diff
@@ -0,0 +1,734 @@
...
+  fields:
+    disabled_on:
+    name: Customers
+    question_id: 26c08878-8dd1-4abf-bc59-84918086bf29
+    order: 100.0
+- model: interaction.serviceansweroption
+  pk: faa5a5c7-5703-46d8-a35c-c92a87850655
+  fields:
+    disabled_on:
+    name: Business Partners (e.g. distributers or manufacturers)
```
reupen: Should it be distributors rather than distributers? (Same for the other one above.)

elcct: Thanks! There was a lot of typos and it seems I didn't catch all.

datahub/metadata/test/test_views.py
elcct: That's neat! I forgot this. Thanks!

datahub/metadata/test/test_views.py
```diff
@@ -313,3 +371,10 @@ def test_list(self, api_client):
         assert set(first_project_stage.keys()) == set(expected_items)
         assert first_project_stage['name'] == 'Prospect'
         assert not first_project_stage['exclude_from_investment_flow']
+
+
+def _format_datetime_field_if_exists(obj, field_name):
+    value = getattr(obj, field_name)
+    if value is None:
```
rafenden: Could be rewritten as:
```
if value is not None:
  return format_date_or_datetime(value)
```

Functions by default return `None` so there is need to explicitly return empty value.

elcct: We prefer to be explicit, so I did this on purpose.

reupen:
This is now unblocked and has been significantly simplified:

- the TAP, ERDF, TDF and Global Growth questions have been removed (TAP and Global Growth will continue to function as they currently do)
- the significant assistance questions have been removed
- `additional_questions` has been removed from the `/metadata/service/` response as it's no longer being used anywhere
- the new services are now added in a migration in the `metadata` app as that is where the Service model is

I will squash most of the commits before merging.

datahub/interaction/migrations/0059_add_service_questions_and_answers.yaml
```diff
@@ -0,0 +1,277 @@
+# Service: Making Introductions (Export)
+- model: interaction.servicequestion
+  pk: 054b29fe-c14b-463d-8177-c378d3a819aa
+  fields:
+    disabled_on:
+    name: Who was the company introduced to?
```
currycoder: I find that `name` is a bit of a misnomer here, could `text` be better?  Appreciate this may be out of scope for this PR.

datahub/metadata/test/test_views.py
```diff
@@ -285,6 +288,79 @@ def test_feature_flagged_services_excluded_if_feature_flag_inactive(self, api_cl
...
+        )
+
+        response_service['contexts'] = sorted(response_service['contexts'])
+
+        assert response_service == {
+            'id': str(service_obj.pk),
+            'name': service_obj.name,
+            'contexts': sorted(service_obj.contexts),
+            'disabled_on': _format_datetime_field_if_exists(service_obj, 'disabled_on'),
+            'interaction_questions': [],
```
currycoder: A thought on this - does the frontend care that this key is still present even though the feature flag is inactive?

datahub/interaction/migrations/0059_add_service_questions_and_answers.yaml
reupen: I agree, it's not ideal in this case. However, as you hinted at, it's how the model is currently defined (as it's a metadata model).

datahub/metadata/test/test_views.py
reupen: The develop branch won't, but the branch with the corresponding work for this feature probably does. Just trying to keep the disruption to a minimum...

PR #1747, opened 2019-05-30
Fix for adviser and team being null

alixedi:
### Description of change

After merging #1721, we found that the following fields are `None` in develop and staging for some records:

1) `InteractionDITParticipant.adviser`
2) `InteractionDITParticipant.team`

This PR reproduces the problem and introduces a fix.

### Checklist

### DIFF
--- a/datahub/activity_stream/interaction/serializers.py
+++ b/datahub/activity_stream/interaction/serializers.py
@@ -54,14 +54,18 @@ class InteractionActivitySerializer(serializers.ModelSerializer):
 
     def _get_dit_participants(self, participants):
         return [
-            {
-                'id': f'dit:DataHubAdviser:{participant.adviser.pk}',
-                'type': ['Person', 'dit:Adviser'],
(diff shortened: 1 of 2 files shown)


datahub/activity_stream/interaction/serializers.py
```diff
@@ -54,15 +54,18 @@ def _get_company(self, company):
...
         return [
-            {
-                'id': f'dit:DataHubAdviser:{participant.adviser.pk}',
-                'type': ['Person', 'dit:Adviser'],
-                'dit:emailAddress': participant.adviser.contact_email or participant.adviser.email,
-                'name': participant.adviser.name,
-            }
-            for participant in participants.all()
+            self._get_adviser(participant.adviser)
+            for participant in participants.filter(adviser__isnull=False)
```
reupen: We'll have the same problem with `.filter()` defeating the prefetching as before – so would do the adviser is not None check in Python.

PR #1748, opened 2019-05-30
Rework email sender verification to use authres library

currycoder:
### Description of change

This reworks the `_verify_authentication()` helper method so that it uses a third party library for parsing the `Authentication-Results` header.  It became clear that the solution that I had written was not fit for purpose for real-world headers as they are not as handily formatted as headers provided by GSuite-managed mailboxes.

The opportunity was also taken to make it so that `_verify_authentication()` could be configured with a kwarg `auth_methods` to verify specific methods only (out of dmarc/spf/dkim).  By default, if the kwarg is not specified, it will verify all of these methods.  This was done in anticipation of some DIT adviser email domains offering patchy sender authentication methods.

### Points of note
 
- I've removed the check on `smtp.mailfrom` as the contents for this field in real-world headers is quite varied and unpredictable.  `spf=pass` should be enough in these cases.

### Checklist


requirements.txt
```diff
@@ -2,17 +2,18 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file=requirements.txt requirements.in
+#    pip-compile --upgrade --output-file requirements.txt requirements.in -v
```
reupen: Not sure why `pip-tools` is changing this – is yours up to date?

currycoder: I've just done some testing of this locally and I believe this line changes to reflect the way that it was called.  In my case, I called with the `--upgrade` flag as specified in https://github.com/uktrade/data-hub-leeloo/blob/8103ed46283ba04a9f24cb1077c414d111640a90/docs/Managing%20dependencies.md

Should the documentation be adjusted here to omit the upgrade flag?

datahub/email_ingestion/validation.py
```diff
@@ -1,25 +1,26 @@
...
+    :param message: mailparse.MailParser object - the message to check
+    :param auth_methods: Optional - An iterable of email authentication methods
+        to check.  Defaults to ('dkim', 'spf', 'dmarc').
+
+    :returns: A boolean for whether or not the Authentication-Results header
+        was verified.
+    """
+    if not auth_methods:
+        auth_methods = ('dkim', 'spf', 'dmarc')
+    header_contents = ' '.join(message.authentication_results.split('\n'))
```
reupen: Any reason for using `.split('\n')` rather than `.splitlines()`?

datahub/email_ingestion/validation.py
```diff
@@ -1,25 +1,26 @@
...
-        if line.startswith('dkim'):
-            auth_results['dkim'] = line.startswith('dkim=pass')
-        if line.startswith('spf'):
-            spf_valid = (
-                line.startswith('spf=pass') and line.endswith(f'smtp.mailfrom={from_email};')
-            )
-            auth_results['spf'] = spf_valid
-        if line.startswith('dmarc'):
-            auth_results['dmarc'] = line.startswith('dmarc=pass')
+def _verify_authentication(message, auth_methods=None):
```
reupen: Is the `auth_methods` argument needed? (Can't see where it's being used.)

requirements.txt
reupen: The current version of `pip-tools` isn't meant to add `--upgrade` and `-v`, that's why I was wondering if yours was up to date (with what's in `requirement.txt`).

datahub/email_ingestion/validation.py
currycoder: Good point!

datahub/email_ingestion/validation.py
currycoder: Not strictly right now.  As I said in the PR description, I've added it in anticipation of us needing to have configurable auth methods per-email domain.  In parallel, we have some ongoing user research work to gather samples of emails from the different DIT adviser domains and I am anticipating that the auth methods provided by emails from these different domains will be patchy.

requirements.txt
currycoder: I think you are right.  I've fully updated the python env and don't see it adjusting that anymore.  
Something that is a little problematic with the docker workflow is that requirements are only installed at the time of container build - which can be quite infrequent if using a tool like docker-compose.  I'll have a think about this in my next round of docker improvements.

In the meantime, I'll commit a fixup to avoid the change on this line :+1:

datahub/email_ingestion/validation.py
reupen: Ah, sorry, missed that. It doesn't seem to be covered by tests and is only a couple of lines, so would suggest removing it until it's needed, but not a big deal.

currycoder: I'll leave it in - but I do appreciate that we could remove until it is strictly needed.

PR #1749, opened 2019-05-30
Add ability to download unmatched rows to the import interactions tool

reupen:
### Description of change

This adds the ability to the import interactions tool to download rows that could not be matched with active contacts following a successful import operation.

The download is a CSV file is in the same file format as input files so it's possible to re-upload it at a later date to try again.

Similar to previous steps, this data is temporarily held in the cache following a successful import.

I would suggest reviewing the commits individually.

#### With unmatched contacts

image

#### Without unmatched contacts

image

### To test

1. Create a valid CSV file according to the spec on the page (containing some rows where `contact_email` matches an active contact and some where it doesn't)
1. Select your CSV file
1. Submit the form to bring up the preview page
1. Click the import button on the preview page
1. Click the download unmatched rows link and save the file
1. Compare the contents of the downloaded file with the one originally uploaded (it should contain the expected rows and data)
1. Try to import the downloaded file (the unmatched records page should be displayed)

```
Digital Data Hub - Live Service,<an adviser name>,,fooimport@bar.com,Email/Website,Account managemenT,interaction,25/02/2019,,some notes
```

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/file_form.py
+++ b/datahub/interaction/admin_csv_import/file_form.py
@@ -1,4 +1,7 @@
+import csv
 import hashlib
+import io
+from codecs import BOM_UTF8
 from secrets import token_urlsafe
 
 import reversion
@@ -18,6 +21,40 @@ from datahub.interaction.admin_csv_import.row_form import InteractionCSVRowForm
 
 
 REVISION_COMMENT = 'Imported from file via the admin site.'
+
+
+class UnmatchedRowCollector:
+    """Holds unmatched rows following an import operation."""
+
+    def __init__(self):
+        """Initialise the instance with an empty list of rows."""
+        self.rows = []
+
+    def append_row(self, row_form):
+        """Add an unmatched row."""
+        self.rows.append(row_form.data)
+
+    def to_raw_csv(self):
+        """
+        Generate a CSV file from collected unmatched rows.
+
+        The CSV file is in the same format as the input file (so that it can be re-uploaded
+        at a later date if required).
+        """
+        if not self.rows:
+            return None
+
+        byte_stream = io.BytesIO()
+        byte_stream.write(BOM_UTF8)
+
+        with byte_stream:
+            text_stream = io.TextIOWrapper(byte_stream, encoding='utf-8', write_through=True)
+            # Use the keys of the first row for the fieldnames argument (all rows should have
+            # the same keys)
+            writer = csv.DictWriter(text_stream, self.rows[0].keys())
+            writer.writeheader()
+            writer.writerows(self.rows)
+            return byte_stream.getvalue()
 
 
 class InteractionCSVForm(BaseCSVImportForm):
@@ -78,14 +115,17 @@ class InteractionCSVForm(BaseCSVImportForm):
         }
 
         matching_counts = {status: 0 for status in ContactMatchingStatus}
+        unmatched_row_collector = UnmatchedRowCollector()
 
         for row_form in self._get_row_form_iterator(raise_error_if_invalid=True):
             if row_form.is_matched():
                 row_form.save(user, source)
+            else:
+                unmatched_row_collector.append_row(row_form)
 
             matching_counts[row_form.cleaned_data['contact_matching_status']] += 1
 
-        return matching_counts
+        return matching_counts, unmatched_row_collector
 
     def save_to_cache(self):
         """
--- a/datahub/interaction/templates/admin/interaction/interaction/import_complete.html
+++ b/datahub/interaction/templates/admin/interaction/interaction/import_complete.html
@@ -13,6 +13,15 @@
 
   {% include 'admin/interaction/interaction/fragment_post_import_counts.html' with num_matched=num_matched num_unmatched=num_unmatched num_multiple_matches=num_multiple_matches only %}
 
+  {% if num_unmatched or num_multiple_matches %}
+    <p>
+      {% url opts|admin_urlname:'import-download-unmatched' token=token as unmatched_download_url %}
+      {% blocktrans %}
+        <a href="{{ unmatched_download_url }}">Download unmatched interactions</a> (link valid for {{ download_timeout_mins }} minutes after importing)
+      {% endblocktrans %}
+    </p>
+  {% endif %}
+
   <p>
     <a href="{% url opts|admin_urlname:'changelist' %}">
       {% trans 'Return to the interaction list' %}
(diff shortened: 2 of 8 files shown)


datahub/interaction/admin_csv_import/file_form.py
```diff
@@ -20,6 +23,38 @@
...
+        """
+        if not self.rows:
+            return None
+
+        byte_stream = io.BytesIO()
+        byte_stream.write(BOM_UTF8)
+
+        with byte_stream:
+            text_stream = io.TextIOWrapper(byte_stream, encoding='utf-8', write_through=True)
+            writer = csv.DictWriter(text_stream, InteractionCSVRowForm.base_fields.keys())
```
marcofucci: Would it be better to get the fieldnames from the row instead of from the form? This way you are guaranteed to have exactly the same format even in case of extra columns.

datahub/interaction/templates/admin/interaction/interaction/import_complete.html
```diff
@@ -13,6 +13,14 @@ <h2>
 
   {% include 'admin/interaction/interaction/fragment_post_import_counts.html' with num_matched=num_matched num_unmatched=num_unmatched num_multiple_matches=num_multiple_matches only %}
 
+  {% if num_unmatched or num_multiple_matches %}
+    <p>
+      <a href="{% url opts|admin_urlname:'import-download-unmatched' token=token %}">
+        {% trans 'Download unmatched interactions' %}
+      </a> (link valid for {{ download_timeout_mins }} minutes after importing)
```
currycoder: This line seems to be missing translation

currycoder: As a side question, how likely do we think it is that Data Hub will offer multiple languages within the next couple of years?

datahub/interaction/test/admin_csv_import/test_file_form.py
```diff
@@ -26,6 +32,78 @@
 )
 
 
+@pytest.mark.django_db
+class TestUnmatchedRowCollector:
```
currycoder: A thought about the tests in general.  Could there be any value for creating a test case which ensures that the files in the cache expire as expected?  I realise that we are using standard/well tested tools for setting cache values, but there is a vector here for using up resources.

datahub/interaction/templates/admin/interaction/interaction/import_complete.html
reupen: Whoops, will fix, thanks!

datahub/interaction/admin_csv_import/file_form.py
reupen: Yes, it would be better. Could get them from the first row since it should be the same for all rows...

datahub/interaction/test/admin_csv_import/test_file_form.py
reupen: Yes, it would be good. Could use `@freeze_time` for that – will probably be easier to test the functions in `cache_utils` directly.

(As an aside, I'm not overly concerned as our current Redis memory usage is pretty low and even if it does hit the memory limit [it's configured to evict the least recently used volatile keys](https://docs.cloud.service.gov.uk/deploying_services/redis/#redis-key-eviction-policy). But do need to make sure that the expiry times are being set.)

currycoder: Sounds good yeah.

(It was your discussion about upgrading redis which made me think about this actually!)

datahub/interaction/admin_csv_import/file_form.py
reupen: Now getting them from the first row.

datahub/interaction/templates/admin/interaction/interaction/import_complete.html
reupen: > As a side question, how likely do we think it is that Data Hub will offer multiple languages within the next couple of years?

I very much doubt it (especially for the admin site).

datahub/interaction/test/admin_csv_import/test_file_form.py
reupen: I've added a test for key expiry for `save_unmatched_rows_csv_contents()`. Will pick up the other functions in another PR.

datahub/interaction/test/admin_csv_import/test_cache_utils.py
```diff
@@ -100,11 +103,23 @@ def test_saves_unmatched_rows(self):
...
+        contents = b'file-contents'
+        token = 'test-token'
+        base_datetime = datetime(2019, 2, 3)
+
+        with freeze_time(base_datetime):
+            save_unmatched_rows_csv_contents(token, contents)
+
+        key = _cache_key_for_token(token, CacheKeyType.unmatched_rows)
+
+        with freeze_time(base_datetime + CACHE_VALUE_TIMEOUT + timedelta(minutes=1)):
```
currycoder: I was surprised that this works, since my understanding is that `freeze_time` only freezes the time for python.  Then I saw that we were using `LocMemCache` for tests, so I suppose that it is an implementation detail that expiry in `LocMemCache` will be based on python time too..

If my understanding here is correct I feel that there may be a risk of us switching to an externally managed cache for tests later, which could cause this test to fail.  I was expecting that the test would modify the expiry time and do some waiting, though I do appreciate it is cleaner to explicitly jump around time with freezegun if we are able to.

reupen: The test explicitly uses `local_memory_cache` (by default, the tests use a dummy cache). The only other alternative I can think of is patching `cache.set` and checking what `timeout` it was called with. Did you have something particular in mind?

reupen: The test explicitly uses `local_memory_cache` (by default, the tests use a dummy cache). We used `freeze_time` in similar tests in https://github.com/uktrade/data-hub-leeloo/blob/dda6b4d47ca9399856c377c6c34db646fe146457/datahub/core/test/test_throttling.py as well.

I would prefer not to rely on a real clock in tests if possible. The only other alternative I can think of is patching `cache.set` and checking what `timeout` it was called with.

currycoder: I think this is cool as is - especially since the test is explicit about using `local_memory_cache`, I didn't realise this was the case!

PR #1755, opened 2019-06-03
Email invite interaction subject

currycoder:
### Description of change

This PR achieves two things:
- Ensures that the sender adviser is not included in the list of `secondary_advisers` returned by the email invite parser class.
- Reworks the interaction subject to be a templated string of `"Meeting between {recipients}"` - this is a business requirement as we cannot be sure that meeting subjects will be suitable as an interaction subject.

### Points of note
- The meeting subject template string is not wrapped with ugettext calls.  I couldn't find a suitable solution for translating a (varied length) list of entities e.g. "{recipient_1}, {recipient_2}, {recipient_3} and {recipient_4}".  I'm certain that different languages will handle "lists of things" in different ways - can anybody suggest a clean way to wrap in translation in this case?

### Checklist

### DIFF
--- a/datahub/interaction/email_processors/processors.py
+++ b/datahub/interaction/email_processors/processors.py
@@ -27,6 +27,31 @@ def _filter_contacts_to_single_company(contacts, company):
     return [contact for contact in contacts if contact.company == company]
 
 
+def _get_meeting_subject(sender, contacts, secondary_advisers):
+    """
+    Construct and return a meeting subject given a sender, contacts and secondary
+    advisers (if present).
+    """
+    adviser_names = [
+        adviser.name
+        for adviser in (sender, *secondary_advisers)
+        if adviser.name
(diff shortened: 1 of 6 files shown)


datahub/interaction/email_processors/processors.py
```diff
@@ -27,6 +27,31 @@ def _filter_contacts_to_single_company(contacts, company):
...
+    if not adviser_names:
+        adviser_names = ['DIT']
+    contact_names = [contact.name for contact in contacts if contact.name]
+    if not contact_names:
+        try:
+            contact_names = [contacts[0].company.name]
+        except IndexError:
+            # This is not going to be a valid interaction, but this helper function
+            # should not raise an exception
+            contact_names = ['A company']
```
reupen: Doesn't look like A needs to be capitalised here.

PR #1762, opened 2019-06-03
Add basic permission logic to activity feed

marcofucci:
### Description of change

This makes sure `/v4/activity-feed` returns an empty list if the authenticated user doesn't have permissions to view all interactions, investment projects or OMIS orders.

I decided to return an empty list instead of status code 403 as the ideal implementation would be to filter out all activities that the user can't see. Open to suggestions though.

### Checklist

### DIFF
--- a/datahub/activity_feed/test/test_views.py
+++ b/datahub/activity_feed/test/test_views.py
@@ -3,7 +3,32 @@ from django.conf import settings
 from rest_framework import status
 from rest_framework.reverse import reverse
 
-from datahub.core.test_utils import APITestMixin
+from datahub.activity_feed.views import ActivityFeedView
+from datahub.core.test_utils import APITestMixin, create_test_user
+
+
+def pytest_generate_tests(metafunc):
+    """
+    Parametrizes the tests that use the `insufficient_activity_permissions` fixture
+    by creating lists with all required permissions except one.
+    """
+    if 'insufficient_activity_permissions' in metafunc.fixturenames:
+        perms = ActivityFeedView.ACTIVITY_MODELS_PERMISSIONS_REQUIRED
+        insufficient_activity_permissions = [
+            pytest.param(
+                [
+                    perm
+                    for perm in perms
+                    if perm != perm_to_exclude
+                ],
+                id=f'without {perm_to_exclude}',
+            )
+            for perm_to_exclude in perms
+        ]
+        metafunc.parametrize(
+            'insufficient_activity_permissions',
+            insufficient_activity_permissions,
+        )
 
 
 class TestActivityFeedView(APITestMixin):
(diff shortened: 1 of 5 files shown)


datahub/activity_feed/views.py
```diff
@@ -15,21 +17,59 @@ class ActivityFeedView(APIView):
...
+                    },
+                },
+            )
+            return HttpResponse(
+                response_content,
+                status=status.HTTP_200_OK,
+                content_type=content_type,
+            )
+
+        # get response from upstream
```
elcct: I think this comment is not necessary

datahub/activity_feed/views.py
```diff
@@ -15,21 +17,59 @@ class ActivityFeedView(APIView):
...
             return HttpResponse(
                 'Please set Content-Type header value to application/json',
                 status=status.HTTP_406_NOT_ACCEPTABLE,
             )
 
+        # if the user doesn't have permissions on all activity models, return an empty list
+        # TODO: instead of returning an empty list, we need to filter out the
+        # activities that the user doesn't have access to
+        if not request.user.has_perms(self.ACTIVITY_MODELS_PERMISSIONS_REQUIRED):
+            response_content = json.dumps(
```
reupen: Is there a response for not using [`JsonResponse` (from Django)](https://docs.djangoproject.com/en/2.2/ref/request-response/#jsonresponse-objects)?

reupen:
@marcofucci Not directly related to this, but I get a failure locally when running the activity feed tests. I assume it needs some of the environment variables to be set, it would be good to set them in `datahub.config.test` instead of the CircleCI config.

datahub/activity_feed/test/test_views.py
```diff
@@ -3,7 +3,34 @@
...
+
+
+def pytest_generate_tests(metafunc):
+    """
+    Parametrizes the tests that use the `insufficient_activity_permissions` fixture
+    by creating lists with all required permissions except one.
+    """
+    if 'insufficient_activity_permissions' in metafunc.fixturenames:
+        perms = ActivityFeedView.ACTIVITY_MODELS_PERMISSIONS_REQUIRED
+        insufficient_activity_permissions = [
```
reupen: It's a bit more indentation in this case, but I'd consider using `pytest.param()` to set the `id`s (to keep them with the params):

```python
        insufficient_activity_permissions = [
            pytest.param(
                [
                    perm
                    for perm in perms
                    if perm != perm_to_exclude
                ],
                id=f'without {perm_to_exclude}',
            )
            for perm_to_exclude in perms
        ]
```

(Up to you though.)

marcofucci: very nice, thanks!

datahub/activity_feed/views.py
```diff
@@ -15,21 +15,55 @@ class ActivityFeedView(APIView):
...
+                    'hits': {
+                        'total': 0,
+                        'hits': [],
+                    },
+                },
+                status=status.HTTP_200_OK,
+                content_type=content_type,
+            )
+
+        upstream_response = self._get_upstream_response(request)
```
currycoder: I find the "next bit" quite interesting here.  Is it possible to rewrite the request body before proxying upstream such that we only request items that the user is allowed to see?  Or will the results be filtered after the fact?

I realise this is peripheral to the PR in question, I'm just interested!

marcofucci: Yeah the idea is to add bits to the request body when received by the proxy and before sending it upstream. It's quite complicated because of how flexible an ES query can be so we'll see how it goes.

PR #1763, opened 2019-06-03
Investment project activity stream

alixedi:
### Description of change

Add activity stream for investment project created for the following design:

Screenshot 2019-06-03 at 15 50 37

I have also promoted common functionality in this and interaction activity views, serialisers etc. to base classes and refactored the tests accordingly.

### Checklist


datahub/activity_stream/tests/test_cursor_pagination.py
```diff
@@ -0,0 +1,40 @@
...
+
+    response = hawk.get(api_client, next_page_url)
+    data = response.json()
+    previous_page_url = data['previous']
+    assert len(data['orderedItems']) == len(interactions) - page_size
+
+    response = hawk.get(api_client, previous_page_url)
+    data = response.json()
+    assert len(data['orderedItems']) == page_size
+    assert data['next'] == next_page_url
```
currycoder: I find the variable name overwriting/reuse here a little difficult to follow, could `data` be changed to be more meaningful for each step?  e.g. `page_1_data` etc?

datahub/activity_stream/views.py
```diff
@@ -0,0 +1,17 @@
...
+    HawkAuthentication,
+    HawkResponseSigningMixin,
+    HawkScopePermission,
+)
+from datahub.core.viewsets import CoreViewSet
+
+
+class ActivityViewSet(HawkResponseSigningMixin, CoreViewSet):
+    """
+    Generic view for activities.
```
currycoder: Could this docstring be extended a little bit?  A new developer would probably want a little more context about what this viewset can enable them to do.

datahub/activity_stream/serializers.py
```diff
@@ -0,0 +1,35 @@
+from rest_framework import serializers
+
+
+class ActivitySerializer(serializers.ModelSerializer):
+    """
+    Generic serializer for activity.
```
currycoder: Please consider extending this docstring a little bit about the goal/uses of the class.

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,215 @@
...
+        'next': None,
+        'orderedItems': [
+            {
+                'id': f'dit:DataHubInvestmentProject:{project.id}:Add',
+                'type': 'Add',
+                'published': format_date_or_datetime(project.created_on),
+                'generator': {'name': 'dit:dataHub', 'type': 'Application'},
+                'object': {
+                    'id': f'dit:DataHubInvestmentProject:{project.id}',
+                    'type': ['dit:Event', 'dit:InvestmentProject'],
```
marcofucci: I think we can take the `dit:Event` out as it's not really an event.

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,215 @@
...
+        'summary': 'Investment Activities',
+        'type': 'OrderedCollectionPage',
+        'id': 'http://testserver/v3/activity-stream/investments',
+        'partOf': 'http://testserver/v3/activity-stream/investments',
+        'previous': None,
+        'next': None,
+        'orderedItems': [
+            {
+                'id': f'dit:DataHubInvestmentProject:{project.id}:Add',
+                'type': 'Add',
```
marcofucci: It would be good to add an `actor` field (the person who created the project).

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,215 @@
...
+                'object': {
+                    'id': f'dit:DataHubInvestmentProject:{project.id}',
+                    'type': ['dit:Event', 'dit:InvestmentProject'],
+                    'dit:name': project.name,
+                    'dit:status': project.status,
+                    'dit:investment_type': project.investment_type.name,
+                    'dit:estimated_land_date': format_date_or_datetime(
+                        project.estimated_land_date,
+                    ),
+                    # 'dit:total_investment_value': project.total_investment,
```
currycoder: What's the context for these comments?  Could do with a TODO if they will be added later?

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,215 @@
...
+        'orderedItems': [
+            {
+                'id': f'dit:DataHubInvestmentProject:{project.id}:Add',
+                'type': 'Add',
+                'published': format_date_or_datetime(project.created_on),
+                'generator': {'name': 'dit:dataHub', 'type': 'Application'},
+                'object': {
+                    'id': f'dit:DataHubInvestmentProject:{project.id}',
+                    'type': ['dit:Event', 'dit:InvestmentProject'],
+                    'dit:name': project.name,
```
marcofucci: There's a `name` in the specs so we don't need the `dit:` namespace I guess?

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,215 @@
...
+                    'dit:status': project.status,
```
marcofucci: I don't think we need `dit:status`

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,215 @@
...
+                    'dit:investment_type': project.investment_type.name,
```
marcofucci: I would have a dict with just a name field as the other ones. This is so that we can easily add the id as well if we need to in the future.

datahub/activity_stream/investment/views.py
```diff
@@ -0,0 +1,34 @@
...
+from datahub.company.models import Contact
+from datahub.investment.project.models import InvestmentProject
+
+
+class InvestmentCursorPagination(ActivityCursorPagination):
+    """
+    Cursor pagination for investment projects.
+    """
+
+    ordering = ('created_on', )
```
reupen: I expect that `created_on` will be messy for projects migrated from the legacy system... Hence, it may be worthwhile adding `pk` as well.

datahub/activity_stream/investment/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+            project['object']['dit:estimated_land_date'] = instance.estimated_land_date
+
+        if instance.total_investment is not None:
+            project['object']['dit:total_investment_value'] = instance.total_investment
+
+        if instance.foreign_equity_investment is not None:
+            project['object']['dit:foreign_equity_investment'] = instance.foreign_equity_investment
+
+        if instance.number_new_jobs is not None:
+            project['object']['dit:number_of_new_jobs'] = instance.number_new_jobs
```
reupen: Do we need to rename the key from `number_new_jobs` to `number_of_new_jobs`?

datahub/activity_stream/investment/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+                ],
+                'url': instance.get_absolute_url(),
+            },
+        }
+
+        if instance.estimated_land_date is not None:
+            project['object']['dit:estimated_land_date'] = instance.estimated_land_date
+
+        if instance.total_investment is not None:
+            project['object']['dit:total_investment_value'] = instance.total_investment
```
reupen: Is there a reason for renaming the key from `total_investment` to `total_investment_value`? (Just thinking of consistency...)

datahub/activity_stream/urls.py
```diff
@@ -1,11 +1,17 @@
...
+from datahub.activity_stream.investment.views import InvestmentActivityViewSet
 
 activity_stream_urls = [
     path(
         'activity-stream/interactions',
         InteractionActivityViewSet.as_view({'get': 'list'}),
         name='interactions',
     ),
+    path(
+        'activity-stream/investments',
```
reupen: Looks like it was missed in the last endpoint, but the rest of the API uses singular nouns – it would be good to keep it consistent for the whole API.

datahub/activity_stream/tests/__init__.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework.reverse import reverse
+
+
+def get_url(name):
```
reupen: Generally we've avoided defining functions directly in `__init__.py` – I think the most common pattern currently used for tests is to define a `utils` module for shared test utility functions and constants.

datahub/activity_stream/investment/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+    pagination_class = InvestmentCursorPagination
+    serializer_class = InvestmentActivitySerializer
+    queryset = InvestmentProject.objects.select_related(
+        'investment_type',
+        'investor_company',
+        'investor_type',
+        'stage',
+    ).prefetch_related(
+        'created_by',
```
reupen: Couldn't `created_by` be in `select_related`? Or are you expecting better performance having it in `prefetch_related`?

datahub/activity_stream/investment/serializers.py
alixedi: I don't think so 👍

datahub/activity_stream/investment/serializers.py
alixedi: I don't think so 👍

datahub/activity_stream/tests/__init__.py
alixedi: Any reason why we have generally avoided defining functions in `__init__.py`?

datahub/activity_stream/investment/views.py
alixedi: No, I agree `created_by` can be moved to `select_related`.

datahub/activity_stream/investment/tests/test_views.py
alixedi: My bad, these got let over.

datahub/activity_stream/investment/tests/test_views.py
alixedi: Discussed IRL, the `dit_Event` was an event like something happened IRL (e.g. a meeting) and not an event as in event sourcing 😸

datahub/activity_stream/tests/__init__.py
reupen: My view is that it tends to have the effect of not making things very discoverable. The Hitchhiker's Guide talks about the file more generally (though I wouldn't say it's entirely clear about what it's trying to say...): https://docs.python-guide.org/writing/structure/#packages

alixedi: I don't quite get the argument about discoverability - I can argues that putting stuff in `__init__.py` makes them more discoverable because of one less namespace.

That said, I do not hold strong opinions about this so I will put the function in `utils.py` 👍

datahub/activity_stream/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+
+    According to the docs (See ref), cursor pagination required an ordering field that amongst
+    other things:
+
+        "Should be an unchanging value, such as a timestamp, slug, or other
+        field that is only set once, on creation."
+
+    Ref: https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination
+    """
+
```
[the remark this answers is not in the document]
alixedi: How about `summary = None`? (see next comment)

datahub/activity_stream/pagination.py
```diff
@@ -0,0 +1,49 @@
...
+    def get_paginated_response(self, data):
+        """
+        Overriding this function to re-format the response according to
+        activity stream spec.
+        """
+        return Response(
+            {
+                '@context': 'https://www.w3.org/ns/activitystreams',
+                'summary': self._get_summary(),
```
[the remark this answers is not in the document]
alixedi: Ah! I thought I would raise a `SummaryNotDefined` exception if `summary is None`. What do you think? (see previous comment)

datahub/activity_stream/investment/views.py
```diff
@@ -0,0 +1,35 @@
+from django.db.models import Prefetch
+
+from datahub.activity_stream.investment.serializers import InvestmentActivitySerializer
+from datahub.activity_stream.pagination import ActivityCursorPagination
+from datahub.activity_stream.views import ActivityViewSet
+from datahub.company.models import Contact
+from datahub.investment.project.models import InvestmentProject
+
+
+class InvestmentCursorPagination(ActivityCursorPagination):
```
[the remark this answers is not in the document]
alixedi: `InvestmentProjectCreatedActivitySerializer` is getting too Java.

Do you have any suggestions for a shorter name?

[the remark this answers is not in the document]
alixedi: how about `IProjectCreatedSerializer`?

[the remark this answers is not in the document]
reupen: I did use `IProject` in a few times class names originally, but everyone else thought the 'I' was Java-y and stood for interface :laughing:

datahub/activity_stream/serializers.py
```diff
@@ -0,0 +1,56 @@
+from rest_framework import serializers
+
+
+class ActivitySerializer(serializers.ModelSerializer):
```
reupen: Is this making use of any `ModelSerializer` functionality or can it derive from `serializers.Serializer` instead?

datahub/investment/project/test/factories.py
```diff
@@ -100,7 +100,7 @@ class AssignPMInvestmentProjectFactory(InvestmentProjectFactory):
 
     stage_id = InvestmentProjectStage.assign_pm.value.id
     client_cannot_provide_total_investment = False
-    total_investment = 100
+    total_investment = 100.0
```
reupen: This (and `foreign_equity_investment`) are decimal fields so if you want to be fully correct I think it should use e.g. `Decimal('100.0')`.

datahub/activity_stream/investment/views.py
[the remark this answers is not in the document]
alixedi: Oh dear! 😂

datahub/activity_stream/tests/__init__.py
elcct: I think it is a bit more descriptive if put in a dedicated file. In `utils.py` you know you should expect some utility stuff, whereas in the `__init__.py` who knows. So if you are browsing unknown code base I think it will take more time to discover utility functions in the `__init__.py` than self-explanatory `utils.py`. That's my take on discoverability.

datahub/investment/project/test/factories.py
alixedi: Not sure what fully correct means - this was `int` to begin with resulting in failing tests which are not failing anymore.

reupen: It was wrong before, but it will typically be a Decimal rather than a `float` or an `int`.

datahub/activity_stream/investment/views.py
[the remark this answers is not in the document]
alixedi: It seems we have a [bunch](https://github.com/uktrade/data-hub-leeloo/search?q=IProject&unscoped_q=IProject) of `IProject..`s so I will use that for consistency

changelog/investment-activity-stream.api.rst
```diff
@@ -0,0 +1,6 @@
+It is now possible to get a list of investment projects created in activity-stream `format.
+<https://www.w3.org/TR/activitystreams-core/>`_ 
+
+The URL for this is:
+
+| /v3/activity-stream/investments
```
reupen: The URL path needs updating here.

reupen:
Don't forget to do some squashing when rebasing!

datahub/activity_stream/investment/serializers.py
```diff
@@ -0,0 +1,52 @@
...
+        Serialize the investment project as per Activity Stream spec:
+        https://www.w3.org/TR/activitystreams-core/
+        """
+        project_id = f'dit:DataHubInvestmentProject:{instance.pk}'
+        project = {
+            'id': f'{project_id}:Add',
+            'type': 'Add',
+            'published': instance.created_on,
+            'generator': self._get_generator(),
+            'actor': self._get_adviser(instance.created_by),
```
marcofucci: Unfortunately `instance.created_by` is sometimes None so we need to exclude it from the response in that case :-/

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,214 @@
...
+from datahub.core.test_utils import format_date_or_datetime
+from datahub.investment.project.test.factories import (
+    AssignPMInvestmentProjectFactory,
+    InvestmentProjectFactory,
+    VerifyWinInvestmentProjectFactory,
+)
+
+
+@pytest.mark.django_db
+def test_investment_project_activity(api_client):
```
marcofucci: Can we add `created` or `added` to the name of these tests?

datahub/activity_stream/investment/views.py
```diff
@@ -0,0 +1,34 @@
...
+from datahub.investment.project.models import InvestmentProject
+
+
+class IProjectCreatedPagination(ActivityCursorPagination):
+    """
+    Investment Project created CursorPagination for activity stream.
+    """
+
+    ordering = ('created_on', 'pk')
+    summary = 'Investment Activities'
```
marcofucci: can we add created or added here as well?

datahub/activity_stream/investment/tests/test_views.py
```diff
@@ -0,0 +1,214 @@
...
+            },
+        ],
+    }
+
+
+@pytest.mark.django_db
+def test_investment_project_with_pm_activity(api_client):
+    """
+    Get a list of investment project and test the returned JSON is valid as per:
+    https://www.w3.org/TR/activitystreams-core/
```
marcofucci: Can you explain what changes compared to a newly-created investment project? Otherwise it's hard to spot how it's different from the previous test. For this specific reason, I usually prefer pytest param so that it's obvious what the different scenarios are. However I'm not sure how doable it is in this case so adding something to the docstring works as well.

PR #1765, opened 2019-06-05
Extend email domain verification to allow configurable auth methods

currycoder:
### Description of change

- This changes email domain verification so that we can have a configuration for authentication methods for each DIT adviser email domain.
- It also relaxes the default authentication so that **any** domain is set as valid, as long as it passes a minimum level of `dmarc=bestguesspass spf=pass compauth=pass`.  This is a temporary measure to make the launch of email ingestion as friction-less for advisers as possible - since it is so difficult to find out the exhaustive list of domains (and their authentication methods) that advisers use in practice.  This is explored in depth in this analysis document: https://docs.google.com/document/d/1p7JzIsSX6KF98DsCuFOGtd52y3pyq5PurS9btEkC4LE/edit

The new `DIT_EMAIL_DOMAIN_..` environment variables replace both the `DIT_EMAIL_DOMAINS` and `DIT_EMAIL_DOMAINS_AUTHENTICATION_EXEMPT` environment variables.  The `DIT_EMAIL_DOMAIN_..` variables are collected in to a revamped `DIT_EMAIL_DOMAINS` django setting which presents a data structure of domains and authentication methods.  This introduces some extra logic in to the `common.py` settings file, but this achieves the aim of keeping the necessary details secret while making the environment variable names relatively readable.
Here is an example set of environment variables that we might use: https://docs.google.com/document/d/1n_sJrMlCi-IZyRt5pXIXXgrN-lJmTzgBrF1FNflFOjk/edit

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the 
 | `MAILBOX_MEETINGS_USERNAME` | No | Username of the inbox for ingesting meeting invites via IMAP (likely to be the same as the email for the mailbox) |
 | `MAILBOX_MEETINGS_PASSWORD` | No | Password for the inbox for ingesting meeting invites via IMAP |
 | `MAILBOX_MEETINGS_IMAP_DOMAIN` | No | IMAP domain for the inbox for ingesting meeting invites via IMAP |
-| `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass|spf:pass|dkim:pass |
+| `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass\|spf:pass\|dkim:pass |
 
 
 ## Management commands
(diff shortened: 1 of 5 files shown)


reupen:
@currycoder 
One of the examples contains `compauth=pass` which I'm finding cannot be parsed correctly (as the example contains two equals signs). Are you seeing the same?

README.md
```diff
@@ -303,6 +303,7 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the
 | `MAILBOX_MEETINGS_USERNAME` | No | Username of the inbox for ingesting meeting invites via IMAP (likely to be the same as the email for the mailbox) |
 | `MAILBOX_MEETINGS_PASSWORD` | No | Password for the inbox for ingesting meeting invites via IMAP |
 | `MAILBOX_MEETINGS_IMAP_DOMAIN` | No | IMAP domain for the inbox for ingesting meeting invites via IMAP |
+| `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass|spf:pass|dkim:pass |
```
reupen: The `|` characters in the example need escaping I think (`\|`).

config/settings/common.py
```diff
@@ -546,5 +546,23 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
     },
 }
 
-DIT_EMAIL_DOMAINS = env.list('DIT_EMAIL_DOMAINS', default=[])
-DIT_EMAIL_DOMAINS_AUTHENTICATION_EXEMPT = env.list('DIT_EMAIL_DOMAINS_AUTHENTICATION_EXEMPT', default=[])
+DIT_EMAIL_DOMAINS = {}
+domain_environ_names = [
+    environ_name 
+    for environ_name in os.environ.keys()
```
reupen: It may be better to use `env.ENVIRON` so we aren't mixing `os.environ` and `env`.

currycoder:
> One of the examples contains `compauth=pass` which I'm finding cannot be parsed correctly (as the example contains two equals signs). Are you seeing the same? Should it have been `compauth:pass`?

My bad @reupen - I've sorted this out now.

PR #1772, opened 2019-06-07
Interaction service answers store

elcct:
### Description of change

When an Interaction is being created, depending on selected Service, a user can be asked to answer a list of questions. This PR enables us to store those answers.

This adds new nullable field `service_answers` of `JSONField` type to `Interaction` model.

The data is expected to be in the following format:

```
    {
        "<uuid>": {  # service question id
            "<uuid>": {  # service answer option id
            }
        },
        ...
    }
```

Interactions `GET`, `POST`, `PATCH` endpoints have been modified to support this field. Additional validation to enforce the format above has been added to the Interaction serializer.
Potential errors in the supplied data are being returned in a standard way. If there is a problem with the `service_answers` field itself (e.g. it is missing when it is required), the error response would look like this:

```
{"service_answers": ["This field is required"]}
```

If an error is within the supplied data, then the error will be provided against immediate key. For example when an answer is provided to a question that doesn't belong to the selected service, the error will look like this:

```
 {"<service_question_id>: ['This question does not relate to selected service.']}
```
This way the errors should easily be matched to the corresponding form fields in the frontend.

Service answers have not been added to the collections, because it is not is scope of this PR.

### Checklist


datahub/core/constants.py
```diff
@@ -147,12 +147,85 @@ class UKRegion(Enum):
     )
 
 
+class ServiceQuestion(Enum):
```
reupen: Perhaps we should put the new enums in the interaction app (under tests if they're just for tests)?

elcct: Good point. I have moved those to the tests. Thanks!

reupen:
I've made some updates:

- added some missing validation logic
- removed handling of additional questions as these aren't currently being used (note: the data structure format is still the same, to avoid FE changes)
- updated the tests
- attempted to simplify the validation logic
- made service answers read-only in the admin site

The original branch is backed up here: https://github.com/uktrade/data-hub-leeloo/tree/feature/interaction-service-qa-store-backup

datahub/interaction/test/views/test_service_answers.py
```diff
@@ -0,0 +1,436 @@
...
+        response = self.api_client.patch(
+            url,
+            data=new_value,
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        assert response.json() == expected_response
+
+    @pytest.mark.parametrize('permissions', NON_RESTRICTED_VIEW_PERMISSIONS)
+    @freeze_time('2017-04-18 13:25:30.986208')
+    def test_non_restricted_user_can_get_interaction_with_service_answers(
```
currycoder: Is it worth also testing the inverse - that a restricted user cannot see something they are not allowed to see?

datahub/interaction/validators.py
```diff
@@ -1,7 +1,124 @@
...
+        if expected_questions and not service_answers:
+            raise serializers.ValidationError({
+                'service_answers': cls.required_message,
+            })
+
+    @classmethod
+    def _validate_questions(cls, expected_questions, service_answers):
+        errors = {}
+
+        # Add errors for any unanswered questions
```
currycoder: I think the comments here make a difficult piece of code more digestible - nice one!

datahub/interaction/test/views/test_service_answers.py
elcct: I think we are a bit time restricted, but that definitely would be worth adding.

elcct:
Replied offline

PR #1773, opened 2019-06-07
Email ingestion adviser whitelist

currycoder:
### Description of change
For the email ingestion project, it was agreed with stakeholders to limit data hub to only ingest emails from advisers on a limited whitelist.  This is to satisfy security concerns during the initial roll out of email ingestion.

This change:
- Introduces an `DIT_EMAIL_INGEST_WHITELIST` setting for the adviser whitelist and enforces it in the `was_email_sent_by_dit` validation function.
- Reverts work done previously to flag received emails that do not meet our minimum level of email validation (as this could be quite noisy in conjunction with the whitelist).  This may be re-instated as we remove the whitelist later.
- Fixes an issue where meeting invites that only specified a date could not be successfully ingested.

### Checklist

### DIFF
--- a/datahub/email_ingestion/validation.py
+++ b/datahub/email_ingestion/validation.py
@@ -51,7 +51,7 @@ def was_email_sent_by_dit(message):
     except IndexError:
         return False
 
-    # TODO: Remove this logic once we are past the pilo5 period for email ingestion
+    # TODO: Remove this logic once we are past the pilot period for email ingestion
     if from_email not in settings.DIT_EMAIL_INGEST_WHITELIST:
         return False
 
--- a/datahub/email_ingestion/validation.py
+++ b/datahub/email_ingestion/validation.py
@@ -46,9 +46,13 @@ def was_email_sent_by_dit(message):
     :returns: True if the email passed our minimal level of email authentication checks.
     """
     try:
-        from_email = message.from_[0][1]
+        from_email = message.from_[0][1].strip()
         from_domain = from_email.rsplit('@', maxsplit=1)[1]
     except IndexError:
+        return False
+
+    # TODO: Remove this logic once we are past the pilo5 period for email ingestion
+    if from_email not in settings.DIT_EMAIL_INGEST_WHITELIST:
         return False
 
     try:
--- a/datahub/interaction/email_processors/parsers.py
+++ b/datahub/interaction/email_processors/parsers.py
@@ -94,16 +94,6 @@ def _get_utc_datetime(localised_datetime):
         return localised_datetime
 
 
-class EmailNotSentByDITException(Exception):
-    """
-    Exception for flagging that an email could not be verified as sent by a DIT
-    adviser.
-
-    TODO: Remove this exception and its uses once we are past calendar invite
-    ingestion pilot stage.
-    """
-
-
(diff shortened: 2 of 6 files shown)


datahub/interaction/email_processors/parsers.py
```diff
@@ -88,20 +89,15 @@ def _extract_calendar_string_from_attachments(message):
 
 def _get_utc_datetime(localised_datetime):
     try:
-        return localised_datetime.astimezone(utc)
-    # When the calendar event is just a date
+        localised_datetime.hour
```
alixedi: Please consider changing this to: `isinstance(localised_datetime, datetime.DateTime)`

datahub/email_ingestion/validation.py
```diff
@@ -46,11 +46,15 @@ def was_email_sent_by_dit(message):
     :returns: True if the email passed our minimal level of email authentication checks.
     """
     try:
-        from_email = message.from_[0][1]
+        from_email = message.from_[0][1].strip()
         from_domain = from_email.rsplit('@', maxsplit=1)[1]
     except IndexError:
         return False
 
+    # TODO: Remove this logic once we are past the pilo5 period for email ingestion
```
reupen:
```suggestion
    # TODO: Remove this logic once we are past the pilot period for email ingestion
```

datahub/interaction/email_processors/parsers.py
```diff
@@ -88,20 +89,15 @@ def _extract_calendar_string_from_attachments(message):
...
-class EmailNotSentByDITException(Exception):
-    """
-    Exception for flagging that an email could not be verified as sent by a DIT
-    adviser.
-
-    TODO: Remove this exception and its uses once we are past calendar invite
-    ingestion pilot stage.
-    """
+        # localised_datetime is a Date object, so convert it to a datetime
+        localised_datetime = datetime.datetime(
```
alixedi: I don't understand what's happening here. Are we converting `DateTime` to `Date` b/c we are not passing in the `Time` bits? If so, can we please rename the function?

datahub/interaction/email_processors/parsers.py
currycoder: Agreed, I've fixed this up.

datahub/interaction/email_processors/parsers.py
currycoder: I've reworked this function and given it an explainer docstring - please see if it makes sense.

datahub/interaction/email_processors/parsers.py
```diff
@@ -86,22 +87,19 @@ def _extract_calendar_string_from_attachments(message):
...
+    and transposes it to a UTC datetime.
     """
+    # If calendar_time is a datetime.date, make it a datetime
+    if not isinstance(calendar_time, datetime.datetime):
+        calendar_time = datetime.datetime(
+            day=calendar_time.day,
+            month=calendar_time.month,
+            year=calendar_time.year,
+        )
+    return calendar_time.astimezone(utc)
```
reupen: I think `calendar_time.astimezone(utc)` will end up using the Django time zone as the source time zone if `calendar_time` does not have one. That might be worth a comment (at least).

PR #1776, opened 2019-06-10
Switch from Raven to Sentry SDK

reupen:
### Description of change

This switches from the deprecated Raven Sentry client to the new Sentry SDK.

See https://docs.sentry.io/platforms/python/ (new client documentation) and https://docs.sentry.io/clients/python/ (old client documentation).

### Checklist

### DIFF
--- a/changelog/sentry-sdk.internal.rst
+++ b/changelog/sentry-sdk.internal.rst
@@ -1,0 +1,1 @@
+The deprecated Raven Sentry client was replaced with the Sentry SDK.
--- a/config/settings/common_sentry.py
+++ b/config/settings/common_sentry.py
@@ -1,7 +1,8 @@
-from config.settings.common import *
+import sentry_sdk
+from sentry_sdk.integrations.celery import CeleryIntegration
+from sentry_sdk.integrations.django import DjangoIntegration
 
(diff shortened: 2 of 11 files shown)


changelog/sentry-sdk.internal.rst
```diff
@@ -0,0 +1 @@
+The deprecated Raven Sentry client with replaced with Sentry SDK.
```
marcofucci: Typo?

config/settings/common_sentry.py
```diff
@@ -33,21 +30,15 @@
...
 }
 
-RAVEN_CONFIG = {
-    'dsn': env('DJANGO_SENTRY_DSN'),
-    'include_paths': ['datahub'],
-    'environment': env('SENTRY_ENVIRONMENT'),
-}
+sentry_sdk.init(
+    dsn=env('DJANGO_SENTRY_DSN'),
+    environment=env('SENTRY_ENVIRONMENT'),
```
marcofucci: Shouldn't it get this value automatically? https://docs.sentry.io/error-reporting/configuration/?platform=python#environment

reupen: Yes, it does. Still, this way we'll get an error if it's not configured. (Same for `SENTRY_DSN` – it will pick that up from an environment variable as well.)

PR #1785, opened 2019-06-11
Fix email auth exemption and whitelist logic

currycoder:
### Description of change
This does the following:
- Fixes the logic for making a domain exempt.
- Ensures that we are dealing with fully lower case emails for the `DIT_EMAIL_INGEST_WHITELIST` setting.

### Checklist

### DIFF
--- a/datahub/email_ingestion/validation.py
+++ b/datahub/email_ingestion/validation.py
@@ -52,7 +52,7 @@ def was_email_sent_by_dit(message):
         return False
 
     # TODO: Remove this logic once we are past the pilot period for email ingestion
-    if from_email not in settings.DIT_EMAIL_INGEST_WHITELIST:
+    if from_email.lower() not in settings.DIT_EMAIL_INGEST_WHITELIST:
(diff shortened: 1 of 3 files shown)


datahub/email_ingestion/validation.py
```diff
@@ -52,7 +52,7 @@ def was_email_sent_by_dit(message):
         return False
 
     # TODO: Remove this logic once we are past the pilot period for email ingestion
-    if from_email not in settings.DIT_EMAIL_INGEST_WHITELIST:
+    if from_email.lower() not in settings.DIT_EMAIL_INGEST_WHITELIST:
```
reupen: Can any of the existing tests be updated to test that it's case-insensitive?

rafenden: How about trailing spaces?

currycoder: Sorry Raf, didn't see this.  The email parser we are using will strip the email of whitespace

PR #1787, opened 2019-06-11
Sync related OMIS orders to ES when a company or a contact name changes

marcofucci:
### Description of change

If/when a company name or a contact name changes, the related ES OMIS order documents are now updated automatically.

I noticed this bug when reverting back DnB changes to a company.

### Checklist

### DIFF
--- a/changelog/omis/es-re-sync.bugfix.rst
+++ b/changelog/omis/es-re-sync.bugfix.rst
@@ -1,0 +1,1 @@
+When a company or a contact name changes, related OMIS orders are now synced to ElasticSearch.
--- a/datahub/search/omis/signals.py
+++ b/datahub/search/omis/signals.py
@@ -1,6 +1,7 @@
 from django.db import transaction
 from django.db.models.signals import post_delete, post_save
 
+from datahub.company.models import Company as DBCompany, Contact as DBContact
 from datahub.omis.order.models import (
     Order as DBOrder,
     OrderAssignee as DBOrderAssignee,
@@ -8,7 +9,7 @@ from datahub.omis.order.models import (
 )
 from datahub.search.omis import OrderSearchApp
 from datahub.search.signals import SignalReceiver
-from datahub.search.sync_object import sync_object_async
+from datahub.search.sync_object import sync_object_async, sync_related_objects_async
 
 
 def order_sync_es(instance):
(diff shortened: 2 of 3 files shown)


datahub/search/omis/signals.py
```diff
@@ -23,10 +24,19 @@ def related_order_sync_es(instance):
     order_sync_es(instance.order)
 
 
+def sync_related_interactions_to_es(instance):
```
reupen: Wrong function name I think – shouldn't it be something like `sync_related_orders_to_es`?

changelog/omis/es-re-sync.bugfix.rst
```diff
@@ -0,0 +1 @@
+When a company or a contact name change, related OMIS orders are now synced to ElasticSearch.
```
reupen: Should that be 'changes' rather than 'change' (or 'is changed')?

datahub/search/omis/test/test_signals.py
```diff
@@ -149,3 +149,44 @@ def test_removing_assignees_syncs_order_to_es(setup_es):
...
     assert len(indexed) == 1
+
+
+def test_updating_company_name_updates_orders(setup_es):
+    """Test that when a company name is updated, the company's orders are synced to ES."""
+    order = OrderFactory()
+    new_company_name = 'exogenous'
+    order.company.name = new_company_name
+    order.company.save()
+    setup_es.indices.refresh()
```
currycoder: Just to talk this through, is there any chance that the asynchronous ES sync call could run **after** this retrieves the ES document?

reupen: I was under the impression that it's synchronous – have you seen something that says otherwise?

marcofucci: @currycoder yeah that's celery with `CELERY_TASK_ALWAYS_EAGER` so it's not async when testing.

currycoder: Just that this eventually calls through to `sync_related_objects_async` which schedules a celery task.  I may be missing some context about how the test suite is set up to run tasks in this case.

reupen: Oh, I misunderstood, I thought we were talking about the `setup_es.indices.refresh()` line.

PR #1802, opened 2019-06-17
Add a mechanism to suppress email ingest connect errors until a threshold is reached

currycoder:
### Description of change

There are some occasional (4 in the past week) connection errors flagged on sentry which are raised from the `ingest_emails` task while connecting to IMAP.  This change aims to suppress these errors until we really need to hear about them.  In our case, I've made it so that we only raise the exception after there has been 5 failures within the last 10 tries.  The suppression mechanism uses redis to track the failures within the period of time of the last 10 runs.

I considered making this a general purpose error suppression utility, but I can't see many uses for it yet.

### Checklist

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -337,6 +337,10 @@ if REDIS_BASE_URL:
         }
     }
 
+# Every 30 seconds, by default
+EMAIL_INGESTION_FREQUENCY_SECONDS = env.float('EMAIL_INGESTION_FREQUENCY_SECONDS', 30.0)
+# Our limit before reporting connection failures in a particular window (out of 10 tries)
+EMAIL_INGESTION_CONNECT_FAILURE_THRESHOLD = env.int('EMAIL_INGESTION_CONNECT_FAILURE_THRESHOLD', 5)
 
 if REDIS_BASE_URL:
     REDIS_CELERY_DB = env('REDIS_CELERY_DB', default=1)
@@ -389,7 +393,7 @@ if REDIS_BASE_URL:
     if env.bool('ENABLE_EMAIL_INGESTION', False):
         CELERY_BEAT_SCHEDULE['email_ingestion'] = {
             'task': 'datahub.email_ingestion.tasks.ingest_emails',
-            'schedule': 30.0, # Every 30 seconds
+            'schedule': EMAIL_INGESTION_FREQUENCY_SECONDS,
         }
 
     CELERY_WORKER_LOG_FORMAT = (
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -21,6 +21,15 @@ class EmailRetrievalError(Exception):
     pass
 
 
+class EmailInboxConnectionError(Exception):
+    """
+    Custom exception for when there was an error in connecting to the email
+    inbox.
+    """
+
+    pass
+
+
 class Mailbox:
     """
     The Mailbox class acts as a gateway to an email inbox - accessed via IMAP.
@@ -87,9 +96,15 @@ class Mailbox:
 
         :yields: An active imaplib server connection object.
         """
-        connection = imaplib.IMAP4_SSL(self.imap_domain, self.imap_port)
-        connection.login(self.username, self.password)
-        connection.select()
+        try:
+            connection = imaplib.IMAP4_SSL(self.imap_domain, self.imap_port)
+            connection.login(self.username, self.password)
+            connection.select()
+        except (connection.error, ConnectionResetError) as exc:
+            exc_message = exc.args[0]
+            raise EmailInboxConnectionError(
+                f'IMAP SSL connection failed with message: "{exc_message}"',
+            )
         try:
             yield connection
         finally:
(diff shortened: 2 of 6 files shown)


datahub/email_ingestion/mailbox.py
```diff
@@ -87,9 +96,15 @@ def _connect(self):
...
         :yields: An active imaplib server connection object.
         """
-        connection = imaplib.IMAP4_SSL(self.imap_domain, self.imap_port)
-        connection.login(self.username, self.password)
-        connection.select()
+        try:
+            connection = imaplib.IMAP4_SSL(self.imap_domain, self.imap_port)
+            connection.login(self.username, self.password)
+            connection.select()
+        except (connection.error, ConnectionResetError) as exc:
```
reupen: Doesn't seem like this is covered by tests – it would be good to (if possible).

reupen: `error` is not specific to the instance, right? It would be good to use `IMAP4.error` if it isn't.

datahub/email_ingestion/mailbox.py
```diff
@@ -87,9 +96,15 @@ def _connect(self):
...
+            exc_message = exc.args[0]
+            raise EmailInboxConnectionError(
+                f'IMAP SSL connection failed with message: "{exc_message}"',
```
reupen: Something like `{exc!r}` should print the class name and args (e.g. `ValueError('message')`) – any reason for not going for that?

datahub/email_ingestion/mailbox.py
```diff
@@ -87,9 +96,15 @@ def _connect(self):
...
+            )
```
reupen: Would be good to add `from exc`.

datahub/email_ingestion/mailbox.py
```diff
@@ -21,6 +21,15 @@ class EmailRetrievalError(Exception):
     pass
 
 
+class EmailInboxConnectionError(Exception):
+    """
+    Custom exception for when there was an error in connecting to the email
+    inbox.
+    """
+
+    pass
```
reupen: Small thing – `pass` is not required here.

datahub/email_ingestion/tasks.py
```diff
@@ -1,14 +1,52 @@
...
+        failures = 1
+        failure_window = int(settings.EMAIL_INGESTION_FREQUENCY_SECONDS) * window_size
+        cache.set(email_ingest_failures_key, failures, failure_window)
+
+    if failures >= settings.EMAIL_INGESTION_CONNECT_FAILURE_THRESHOLD:
+        # Delete the failure tracker cache entry - so that we need to hit the threshold
+        # again before raising the next error
+        cache.delete(email_ingest_failures_key)
+        original_exc_message = exc.args[0]
+        raise EmailInboxConnectionError(
```
reupen: I'm not sure about re-using `EmailInboxConnectionError` again here (that's the same exception class as `exc` if I followed the logic correctly). Perhaps we could defined a new exception class?

datahub/email_ingestion/tasks.py
```diff
@@ -1,14 +1,52 @@
...
+            f'Connecting to mailbox {mailbox_identifier} has failed {failures} times out of '
+            f'{window_size} connection attempts. The latest failure was: '
+            f'{original_exc_message}',
+        ) from exc
```
reupen: If I haven't missed it, it would be good to log a warning even if it doesn't hit the threshold (just so there is still some visibility of intermittent problems).

datahub/email_ingestion/tasks.py
```diff
@@ -1,14 +1,52 @@
...
+
+    if failures >= settings.EMAIL_INGESTION_CONNECT_FAILURE_THRESHOLD:
+        # Delete the failure tracker cache entry - so that we need to hit the threshold
+        # again before raising the next error
+        cache.delete(email_ingest_failures_key)
+        original_exc_message = exc.args[0]
+        raise EmailInboxConnectionError(
+            f'Connecting to mailbox {mailbox_identifier} has failed {failures} times out of '
+            f'{window_size} connection attempts. The latest failure was: '
+            f'{original_exc_message}',
```
reupen: As earlier – would be good to go for something like `{exc!r}`.

config/settings/common.py
```diff
@@ -337,6 +337,10 @@ def _build_redis_url(base_url, db_number, **query_args):
         }
     }
 
+# Every 30 seconds, by default
+EMAIL_INGESTION_FREQUENCY_SECONDS = env.float('EMAIL_INGESTION_FREQUENCY_SECONDS', 30.0)
+# Our limit before reporting connection failures in a particular window (out of 10 tries)
+EMAIL_INGESTION_CONNECT_FAILURE_THRESHOLD = env.int('EMAIL_INGESTION_CONNECT_FAILURE_THRESHOLD', 5)
```
reupen: It looks like it would be possible to set this higher than the window (10) which should probably be an error.

One option might be to define the window here and add a bit of simple validation so that an error is raised if the combination doesn't make sense.

reupen:
@currycoder Sentry would seem to have a similar feature built-in:

image

I'm wondering if we should make use of that rather than writing our own implementation.

currycoder:
> @currycoder Sentry would seem to have a similar feature built-in:
> 
> image
> 
> I'm wondering if we should make use of that rather than writing our own implementation.

Welp that renders this redundant.  Thanks Reupen.  As discussed I'm going to put this branch on ice for now and see how sentry's error suppression works out.

PR #1816, opened 2019-06-25
Docs for adding non-nullable fields and incremental PRs

currycoder:
### Description of change

These are a couple of things that I found new/unfamiliar having come from a dev team with different approaches/values.  I'm hoping that this documentation can help onboard other new developers from similar backgrounds.

### Checklist

### DIFF
--- a/docs/How to add a non-nullable field.md
+++ b/docs/How to add a non-nullable field.md
@@ -3,15 +3,19 @@
 ## Overview
 
 When tables are large in PostgreSQL, adding a text field which is not null can be 
-problematic.  Setting a DB-enforced default value will cause Postgres to 
-set the default value for every existing row in the table.  Since our migrations
+problematic. Setting a DB-enforced default value will cause Postgres to set the default value for 
+every existing row in the table.  Since our migrations
 run in a transaction, the table in question will become locked during the 
-migration which could cause the application to grind to a halt.
+migration which could cause the application to grind to a halt. This is problematic 
+if there could be writes happening to the table during deployment.
+Adding a (nullable or non-nullable) column with a default can be problematic for 
+large tables because the ALTER TABLE statement will cause a full table rewrite 
+(for PostgreSQL < 11) during which time the table will be locked (pretty much completely...).
 
 ## A Multi-step Approach
 
 Instead of adding a non-nullable field with a single DB migration, we must take
-a multiple-step approach.  These steps should merged and released separately:
+a multiple-step approach.  These steps should be merged and released separately:
 
 1) Add the new DB field as a nullable field in a django migration.  Ensure that
    any code which can add new entries (e.g. API views) adds them with a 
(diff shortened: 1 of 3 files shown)


docs/How to add a non-nullable field.md
```diff
@@ -0,0 +1,40 @@
...
+When tables are large in PostgreSQL, adding a text field which is not null can be 
+problematic.  Setting a DB-enforced default value will cause Postgres to 
+set the default value for every existing row in the table.  Since our migrations
+run in a transaction, the table in question will become locked during the 
+migration which could cause the application to grind to a halt.
+
+## A Multi-step Approach
+
+Instead of adding a non-nullable field with a single DB migration, we must take
+a multiple-step approach.  These steps should merged and released separately:
```
marcofucci: Should **be** merged and released?

docs/How to add a non-nullable field.md
```diff
@@ -0,0 +1,40 @@
...
+2) Manually run our `datahub.dbmaintenance.tasks.replace_null_with_default` task
+   for the field in question.  e.g. `replace_null_with_default.apply_async(args=("interaction", "status", "complete"))`
+   This function is a recursive celery task - it will set the default value in
+   small batches which ensure that the table is responsive while the default is
+   added.
+   **Note:** This task will need to be run in at least the dev, staging and 
+   production environments.
+3) Set the DB field as not nullable with a sensible default in a django migration.
+   Adding a default value will ensure that any local dev environments are also 
+   picked up.
```
marcofucci: Worth saying that in case of multiple fields, one migration per field must be used because of this bug https://code.djangoproject.com/ticket/25105

docs/Incremental Pull Requests.md
```diff
@@ -0,0 +1,25 @@
...
+cleaned up.
+
+## Why?
+
+Building features incrementally in this way might seem to add overhead for the
+developer building the feature.  Actually, there are a number of benefits:
+
+- Each incremental step for the feature is peer reviewed in great detail; the
+  chunks are much easier for other devs in the team to digest and give meaningful
+  reviews for. This also allows us to better adhere to our goal of 400 additions
```
marcofucci: _400 additions or changes_. It might also be worth linking to the _Code review guidelines_ doc so that people know where it comes from.

docs/How to add a non-nullable field.md
```diff
@@ -0,0 +1,40 @@
+# How to add a non-nullable field to a large table
+
+## Overview
+
+When tables are large in PostgreSQL, adding a text field which is not null can be
```
reupen: I think there are two (related, but) separate concerns to consider here:

- adding a non-nullable column is problematic if there could be writes happening to the table during deployment
- adding a (nullable or non-nullable) column with a default can be problematic for large tables because the ALTER TABLE statement will cause a full table rewrite (for PostgreSQL < 11) during which time the table will be locked (pretty much completely...)

So it would be good to clarify that a bit as even if the ALTER TABLE doesn't take not too long, NOT NULL could still be problematic if there could be writes happening.

PR #1817, opened 2019-06-28
Update interaction services

elcct:
### Description of change

This changes service names so that the front-end can display them in a hierarchy.
Hierarchy is marked by a colon character.

This also disables services that are no longer required.

### Checklist


datahub/metadata/migrations/0031_update_services.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import PurePath
+
+from django.db import migrations
+
+from datahub.core.migration_utils import load_yaml_data_in_migration
+
+
+def load_administrative_areas(apps, schema_editor):
```
reupen: Small thing – the function name here needs updating.

datahub/metadata/migrations/0031_update_services.yaml
```diff
@@ -0,0 +1,176 @@
+- model: metadata.service
```
reupen: I think it would be better to include all the fields (i.e. also `disabled_on` and `contexts`) for each service because otherwise you have to look at the previous migration to find out what they are.

(Also might help with squashing later on.)

datahub/metadata/migrations/0031_update_services.yaml
```diff
@@ -0,0 +1,176 @@
...
+    disabled_on:
+    name: 'Enquiry received : HPO High Potential Opportunity Investment Enquiry via IIGB'
+- model: metadata.service
+  pk: 9984b82b-3499-e211-a939-e4115bead28a
+  fields:
+    name: 'Enquiry or Referral Received : Other Inbound Export Referral'
+- model: metadata.service
+  pk: 5959ba2b-3499-e211-a939-e4115bead28a
+  fields:
+    disabled_on: '2019-07-04T00:00:00.000Z'
```
reupen: If we disable this it will also not be possible to use it in investment project interactions. Is that correct?

elcct: Good call! That would indeed remove them from the investment interactions. I have removed these from the update.

datahub/metadata/migrations/0031_update_services.py
elcct: Updated. Thanks!

datahub/metadata/migrations/0031_update_services.yaml
elcct: Added the `disabled_on` and `contexts` fields. `contexts` are also revised.

reupen:
Are we able to merge this one now @elcct ?

elcct:
@reupen We should be okay, but it will be good to follow it shortly with the other two. I need to check the feature flag on environments as well.

PR #1831, opened 2019-07-05
Improve approach for matching contacts for meeting invite ingestion

currycoder:
### Description of change
It turns out that the data quality for Contacts in production is not the best - we have a lot of instances where multiple Contact models (for different companies) share the same email address.  For the meeting invite ingestion feature, this is problematic as using the contact matching utility would fail to select a suitable Contact for a draft interaction.  

This would often mean that meeting emails were ignored as one suitable matching contact could not be found in Data Hub to attribute the interaction to.

As a quick fix, this PR ensures that in a situation where there are multiple matching contacts for an email address, the contact with the highest number of interactions attributed to it is selected.  This is not a perfect solution by any means - but it strives to ensure that we save a draft interaction in Data Hub when we can.

Once the email notification feature is in, advisers will be directed to the actual interaction that was saved for their meeting invite and they will have an opportunity to move it to the correct contact if the contact selected automatically was not right.

### Things of note

- I've adapted the contact matching utility to allow the use of different strategies - who's identifier can be specified in a kwarg.  This allows us to customise the matching logic used for finding contacts according to different business needs.

### Checklist

### DIFF
--- a/datahub/company/contact_matching.py
+++ b/datahub/company/contact_matching.py
@@ -1,4 +1,4 @@
-from enum import IntEnum
+from enum import auto, Enum, IntEnum
 from typing import Optional, Tuple
 
 from django.db.models import Count
@@ -6,48 +6,13 @@ from django.db.models import Count
 from datahub.company.models import Contact
 
 
-def _match_contact(filter_criteria):
+class MatchStrategyName(Enum):
     """
-    This default matching strategy function will attempt to get a single result
-    for the specified criteria.
-    It will fail with an `unmatched` result if there are no matching contacts.
-    It will fail with a `multiple_matches` result if there are multiple matches
-    for this criteria.
+    Enum of contact match strategy identifiers.
     """
-    contact = None
-    try:
-        contact = Contact.objects.get(**filter_criteria)
-        contact_matching_status = ContactMatchingStatus.matched
-    except Contact.DoesNotExist:
-        contact_matching_status = ContactMatchingStatus.unmatched
-    except Contact.MultipleObjectsReturned:
-        contact_matching_status = ContactMatchingStatus.multiple_matches
 
-    return contact, contact_matching_status
-
-
-def _match_contact_max_interactions(filter_criteria):
-    """
-    This matching strategy function is the same as the default strategy, except
-    that it will prefer to return the contact with the most interactions in the
-    case where there are multiple contacts that match the criteria.
-    """
-    contact = None
-    try:
-        contact = Contact.objects.filter(**filter_criteria)\
-            .annotate(interactions_count=Count('interactions'))\
-            .order_by('-interactions_count')[0]
-        contact_matching_status = ContactMatchingStatus.matched
-    except IndexError:
-        contact_matching_status = ContactMatchingStatus.unmatched
-
-    return contact, contact_matching_status
-
-
-MATCH_STRATEGIES = {
-    'max_interactions': _match_contact_max_interactions,
-    'default': _match_contact,
-}
+    MAX_INTERACTIONS = auto()
+    DEFAULT = auto()
 
 
 class ContactMatchingStatus(IntEnum):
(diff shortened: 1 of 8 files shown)


datahub/company/contact_matching.py
```diff
@@ -1,9 +1,55 @@
...
+
+def _match_contact_max_interactions(filter_criteria):
+    """
+    This matching strategy function is the same as the default strategy, except
+    that it will prefer to return the contact with the most interactions in the
+    case where there are multiple contacts that match the criteria.
+    """
+    contact = None
+    try:
+        contact = Contact.objects.filter(**filter_criteria)\
```
reupen: As PEP 8 says to avoid \ for line continuation, the preferred style would be something like:

```python
        contact = Contact.objects.filter(
            **filter_criteria,
        ).annotate(
            interactions_count=Count('interactions'),
        ).order_by(
            '-interactions_count',
        )[0]
```

datahub/company/contact_matching.py
```diff
@@ -1,9 +1,55 @@
...
+            .annotate(interactions_count=Count('interactions'))\
+            .order_by('-interactions_count')[0]
```
reupen: Would be good to add on `pk` to the ordering make sure it's deterministic (when multiple matched contacts have the same interaction count e.g. 0).

(Could consider adding something date-related in the middle as well.)

datahub/company/contact_matching.py
```diff
@@ -1,9 +1,55 @@
...
+        contact_matching_status = ContactMatchingStatus.matched
+    except IndexError:
+        contact_matching_status = ContactMatchingStatus.unmatched
+
+    return contact, contact_matching_status
+
+
+MATCH_STRATEGIES = {
+    'max_interactions': _match_contact_max_interactions,
```
reupen: Would be good to use an enum for `'max_interactions'` and `'default'` – could use `enum.auto()` for the values (or maybe it would be possible to use the functions).

datahub/company/contact_matching.py
```diff
@@ -1,9 +1,55 @@
 from enum import IntEnum
 from typing import Optional, Tuple
 
+from django.db.models import Count
+
 from datahub.company.models import Contact
 
 
+def _match_contact(filter_criteria):
```
reupen: Would be good to keep the order of objects consistent in this file (previously the public API was at the top, now it's in the middle).

datahub/company/contact_matching.py
```diff
@@ -12,7 +23,10 @@ class ContactMatchingStatus(IntEnum):
     multiple_matches = 3
 
 
-def find_active_contact_by_email_address(email) -> Tuple[Optional[Contact], ContactMatchingStatus]:
+def find_active_contact_by_email_address(
+    email,
+    match_strategy=MatchStrategyName.DEFAULT,
```
alixedi: 1. Why do we need multiple matching strategies in the codebase? 

2. Is there a mechanism for switching between matching strategies? If not, can we just have one?

3. If yes to both Qs above, I would prefer implementing matching strategies as functions and passing in the function here instead of the `enum`.

datahub/company/contact_matching.py
```diff
@@ -12,7 +23,10 @@ class ContactMatchingStatus(IntEnum):
...
+) -> Tuple[Optional[Contact], ContactMatchingStatus]:
```
alixedi: The docs say that this function attempts to find a contact. Why are we returning the `matching_status` here? Can the matching status be anything other than `matched`?

datahub/company/contact_matching.py
currycoder: > Why do we need multiple matching strategies in the codebase?
Different components of the code have different needs due to business logic: 
- one consumer of this utility is for matching users from an automated CSV import and this has strict correctness requirements.
- In the case catered for here, we are ingesting user-submitted data to build a draft interaction.  The final details of the interaction (including contact/company) can be adjusted at the time when the interaction is completed so in this case getting *something* saved is preferable to having complete confidence in a correct match.

> Is there a mechanism for switching between matching strategies? If not, can we just have one?
Yes - it's this line that has been commented on.  The caller can choose which strategy to use.

> If yes to both Qs above, I would prefer implementing matching strategies as functions and passing in the function here instead of the enum.

I think this is a good idea.  How about if I adjust the enum so that it's values are the matching functions instead of using the redundant `MATCH_STRATEGIES` dictionary as an intermediate?  We then still have the well defined contract that the enum offers without the level of indirection.

datahub/company/contact_matching.py
currycoder: Good question.  I'll make the docs clearer.

datahub/company/contact_matching.py
currycoder: I've done this btw - let me know what you think.

datahub/company/contact_matching.py
currycoder: So, when I adjusted the proposed enum to use the strategy functions as values directly, it necessitated a re-ordering of private/public objects.  Private are first, public are last now.

PR #1832, opened 2019-07-05
Enable the Django Rest Framework built-in documentation feature

reupen:
### Description of change

This enables the [Django Rest Framework built-in documentation feature](https://www.django-rest-framework.org/topics/documenting-your-api/) at the URL path `/docs`.

This is currently only enabled if the `ENABLE_API_DOCUMENTATION` environment variable is set to `True` as the documentation is not fully functional as yet.

Currently, a user must log into Django admin prior to accessing the documentation.

Some minor changes had to be made to a few views in order to get the view introspection working. (In the case of the D&B views, they were not really generic views and so have been changed to use APIView instead of GenericAPIView.)

The next version of DRF moves more towards OpenAPI-based schema generation and these changes will help us in switching to that when we upgrade.

I'm planning to fix a few things in further PRs, and then will probably enable the docs by default for (at least) local and the dev environments. Will also add some docs to make it clear what needs to be done for new endpoints.

### Screenshots

These show the docs functionality in DRF 3.9 (this may change when we upgrade to 3.10).

image

image

### Checklist


datahub/omis/invoice/views.py
```diff
@@ -21,7 +21,7 @@ def get_object(self):
 
         :raises Http404: if the invoice doesn't exist
         """
-        invoice = self.get_order().invoice
+        invoice = self.get_order_or_404().invoice
         if not invoice:
             raise Http404('The specified invoice does not exist.')
```
currycoder: Do we still need this if we are using `get_order_or_404`?

reupen: Yes, the order may not have an invoice as well.

reupen:
> Also a question - with the OpenAPI changes proposed, will the swagger definition go as far as explaining field validation criteria..?

I expect not, but it should hopefully still be possible to add a description (with general comments) for each view.

You can have a look at it looks like in DRF 3.10 via the docs on their master branch:

- https://github.com/encode/django-rest-framework/blob/master/docs/topics/documenting-your-api.md
- https://github.com/encode/django-rest-framework/blob/master/docs/api-guide/schemas.md

They seem to be in the process of releasing DRF 3.10.0 right now so it shouldn't be far off at all (but we will need an update to django-filter as well for https://github.com/carltongibson/django-filter/pull/1086).

datahub/dnb_match/views.py
```diff
@@ -15,18 +16,22 @@
...
-class BaseMatchingInformationAPIView(GenericAPIView):
+class BaseMatchingInformationAPIView(APIView):
     """Base matching information APIView."""
 
     required_scopes = (Scope.internal_front_end,)
 
     queryset = Company.objects.select_related('dnbmatchingresult')
 
-    lookup_url_kwarg = 'company_pk'
+    def _get_company(self):
```
alixedi: Can we please pass down `self.kwargs['company_pk']` as an argument to this method? The same for `_get_matching_information`. I don't think utility functions should concern themselves with the existence of `self.kwargs['company_pk']`.

reupen: I'll change `_get_matching_information()` to take a company as that one does seem a bit weird.

However, I think `_get_company()` fits in with how `get_object()` works in generic views and it would be a bit repetitive if we started writing `self.kwargs['company_pk']` everywhere.

PR #1811, opened 2019-06-20
Spike: notify advisers

currycoder:
### Description of change

**This PR will not be merged**

This PR adds a new notification app for accessing the GOVUK notification service.  This PR is designed to be reviewed in tandem with a short spec document: https://docs.google.com/document/d/1Zfyu3esSsyP-Hpvln_PYEj5pW_1k4eUro7NarhWTxDw/edit?usp=sharing
This doc explains the business rationale, walks through the design approach and differences with OMIS notifications.  It also explores possible extensions that we could make in the future.

Broadly, a lot of the approaches here are lifted from OMIS notifications as they stand - with the idea that the notification-specific logic can be extricated from that django app to provide some reusable tools.

### Checklist

* [ ] Has a new newsfragment been created? Check [changelog/README.rst](https://github.com/uktrade/data-hub-leeloo/blob/master/changelog/README.rst) for instructions
* [ ] Have any relevant search models been updated?
* [ ] Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?
* [ ] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?
* [ ] Has the admin site been updated (for new models, fields etc.)?
* [ ] Has the README been updated (if needed)?

PR #1843, opened 2019-07-08
Add initial notification django app

currycoder:
### Description of change

Following on from the spike (https://github.com/uktrade/data-hub-leeloo/pull/1811), this PR adds the initial `notification` django app that will enable us to notify advisers and contacts from within Data Hub.  

There will be further work in subsequent PRs to use this framework for notifying advisers about receipt/bounce of their email invites and also work to replace the notification logic within the omis app with this generalised framework.

### Checklist


datahub/notification/__init__.py
```diff
@@ -0,0 +1,3 @@
+from datahub.notification import core
+
+client = core.client
```
reupen: Would be good to set `__all__` – something like this should work:

```python
from datahub.notification import client

__all__ = ('client',)
```

datahub/notification/core.py
```diff
@@ -0,0 +1,49 @@
+import warnings
+from unittest import mock
+
+from django.conf import settings
+from notifications_python_client.notifications import NotificationsAPIClient
+
+
+class NotificationClient:
+    """
+    Class for accessing underlying GOVUK notification service.
```
elcct: I think `Class` is redundant in this comment.

datahub/notification/test/test_core.py
```diff
@@ -0,0 +1,28 @@
...
+)
+def test_send_email_notification(context):
+    """
+    Test that NotificationClient.send_email_notification method
+    works calls through
+    to the underlying notify library as expected.
+    """
+    client.send_email_notification('john.smith@example.net', 'foobar', context)
+    expected_context = context
+    if not context:
```
elcct: That could be something like `expected_context = context or {}`

datahub/notification/tasks.py
```diff
@@ -0,0 +1,17 @@
+from celery import shared_task
+
+from datahub.notification.core import client
+
+
+@shared_task(acks_late=True, priority=9)
```
reupen: Notify has gone down before, so it might be useful to make use of the retry logic that Celery has e.g. pass `max_retries=5, autoretry_for=(Exception,), retry_backoff=1` (will retry after approx 60, 2\*60, 4\*60, 8\*60 and 16\*60 seconds).

currycoder: Thanks Reupen - nice suggestion.

datahub/notification/tasks.py
```diff
@@ -0,0 +1,23 @@
...
+
+from datahub.notification.core import client
+
+
+@shared_task(
+    acks_late=True,
+    priority=9,
+    max_retries=5,
+    autoretry_for=(Exception,),
+    retry_backoff=1,
```
reupen: I think 1 is a bit low as it's an external service so would consider using a larger value.

datahub/notification/notify.py
```diff
@@ -0,0 +1,18 @@
+from datahub.notification.tasks import send_email_notification
+
+
+def notify_adviser_by_email(adviser, template_identifier, context):
+    """
+    Notify an adviser by email, using a GOVUK notify template and some template
+    context.
+    """
+    email_address = adviser.get_current_email()
+    send_email_notification.apply_async(args=(email_address, template_identifier, context))
```
reupen: Would use `kwargs` for `context` (here and below) since it's defined as a keyword argument.

config/settings/common.py
```diff
@@ -568,3 +568,5 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
             for domain, auth in domain_details.items()
         }
     )
+
+DATAHUB_NOTIFICATION_API_KEY = env('DATAHUB_NOTIFICATION_API_KEY', default=None)
```
reupen: Do we need to add `DATAHUB_NOTIFICATION_API_KEY` to the list of environment variables in the README?

PR #1853, opened 2019-07-11
Add order to services

elcct:
### Description of change

This is branched off `feature/interaction-service-hierarchy`.

It adds `order` column to metadata service so that we can specify the order in which the services are being returned to the client.

Unfortunately this resulted in creating another migration to set the order for all current services. Hopefully we will be able to squash this at some point.

### Checklist

### DIFF
--- a/changelog/metadata-service.db.rst
+++ b/changelog/metadata-service.db.rst
@@ -1,0 +1,1 @@
+An ``order double precision`` column was added to the ``metadata_service`` table to store the order of services.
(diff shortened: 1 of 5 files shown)


reupen:
`requires_service_answers_flow_feature_flag` seems to be set a bit inconsistently for the making introduction services – I suppose it shouldn't matter as we should have the feature flag turned on when we deploy this, but it might be worth reviewing.

elcct:
Great find about the feature flag. I have updated them now. Thanks!

PR #1855, opened 2019-07-12
Add receive/bounce notification

alixedi:
### Description of change

This PR brings in notification emails for failure and success of inventing the calendar invite emails. Much of the work here was done by @currycoder in the #1811 spike.

### Checklist


datahub/interaction/email_processors/notify.py
```diff
@@ -0,0 +1,63 @@
...
+logger = get_task_logger(__name__)
+
+
+class Templates(Enum):
+    """
+    GOV.UK notifications template ids.
+    """
+
+    meeting_ingest_failure = 'fc2c07d5-9f3b-4647-853e-68c324565577'
+    meeting_ingest_success = '47418011-3f53-4b39-860a-30969b29781b'
```
alixedi: This one is from [here](https://www.notifications.service.gov.uk/services/e7a0ce16-6352-4b5e-b5a1-8b9221296d2b/templates/47418011-3f53-4b39-860a-30969b29781b).

datahub/interaction/email_processors/notify.py
```diff
@@ -0,0 +1,63 @@
...
+
+logger = get_task_logger(__name__)
+
+
+class Templates(Enum):
+    """
+    GOV.UK notifications template ids.
+    """
+
+    meeting_ingest_failure = 'fc2c07d5-9f3b-4647-853e-68c324565577'
```
alixedi: This one lives [here](https://www.notifications.service.gov.uk/services/e7a0ce16-6352-4b5e-b5a1-8b9221296d2b/templates/fc2c07d5-9f3b-4647-853e-68c324565577).

I fiddled with the idea of putting these in an environment variable, a config etc. but stuck with this because the templates are editable and this UUID is unlikely to change.

datahub/interaction/email_processors/notify.py
```diff
@@ -0,0 +1,63 @@
...
+
+from datahub.feature_flag.utils import is_feature_flag_active
+from datahub.interaction import INTERACTION_EMAIL_NOTIFICATION_FEATURE_FLAG_NAME
+from datahub.notification.notify import notify_adviser_by_email
+
+
+logger = get_task_logger(__name__)
+
+
+class Templates(Enum):
```
reupen: Would suggest a singular noun here – seems to be the [usual convention](https://docs.python.org/3.7/library/enum.html#creating-an-enum), and also in line with other enums in the code base.

datahub/interaction/email_processors/processors.py
```diff
@@ -59,6 +67,17 @@ class CalendarInteractionEmailProcessor(EmailProcessor):
     instance for it if the information is valid.
     """
 
+    def _notify_meeting_ingest_failure(self, message, errors):
+        try:
+            sender_email = message.from_[0][1]
+        except IndexError:
+            return
```
reupen: Do we need to be notified if this happens?

datahub/interaction/email_processors/processors.py
```diff
@@ -59,6 +67,17 @@ class CalendarInteractionEmailProcessor(EmailProcessor):
...
+        sender_adviser = get_best_match_adviser_by_email(sender_email)
+        if not sender_adviser:
```
reupen: Why would this happen? It would be good to add a comment to explain. Also do we need to be notified if it happens?

datahub/interaction/email_processors/processors.py
```diff
@@ -121,6 +140,8 @@ def process_email(self, message):
         try:
             interaction_data = email_parser.extract_interaction_data_from_email()
         except ValidationError as exc:
+            error_message = exc.args[0]
+            self._notify_meeting_ingest_failure(message, [error_message])
```
reupen: Since it's expecting a list already it may be better to pass through `exc.messages`.

datahub/interaction/email_processors/utils.py
```diff
@@ -0,0 +1,33 @@
...
+            return Advisor.objects.filter(**criteria).earliest('date_joined')
+        except Advisor.DoesNotExist:
+            continue
+    return None
+
+
+def get_all_recipients(message):
+    """
+    Get all of the recipient emails from a MailParser message object.
+    :returns: a set of all recipient emails
```
reupen: The docstring needs updating here as it's returning a list (rather than a set) if I'm not mistaken.

datahub/interaction/email_processors/utils.py
```diff
@@ -0,0 +1,33 @@
+from datahub.company.models import Advisor
+
+
+def get_best_match_adviser_by_email(email):
```
reupen: Do we need to keep the equivalents to these in `parsers.py`?

datahub/interaction/email_processors/processors.py
alixedi: I think the user cannot be notified in cases where we cannot find the sender's (dit-adviser) email. But I agree, we should at least log this.

datahub/interaction/email_processors/processors.py
alixedi: Same as above.

datahub/interaction/email_processors/utils.py
alixedi: Well spotted! I reckon we remove the private functions in `parsers.py`. Does that sound sensible @currycoder?

datahub/interaction/email_processors/processors.py
elcct: Let's log these. We shouldn't typically be silently failing.

datahub/interaction/email_processors/notify.py
```diff
@@ -0,0 +1,63 @@
+from enum import Enum
+
+from celery.utils.log import get_task_logger
+
+from datahub.feature_flag.utils import is_feature_flag_active
+from datahub.interaction import INTERACTION_EMAIL_NOTIFICATION_FEATURE_FLAG_NAME
+from datahub.notification.notify import notify_adviser_by_email
+
+
+logger = get_task_logger(__name__)
```
currycoder: Should this strictly be a task logger?  My understanding is that this will call a function which in turn will trigger a celery task to be queued.

datahub/interaction/email_processors/processors.py
currycoder: Yeah, if the sender is not somebody that Data Hub knows about, we shouldn't send anything to them.  They may have an email domain that makes them look like a valid DIT person, but there's a security question here that we decided to err on the side of caution for.

datahub/interaction/email_processors/utils.py
currycoder: Agreed!

currycoder: This makes me think, there were a number of changes made to `parsers.py` in the spike in order to improve error messages ready for consumption by users, see: https://github.com/uktrade/data-hub-leeloo/pull/1811/files#diff-b6ce04a93e95085dfa6ed7aded5fcea3

Will these follow in a subsequent PR or were they just missed..?

alixedi:
I have resolved the comments but Michael has added a few things to the templates so will update the code to reflect that in this PR. Shouldn't be anything major.

config/settings/common.py
```diff
@@ -575,3 +575,4 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
     )
 
 DATAHUB_NOTIFICATION_API_KEY = env('DATAHUB_NOTIFICATION_API_KEY', default=None)
+DATAHUB_SUPPORT_EMAIL_ADDRESS = 'datahubsupport@uktrade.zendesk.com'
```
reupen: Can we use an environment variable for this? It doesn't seem like something that should be hard-coded. (Also, there is a newline missing at the end of the file here.)

datahub/interaction/test/email_processors/test_processors.py
```diff
@@ -192,14 +193,19 @@ def test_process_email_successful(
         mock_notify_adviser_by_email.assert_called_once_with(
             interaction.dit_adviser,
             Template.meeting_ingest_success.value,
-            context={'interaction_url': interaction.get_absolute_url()},
+            context={
+                'interaction_url': interaction.get_absolute_url(),
+                'recipients': 'bill.adama@example.net, saul.tigh@example.net',
+                'support_team_email': 'datahubsupport@uktrade.zendesk.com',
```
reupen: Would be good to use `settings.DATAHUB_SUPPORT_EMAIL_ADDRESS` here (and in the tests below).

PR #1856, opened 2019-07-12
Add company list app and endpoint to add company list item

reupen:
### Description of change

This adds some initial bits of functionality that will allow users to build up their own personal list of companies.

This PR:

- adds a new `company_list` Django app under `user`
- adds a new `CompanyListItem` model
- adds an endpoint (``PUT /v4/user/company-list/<company ID>``) to add a company to the authenticated user's list

I've put things in the places I thought they best fitted but everything's open for discussion.

I haven't used the DRF generic views for this endpoint as the desired behaviour is somewhat different in this case.

I'd recommend reviewing the commits individually.

Work in future PRs will include:

- changing the merge company tool to update companies on users' lists
- not allowing archived companies to be added to a list
- an endpoint to remove companies from the authenticated user's list
- an endpoint to check if a company is on the authenticated user's list
- an endpoint to list all companies on the authenticated user's list

### Checklist


changelog/adviser/company-list-item.db.rst
```diff
@@ -0,0 +1,11 @@
+A ``company_list_companylistitem`` table was created with the following columns:
+
+- ``"id" bigserial NOT NULL PRIMARY KEY``
+- ``"adviser_id" uuid NOT NULL``
+- ``"company_id" uuid NOT NULL``
```
alixedi: Can we denormalise this and store the list of companies as a `jsonb`?

datahub/cleanup/management/commands/delete_orphans.py
```diff
@@ -32,6 +32,7 @@ class Command(BaseCleanupCommand):
             ),
             # We want to delete the relations below along with any orphaned companies
             excluded_relations=(
+                Company._meta.get_field('companylistitem'),
```
alixedi: I am very confused by the comment that reads "We want to delete..." and the name of the property `excluded_relations`. If we are deleting these, surely the property should be called `included_relations` b/c we are including them in the cleanup operation.

reupen: Excluded here means it does not do any checks on those relations. If we need to make it clearer, we can, but it's not related to this PR.

changelog/adviser/company-list-item.db.rst
reupen: Yes we could, but I expect that will make things more difficult in terms of working with the framework we are using. It would also be a list of objects as we aren't just storing the company IDs.

fixtures/metadata/teams.yaml
```diff
@@ -17,6 +17,18 @@
     - - view_company
       - company
       - company
+    - - view_companylistitem
```
currycoder: On this note, is it worth adding a test fixture which contains some lists for some advisers?

datahub/user/company_list/tests/test_views.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        company = CompanyFactory()
+        CompanyListItem.objects.create(adviser=self.user, company=company)
+
+        url = reverse('api-v4:company-list:replace-item', kwargs={'company_pk': company.pk})
+        response = self.api_client.put(url)
+
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert response.content == b''
+        assert CompanyListItem.objects.filter(adviser=self.user, company=company).exists()
```
currycoder: Might also be worth checking that the item is the same item as it was before the request - i.e. it was updated and not deleted/recreated?

reupen: Have updated the test :+1:

fixtures/metadata/teams.yaml
reupen: I think the difficulty is that you won't be able to access it unless you authenticate as the same adviser in the fixture (which we aren't set up to do at the moment...)

So, probably not worth it for now but we could revisit later.

currycoder: Fair enough.  I do think this would be a good idea though - to have the option to impersonate a real adviser and see some representative things; especially with the personalisation roadmap.

currycoder: In a later PR of course.

PR #1857, opened 2019-07-12
Enable the use of TAP fields for interaction.

elcct:
### Description of change

This change removes validation of TAP fields so that they can be accepted when creating an interaction (before the change, these fields were only allowed for service delivery).

This is branched off `feature/service-order`.

### Checklist


datahub/interaction/test/views/test_interaction.py
```diff
@@ -342,8 +342,6 @@ def test_add(self, extra_data, permissions):
                     'service_delivery_status': [
```
reupen: It would be good to move 

```
                    'grant_amount_offered': '1111.11',
                    'net_company_receipt': '8888.11',
```

up above so that they aren't under the `# fields not allowed` comment.

PR #1863, opened 2019-07-15
Add DNB search API endpoint (proxying dnb-service)

currycoder:
### Description of change

This change adds an endpoint `/v4/dnb/company-search` which proxies through to the search endpoint on dnb-service (https://github.com/uktrade/dnb-service/).  The general approach has been copied from the approach in the `activity_feed` app.

### Points of note
- This is currently behind a feature flag, as it is quite early days for the feature (and proxied service).
- Some code in the activity feed endpoint to enforce a request content type has been broken out in to a separate utility to keep things DRY.

### Checklist


datahub/core/api_client.py
```diff
@@ -42,6 +42,26 @@ def __call__(self, request):
...
+    """
+    Token authentication class.
+    """
+
+    def __init__(self, token, token_keyword=None):
+        """
+        Initialise the class with the token.
+        """
+        self.token = token
+        self.token_keyword = token_keyword or 'Token'
```
reupen: Perhaps we could set 'Token' as the default value for the `token_keyword` keyword argument instead?

datahub/dnb_api/test/test_views.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_post(self, requests_mock, request_data, response_status_code, response_content):
+        """Test for GET proxy."""
+        requests_mock.post(
+            settings.DNB_SERVICE_BASE_URL + 'companies/search/',
+            status_code=response_status_code,
+            content=response_content,
+        )
+
+        url = reverse('api-v4:dnb-api:company-search')
+        # api_client.get transforms data into querystring hence generic is used here
```
reupen: Is this comment right? (It's using a POST here.)

datahub/dnb_api/test/test_views.py
```diff
@@ -0,0 +1,52 @@
...
+            ),
+            (
+                b'{"arg": "value"}',
+                500,
+                b'{"error":"msg"}',
+            ),
+        ),
+    )
+    def test_post(self, requests_mock, request_data, response_status_code, response_content):
+        """Test for GET proxy."""
```
reupen: Is this docstring right (GET vs POST)?

datahub/dnb_api/test/test_views.py
```diff
@@ -0,0 +1,52 @@
...
+
+class TestDNBCompanySearchAPI(APITestMixin):
+    """
+    DNB Company Search view test case.
+    """
+
+    @pytest.mark.parametrize(
+        'request_data,response_status_code,response_content',
+        (
+            (
```
reupen: It would be good to use `pytest.param` and set the `id` for each one to make it clear what each case is testing.

datahub/dnb_api/views.py
```diff
@@ -0,0 +1,57 @@
...
+        upstream_response = self._get_upstream_response(request)
+        return HttpResponse(
+            upstream_response.text,
+            status=upstream_response.status_code,
+            content_type=upstream_response.headers.get('content-type'),
+        )
+
+    def _get_upstream_response(self, request):
+
+        search_endpoint = settings.DNB_SERVICE_BASE_URL + 'companies/search/'
```
reupen: Something like:

```
from urllib.parse import urljoin

urljoin(f'{settings.DNB_SERVICE_BASE_URL}/', 'companies/search/')
```

 should be more robust I think (it shouldn't care about whether `DNB_SERVICE_BASE_URL` has a trailing slash).

datahub/core/view_utils.py
```diff
@@ -0,0 +1,27 @@
...
+def enforce_request_content_type(content_type):
+    """
+    Decorator to enforce request content types to be a certain value.  Returns
+    a 406 HttpResponse if the value is not allowed.
+
+    This should be used on rest framework view methods which have `request` as
+    the first argument
+    """
+    def _enforce_request_content_type(f):
+        def wrapper(self, *args, **kwargs):
```
reupen: We might be able to use `django.utils.decorators.method_decorator` (where this is used) rather than adding `self` here and restricting this to methods.

datahub/core/view_utils.py
```diff
@@ -0,0 +1,27 @@
...
+            request = args[0]
```
reupen: Should we add `request` to the signature?

datahub/dnb_api/views.py
```diff
@@ -19,6 +22,10 @@ class DNBCompanySearchView(APIView):
     @enforce_request_content_type('application/json')
     def post(self, request):
         """Proxy for POST requests."""
+        if not is_feature_flag_active(FEATURE_FLAG_DNB_COMPANY_SEARCH):
```
reupen: Does the `feature_flagged_view` decorator work (with `method_decorator`)?

README.md
```diff
@@ -306,6 +306,8 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the
 | `MAILBOX_MEETINGS_IMAP_DOMAIN` | No | IMAP domain for the inbox for ingesting meeting invites via IMAP |
 | `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass\|spf:pass\|dkim:pass |
 | `DATAHUB_NOTIFICATION_API_KEY` | No | The GOVUK notify API key to use for the `datahub.notification` django app. |
+| `DNB_SERVICE_BASE_URL` | No | The base URL of the DNB service. |
```
reupen: The existing environment variables are sorted alphabetically (save for the ones at the end) – would be good to keep it consistent.

datahub/dnb_api/test/test_views.py
```diff
@@ -0,0 +1,94 @@
...
+            url,
+            data=request_data,
+            content_type='application/json',
+        )
+
+        assert response.status_code == response_status_code
+        assert response.content == response_content
+        assert requests_mock.last_request.body == request_data
+
+    @pytest.mark.parametrize(
```
reupen: Would be good to add a test that checks that a 401 is returned if the user is unauthenticated (you should be able to use the `api_client` fixture to make an unauthenticated request).

config/api_urls.py
```diff
@@ -59,6 +60,7 @@
 v4_urls = [
     path('', include((ch_company_urls.urls, 'ch-company'), namespace='ch-company')),
     path('', include((company_urls.urls, 'company'), namespace='company')),
+    path('dnb/', include((dnb_api_urls, 'dnb_api'), namespace='dnb-api')),
```
alixedi: Seeing this next to `dnb-match/`, it seems like `dab-search/` would be a better name.

datahub/activity_feed/views.py
```diff
@@ -29,18 +29,10 @@ class ActivityFeedView(APIView):
         'order.view_order',
     )
 
+    @enforce_request_content_type('application/json')
```
alixedi: Very nice 👍

datahub/core/api_client.py
alixedi: ✅ The: `var=None` followed by `self.var = var or "default"` pattern is only [necessary](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) when the argument is a mutable type. The following behaves like it should:

```
In [3]: def ali(a, x=''): 
   ...:     return x+a 
   ...:                                                                         

In [4]: ali('ali')                                                              
Out[4]: 'ali'

In [5]: ali('zaidi')                                                            
Out[5]: 'zaidi'
```

datahub/core/test/test_api_client.py
```diff
@@ -53,6 +53,30 @@ def test_response_not_verified_if_verification_disabled(self):
...
+            ),
+            (
+                ['foobarbaz', 'Key'], 'Key foobarbaz',
+            ),
+        ),
+    )
+    def test_requests_with_no_body_are_signed(self, token_auth_args, expected_authorization):
+        """Tests that requests without a body are signed."""
+        auth = TokenAuth(*token_auth_args)
+        request = Mock(method='GET', url='http://test.com/test', body=None, headers={})
```
alixedi: You prolly only need to specify `headers={}` in the Mock object.

datahub/core/api_client.py
currycoder: Thanks, I had a bit of a foggy brain when I was writing this kwarg.

config/api_urls.py
currycoder: I initially called it `dnb-search`! But then I got thinking that we were likely to add an endpoint for getting company detail in to this app, and possibly some logic which worked with Data Hub companies (for saving or hydrating searches etc), so I felt like it was best to keep it generic for now.

Similarly I almost added it to the `company` app, but that app is already feeling fairly bloated - so here we are.  If you have any other ideas I'd be keen to hear them.

datahub/dnb_api/test/test_views.py
currycoder: I spotted this and fixed it in a subsequent commit. :+1:

datahub/dnb_api/views.py
currycoder: Yes it does!

datahub/dnb_api/test/test_views.py
```diff
@@ -0,0 +1,140 @@
...
+            content_type='application/json',
+        )
+
+        assert response.status_code == response_status_code
+        assert response.content == response_content
+        assert requests_mock.last_request.body == request_data
+
+    def test_post_no_feature_flag(self, requests_mock):
+        """
+        Test for POST proxy.
```
reupen: This docstring could be updated to mention that it's without the feature flag.

datahub/dnb_api/views.py
```diff
@@ -0,0 +1,53 @@
...
+        upstream_response = self._get_upstream_response(request)
+        return HttpResponse(
+            upstream_response.text,
+            status=upstream_response.status_code,
+            content_type=upstream_response.headers.get('content-type'),
+        )
+
+    def _get_upstream_response(self, request):
+
+        search_endpoint = urljoin(settings.DNB_SERVICE_BASE_URL, 'companies/search/')
```
reupen: OK, but I don't think this will protect against a missing trailing slash in `settings.DNB_SERVICE_BASE_URL` unless one is explicitly added. Another option would be to write a small function that does the right thing in either case.

currycoder: That is a subtle difference in the way that I had assumed `urljoin` works - thank you.

PR #1872, opened 2019-07-17
Update Django Rest Framework and django-filter

reupen:
### Description of change

This updates Django Rest Framework to version 3.10.1 and django-filter to version 2.2.0.

Release notes:

- DRF 3.10.0: https://www.django-rest-framework.org/community/3.10-announcement/
- django-filter 2.2.0: https://github.com/carltongibson/django-filter/blob/master/CHANGES.rst#version-22-2019-7-16

This keeps using the Core API schema features of DRF (rather than switching to OpenAPI schema generation). Switching to OpenAPI will be tackled separately.

There [was a slight change to the order DRF saves fields during updates](https://github.com/encode/django-rest-framework/pull/6752). Previously, it saved to-one and to-many fields in a somewhat arbitrary order. Now, it consistently saves to-many fields last. 

That change revealed a problem with the investment project GVA updates. One of the investment project GVA tests was failing after the DRF update as there was an assumption that the business activities to-many field would be saved before the model object.

This has been fixed by adding an additional signal receiver that specifically monitors changes to the `business_activities` many-to-many field.

A workaround for encode/django-rest-framework#6509 has also been removed as that bug has now been fixed.

### Checklist

### DIFF
--- a/datahub/investment/project/signals.py
+++ b/datahub/investment/project/signals.py
@@ -1,4 +1,4 @@
-from django.db.models.signals import post_save, pre_save
+from django.db.models.signals import m2m_changed, post_save, pre_save
 from django.dispatch import receiver
 
 from datahub.investment.project.gva_utils import set_gross_value_added_for_investment_project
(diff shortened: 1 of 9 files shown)


datahub/investment/project/signals.py
```diff
@@ -42,10 +42,30 @@ def set_gross_value_added_for_investment_project_pre_save(sender, instance, **kw
...
     """
     set_gross_value_added_for_investment_project(instance)
 
 
+@receiver(
+    m2m_changed,
+    sender=InvestmentProject.business_activities.through,
+    dispatch_uid='update_gross_value_added_on_project_business_activities_pre_save',
+)
+def update_gross_value_added_on_project_business_activities_pre_save(
```
elcct: I think the function name needs to be changed as it is no longer `pre_save` ?

reupen: Good spot! I've updated it.

PR #1874, opened 2019-07-17
Add endpoint to check if a company is on a user's personal list

reupen:
### Description of change

This adds an endpoint at `GET /v4/user/company-list/<company ID>` which checks if a specific company is on the authenticated user's personal list of companies.

(The endpoint should also support `HEAD` although we've seen some weirdness with it that's not in our control. Also note that `HEAD` is omitted from the docs generated by DRF.)

### Checklist

### DIFF
--- a/datahub/user/company_list/views.py
+++ b/datahub/user/company_list/views.py
@@ -1,3 +1,4 @@
+from django.http import Http404
 from django.shortcuts import get_object_or_404
 from oauth2_provider.contrib.rest_framework import IsAuthenticatedOrTokenHasScope
 from rest_framework import status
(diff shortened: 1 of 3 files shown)


datahub/user/company_list/views.py
```diff
@@ -45,6 +54,13 @@ class CompanyListItemView(APIView):
     # Note: A query set is required for CompanyListItemPermissions
     queryset = CompanyListItem.objects.all()
 
+    def get(self, request, format=None, company_pk=None):
+        """Check if a CompanyListItem exists for the authenticated user and specified company."""
+        if not CompanyListItem.objects.filter(adviser=request.user.pk, company_id=company_pk):
```
currycoder: Could it be good to use `.exists()` here?

reupen: Good spot! :+1:

PR #1876, opened 2019-07-19
Switch to OpenAPI-based schema generation

reupen:
### Description of change

This switches to the [OpenAPI-based schema generation added in DRF 3.10.0](https://www.django-rest-framework.org/community/3.10-announcement/).

This also adds a Swagger UI view at `/docs` in a similar vein to the example at https://www.django-rest-framework.org/topics/documenting-your-api/. (Sub-resource integrity checking has been added to the example template).

There are a few more limitations than I'd like:

1. The new OpenAPI-based `AutoSchema` class in DRF is not easily customisable (without using internal methods). 
    
    For now, `ExplicitSerializerSchema` has been removed due to this. A simpler `StubSchema` was added which can be used to suppress default introspection when needed. As a result, there are some missing details for endpoints that were using `ExplicitSerializerSchema`, but we should be able to make improvements if we give some feedback to DRF and related improvements are made there.

2. ~Some endpoints are displaying an error in Swagger UI when you click on them – this will be fixed imminently when DRF release a new version (containing https://github.com/encode/django-rest-framework/pull/6819).~ The update has now been released.

3. There was also a problem with specifying a different schema class on a view set action method that's part of a mixin class – the schema instance ends up being shared between subclasses and schema generation then misbehaves. There's a mechanism in DRF to avoid that for some situations, but it doesn't apply in this one. I put in a simple workaround for it.
    
4. DRF can generate duplicate `operationId` values if there are multiple view sets for the same model. The only side effect I've seen is that operations sharing the same `operationId` share the same expand and collapse state in Swagger UI. Will open a bug in DRF for this.
    
5. Currently `NestedRelatedField` fields are showing as strings – DRF doesn't allow us to easily specify schemas for custom fields. I started a discussion on their mailing list about this (no replies as yet though...)

I'd recommend reviewing each commit individually.

(I'm planning on adding some docs and updating the PR checklist in a follow-up PR.)

### Screenshot

Partial screenshot:

image

### Checklist


datahub/core/schemas.py
```diff
@@ -1,106 +1,17 @@
...
-    return [
-        coreapi.Field(
-            name=field.field_name,
-            location=location,
-            required=field.required and not is_partial,
-            schema=field_to_schema(field),
-        )
-        for field in writable_fields
-    ]
+    def get_operation(self, path, method):
```
mcbhenwood: Is the purpose of this function solely to remove all keys other than 'operationId' from the return value of `AutoSchema.get_operation()`?

datahub/core/templates/core/docs/swagger-ui.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Swagger</title>
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link
+        rel="stylesheet"
+        type="text/css"
+        href="https://unpkg.com/swagger-ui-dist@3.23.1/swagger-ui.css"
```
mcbhenwood: Do you think it's worth getting the version number from elsewhere and inserting it here as a variable? I know it's only a path to a CSS file, but....

datahub/core/test/test_docs.py
```diff
@@ -2,7 +2,31 @@
...
+        ]
+
+    def test_returns_200_if_authenticated(self, client):
+        """Test that a 200 is returned if the user is authenticated via the admin site."""
+        url = reverse('api-docs:swagger-ui')
+        user = create_test_user(is_staff=True, password=self.PASSWORD)
+
+        client = self.create_client(user=user)
+        response = client.get(url)
+        assert response.status_code == status.HTTP_200_OK
```
mcbhenwood: At the risk of being picky, the previous test `test_redirects_to_login_page_if_not_logged_in also sees 200 as success. Seeing as this test is supposed to complement that previous test, perhaps you should also assert here that `len(response.redirect_chain) == 0` ?

reupen: There's another small difference between the two requests – the first one specifies `follow=True` while the second one doesn't.

In the second test any redirects won't be followed and `response.redirect_chain` won't exist.

Usually, we only specify `follow=True` if we're expecting redirects. (If there is a redirect in the second test, it will fail as the response status code won't be a 200.)

datahub/core/templates/core/docs/swagger-ui.html
reupen: I wasn't too much of a fan of it either, but the main other alternative I was thinking of was adding a build process using `npm` and serving the files ourselves (which seemed like too much complexity for this).

Perhaps we could define these in the settings file so they at least have some more visibility? There are two URLs and two hashes that would need adding.

datahub/core/schemas.py
reupen: Yes – perhaps I could update the docstring to make it clearer.

datahub/core/templates/core/docs/swagger-ui.html
mcbhenwood: I'd be happy with putting this in the settings file. Agreed that serving them ourselves is too heavy.

datahub/core/test/test_docs.py
mcbhenwood: I missed that, apologies.

datahub/core/viewsets.py
```diff
@@ -47,7 +49,14 @@ def archive(self, request, pk):
         """
         method = getattr(cls, action)
         mapping = dict(method.mapping)
-        return cls.as_view(mapping, **method.kwargs)
+        initkwargs = method.kwargs
+        # If the method is defined on a mixin class,  a single schema instance will end up
```
elcct: `class,  a` - extra space

config/api_docs_urls.py
```diff
@@ -0,0 +1,43 @@
...
+from django.views.generic import TemplateView
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.renderers import JSONOpenAPIRenderer
+from rest_framework.schemas import get_schema_view
+
+
+API_DOCUMENTATION_TITLE = 'Data Hub API'
+API_DOCUMENTATION_DESCRIPTION = """Auto-generated API documentation for Data Hub.
+
+There are currently some missing or incorrect details as we are dependent on the web framework
```
elcct: Maybe it would be good to add a small note why being dependent on the web framework we are using makes details missing or incorrect.

datahub/investment/project/views.py
```diff
@@ -188,9 +188,15 @@ class IProjectTeamMembersViewSet(CoreViewSet):
     lookup_url_kwarg = 'adviser_pk'
     queryset = _team_member_queryset
 
+    def initial(self, request, *args, **kwargs):
+        """Raise an Http404 if the project referenced in the URL path does not exist."""
+        super().initial(request, *args, **kwargs)
+
+        if not InvestmentProject.objects.filter(pk=self.kwargs['project_pk']).exists():
```
alixedi: 1) Can we do"

```
try
    InvestmentProjct.objects.get(pk=self.kwargs['project_pk'])
except ...
```

2) What is the relationship of this to `get_project` https://github.com/uktrade/data-hub-leeloo/blob/48205b19558a2d0a25b73ff1a4b4a84518f5852f/datahub/investment/project/views.py#L251-L254? Is there potential for a method for code reuse?

3) IIRC, there is `self.get_object`. Why can't we use/override that?

reupen: 1. It isn't using the project so calling `InvestmentProjct.objects.get()` and handling the exception is the wrong choice IMO. (Trying to do something and handling the exception applies when that was the thing you were trying to do in the first place.)

2. We could use that (if we update its docstring).

3. `get_object()` will try to get a team member object, so it won't work for this (and isn't used when e.g. creating a team member).

alixedi: Talked to @reupen IRL. 

1. Agreed
2. Lets not
3. This was a weird one b/c we are not really using retrieve endpoint for team members 🤷‍♂

PR #1885, opened 2019-07-23
Migrate services to a tree structure

elcct:
### Description of change

This add hierarchy to the services using `django-mptt`. This more or less follows the work we did for Sectors.

The service's `name` field has been deprecated and will be removed on or after 5th of August. 

Service name is now being created on the fly by joining segments. 

We decided to output services the same way as before the change (with colon separator), but we may change it in the future.

Services without children have no contexts, as we display these depending if the children match the selected context. There is a test that ensure no service with children has contexts.

This PR will be followed by a PR that deletes the service's `name` column from the database.

### Checklist


mcbhenwood:
This PR talks in terms of migrating Sectors to a tree structure, but I thought we were migrating _Services_ to a tree structure. Am I being thick?

mcbhenwood:
OK so @reupen has kindly explained what's happening - There are legit mentions of sectors in the code but the PR and the commit message talk mistakenly about "sectors" when they do actually mean "services". So I'm half-right.

elcct:
It's the heat :) Of course I meant to write services :)

mcbhenwood:
I really like the implementation internally though. Employing the Django-MPTT package was a really neat, unobtrusive solution.

I just want to make sure that a) we'll gain the value we’d hoped to in the front-end; and b) the data produced by the API is sensibly generic rather than sculpted for the purpose of a specific front-end rendering.

On that second point, what I mean is….. At an API design level it might be preferable to produce properties which look like this:
```name_hierarchy: ['Level1a', 'Level2c', 'Level3a']```
rather than this:
```name: 'Level1a : Level2c : Level3a'```
just in the interests of being more _reflective_ of the data’s nature. However it may be in this case that the flat version is actually preferable in general. Something to discuss tomorrow.

**Comment by reupen** [comment]
A few comments from a quick general look:

1. Could you clarify a bit regarding `is_category_node`? It sounds like these are the leaf nodes (nodes without children). When would the two concepts differ?

1. We should update the validation for interactions and events to make sure that they don't accept category services.

1. It would be good to update the docstring for `Service` to give a bit of an overview of how the tree is organised and used, and how to maintain the services (add services, disable them etc.) in the context of the hierarchy.

1. We might need to think about `contexts` as you could have mismatches at different levels in the tree. 
  
     Perhaps they don't have much utility for categories (since it could be inferred from the contexts of the children)?

    But, if we keep them, it would be good to add a test to make sure that a terminal service does not have a context that its parents node don't have.

**Comment by elcct** [comment]
1) It is when a node has children but itself is not being a service. For example `A specific DIT export service or funding`. It has multiple children, but itself cannot be selected as a service, as we called it "a theme", "a category" etc. but this is not a service. As I am thinking about it, we can only select a leaf node, so maybe this flag is not actually needed, but could help parsing the tree.

2) Fair point

3) Good idea, I'll add some more information

4) I think this is tricky. The front end will display the tree depending on the context and what counts is the context of the leaf node. If leaf node context matches, then front end will display the parent node regardless of its context.

**Comment by reupen** [comment]
1. My feeling is that if we don't have a logical reason for the two concepts diverging, it would be better to compute it. django-mptt provides an `is_leaf_node()` method, we could use that for an `is_category_node()` method and it should be easy to do a similar query set annotation.

4. I think I missed that the contexts were empty for the category nodes. I think that's fine (for what we store), but we should enforce it somehow. As the services are being loaded via a fixture, a test is the only thing that's coming to mind. We could also set the help text for the field to make it clear (to us) that it's only used for leaf nodes.

**Comment by elcct** [comment]
1. I have been thinking about it and realised I am probably duplicating something I could compute off of the MPTT columns. I wanted to have this working first and then look at improvements, so I am going to update it.

2. I'll look into this too. Thank you!

**Comment by elcct** [comment]
1. I removed `is_category_node` field. Filtering with `children__isnull` was sufficient.

2. I added a test to endure services with children have no contexts.

File: datahub/metadata/models.py [hunk 1]
```diff
@@ -206,16 +208,48 @@ class Service(BaseOrderedConstantModel):
...
+    def name(self):
+        """
+        Full name of the service in the form of a path.
+
+        self.get_ancestors() is not used as it's incompatible with pre-fetching.
+        """
+        ancestors = self._get_ancestors_using_parent(include_self=True)
+        return self.PATH_SEPARATOR.join(ancestor.segment for ancestor in ancestors)
+
+    def _get_ancestors_using_parent(self, include_self=False):
```
reupen: Do we need the `include_self` argument? Also this is the same as for `Sector`, should we define a utility function somewhere for this?

datahub/search/interaction/tests/test_views.py
```diff
@@ -919,7 +920,9 @@ def test_interaction_export(
             'attachment', {'filename': 'Data Hub - Interactions - 2018-01-01-11-12-13.csv'},
         )
 
-        sorted_interactions = Interaction.objects.all().order_by(
+        sorted_interactions = Interaction.objects.annotate(
+            service__name=get_service_name_subquery('service'),
```
reupen: Is this annotation being used?

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -773,13 +773,13 @@ def test_common_non_relations(self, field, input_value, expected_value):
...
                 'service',
-                lambda: ServiceFactory(name='UNIQUE EXPORT DEAL'),
+                lambda: ServiceFactory(segment='UNIQUE EXPORT DEAL'),
                 lambda obj: obj.name,
                 id='service look-up (same case)',
             ),
             pytest.param(
                 'service',
-                lambda: ServiceFactory(name='UNIQUE EXPORT DEAL'),
+                lambda: ServiceFactory(segment='UNIQUE EXPORT DEAL'),
```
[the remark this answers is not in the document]
elcct: Added, thanks!

datahub/metadata/models.py
```diff
@@ -206,16 +208,48 @@ class Service(BaseOrderedConstantModel):
         ('service_delivery', 'Service delivery (deprecated)'),
     )
 
+    segment = models.CharField(max_length=settings.CHAR_FIELD_MAX_LENGTH)
+
+    parent = TreeForeignKey(
+        'self', null=True, blank=True, related_name='children',
```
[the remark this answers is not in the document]
elcct: It was. I update the formatting now. Thanks!

datahub/metadata/query_utils.py
```diff
@@ -34,12 +29,36 @@ def get_sector_name_subquery(relation_name=None):
...
-    subquery = Sector._base_manager.annotate(
-        name=_SectorStringAgg(F('segment'), Value(Sector.PATH_SEPARATOR)),
+    subquery = model._base_manager.annotate(
+        name=_SectorStringAgg(F('segment'), Value(model.PATH_SEPARATOR)),
     ).filter(
         lft__lte=OuterRef(f'{outer_ref_prefix}lft'),
         rght__gte=OuterRef(f'{outer_ref_prefix}rght'),
         tree_id=OuterRef(f'{outer_ref_prefix}tree_id'),
-    ).values('name')
+    ).order_by().values('name')
```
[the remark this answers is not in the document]
elcct: Yes, `model.objects` should be used. Updated.

datahub/metadata/query_utils.py
```diff
@@ -1,30 +1,25 @@
 from django.db.models import F, Func, OuterRef, Subquery, TextField, Value
 
-from datahub.metadata.models import Sector
+from datahub.metadata.models import Sector, Service
 
 
 class _SectorStringAgg(Func):
```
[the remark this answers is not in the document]
elcct: Oops... renamed :)

datahub/metadata/test/test_views.py
```diff
@@ -96,7 +96,15 @@ def test_ordered_metadata_order_view(ordered_mapping, api_client):
 
     assert response.status_code == status.HTTP_200_OK
     response_names = [value['name'] for value in response.json()]
-    assert response_names == list(model.objects.order_by('order').values_list('name', flat=True))
+
+    model_filter = {}
+    if model == Service:
```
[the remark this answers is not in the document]
elcct: That's a good idea. I changed it to use `queryset` instead. Thanks!

datahub/metadata/models.py
elcct: This argument indeed has not been used. I removed it and created a class that `Sector` and `Service` inherit. I am not sure about its name though.

datahub/search/interaction/tests/test_views.py
elcct: It's not! I was trying something and then forgot to remove. Thanks!

elcct:
I think I addressed your points. For the last one I also added a validation for Event and Interaction, so that Services without children are not accepted. I am not sure about wording of the error.

datahub/event/serializers.py
```diff
@@ -51,6 +53,14 @@ def validate(self, data):
 
         return data
 
+    def _validate_service(self, combiner):
```
reupen: We should be do this directly in a `validate_service` method since it only looks at one field. Is there any particular reason you went for doing it in `validate`?

changelog/interaction/services-name.db.rst
```diff
@@ -0,0 +1 @@
+The ``metadata_service`` table column ``name`` is deprecated and will be removed on or after the 5th of August 2019.
```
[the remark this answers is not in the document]
reupen: Have we settled on a date to put in the news fragment?

datahub/interaction/serializers.py
```diff
@@ -370,6 +371,7 @@ class Meta:
             HasAssociatedInvestmentProjectValidator(),
             ContactsBelongToCompanyValidator(),
             StatusChangeValidator(),
+            ServiceValidator(),
```
reupen: Same comment as for events, but writing a shared validator for the field might be useful.

datahub/metadata/models.py
```diff
@@ -180,7 +186,7 @@ class Meta(BaseConstantModel.Meta):
         ]
 
 
-class Service(BaseOrderedConstantModel):
+class Service(MPTTModel, _MPTTObjectName, BaseOrderedConstantModel):
     """Service."""
```
reupen: Are we able to add a few notes here about how the hierarchy is used and things to watch out for when adding and removing services etc.?

reupen:
Looks good generally! Just those four comments above.

changelog/interaction/services-name.db.rst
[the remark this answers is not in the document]
elcct: I thought I would update that as the last thing, but forgot to post a comment about it :) I think a month is okay, but I could ask data science if that's enough.

datahub/event/serializers.py
elcct: That's a good point. I thought it will be "cleaner" with adding a method, but maybe it isn't. I added this check to `validate`.

datahub/metadata/models.py
elcct: I am going to write it. It will be useful for sure. Thanks!

elcct: Added some extra notes.

PR #1888, opened 2019-07-25
Add endpoint to retrieve the contents of a user's company list

reupen:
### Description of change

This adds an endpoint at `GET /v4/user/company-list` to retrieve all the companies on the authenticated user's list.

Details of the latest interaction for each company is also included. There wasn't a perfect way to do this in an efficient manner with the Django ORM – I've gone with something that still works well. (There is a need to be fairly efficient here as the list will be displayed on users' home pages.)

I've added an index to keep the looking up of the latest interaction for each company efficient. 
The index takes a bit of time to create (about 17 seconds on staging) so I've used CREATE INDEX CONCURRENTLY in this case (to avoid blocking inserts).

Lastly, we did not want pagination in the initial version of the feature, but removing pagination from the view set will cause DRF to return an array instead of an object with a `results` key (which would be inconsistent, and make it difficult to add pagination later on). So I have left it in – the front end can pass a large value for the `limit` query parameter instead (which is already done elsewhere, and gives us a bit of safety too).

I'd recommend reviewing each commit individually.

### Checklist


datahub/user/company_list/tests/test_views.py
```diff
@@ -9,12 +12,110 @@
...
+                'latest_interaction': {
+                    'id': str(latest_interaction.pk),
+                    'date': format_date_or_datetime(latest_interaction.date),
+                    'subject': latest_interaction.subject,
+                } if latest_interaction else None,
+            },
+        ]
+
+    def test_sorting(self):
+        """Test that list items are sorted by company name."""
```
mcbhenwood: I seem to remember us all discussing sorting and concluding that the results of the list retrieval should be unordered, so that the front end can do the sorting itself (since in the future it will want to introduce its own sorting options on the data once loaded).

reupen: Yes, I can remove the sorting. Just put something in so that's it's in an order until there's sorting in the front end, but can remove it if you prefer.

mcbhenwood: Much as it's a pain to remove tested code, I think it better to declare this an unsorted list from the outset. Really the ordering is the concern of the user (and thus the FE) and seeing as there is a fair amount of db work and computation to produce the result set, shaving off a few cycles of effort by not needing to order it will probably do us a favour.

mcbhenwood: [I approve v much of your query optimisations by the way - but it will take work to produce this data at query time.]

reupen:
As discussed offline, default sorting by date of the latest interaction (in reverse order), has been added.

datahub/user/company_list/queryset.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Returns an annotated query set used by CompanyListViewSet.
+
+    The annotations are supported by an index on the Interaction model.
+
+    (Note that getting all three interaction fields in one expression currently is not easily
+    done with the Django ORM, hence three annotations are used.)
+    """
+    return CompanyListItem.objects.annotate(
+        latest_interaction_id=_get_field_of_latest_interaction('pk'),
```
currycoder: I had a good think (and google) on how we could do this differently and I couldn't really come up with anything that allowed us to use django ORM.

Neat solution!

datahub/core/test/test_query_utils.py
```diff
@@ -67,33 +66,71 @@ def test_get_aggregate_subquery_raises_error_on_on_aggregate_expression(self):
...
+
+        queryset = PersonListItem.objects.annotate(
+            name_of_latest_book=get_top_related_expression_subquery(
+                Book.proofreader.field,
+                'name',
+                ('-published_on',),
+                outer_field='person__pk',
+            ),
+        )
+        assert queryset.first().name_of_latest_book == 'newest'
```
currycoder: Actually, I have a further suggestion.  It might be reasonable to test what happens when a Person has proofread no books.  In the business case for this feature, this would be when a company has no interactions I believe.

reupen: I've added this :+1:

reupen:
Just spotted that the logic to filter to the current user was missing :grimacing: – just pushed a fix for that (also addressed @currycoder 's comment).

datahub/user/company_list/queryset.py
reupen: [`FilteredRelation`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#filteredrelation-objects) was very close, but it has limitations on the condition in Q object which seem to prevent it being used for this.

Still, it does seem possible to use it to annotate a query set with a relation, and then add additional annotations that use the first one (the annotation using `FilteredRelation` only seems to be accessible within the query itself and not on the objects returned by the query).

currycoder: Yeah I did have a play with FilteredRelation in a shell, but I couldn't quite get there.

PR #1896, opened 2019-07-30
Spike: Add proposed DNB API contract

currycoder:
### Description of change

This is a proposal for the API contracts for searching/saving companies from DNB.  What do we think?

*Note:* This branch will not merge, it's just used for easy review and critique of the backend/frontend contract.

### Background

For context, the overall user flow that this will assist with is as follows:
- User goes to record an interaction.
- User searches for the contact on Data Hub, they don't exist.
- User searches for the company - through DNB search, it does not yet exist on Data Hub.
- User chooses the company from a DNB search result.
- User enters data for the contact that they spoke to.
- User submits the information and the Company/Contact/Interaction are saved on Data Hub.

In the case for these API endpoints, we are aiming to provide a solution for DNB-based company search and saving a new company that exists in DNB but does not yet exist in Data Hub.

### Checklist

### DIFF
--- a/datahub/dnb_api/README.md
+++ b/datahub/dnb_api/README.md
@@ -1,0 +1,220 @@
+# DNB Search and Save API Contracts
+
+## DNB Company Search (with DH hydration)
+
+`POST /v4/dnb/company-search`
+
+**Request JSON Body:**
+- `search_term` - required - The string search term to use to find the company e.g. `"siemens"`
+- `address_counrty` - optional - A string ISO Alpha 2 code representing the country to use to filter search results e.g. `"GB"`
+- `address_postcode` - optional - A string representing the postcode of the business address/registered address to filter search results e.g. `"BN1 4SE"`
+- `page_size` - optional - Integer number of results to show per page of results, defaults to `10`
+- `page_number` - optional - Integer of the page of results to return, defaults to `1`
+
+**Response:**
+
+```
+{
+  "total_matches": 271,
+  "total_returned": 2,
+  "page_size": 2,
+  "page_number": 1,
+  "results": [
+    {
+      "dnb_company": {
+        "duns_number": "12345678",
+        "primary_name": "Some company name",
+        "trading_names": [
+          "Some trading name"
+        ],
+        "registration_numbers": [
+          {
+            "registration_type": "uk_companies_house_number",
+            "registration_number": "1234567"
+          }
+        ],
+        "global_ultimate_duns_number": "123456789",
+        "global_ultimate_primary_name": "Some parent company name",
+        "domain": "example.co.uk",
+        "is_out_of_business": false,
+        "address_line_1": "123 Fake Street",
+        "address_line_2": "",
+        "address_town": "Brighton",
+        "address_county": "",
+        "address_postcode": "BN1 4SE",
+        "address_country": "GB",
+        "registered_address_line_1": "",
+        "registered_address_line_2": "",
+        "registered_address_town": "Brighton",
+        "registered_address_county": "",
+        "registered_address_postcode": "BN1 4SE",
+        "registered_address_country": "GB",
+        "annual_sales": 1860000000,
+        "annual_sales_currency": "USD",
+        "is_annual_sales_estimated": null,
+        "employee_number": 2000,
+        "is_employees_number_estimated": true,
+        "industry_codes": [
+          {
+            "usSicV4": "1623",
+            "usSicV4Description": "Water/sewer/utility construction"
+          }
+        ],
+        "legal_status": "corporation"
+      },
+      "datahub_company": {
+        "id": "0fb3379c-341c-4da4-b825-bf8d47b26baa",
+        "latest_interaction": {
+          "id": "ec4a46ef-6e50-4a5c-bba0-e311f0471312",
+          "created_on": "2019-08-01T18:10:00",
+          "date": "2019-08-01",
+          "subject": "Meeting between DIT and Joe Bloggs"
+        }
+      }
+    },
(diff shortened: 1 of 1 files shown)


datahub/dnb_api/README.md
```diff
@@ -0,0 +1,220 @@
...
+## DNB Company Search (with DH hydration)
+
+`POST /v4/dnb/company-search`
+
+**Request JSON Body:**
+- `search_term` - required - The string search term to use to find the company e.g. `"siemens"`
+- `address_counrty` - optional - A string ISO Alpha 2 code representing the country to use to filter search results e.g. `"GB"`
+- `address_postcode` - optional - A string representing the postcode of the business address/registered address to filter search results e.g. `"BN1 4SE"`
+- `page_size` - optional - Integer number of results to show per page of results, defaults to `10`
+- `page_number` - optional - Integer of the page of results to return, defaults to `1`
```
paulgain: Looking at the design we also have an option for city or town.

datahub/dnb_api/README.md
```diff
@@ -0,0 +1,220 @@
...
+        "is_employees_number_estimated": true,
+        "industry_codes": [
+          {
+            "usSicV4": "3799",
+            "usSicV4Description": "Mfg transportation equipment"
+          }
+        ],
+        "legal_status": "corporation"
+      },
+      "datahub_company": {}
```
paulgain: From a frontend perspective and when there is not a company within Data Hub I would prefer to see this field omitted or set to null. Trying to avoid scenarios like this: 
```
if (company.datahub_company) { // This will be true even though the object is empty}
```

currycoder:
@rafenden, @paulgain  and I just had a discussion about the "Saving a DNB Company to Data Hub" approaches.  The broad consensus was that Approach 2 probably solves a lot of problems and prevents some unnecessary saving of search results in a session (or similar) on the frontend along with some transposing of data between the format DNB search returns and the format that Data Hub's company API expects.

One drawback with this approach is that we are making (at least) two calls to the DNB API for each company that we save - one to search, one to save the details on Data Hub.  We can look at how to improve upon this in a future iteration though; perhaps with a caching solution (in Data Hub API or in dnb-service) or using the CMPELK endpoint for company detail requests when saving the company to Data Hub in order to distribute our API calls better and avoid rate limits.

Another thing we discussed was that this pushes us towards making 3 sequential API calls at the end of the process:
1) Save the company from a DNB duns number.
2) Save the contact to the newly created company.
3) Save the interaction.

This could prove to be slow and we might run in to problems to do with errors occurring in the middle of this process.  An alternative could be to provide a single endpoint which could allow a user to save all these details in a single call.  We agreed to go ahead with the three separate calls (as there are already endpoints available for two of them) and review it again later.

datahub/dnb_api/README.md
currycoder: Will do!  Good observation.

datahub/dnb_api/README.md
currycoder: I've asked Michael about this - there's a chance that we won't need it, but we are waiting on more user feedback before making that decision.  Let's check in on it later.

paulgain:
**Things we need to consider:** 

When the user completes the journey and then selects _Add Interaction_  we will need to POST up to 3 entities:

1. Contact
2. Company
3. Interaction

As discussed there are two approaches:

1. Make a single API call posting a chunk of JSON that encapsulates 1 - 3.

```
{
 contact: { ... }
 company: { ... }
 interaction: { ... }
}
```

2. Make 3 separate synchronised API calls to each corresponding endpoint. 

There are pros and conns to each, however, from a FE perspective I would prefer the first option as it's far simpler, we don't have to guard against one of the 3 API calls failing and trying to recover from that scenario.

datahub/dnb_api/README.md
```diff
@@ -0,0 +1,220 @@
+# DNB Search and Save API Contracts
+
+## DNB Company Search (with DH hydration)
+
+`POST /v4/dnb/company-search`
+
+**Request JSON Body:**
+- `search_term` - required - The string search term to use to find the company e.g. `"siemens"`
+- `address_counrty` - optional - A string ISO Alpha 2 code representing the country to use to filter search results e.g. `"GB"`
```
elcct: `address_country`

datahub/dnb_api/README.md
currycoder: Michael has said to leave it for now.

paulgain: OK, I've just spoken to Rogier and I guess he'll modify the design.

currycoder:
@paulgain Just looping back on this after our change in feature scope that we discussed as a team yesterday.

I think this pushes us towards the initial proposal for "Approach 2" mentioned here: https://github.com/uktrade/data-hub-leeloo/pull/1896/files#diff-d98c4ace8935262b985c9c4644edcf81R136

We can revisit the question of "do we provide an endpoint to save DNB company/contact/interaction all at the same time" later when that is in scope for the product again.

paulgain:
> @paulgain Just looping back on this after our change in feature scope that we discussed as a team yesterday.
> 
> I think this pushes us towards the initial proposal for "Approach 2" mentioned here: https://github.com/uktrade/data-hub-leeloo/pull/1896/files#diff-d98c4ace8935262b985c9c4644edcf81R136
> 
> We can revisit the question of "do we provide an endpoint to save DNB company/contact/interaction all at the same time" later when that is in scope for the product again.

👍

currycoder:
Closing for now.  The contract has been agreed in principle and will be documented as we go while building the API endpoints.

PR #1899, opened 2019-07-31
Add calendar invite monitoring

alixedi:
### Description of change

This PR puts in the minimal thing that would let us publish metrics to Prometheus for monitoring calendar invite ingestion. Because the calendar invite ingestion happens inside an Celery task we have to go via [pushgateway](https://github.com/prometheus/pushgateway/) which is the canonical solution for connecting Prometheus to ephemeral processes.

In order to test this locally, I did something along the following lines:

```
$ docker run -p 9091:9091 prom/pushgateway
$ PUSHGATEWAY_URL=localhost:9091 pytest datahub/interaction/test/email_processors
$ curl localhost:9091/metrics | grep email_ingest_failure
```

The `webops` team is working on getting us a `pushgateway` in production. Rolling this feature out would require configuring `PUSHGATEWAY_URL` to point to the thing.

### Checklist

### DIFF
--- a/datahub/monitoring/interactions/email_processors.py
+++ b/datahub/monitoring/interactions/email_processors.py
@@ -1,0 +1,44 @@
+from prometheus_client import Counter
+
+from datahub.monitoring import registry
+from datahub.monitoring.utils import push_to_gateway
+
+
+JOB_NAME = 'calendar-invite-ingest'
+
+failure_counter = Counter(
+    'email_ingest_failure',
+    'Number of failed calendar invite ingestions',
+    ('domain', ),
+    registry=registry,
+)
+
+success_counter = Counter(
+    'email_ingest_success',
+    'Number of successful calendar invite ingestions',
+    ('domain', ),
+    registry=registry,
+)
+
+
+def get_adviser_domain(adviser):
+    """
+    Return domain for the adviser's email.
+    """
+    return adviser.get_current_email().split('@')[-1]
+
+
+def record_failure(adviser):
+    """
+    Increment the failure counter.
+    """
+    failure_counter.labels(domain=get_adviser_domain(adviser)).inc()
+    push_to_gateway(JOB_NAME)
+
+
+def record_success(adviser):
+    """
+    Increment the success counter.
+    """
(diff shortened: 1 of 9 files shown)


datahub/monitoring/interactions/email_processors.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_adviser_domain(adviser):
+    """
+    Return domain for the adviser's email.
+    """
+    return adviser.get_current_email().split('@')[-1]
+
+
+def record_failure(adviser):
```
mcbhenwood: Because both `record_failure` and `record_success` only employ the _domain_ of the adviser, do you think it would be more reflective to do the domain determination outside these two functions, and make the argument into `domain`? When I read the code that calls `record_{failure,success}` I felt that it was going to store some data pertaining to the specific adviser, which is not actually the case.

datahub/monitoring/utils.py
```diff
@@ -0,0 +1,24 @@
...
+logger = get_task_logger(__name__)
+
+
+def push_to_gateway(job):
+    """
+    Wrapper around push_to_gateway that handles the exceptions gracefully.
+    """
+    pushgateway_url = settings.PUSHGATEWAY_URL
+    if pushgateway_url is None:
+        logger.info('PUSHGATEWAY_URL have not been configured.')
```
mcbhenwood: "has not" :)

alixedi:
This didn't quite work out 😄 

We decided to go with [Pushgateway](https://github.com/prometheus/pushgateway/) which seemed like the canonical Prometheus solution to monitoring ephemeral processes (e.g. Celery tasks). 

After having written this code, I realized that metrics from difference processes were overwriting each other and upon reading more it seemed like it was by design. In hindsight, the Pushgateway example in the Python client’s documentation was a bit misleading as it used a `last_success_unixtime` metric - which incidentally is something you would want to overwrite.

There is a way for metrics to not overwrite each other by tagging them with PIDs as shown [here](https://groups.google.com/forum/#!msg/prometheus-users/qC5gakQPiMQ/OHxKstYOAQAJ) but this is hardly an elegant solution as it creates as many metrics as Celery processes.

PR #1903, opened 2019-07-31
Hydrate DNB search API responses with Data Hub company data if it exists

currycoder:
### Description of change

This PR aims to give DNB search results some context for the user by hydrating the results with Data Hub company data. Search results will be hydrated if the corresponding Company exists on Data Hub and if it has a duns number which matches the DNB duns number.

We have a requirement for the last interaction to show against each matched search result in the frontend, so this data is exposed using @reupen's pattern for getting the latest related entity to a model from a one to many relationship.  It's possible that we will need other company/related data to be exposed later, but for now this gives us a basis to get started with.

### Checklist

### DIFF
--- a/datahub/dnb_api/queryset.py
+++ b/datahub/dnb_api/queryset.py
@@ -1,0 +1,31 @@
+from django.db.models import F
+
+from datahub.company.models import Company
+from datahub.core.query_utils import get_top_related_expression_subquery
+from datahub.interaction.models import Interaction
+
+
+def get_company_queryset():
+    """
+    Returns an annotated query set used by DNBCompanySearchView.
+
+    The annotations are supported by an index on the Interaction model.
+
+    (Note that getting all four interaction fields in one expression currently is not easily
+    done with the Django ORM, hence four annotations are used.)
+    """
+    return Company.objects.annotate(
+        latest_interaction_id=_get_field_of_latest_interaction('pk'),
+        latest_interaction_created_on=_get_field_of_latest_interaction('created_on'),
+        latest_interaction_date=_get_field_of_latest_interaction('date'),
+        latest_interaction_subject=_get_field_of_latest_interaction('subject'),
+    )
+
+
+def _get_field_of_latest_interaction(field):
+    return get_top_related_expression_subquery(
+        Interaction.company.field,
+        F(field),
+        ('-date', '-created_on', 'pk'),
+        outer_field='id',
+    )
--- a/datahub/dnb_api/test/test_views.py
+++ b/datahub/dnb_api/test/test_views.py
@@ -9,10 +9,12 @@ from django.utils.timezone import make_aware
 from rest_framework import status
 from rest_framework.reverse import reverse
 
+from datahub.company.models import CompanyPermission
 from datahub.company.test.factories import CompanyFactory
-from datahub.core.test_utils import APITestMixin
+from datahub.core.test_utils import APITestMixin, create_test_user
 from datahub.dnb_api.constants import FEATURE_FLAG_DNB_COMPANY_SEARCH
 from datahub.feature_flag.test.factories import FeatureFlagFactory
+from datahub.interaction.models import InteractionPermission
 from datahub.interaction.test.factories import CompanyInteractionFactory
 
 
@@ -142,8 +144,16 @@ class TestDNBCompanySearchAPI(APITestMixin):
             content=upstream_response_content,
         )
 
+        user = create_test_user(
+            permission_codenames=[
+                CompanyPermission.view_company,
+                InteractionPermission.view_all,
+            ],
+        )
+        api_client = self.create_api_client(user=user)
+
         url = reverse('api-v4:dnb-api:company-search')
-        response = self.api_client.post(
+        response = api_client.post(
             url,
             data=request_data,
             content_type='application/json',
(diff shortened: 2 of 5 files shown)


datahub/dnb_api/queryset.py
```diff
@@ -0,0 +1,31 @@
+from django.db.models import F
+
+from datahub.company.models import Company
+from datahub.core.query_utils import get_top_related_expression_subquery
+from datahub.interaction.models import Interaction
+
+
+def get_company_queryset():
+    """
+    Returns an annotated query set used by DNBCompnaySearchView.
```
mcbhenwood: "DNBCompanySearchView" :)

datahub/dnb_api/views.py
```diff
@@ -26,15 +30,50 @@ class DNBCompanySearchView(APIView):
...
+
+        hydrated_results = []
+
+        for dnb_result in dnb_results:
+            duns_number = dnb_result['duns_number']
+            hydrated_result = {'dnb_company': dnb_result, 'datahub_company': None}
+            datahub_company = datahub_companies_by_duns.get(duns_number)
+            if datahub_company:
+                datahub_company_data = DNBMatchedCompanySerializer(datahub_company).data
+                hydrated_result['datahub_company'] = datahub_company_data
```
reupen: I'd try to define `hydrated_result` in one go and separate it from the conditional logic to make it a bit easier to follow. In other words, define `datahub_company_data` first and then `hydrated_result`.

reupen:
There are some groups of users that don't have access to company interactions – will they have access to the D&B search?

Additionally, the search should require the `company.view_company` permission now it includes Data Hub company details. (There is a `HasPermissions` permission class that could be used for that.)

datahub/dnb_api/views.py
```diff
@@ -26,15 +30,50 @@ class DNBCompanySearchView(APIView):
...
+            response_body_text = JSONRenderer().render(response_body)
+
         return HttpResponse(
-            upstream_response.text,
+            response_body_text,
             status=upstream_response.status_code,
             content_type=upstream_response.headers.get('content-type'),
         )
 
+    def _transpose_and_hydrate(self, dnb_results):
```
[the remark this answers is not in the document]
currycoder: Good point.  It's more just changing the format of the results, I've changed it to `_format_and_hydrate()`

datahub/dnb_api/views.py
currycoder: Nice observation, thanks.

currycoder:
> There are some groups of users that don't have access to company interactions – will they have access to the D&B search?
> 
> Additionally, the D&B search view should probably require the `company.view_company` permission now it includes Data Hub company details. (There is a `HasPermissions` permission class that could be used for that.)

Thanks Reupen.  I've added in permissions for these parts.  As we discussed, I did look at `PermittedFieldsModelSerializer` for ensuring that `last_interaction` data was not shown to users that do not have permissions for interactions. Unfortunately it looked like it omitted the field entirely - for ease of frontend development (and to avoid us missing the edge case later) I've added my own code which just sets the data to `None` if the user does not have permission, while ensuring that the API still responds with a `last_interaction` key.

currycoder:
> I think there is an advantage of omitting the field: it more clearly differentiates between 'this company has no interactions and 'I can't tell you anything about the company's interactions'. But, the permission can also be used to work the latter out if needed so either should work.

I like the rationale here.  I think I'll go ahead and use the `PermittedFieldsModelSerializer` for the benefit of code reuse. Although using it does necessitate the serializer knowing about the request, which is at odds with your other observation.

reupen:
> Although using it does necessitate the serializer knowing about the request, which is at odds with your other observation.

I know :disappointed:

datahub/dnb_api/views.py
```diff
@@ -20,21 +25,92 @@ class DNBCompanySearchView(APIView):
...
+              "id": "e8c3534f-4f60-4c93-9880-09c22e4fc011",
+              "created_on": "2018-04-08T14:00:00Z",
+              "date": "2018-06-06",
+              "subject": "Meeting with Joe Bloggs"
+            }
+          }
+        }
+
+        """
+        duns_numbers = [result['duns_number'] for result in dnb_results]
```
alixedi: I feel the code here can be a bit more succinct. I would propose something along the following lines to give you some ideas (prolly not all good ones):

```python
def get_matched_datahub_company(company):
    return DNBMatchedCompanySerializer(
        company, 
        context={'request': self.request})
    }.data

def format_dnb_companies(companies):
    return {
        company['duns_number']: {'dnb_company': company} 
        for company in companies
    }

def get_datahub_companies(dnb_numbers):
     return get_company_queryset().filter(duns_number__in=dnb_numbers)

def format_datahub_companies(companies):
    return {
        company.duns_number: {'datahub_company': get_matched_datahub_company(company)
        for company in companies
    }

def format_and_hydrate(self, dnb_results):
    dnb_companies = format_dnb_companies(dnb_results)
    datahub_companies = format_datahub_companies(
        get_datahub_companies(dnb_companies.keys())
    )

    return {
        # All companies found in DNB 
        **dnb_companies, 
        # Override ones that were matched to DataHub with richer matched objects
        **{
            k: {**datahub_companies[k], **dnb_companies[k]} 
            for k in data hub_companies.keys()
        }
    }.values()
```

Essentially, trying to break it down into small testable units.

currycoder: Honestly, I feel like the specific suggestion here is a little over the top - to me, it replaces complexity in the original solution with new complexity from dealing with intermediate dictionaries and then merging them in a complex comprehension.

I do think there's a good point about breaking the method down in to separate chunks for readability/maintainability though.  I've broken parts out of the original method in to helpers with the hope that the overall `_format_and_hydrate` method is more readable.

datahub/dnb_api/views.py
```diff
@@ -20,21 +25,98 @@ class DNBCompanySearchView(APIView):
...
+            company.duns_number: company for company in datahub_companies
+        }
+
+    def _get_datahub_company_data(self, datahub_company):
+        if datahub_company:
+            return DNBMatchedCompanySerializer(
+                datahub_company,
+                context={'request': self.request},
+            ).data
+        else:
```
reupen: This `else:` can be removed (and `return None` brought up an indentation level) now.

datahub/dnb_api/test/test_views.py
```diff
@@ -18,60 +26,141 @@ def dnb_company_search_feature_flag():
...
-        response = self.api_client.post(
+        response = api_client.post(
             url,
             data=request_data,
             content_type='application/json',
         )
 
         assert response.status_code == response_status_code
-        assert response.content == response_content
+        assert json.loads(response.content) == response_data
```
reupen: Are we able to use `response.json()` (and in similar lines elsewhere)?

datahub/dnb_api/views.py
```diff
@@ -20,21 +25,97 @@ class DNBCompanySearchView(APIView):
...
+          }
+        }
+
+        """
+        duns_numbers = [result['duns_number'] for result in dnb_results]
+        datahub_companies_by_duns = self._get_datahub_companies_by_duns(duns_numbers)
+
+        hydrated_results = []
+
+        for dnb_result in dnb_results:
```
reupen: It might be worth having a go at moving the loop body to a separate function and using a list comprehension – I'll leave it with you to have a look.

datahub/dnb_api/test/test_views.py
```diff
@@ -18,60 +26,141 @@ def dnb_company_search_feature_flag():
...
+def dnb_company_search_datahub_companies():
+    """
+    Creates Data Hub companies for hydrating DNB search results with.
+    """
+    # Company with no interactions
+    CompanyFactory(duns_number='1234567', id='6083b732-b07a-42d6-ada4-c8082293285b')
+    # Company with two interactions
+    company = CompanyFactory(duns_number='7654321', id='6083b732-b07a-42d6-ada4-c99999999999')
+
+    interaction_date = make_aware(
```
reupen: Elsewhere we pass `tzinfo` explicitly instead of using `make_aware()` – it would be good to keep it consistent.

datahub/dnb_api/test/test_views.py
```diff
@@ -18,60 +26,141 @@ def dnb_company_search_feature_flag():
...
+        datetime.datetime(year=2019, month=8, day=1, hour=16, minute=0, second=0),
+    )
+    latest_interaction = CompanyInteractionFactory(
+        id='6083b732-b07a-42d6-ada4-222222222222',
+        date=interaction_date,
+        subject='Meeting with Joe Bloggs',
+        company=company,
+    )
+    latest_interaction.created_on = interaction_date
```
reupen: Although this should work, I think I would use `freeze_time()` since overwriting `created_on` wouldn't normally be done.

datahub/dnb_api/test/test_views.py
```diff
@@ -18,60 +26,141 @@ def dnb_company_search_feature_flag():
...
+    latest_interaction.save()
+
+    older_interaction_date = make_aware(datetime.datetime(year=2018, month=8, day=1))
+    older_interaction = CompanyInteractionFactory(
+        id='6083b732-b07a-42d6-ada4-111111111111',
+        date=older_interaction_date,
+        subject='Meeting with John Smith',
+        company=company,
+    )
+    older_interaction.created_on = older_interaction_date
```
reupen: This change doesn't seem to be saved (but see previous comment too).

datahub/dnb_api/views.py
elcct: I think this is a good suggestion to move the body to a separate function.

PR #1900, opened 2019-07-31
Explicitly order multi-value fields in search CSV exports

reupen:
### Description of change

This sets StringAgg orderings in CSV export query sets now that https://code.djangoproject.com/ticket/30315 has been fixed.

The ordering used is alphabetical as this is the only real ordering available.

### Checklist

* [x] Has a new newsfragment been created? Check [changelog/README.rst](https://github.com/uktrade/data-hub-leeloo/blob/master/changelog/README.rst) for instructions
* [ ] Have any relevant search models been updated?
* [ ] Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?
* [ ] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?
* [ ] Has the admin site been updated (for new models, fields etc.)?
* [ ] Has the README been updated (if needed)?

PR #1905, opened 2019-08-01
Update contact search CSV export for multiple interaction advisers

reupen:
### Description of change

Dependent on https://github.com/uktrade/data-hub-leeloo/pull/1900.

This updates the 'Team of latest interaction' field in the contact search CSV export to use `Interaction.dit_participants` instead of the old `Interaction.dit_team` field.

As a result, it now can contain multiple (comma-separated) values and the output column has been renamed 'Teams of latest interaction'.

Only unique values are included (as one team could be on an interaction multiple times if multiple advisers from the team were added to the interaction).

### Checklist

### DIFF
--- a/datahub/core/test/test_query_utils.py
+++ b/datahub/core/test/test_query_utils.py
@@ -1,5 +1,5 @@
 from datetime import date
-from random import shuffle
+from random import sample, shuffle
 
 import factory
 import pytest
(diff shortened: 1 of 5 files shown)


datahub/core/test/test_query_utils.py
```diff
@@ -22,23 +23,42 @@
...
+        actual_author_names = queryset.first().author_names or ''
+        expected_author_names = join_attr_values(
+            book.authors.order_by('first_name'),
+            'first_name',
+        )
+        assert actual_author_names == expected_author_names
+
+    def test_distinct(self):
+        """Test that duplicate values are omitted if distinct=True is specified."""
+        names = ['John', 'John', 'Barbara', 'Claire']
```
elcct: I think this is a bit too simple. I'd add some more duplicated, e.g. two more of `John` and one more `Barbara`.

datahub/search/contact/views.py
```diff
@@ -105,9 +106,9 @@ class SearchContactExportAPIView(SearchContactAPIViewMixin, SearchExportAPIView)
             DBContact,
             Max('interactions__date'),
         ),
-        team_of_latest_interaction=get_top_related_expression_subquery(
+        teams_of_latest_interaction=get_top_related_expression_subquery(
             DBInteraction.contacts.field,
-            F('dit_team__name'),
+            get_string_agg_subquery(DBInteraction, 'dit_participants__team__name', distinct=True),
```
currycoder: This is essentially grouping teams by the `name` field.  Looking at the models, there appears to be no unique constraint on `Team.name` - is it possible in practice for the same name string to be shared between multiple teams?  e.g.

"Investments" (North East), "Investments" (South West), ...

reupen: The most likely scenario it would happen in is when multiple advisers from the same team are added to an interaction. In that case, returning the team name multiple names is undesirable.

In theory teams with duplicate names could also be created, but that would be unintentional and create other problems (not being able to tell them apart etc.)

datahub/core/test/test_query_utils.py
```diff
@@ -22,23 +23,42 @@
...
+            first_name=factory.Iterator(
+                sample(names, len(names)),
+            ),
+        )
+        BookFactory(authors=authors)
+        queryset = Book.objects.annotate(
+            author_names=get_string_agg_subquery(Book, 'authors__first_name', distinct=True),
+        )
+        actual_author_names = queryset.first().author_names
+        assert actual_author_names == 'Barbara, Claire, John, Samantha'
```
mcbhenwood: I feel that to expect the team names to be in alphabetical order is conflating the concept of removing duplicates with the known behaviour of the overall implementation. It's really picky but I feel that if the test is solely concerned with the distinction taking place, it should use different assertions. Probably easiest would be to assert the `len` is 4, then that each of these names are _somewhere_ in the `author_names`.

reupen:
> Do you think that commit 8b23355 should be a fixup? Seeing as it addresses a comment which came up in this review.

Yes, it wasn't marked as one but was meant as one. I'll squash it when rebasing.

datahub/core/test/test_query_utils.py
reupen: Did you mean to split the string on `', '` and then do those two assertions?

The intention of the two tests was to test the overall behaviour of `get_string_agg_subquery()` in the two main cases (`distinct=True` and `distinct=False`).

So, we would need at least one more test to also test the ordering when `distinct=True` if we removed that from this test.

reupen: I've rewritten the tests to be more comprehensive and in line with my original intention.

PR #1908, opened 2019-08-02
Add statsd monitoring for calendar invite ingest

alixedi:
### Description of change

This puts in the minimal code required to send metrics to `statsd` or specifically [statsd_exporter](https://github.com/prometheus/statsd_exporter). 

There are a few caveats around the DSL used by `statsd_exporter` to map metrics. Happy to hear ideas for how to solve this more elegantly.

To run this, you can do something along the following lines:

```
$ docker run -d -p 9102:9102 -p 9125:9125 -p 9125:9125/udp \
-v $PWD/statsd_mapping.yml:/tmp/statsd_mapping.yml \
prom/statsd-exporter --statsd.mapping-config=/tmp/statsd_mapping.yml

$ pytest datahub/interaction/test/email_processors/test_processors.py

$ curl localhost:9102/metrics

💰 
```

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the 
 | `DATABASE_URL`  | Yes | PostgreSQL server URL (with embedded credentials). |
 | `DATAHUB_FRONTEND_BASE_URL`  | Yes | |
 | `DATAHUB_NOTIFICATION_API_KEY` | No | The GOVUK notify API key to use for the `datahub.notification` django app. |
+| `DATAHUB_SUPPORT_EMAIL_ADDRESS` | No | Email address for DataHub support team. |
 | `DEBUG`  | Yes | Whether Django's debug mode should be enabled. |
 | `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass\|spf:pass\|dkim:pass |
 | `DJANGO_SECRET_KEY`  | Yes | |
@@ -305,15 +306,13 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the 
 | `RESTRICT_ADMIN` | No | Whether to restrict access to the admin site by IP address. |
 | `SENTRY_ENVIRONMENT`  | Yes | Value for the environment tag in Sentry. |
 | `SSO_ENABLED` | Yes | Whether single sign-on via RFC 7662 token introspection is enabled |
+| `STATSD_HOST` | No | StatsD host url. |
+| `STATSD_PORT` | No | StatsD port number. |
+| `STATSD_PREFIX` | No | Prefix for metrics being pushed to StatsD. |
 | `VCAP_SERVICES` | No | Set by GOV.UK PaaS when using their backing services. Contains connection details for Elasticsearch and Redis. |
 | `WEB_CONCURRENCY` | No | Number of Gunicorn workers (set automatically by Heroku, otherwise defaults to 1). |
 | `ENABLE_EMAIL_INGESTION` | No | True or False.  Whether or not to activate the celery beat task for ingesting emails |
-| `MAILBOX_MEETINGS_USERNAME` | No | Username of the inbox for ingesting meeting invites via IMAP (likely to be the same as the email for the mailbox) |
-| `MAILBOX_MEETINGS_PASSWORD` | No | Password for the inbox for ingesting meeting invites via IMAP |
-| `MAILBOX_MEETINGS_IMAP_DOMAIN` | No | IMAP domain for the inbox for ingesting meeting invites via IMAP |
-| `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass\|spf:pass\|dkim:pass |
 | `DATAHUB_NOTIFICATION_API_KEY` | No | The GOVUK notify API key to use for the `datahub.notification` django app. |
-| `DATAHUB_SUPPORT_EMAIL_ADDRESS` | No | Email address for DataHub support team. |
 
 
 ## Management commands
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -567,13 +567,13 @@ MAILBOXES = {
 DIT_EMAIL_INGEST_WHITELIST = env.list('DIT_EMAIL_INGEST_WHITELIST', default=[])
 DIT_EMAIL_DOMAINS = {}
 domain_environ_names = [
-    environ_name 
+    environ_name
     for environ_name in env.ENVIRON.keys()
     if environ_name.startswith('DIT_EMAIL_DOMAIN_')
 ]
 
 # Go through all DIT_EMAIL_DOMAIN_* environment variables and extract
-# dictionary with key email domain and value consisting of 
+# dictionary with key email domain and value consisting of
 # authentication method/minimum pass result pairs e.g.
 # example.com=dmarc:pass|spf:pass|dkim:pass becomes
 # {'example.com': [['dmarc', 'pass'], ['spf', 'pass'], ['dkim', 'pass']]}
@@ -592,3 +592,7 @@ DNB_SERVICE_BASE_URL = env('DNB_SERVICE_BASE_URL', default=None)
 DNB_SERVICE_TOKEN = env('DNB_SERVICE_TOKEN', default=None)
 
 DATAHUB_SUPPORT_EMAIL_ADDRESS = env('DATAHUB_SUPPORT_EMAIL_ADDRESS', default=None)
+
+STATSD_HOST = env('STATSD_HOST', default='localhost')
+STATSD_PORT = env('STATSD_PORT', default='9125')
+STATSD_PREFIX = env('STATSD_PREFIX', default='datahub-api')
--- a/datahub/company/models/adviser.py
+++ b/datahub/company/models/adviser.py
@@ -121,6 +121,17 @@ class Advisor(AbstractBaseUser, PermissionsMixin):
         """
         return self.contact_email or self.email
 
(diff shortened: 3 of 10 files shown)


config/settings/common.py
```diff
@@ -583,3 +583,7 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
 DNB_SERVICE_TOKEN = env('DNB_SERVICE_TOKEN', default=None)
 
 DATAHUB_SUPPORT_EMAIL_ADDRESS = env('DATAHUB_SUPPORT_EMAIL_ADDRESS', default=None)
+
+STATSD_HOST = env('DATAHUB_STATSD_HOST', default='localhost')
```
reupen: Do we want these as defaults? We could also make it required and put these in the sample env file (so we don't inadvertently end up with a broken configuration in production).

Also most of the environment variables do not have a `DATAHUB_` prefix, it would be good to be consistent.

reupen:
We also need to add the new environment variables to the README.

datahub/company/models/adviser.py
```diff
@@ -121,6 +121,12 @@ def get_current_email(self):
         """
         return self.contact_email or self.email
 
+    def get_email_domain(self):
+        """
+        :returns: Domain for adviser's current email.
+        """
+        return self.get_current_email().split('@')[-1]
```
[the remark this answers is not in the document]
alixedi: > We are doing something similar in was_email_sent_by_dit() – should we write a utility function for getting a domain from an email address, so we are only doing it in one way?

I have taken a look at `was_email_sent_by_dit`. It is also trying to find the domain in context of an `adviser`'s email and so I reckon instead of moving this to a utility function, we should refactor the code that calls `was_email_sent_by_dit`.

This should happen in a separate PR.

> Also, should we normalise it to lower case?

Yup.

> Also, should we have a generic fallback in case it doesn't find an @ in the email address (due to dodgy data)?

At this moment, we have the following behaviour:

```
>>> adviser = AdviserFactory(contact_email='adviserdit')
>>> adviser.get_email_domain()
`adviserdit`
```

Are you suggesting something different?

config/settings/common.py
alixedi: > Do we want these as defaults? 

Can't see any problem with it.

> We could also make it required and put these in the sample env file (so we don't inadvertently end up with a broken configuration in production).

I want to stress that this is monitoring config and while I want to place a certain amount of importance to this, it is not important enough to bring down production in case of misconfiguration.

> Also most of the environment variables do not have a DATAHUB_ prefix, it would be good to be consistent.

👍

datahub/interaction/email_processors/notify.py
```diff
@@ -20,11 +21,21 @@ class Template(Enum):
     meeting_ingest_success = '47418011-3f53-4b39-860a-30969b29781b'
 
 
+def get_domain_label(domain):
```
currycoder: Is it worth making this a core utility e.g. `datahub.core.monitoring_utils.py:get_prometheus_label()`

datahub/interaction/email_processors/notify.py
```diff
@@ -20,11 +21,21 @@ class Template(Enum):
...
+    return domain.replace('.', '_')
+
+
 def notify_meeting_ingest_failure(adviser, errors, recipients):
     """
     Notify an adviser that a meeting ingest has failed - including error
     details and intended recipients.
     """
+    domain_label = get_domain_label(adviser.get_email_domain())
+    statsd.incr(f'celery.calendar-invite-ingest.failure.{domain_label}')
```
currycoder: To double check, these are all just fire and forget UDP requests right? Is there any chance they could block for any significant time?

datahub/interaction/email_processors/notify.py
```diff
@@ -49,6 +60,8 @@ def notify_meeting_ingest_success(adviser, interaction, recipients):
     Notify an adviser that a meeting ingest has succeeeded - including a link
     to the interaction and intended recipients.
     """
+    domain_label = get_domain_label(adviser.get_email_domain())
```
currycoder: I assume that these effectively create a Prometheus counter per adviser email domain? Is it easy to aggregate multiple counters together in graphana et al, with glob syntax? It would be unfortunate if we had to add all the domains in to the aggregation manually.

statsd_mapping.yml
```diff
@@ -0,0 +1,8 @@
+mappings:
```
currycoder: I would like to understand what this is for a bit better... Could it be handy to have a barebones "monitoring in data hub" `docs/` entry? That could explain what the monitoring stack is, a simple tutorial on how to monitor with a counter and links out to further reading on the stasd/Prometheus documentation.

elcct: I agree, we definitely need a document explaining this.

datahub/interaction/email_processors/notify.py
alixedi: These are fire and forget. E.g. they do not raise any exception if you run:

```
pytest datahub/interaction/test/email_processors/test_processors.py
```

without a `statsd-exporter` running.

datahub/interaction/email_processors/notify.py
alixedi: It is creating a single counter. Failure/success and domain are prometheus labels attached to this counter. Aggregating across labels in PromQL is straight-forward.

statsd_mapping.yml
alixedi: At this moment, this is experimental and subject to revision. Happy to put down docs as soon as we have figured out that this is fit for purpose.

alixedi: For the purpose of reviewing this rn, please refer: https://github.com/prometheus/statsd_exporter#metric-mapping-and-configuration

datahub/interaction/email_processors/notify.py
alixedi: We will as soon as this is needed elsewhere.

datahub/company/models/adviser.py
```diff
@@ -121,6 +121,17 @@ def get_current_email(self):
         """
         return self.contact_email or self.email
 
+    def get_email_domain(self):
+        """
+        :returns: Domain for adviser's current email.
+        """
+        email = self.get_current_email()
+        try:
+            _, domain = email.rsplit('@', 1)
```
reupen: Small thing, but it would be good to use `maxsplit=1` as it's a keyword argument [and the examples in the documentation are in this style](https://docs.python.org/3/library/stdtypes.html#str.split).

datahub/interaction/test/email_processors/test_notify.py
```diff
@@ -0,0 +1,62 @@
...
+def test_get_domain_label(domain, label):
+    """
+    Test if the `get_domain_label` function converts
+    a given domain (with "." characters) to a Prometheus
+    label (without "." characters).
+    """
+    assert get_domain_label(domain) == label
+
+
+def test_notify_email_ingest_failure(mock_statsd, monkeypatch, db):
```
reupen: This doesn't seem to be using `monkeypatch` directly, can it be removed?

datahub/interaction/test/email_processors/test_notify.py
```diff
@@ -0,0 +1,62 @@
...
+    right counters in StatsD.
+    """
+    adviser = AdviserFactory(contact_email='adviser@dit.gov.uk')
+    notify_meeting_ingest_failure(adviser, (), ())
+    mock_statsd.incr.assert_called_once_with(
+        f'celery.calendar-invite-ingest.failure.dit_gov_uk',
+    )
+
+
+def test_notify_email_ingest_success(mock_statsd, monkeypatch, db):
```
reupen: This doesn't seem to be using `monkeypatch` directly, can it be removed?

README.md
```diff
@@ -315,6 +315,9 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the
 | `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass\|spf:pass\|dkim:pass |
 | `DATAHUB_NOTIFICATION_API_KEY` | No | The GOVUK notify API key to use for the `datahub.notification` django app. |
 | `DATAHUB_SUPPORT_EMAIL_ADDRESS` | No | Email address for DataHub support team. |
+| `STATSD_HOST` | No | StatsD host url. |
```
reupen: The environment variables were previously sorted alphabetically – it would be good to maintain something consistent.

(Also some of the ones at the end are duplicates – probably from a missed conflict at some point...)

datahub/interaction/test/email_processors/test_notify.py
```diff
@@ -0,0 +1,62 @@
...
+def test_get_domain_label(domain, label):
+    """
+    Test if the `get_domain_label` function converts
+    a given domain (with "." characters) to a Prometheus
+    label (without "." characters).
+    """
+    assert get_domain_label(domain) == label
+
+
+def test_notify_email_ingest_failure(mock_statsd, db):
```
reupen: Elsewhere `pytest.mark.django_db` is used for test functions and [it's what is recommended](https://pytest-django.readthedocs.io/en/latest/helpers.html#db) so it would be good to stick with that (as opposed to the `db` fixture).

config/settings/common.py
```diff
@@ -583,3 +583,7 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
 DNB_SERVICE_TOKEN = env('DNB_SERVICE_TOKEN', default=None)
 
 DATAHUB_SUPPORT_EMAIL_ADDRESS = env('DATAHUB_SUPPORT_EMAIL_ADDRESS', default=None)
+
+STATSD_HOST = env('STATSD_HOST', default='localhost')
+STATSD_PORT = env('STATSD_PORT', default='9125')
+STATSD_PREFIX = env('STATSD_PREFIX', default='datahub')
```
reupen: Should we use a prefix of `datahub-api` in case we start using statsd in another Data Hub service?

PR #1920, opened 2019-08-05
Add API documentation maintenance details

reupen:
### Description of change

This adds a few notes about the API documentation and schema generation in DRF.

(I've left the wording of the PR checklist item a bit loose as we know the generated documentation isn't perfect.)

### Checklist

### DIFF
--- a/docs/Maintaining the API documentation.md
+++ b/docs/Maintaining the API documentation.md
@@ -1,0 +1,32 @@
+# Maintaining the API documentation
+
+We make use of the built-in OpenAPI schema generation feature of Django Rest Framework. The feature is described in the following articles:
+
+- https://www.django-rest-framework.org/topics/documenting-your-api/
+- https://www.django-rest-framework.org/api-guide/schemas/
+
+Swagger UI is served at ``/docs`` and a JSON OpenAPI schema at ``/docs/schema``.  
+
+## Things to look out for when writing API views
+
+The feature in DRF largely works by trying to introspect views. There are a few things to be aware of:
+
+1. During introspection, generic view sets (in our code base, usually inheriting from `CoreViewSet`) will have their `get_serializer()` method (and, hence, `get_serializer_context()`) called without `self.kwargs` being populated. 
+   
+1. Where action methods are added to a generic view set (e.g. the `archive` and `unarchive` methods on many of our view sets), DRF will assume that the view set’s serializer is used for requests and responses.
+    
+    If this is wrong, the simplest thing to do is use the `datahub.core.schemas.StubSchema` to suppress request and response schema generation. See `datahub.omis.order.views.OrderViewSet` for an example.
+
+1. [Generic view sets must specify a serializer class.](https://github.com/encode/django-rest-framework/issues/6535)
+    
+    If the view doesn’t, it’s not a generic view set and so you should use `APIView` instead. 
+
+## Known limitations
+
+1. `NestedRelatedField` is currently rendered as a string. Currently, all the logic pertaining to particular fields lies in DRF’s `AutoSchema` so this can’t be changed without replacing `AutoSchema`.
+
+1. Pagination properties are missing from responses. This is a DRF limitation. 
+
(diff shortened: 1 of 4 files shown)


docs/Maintaining the API documentation.md
```diff
@@ -0,0 +1,22 @@
...
+
+1. For generic view sets (in our code base, typically inheriting from `CoreViewSet`), `get_serializer()` (and, hence, `get_serializer_context()`) will be called without `self.kwargs` being populated. 
+   
+2. For action methods are added to a generic view set (e.g. the `archive` and `unarchive` methods on many of our view sets), DRF will assume that the view set's serializer is used for requests and responses.
+    
+    If this is wrong, the simplest thing to do is use the `datahub.core.schemas.StubSchema` to suppress request and response schema generation. See `datahub.omis.order.views.OrderViewSet` for an example.
+
+3. [Generic view sets must specify a serializer class.](https://github.com/encode/django-rest-framework/issues/6535)
+    
+    If the view doesn't, it's not a generic view set and so you should use `APIView` instead.
```
currycoder: I have a question about views which can't be automatically introspected.  How do we manage documentation for these views?  I realise this is probably out of scope for the initial feature, but it might be nice to call it out here and explain any future path we could take for documenting them (if any).

e.g. Views which proxy to other API endpoints, we often don't have class level serializers for etc.

reupen: I think in the reverse proxy case the only option would be for us to write a DRF schema class that lets us specify a schema (in OpenAPI) directly.

The DRF docs linked at the top cover the feature generally (in particular the second link), so I don't want to replicate those. I'll add some more general points here though.

currycoder: I did have a read through those docs, but I couldn't find a particular example which solved what I was thinking about.  It seemed like the docs were geared towards functionality for guiding the automatic introspection rather than highlighting how developers could specify a schema directly.  I obviously missed it!

It would be great if you could call out the section of the docs that would help here.

docs/Maintaining the API documentation.md
```diff
@@ -0,0 +1,22 @@
...
+- https://www.django-rest-framework.org/topics/documenting-your-api/
+- https://www.django-rest-framework.org/api-guide/schemas/
+
+Swagger UI is served at ``/docs`` and a JSON OpenAPI schema at ``/docs/schema``.  
+
+## Things to look out for
+
+The feature in DRF largely works by trying to introspect views. There are a few things to be aware of:
+
+1. For generic view sets (in our code base, typically inheriting from `CoreViewSet`), `get_serializer()` (and, hence, `get_serializer_context()`) will be called without `self.kwargs` being populated.
```
alixedi: Minor point but as per DRF docs, `Viewset` is one word.

docs/Maintaining the API documentation.md
```diff
@@ -0,0 +1,22 @@
...
+
+Swagger UI is served at ``/docs`` and a JSON OpenAPI schema at ``/docs/schema``.  
+
+## Things to look out for
+
+The feature in DRF largely works by trying to introspect views. There are a few things to be aware of:
+
+1. For generic view sets (in our code base, typically inheriting from `CoreViewSet`), `get_serializer()` (and, hence, `get_serializer_context()`) will be called without `self.kwargs` being populated. 
+   
+2. For action methods are added to a generic view set (e.g. the `archive` and `unarchive` methods on many of our view sets), DRF will assume that the view set's serializer is used for requests and responses.
```
alixedi: ~For~ Action methods are added to... ?

docs/Maintaining the API documentation.md
reupen: Then the class shouldn't be named `ViewSet` :slightly_smiling_face:

docs/Maintaining the API documentation.md
reupen: No, the sentence doesn't make sense as a whole then. I'll try to reword it.

PR #1926, opened 2019-08-06
Fix high memory usage with blank contact email addresses in import interaction tools

reupen:
### Description of change

This fixes a problem that arose in non-development environments when using the import interactions admin site tool.

This occurred when the `contact_email` field was left blank in the CSV file being imported and there are a large number of contacts without email addresses in the database (despite it being a mandatory field).

The problem was twofold:

- the tool was attempting to look up contacts when the `contact_email` was blank
- `QuerySet.get()` was being used to look up contacts and it loads all results into memory despite it raising an exception when multiple matches are found

This fixes the problem at both points of failure so it should not be possible for a similar problem to reoccur in this scenario.

I also added a utility function, `get_queryset_object()` in case we need to do similar query set look-ups elsewhere in a safe fashion, and also as it's a bit easier to check that it's behaving correctly that way. However, the function could be removed (and the code inlined) if we don't think it's adding much.

### Checklist

### DIFF
--- a/datahub/core/query_utils.py
+++ b/datahub/core/query_utils.py
@@ -229,3 +229,21 @@ def get_front_end_url_expression(model_name, pk_expression, url_suffix=''):
         pk_expression,
         Value(url_suffix),
     )
+
+
+def get_queryset_object(queryset, **filters):
+    """
+    Safer version of QuerySet.get() that avoids loading all objects into memory when there
+    are multiple results.
+
+    (When there are multiple results, QuerySet.get() loads all of them into memory even though
+    it raises an exception when multiple matches are found. This can lead to out-of-memory
+    situations when there are a very large number of matches. This function avoids that by
+    applying a LIMIT to the query.)
+
+    TODO: This will be fixed in Django 3.0 and we should be able to remove this function once
+    Django 3.0 has been released and we have updated to it.
+
+    See https://code.djangoproject.com/ticket/6785 for more information.
+    """
+    return queryset.filter(**filters)[:2].get()
--- a/datahub/interaction/admin_csv_import/row_form.py
+++ b/datahub/interaction/admin_csv_import/row_form.py
@@ -214,10 +214,7 @@ class InteractionCSVRowForm(forms.Form):
         if not subject and service:
             data['subject'] = service.name
 
-        # Attempt to look up the contact
-        data['contact'], data['contact_matching_status'] = find_active_contact_by_email_address(
-            data.get('contact_email'),
-        )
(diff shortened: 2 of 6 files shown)


datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -299,6 +296,20 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+        """Attempt to look up the contact using the provided email address."""
+        contact_email = data.get('contact_email')
+
+        if not contact_email:
+            # No contact email address was provided, or it did not pass validation.
+            # Skip the look-up in this case.
+            return
+
+        data['contact'], data['contact_matching_status'] = find_active_contact_by_email_address(
+            data.get('contact_email'),
```
mcbhenwood: _Marginally_ faster to use the local variable `contact_email` here rather than repeat the `.get()` operation.

datahub/core/query_utils.py
```diff
@@ -229,3 +229,15 @@ def get_front_end_url_expression(model_name, pk_expression, url_suffix=''):
...
+def get_queryset_object(queryset, **filters):
+    """
+    Safer version of QuerySet.get() that avoids loading all objects into memory when there
+    are multiple results.
+
+    (When there are multiple results, QuerySet.get() loads all of them into memory even though
+    it raises an exception when multiple matches are found. This can lead to out-of-memory
+    situations when there are a very large number of matches.)
+    """
+    return queryset.filter(**filters)[:2].get()
```
alixedi: Might be useful to put in a comment saying that `[:2]` essentially adds a `LIMIT` clause to the generated SQL.

datahub/interaction/admin_csv_import/row_form.py
reupen: Thanks wasn't intentional, thanks!

datahub/core/query_utils.py
```diff
@@ -229,3 +229,16 @@ def get_front_end_url_expression(model_name, pk_expression, url_suffix=''):
         pk_expression,
         Value(url_suffix),
     )
+
+
+def get_queryset_object(queryset, **filters):
```
currycoder: I've just been digging in to the history of this issue.  
https://code.djangoproject.com/ticket/6785

Looks like this was fixed, unfixed and then fixed again quite recently. I believe that the next minor version of Django should cover it? Could be worth putting a TODO in to remove this when the next version is released?

reupen: Good spot! Will add a note mentioning that.

PR #1928, opened 2019-08-09
Improved notify messages for meeting invite ingest

currycoder:
### Description of change

This PR achieves the following business goals:
- It improves the validation error messages that we notify our users with when there was some problem ingesting their meeting invite.  These are almost always user errors, so the messages should be meaningful enough for them to understand the path to resolution.
- Logs `info` level messages for these errors which should be meaningful to developers reviewing the logs on ELK.
- Certain errors are not sent to users as notifications; errors relating to malformed emails and unverified senders.  This is because we should only notify users that we have verified as Data Hub advisers.

### Checklist


reupen:
Not too fond of wrapping every `raise` by `_log_and_raise_validation_error` (it makes it harder to see what the exit points of the function are, and also confuses the IDE).

What about if we attached the relevant context to the exception and then the exception handler can process it as desired? `ValidationError` supports a code (and params), but I think our own exception class would be more appropriate.

datahub/interaction/email_processors/parsers.py
```diff
@@ -79,6 +99,21 @@ def _convert_calendar_time_to_utc_datetime(calendar_time):
...
+    """
+    Log an email format error as an info level message and raise a ValidationError.
+    The ValidationError can be raised with a user readable string if
+    `user_readable_error` is set, otherwise it will be raised with the `internal_error`.
+    """
+    error_with_message_info = (
+        f'Ingested email with ID "{email.message_id}" (received {email.received[0]["date_utc"]}) '
+        f'was not valid: {internal_error}'
+    )
+    logger.info(error_with_message_info)
```
reupen: I'd make these warnings (if a 404 is a warning, these should be as well, and also they'll be easier to find).

currycoder:
> What about if we attached the relevant context to the exception and then the exception handler can process it as desired? `ValidationError` supports a code (and params), but I think our own exception class would be more appropriate.

This is a neat idea - thanks.  I'll investigate how this can be done.

datahub/interaction/email_processors/parsers.py
```diff
@@ -17,11 +18,30 @@
...
 
 
+logger = get_task_logger(__name__)
+
 ICALENDAR_CONTENT_TYPE = 'application/ics'
 BEGIN_VCALENDAR = 'BEGIN:VCALENDAR'
 CALENDAR_STATUS_CONFIRMED = 'CONFIRMED'
 CALENDAR_COMPONENT_VEVENT = 'VEVENT'
 
+USER_READABLE_ERROR_MESSAGES = {
```
mcbhenwood: I kind of feel that mapping from internal error shorthand into copy which will be sent to a human is a task which ought to be undertaken in a template layer somewhere.

I appreciate this is probably not straightforward since we are using Notify and its templates rather than standard Django templates, but I still feel that the contract between the error code and the human content should live closer to the email creation point than inside parsers.py.

currycoder: Nice observation.  I've broken this out in to `constants.py` for now as this is a well established pattern across DH apps.  I did consider it living in `processors.py` but decided against it for reasons that you have already illustrated.

currycoder:
I've responded to the comments by introducing a new custom exception `InvalidInviteError` which is raised by the parser and handled appropriately by the processor.

That new custom exception takes a message string and `error_code` (from the new `InvalidInviteErrorCode` enum) which groups similar errors under a common set of keys.

I've also added some extra business functionality, to ensure that we do not notify users when the email is malformed or the sender is unverified.  This is because we should only notify users that we have verified as Data Hub advisers.

datahub/interaction/email_processors/parsers.py
```diff
@@ -79,6 +79,19 @@ def _convert_calendar_time_to_utc_datetime(calendar_time):
     return calendar_time.astimezone(utc)
 
 
+class InvalidInviteError(Exception):
+    """
+    A custom exception for when the email invite is not valid.
+    """
+
+    def __init__(self, message, error_code):
```
reupen: Should we declare the class in a similar way to [the example in the Python docs](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions) (specifically in relation to what `__init__()` does)?

datahub/interaction/email_processors/constants.py
```diff
@@ -0,0 +1,31 @@
...
+class InvalidInviteErrorCode(Enum):
+    """
+    Error codes collect the reasons that an invite can be invalid in to groups -
+    the codes are the keys to these groups.
+    """
+
+    malformed_email = 'malformed_email'
+    sender_unverified = 'sender_unverified'
+    no_known_contacts = 'no_known_contacts'
+    bad_calendar_format = 'bad_calendar_format'
```
alixedi: Why can we not have 4 custom exceptions instead of this `InvalidInviteErrorCode`, `USER_READABLE_ERROR_MESSAGES`, `NOTIFIABLE_ERROR_CODES` etc.

It might make the code tidier b/c the exception will contain the entire context of what needs to be done when it is raised instead of a bunch of objects spread around different modules.

currycoder: Hmmm.  Just thinking this through.

To unpack your suggestion, I think what you are describing is something like:

```
class InvalidInviteError(Exception):
    user_readable_error = 'foo bar'
    visible_to_user = False

class InvalidInviteBadCalendarError(Exception):
    user_readable_error = (
        'It looks like you sent something to Data Hub that was not a calendar invitation. '
        'Please send to Data Hub a calendar invitation (not an email or another type of '
        'message). If you still see this error message after sending a calendar email, '
        'please contact the Data Hub support team.'
    )
    visible_to_user = True

...
```

And in `processors.py`:

```
def _handle_invalid_invite(self, exception, message):
        """
        Given an InvalidInviteError and an email message, log the error that
        the user triggered and notify them with an explanation (if applicable).
        """
        error_message = exception.args[0]
        error_with_email_info = (
            f'Ingested email with ID "{message.message_id}" (received '
            f'{message.received[0]["date_utc"]}) was not valid: {error_message}'
        )
        logger.warning(error_with_email_info)
        # Only notify users if the error code is one that we can notify users about
        if exception.visible_to_user:
            self._notify_meeting_ingest_failure(message, [exception.readable_error])
```

I do like some of the elements here, but I'm thinking back to @mcbhenwood's comment
> I still feel that the contract between the error code and the human content should live closer to the email creation point than inside parsers.py 

My problem with it is that we are making our exception classes know about whether a user should see a particular message or not.  As a compromise, I would be tempted to change `NOTIFIABLE_ERROR_CODES` to `NOTIFIABLE_INVALID_INVITE_EXCEPTIONS` and keep that determination within `processors.py`.  What do you think?

reupen: Was this resolved?

We could use different subclasses without tying it specifically to the notifications. So e.g. `VertificationError` for `malformed_email` and `sender_unverified` and `ProcessingError` for `no_known_contacts` and `bad_calendar_format` (just the first names that came to mind without thinking too hard about it).

datahub/interaction/email_processors/constants.py
```diff
@@ -0,0 +1,31 @@
+from enum import Enum
+
+
+class InvalidInviteErrorCode(Enum):
+    """
+    Error codes collect the reasons that an invite can be invalid in to groups -
+    the codes are the keys to these groups.
+    """
+
+    malformed_email = 'malformed_email'
```
reupen: If we keep this Enum then there's some repetition in the names and values that we might be able to avoid (e.g. currently `InvalidInviteErrorCode.malformed_email.name == InvalidInviteErrorCode.malformed_email.value`).

We may be able to use `enum.auto()` for the values instead (and we can use the `name` attribute if we need the codes somewhere).

datahub/interaction/email_processors/constants.py
mcbhenwood: Yes I agree @currycoder that the "should the user see this?" determination should be outside the scope of the code which raises the errors. It's kind of irrelevant at that point.

So yes I like very much the idea of a base exception class from which the specific exceptions are extended, then catching that general base class and comparing the actual exception class with some list of NOTIFIABLE_INVALID_INVITE_EXCEPTIONS. That comparison should take place in the same scope of code which initiates the email / notification call. Then I think all the concerns are separated.

By the same token I think that the human text attributable to each exception should be decided at the point of notification and not within the definition of the exception classes themselves.

datahub/interaction/email_processors/processors.py
```diff
@@ -151,9 +195,9 @@ def process_email(self, message):
         email_parser = CalendarInteractionEmailParser(message)
         try:
             interaction_data = email_parser.extract_interaction_data_from_email()
-        except ValidationError as exc:
-            self._notify_meeting_ingest_failure(message, exc.messages)
-            return (False, exc.message)  # noqa: B306
+        except InvalidInviteError as exc:
```
currycoder: Note: I decided it would be better to wholly depend on a helper method than using separate except statements as this was the most DRY approach.

currycoder:
I've tried to find a compromise in the feedback from what we discussed about where readable error messages/notify logic should live along with using more custom exceptions to better telegraph meaning for exception handling code.

As a result, I've done away with the `constants.py` file and the `InvalidInviteErrorCodes` enum that it contained (phew!).

Note that since we don't have a template layer within our control while using the notify service, the easiest place to keep readable errors and logic determining whether or not a user should be notified is currently `processors.py`.  In the future, we might want to move this over to `notify.py` along with some other related logic but I feel like this wider refactor is out of scope for what I'm trying to achieve in this PR.

datahub/interaction/email_processors/constants.py
currycoder: Agreed - nice suggestion.  Thankfully, it looks like we will drop this enum.

datahub/interaction/email_processors/parsers.py
```diff
@@ -96,15 +101,17 @@ def _extract_and_validate_sender_adviser(self):
...
-            raise ValidationError(
+            raise SenderUnverifiedError(
                 'The meeting email did not pass our minimal checks to be verified as '
                 'having been sent by a valid DIT Adviser email domain.',
             )
         sender_adviser = get_best_match_adviser_by_email(sender_email)
         if not sender_adviser:
-            raise ValidationError('Email not sent by recognised DIT Adviser')
+            raise SenderUnverifiedError(
+                'Email was not sent by a recognised DIT Adviser.',
```
elcct: Why we have two different messages for the same exception? If the exceptions are different maybe we should have a different type for those?

datahub/interaction/email_processors/parsers.py
```diff
@@ -136,27 +142,29 @@ def _extract_secondary_advisers(self, all_recipients, sender_adviser):
...
-            'No calendar event could be extracted - the iCalendar attachment was either '
-            'missing or badly formatted.',
-        )
         calendar_string = _extract_calendar_string_from_text(self.message)
         if not calendar_string:
             calendar_string = _extract_calendar_string_from_attachments(self.message)
         if not calendar_string:
-            raise error
+            raise BadCalendarInviteError(
+                'There was no iCalendar attachment on the email.',
```
elcct: Same as exceptions above - maybe we should have `BadCalendarInviteBadAttachmentError` and `BadCalendarInviteBadFormattingError` etc.

datahub/interaction/email_processors/parsers.py
currycoder: Using separate exception classes was a way to group together similar issues under a single class which can be acted on programmatically.

In this case, the error handling code doesn't care what the specific issue was - it just cares that the adviser was unverified or the calendar format was bad etc.  It can then make a determination about whether the user can be told about the problem and what simplified message to use to tell them about it.

We don't need the granularity of one exception per scenario and my feeling is that adding these exceptions would add to the noise and complicate the error handling without giving us a defined benefit at the moment.  We could bear this in mind for future iterations or if the requirements change in terms of how much detail we need to share with our users.

datahub/interaction/email_processors/processors.py
```diff
@@ -20,6 +24,29 @@
...
+    NoContactsError: (
+        'The calendar invitation you sent to Data Hub did not include an email address '
+        'that is recognised as a contact on Data Hub. The invitation must also be sent '
+        'to Data Hub at the same time (and not forwarded to Data Hub later). Please '
+        'check both of these things when resending your invitation and if you still '
+        'see this message after resending the invitation, contact the Data Hub support '
+        'team.'
+    ),
+}
+NOTIFIABLE_EXCEPTIONS = (BadCalendarInviteError, NoContactsError)
```
alixedi: If we rename `READABLE_ERROR_MESSAGES` to `EXCEPTION_MESSAGES` we will not need this b/c:

```
exc in EXCEPTION_MESSAGES
```

datahub/interaction/email_processors/processors.py
alixedi: Perhaps following is a better way handling exceptions instead of `_handle_invalid_invite`:

```
def _extract_interaction_from_email(email_paresr):
    try:
        interaction_data = email_parser.extract_interaction_data_from_email()
    except InvalidInviteError as exc:
        # handle the exception..
        pass
```

Essentially encapsulate the function that is raising the exception in one that catches and handles it.

datahub/interaction/email_processors/parsers.py
elcct: I was about to write that it's maybe not needed this way but had to go to a meeting. I agree with you. Thanks!

datahub/interaction/email_processors/processors.py
```diff
@@ -151,9 +195,9 @@ def process_email(self, message):
         email_parser = CalendarInteractionEmailParser(message)
         try:
             interaction_data = email_parser.extract_interaction_data_from_email()
-        except ValidationError as exc:
-            self._notify_meeting_ingest_failure(message, exc.messages)
-            return (False, exc.message)  # noqa: B306
+        except InvalidInviteError as exc:
+            self._handle_invalid_invite(exc, message)
+            return (False, exc.args[0])
```
reupen: If we use `exc.args[0]` like this then it will fail if a message isn't passed into the exception (which doesn't seem to be enforced anywhere). So we should probably enforce that (and I think avoiding using the `args` attribute would be good as well).

datahub/interaction/email_processors/processors.py
reupen: I agree with this. If the two things need to diverge for some reason, that can be added when they actually do diverge.

datahub/interaction/email_processors/processors.py
currycoder: I agree with that approach, but it gets a bit complicated when we need to transform the exception in to an early return value in `process_email()`.

I think this loops back to your earlier comment in a previous PR that we should change the contract for email processors to raise exceptions for exceptional cases rather than returning a pair tuple of `(succeeded, message)`.  I'd still like to dedicate some time to this but my feeling is that it's out of scope for this PR.

datahub/interaction/email_processors/processors.py
currycoder: Agreed!

PR #1930, opened 2019-08-09
Add additional Elasticsearch sub-fields to mapping types

reupen:
### Description of change

This adds additional sub-fields to our Elasticsearch mapping types. This is part of the migration away from (unnecessarily) using the `copy_to` mapping parameter (and [using multi-fields instead](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/multi-fields.html)) that was started some time ago.

These fields represent the last few places `copy_to` was still being used. Once all environments have been migrated to the new mapping types, we can update search queries to use the new sub-fields and remove the `copy_to` usage and target fields.

This is the most recent similar PR: https://github.com/uktrade/data-hub-leeloo/pull/1595

### Checklist

### DIFF
--- a/datahub/search/event/models.py
+++ b/datahub/search/event/models.py
@@ -15,7 +15,9 @@ class Event(BaseESModel):
     address_2 = Text()
     address_town = fields.NormalizedKeyword()
     address_county = fields.NormalizedKeyword()
-    address_postcode = Text(copy_to='address_postcode_trigram')
+    # TODO: Update queries to use the trigram sub-field (once indexed) and remove
+    #  address_postcode_trigram
+    address_postcode = fields.TextWithTrigram(copy_to='address_postcode_trigram')
     address_postcode_trigram = fields.TrigramText()
     address_country = fields.id_name_partial_field()
     created_on = Date()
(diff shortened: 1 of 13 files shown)


datahub/search/event/models.py
```diff
@@ -15,7 +15,9 @@ class Event(BaseESModel):
     address_2 = Text()
     address_town = fields.NormalizedKeyword()
     address_county = fields.NormalizedKeyword()
-    address_postcode = Text(copy_to='address_postcode_trigram')
+    # TODO: Update queries to use the trigram sub-field (once indexed) and remove
+    #  address_postcode_trigram
+    address_postcode = fields.TextWithTrigram(copy_to='address_postcode_trigram')
```
currycoder: I want to double check my understanding here.  `copy_to` will copy the values of all sub-fields in to a single de-normalised field?  In this case, there's only one sub-field for `TextWithTrigram`.

reupen: No, `copy_to` will add the original value to the target field. It won't use what comes out following analysis, so the sub-field should have no bearing on that.

currycoder: Gotcha.  Thanks.

PR #1939, opened 2019-08-13
Add service_answers support to interaction upload tool

elcct:
### Description of change

This adds support of `service_answers` field for interaction upload tool.

At the moment we only have one question at most that requires only one answer, so that this can be provided with a `service_answer` column in the spreadsheet.

If we get to have more questions, we could extend this to have `service_answer_1`, `service_answer_2` and so on. 

If the service answer is not found, we should get an error `A service answer could not be found with the specified name and service.`.

Service answer is only required when a service that has questions is provided in the CSV file.

Provided answer once found valid it is then transformed into a dictionary like below:

```
{
  <question_id_str>: {
    <answer_id_str>: {}
  }
}
```

the above dictionary is then stored in the Interaction's JSONField `service_answers`.

### Checklist

### DIFF
--- a/datahub/interaction/admin_csv_import/row_form.py
+++ b/datahub/interaction/admin_csv_import/row_form.py
@@ -21,7 +21,12 @@ from datahub.event.models import Event
 from datahub.interaction.admin_csv_import.duplicate_checking import (
     is_duplicate_of_existing_interaction,
 )
-from datahub.interaction.models import CommunicationChannel, Interaction, InteractionDITParticipant
+from datahub.interaction.models import (
+    CommunicationChannel,
+    Interaction,
+    InteractionDITParticipant,
+    ServiceAnswerOption,
+)
 from datahub.interaction.serializers import InteractionSerializer
 from datahub.metadata.models import Service, Team
 from datahub.metadata.query_utils import get_service_name_subquery
@@ -47,6 +52,15 @@ DUPLICATE_OF_EXISTING_INTERACTION_MESSAGE = gettext_lazy(
 DUPLICATE_OF_ANOTHER_ROW_MESSAGE = gettext_lazy(
     'This interaction appears to be a duplicate as there is another row in this file with the '
     'same service, date and contact.',
+)
+SERVICE_ANSWER_NOT_FOUND = gettext_lazy(
+    'A service answer could not be found with the specified name and service.',
+)
+SERVICE_ANSWER_NOT_REQUIRED = gettext_lazy(
+    'A service answer was provided when the selected service does not require it.',
+)
+SERVICE_ANSWER_REQUIRED = gettext_lazy(
+    'A service answer is required for the specified service.',
 )
 
 
@@ -146,6 +160,8 @@ class InteractionCSVRowForm(forms.Form):
         to_field_name='name__iexact',
         validators=[_validate_not_disabled],
     )
+    service_answer = forms.CharField(required=False)
+
     communication_channel = NoDuplicatesModelChoiceField(
         CommunicationChannel.objects.all(),
         to_field_name='name__iexact',
@@ -199,6 +215,7 @@ class InteractionCSVRowForm(forms.Form):
         kind = data.get('kind')
         subject = data.get('subject')
         service = data.get('service')
+        self._populate_service_answers(data)
 
         # Ignore communication channel for service deliveries (as it is not a valid field for
         # service deliveries, but we are likely to get it in provided data anyway)
@@ -296,6 +313,55 @@ class InteractionCSVRowForm(forms.Form):
         except ValidationError as exc:
             self.add_error(adviser_field, exc)
 
+    def _populate_service_answers(self, data):
+        """Transform service_answer into service_answers dictionary."""
+        service = data.get('service')
+        if not service:
+            return
+
+        service_answer = data.get('service_answer')
+
+        if not service.interaction_questions.exists():
+            if service_answer:
+                self.add_error(
+                    'service_answer',
+                    ValidationError(
+                        SERVICE_ANSWER_NOT_REQUIRED,
+                        code='service_answer_not_required',
+                    ),
+                )
+            # if service has no questions and answer is not provided, there is nothing to do
+            return
+
+        if not service_answer:
+            self.add_error(
+                'service_answer',
+                ValidationError(
+                    SERVICE_ANSWER_REQUIRED,
+                    code='service_answer_required',
+                ),
+            )
+            return
+
+        try:
+            service_answer_option_db = ServiceAnswerOption.objects.get(
+                name__iexact=service_answer,
+                question__service=service,
+            )
+            data['service_answers'] = {
+                str(service_answer_option_db.question.pk): {
+                    str(service_answer_option_db.pk): {},
+                },
+            }
+        except ServiceAnswerOption.DoesNotExist:
+            self.add_error(
+                'service_answer',
+                ValidationError(
+                    SERVICE_ANSWER_NOT_FOUND,
+                    code='service_answer_not_found',
+                ),
+            )
+
     @staticmethod
     def _populate_contact(data):
         """Attempt to look up the contact using the provided email address."""
@@ -365,6 +431,7 @@ class InteractionCSVRowForm(forms.Form):
             'kind': data['kind'],
             'notes': data.get('notes'),
             'service': data['service'],
+            'service_answers': data.get('service_answers'),
             'status': Interaction.STATUSES.complete,
             'subject': subject,
             'theme': data['theme'],
--- a/datahub/interaction/test/admin_csv_import/test_row_form.py
+++ b/datahub/interaction/test/admin_csv_import/test_row_form.py
@@ -29,6 +29,7 @@ from datahub.interaction.test.admin_csv_import.utils import random_communication
 from datahub.interaction.test.factories import (
     CommunicationChannelFactory,
     CompanyInteractionFactory,
+    ServiceAnswerOptionFactory,
 )
 from datahub.interaction.test.utils import random_service
 from datahub.metadata.test.factories import ChildServiceFactory, ServiceFactory, TeamFactory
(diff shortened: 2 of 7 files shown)


datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -218,7 +229,6 @@ def clean(self):
 
         self._validate_not_duplicate_of_prior_row(data)
         self._validate_not_duplicate_of_existing_interaction(data)
-
```
reupen: Was this change intended?

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +306,32 @@ def _populate_adviser(self, data, adviser_field, team_field):
         except ValidationError as exc:
             self.add_error(adviser_field, exc)
 
+    def _populate_service_answers(self, data):
+        """Transform service_answer into service_answers dictionary."""
+        service = data.get('service')
+        if service and service.interaction_questions.count() > 0:
```
reupen: Shouldn't we also return an error if a `service_answer` is provided but the service has no service answers?

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -370,12 +406,17 @@ def cleaned_data_as_serializer_dict(self):
...
         }
+        if data.get('service_answers'):
+            creation_data['service_answers'] = data['service_answers']
 
         if data['kind'] == Interaction.KINDS.service_delivery:
             creation_data['is_event'] = bool(data.get('event_id'))
 
         return creation_data
 
+    def _validate_service_answer(self, data, service_answer_field):
```
reupen: Is this intended to be here?

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -323,6 +325,20 @@ class TestInteractionCSVRowFormValidation:
                 },
                 id='cannot specify event_id for an interaction',
             ),
+            pytest.param(
+                {
+                    'kind': Interaction.KINDS.interaction,
+                    'service':
```
reupen: I think it would be better to avoid having a dependency on the service and service answer fixtures in these tests. 

It would be good to follow the same pattern as the other tests using factories – there is a similar dependency between the `adviser_1` and `team_1` fields so it should be possible to do it in a similar way.

datahub/interaction/admin_csv_import/row_form.py
elcct: Oh no, reverted. Thanks!

datahub/interaction/admin_csv_import/row_form.py
elcct: I forgot to delete that one, thank you!

datahub/interaction/admin_csv_import/row_form.py
elcct: Good idea - updated. Thanks!

datahub/interaction/test/admin_csv_import/test_row_form.py
elcct: I agree. I have removed the dependency now. Thank you!

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +310,41 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+        service_answer = data.get('service_answer')
+
+        if service and service.interaction_questions.count() > 0:
+
+            try:
+                service_answer_option_db = ServiceAnswerOption.objects.get(
+                    name=service_answer,
+                    question__service__pk=service.pk,
+                )
+                data['service_answers'] = {
```
currycoder: Could it be good to call this something distinct rather than overwriting the existing `data['service_answers']`?

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +310,41 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
             self.add_error(adviser_field, exc)
 
+    def _populate_service_answers(self, data):
+        """Transform service_answer into service_answers dictionary."""
+        service = data.get('service')
+        service_answer = data.get('service_answer')
+
+        if service and service.interaction_questions.count() > 0:
+
+            try:
```
currycoder: It might be helpful for readability to break the body of this `if` out in to another helper method.  There's a fair amount of indentation.

datahub/interaction/admin_csv_import/row_form.py
currycoder: Oh wait this is a new key!  Is it plural because you are anticipating multiple answers in the future?

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +310,41 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+
+            except ServiceAnswerOption.DoesNotExist:
+                self.add_error(
+                    'service_answer',
+                    ValidationError(
+                        SERVICE_ANSWER_NOT_FOUND,
+                        code='service_answer_not_found',
+                    ),
+                )
+        elif service_answer:
```
currycoder: Might be nice to have a comment here like "We have a service answer, but there was no service or the service did not require an answer"

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +310,41 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+                    question__service__pk=service.pk,
+                )
+                data['service_answers'] = {
+                    str(service_answer_option_db.question.pk): {
+                        str(service_answer_option_db.pk): {},
+                    },
+                }
+
+            except ServiceAnswerOption.DoesNotExist:
+                self.add_error(
```
currycoder: My understanding is that this will also cover the case where no service answer was given, but it was required for the service.  It might be good to call that situation out with a separate error message.

datahub/interaction/admin_csv_import/row_form.py
elcct: It is plural on the Interaction model, because it can hold multiple questions and answers and in the CSV file we support only one, so the name is singular.

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -1326,6 +1385,56 @@ def test_save_rolls_back_on_error(self, monkeypatch):
...
+            'service_answer': service_answer.name,
+            'communication_channel': communication_channel.name,
+        }
+        form = InteractionCSVRowForm(data=data)
+
+        assert form.is_valid()
+        interaction = form.save(user, source=source)
+        interaction.refresh_from_db()
+
+        assert interaction.theme == data['theme']
```
reupen: We could probably make this a little bit more specific and just check the relevant service answer attributes.

Also would be good to move the test up a tiny bit (near the other similar tests).

datahub/interaction/admin_csv_import/row_form.py
elcct: I have reorganised the logic so now it should be more readable now.

datahub/interaction/admin_csv_import/row_form.py
elcct: I have reorganised it a bit so maybe now the comment will not be needed?

datahub/interaction/admin_csv_import/row_form.py
elcct: I have added this case. Thanks!

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +310,41 @@ def _populate_adviser(self, data, adviser_field, team_field):
         except ValidationError as exc:
             self.add_error(adviser_field, exc)
 
+    def _populate_service_answers(self, data):
+        """Transform service_answer into service_answers dictionary."""
+        service = data.get('service')
+        service_answer = data.get('service_answer')
+
+        if service and service.interaction_questions.count() > 0:
```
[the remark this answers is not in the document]
elcct: This is a good idea so I reorganised this method a bit using your suggestions. Thank you!

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +310,41 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+
+            try:
+                service_answer_option_db = ServiceAnswerOption.objects.get(
+                    name=service_answer,
+                    question__service__pk=service.pk,
```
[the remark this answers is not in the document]
elcct: Good point! Fixed.

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -370,6 +419,8 @@ def cleaned_data_as_serializer_dict(self):
             'theme': data['theme'],
             'was_policy_feedback_provided': False,
         }
+        if data.get('service_answers'):
+            creation_data['service_answers'] = data['service_answers']
```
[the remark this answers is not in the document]
elcct: I think it was breaking few tests if without the `if` and I didn't want to change those at the time, then I forgot to remove it. It is now updated. Thank you!

datahub/interaction/test/admin_csv_import/test_row_form.py
elcct: That's a good suggestion. I moved the test a little and removed irrelevant checks. Thank you!

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +313,54 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+    def _populate_service_answers(self, data):
+        """Transform service_answer into service_answers dictionary."""
+        service = data.get('service')
+        if not service:
+            return
+
+        service_answer = data.get('service_answer')
+
+        if service.interaction_questions.count() == 0:
+            if service_answer:
```
reupen: Would be good to combine these two if statements e.g. `if service_answer and not service.interaction_questions.exists()`.

datahub/interaction/admin_csv_import/row_form.py
```diff
@@ -296,6 +313,54 @@ def _populate_adviser(self, data, adviser_field, team_field):
...
+                ValidationError(
+                    SERVICE_ANSWER_REQUIRED,
+                    code='service_answer_required',
+                ),
+            )
+            return
+
+        try:
+            service_answer_option_db = ServiceAnswerOption.objects.get(
+                name=service_answer,
```
reupen: Would be good to use `name__iexact` to make it case-insensitive (in line with most of the other fields). (Plus add tests for case sensitivity...)

datahub/interaction/test/admin_csv_import/test_row_form.py
```diff
@@ -1016,6 +1058,36 @@ def test_contact_lookup(self, input_email, matching_status, match_on_alternative
         else:
             assert not contact
 
+    @pytest.mark.parametrize(
+        'service_answer',
```
reupen: Is this parameter used?

datahub/interaction/admin_csv_import/row_form.py
elcct: Good point. I changed it so now it is not case sensitive and I also adjusted the test.

datahub/interaction/test/admin_csv_import/test_row_form.py
elcct: It wasn't, but now it is being used for checking case sensitivity.

datahub/interaction/admin_csv_import/row_form.py
elcct: Maybe the indent in the return is not too visible, but this is testing two things - if there is no questions, then we should exit early, but in case there is actually an answer provided, that is an error. If I combine the two, I would have to check again if there is no questions and then exit?

reupen: Oh, yes, sorry, my mistake. Maybe it could deal with the return before adding the error to make it a bit more obvious?

datahub/interaction/test/admin_csv_import/test_row_form.py
reupen: It would be good to not use `service_answer` twice in the function – maybe the argument could be renamed `service_answer_name`.

datahub/interaction/admin_csv_import/row_form.py
reupen: Although I guess you'd have two returns then. Maybe just a comment before the return...

elcct: I added a comment. Thank you!

datahub/interaction/test/admin_csv_import/test_row_form.py
elcct: Yes! Nice one. I have renamed it now. Thank you.

PR #1943, opened 2019-08-15
Add company-create endpoint

alixedi:
### Description of change

This adds an endpoint that creates a company using data from the `dnb_service` given a `duns_number`.

Following PRs would primarily deal with extracting more data from the DNB response to create richer company records in DataHub.

### Checklist


datahub/dnb_api/views.py
```diff
@@ -28,29 +30,62 @@ def post(self, request):
         """
         Proxy for POST requests.
         """
-        upstream_response = self._get_upstream_response(request)
+        upstream_response = search_dnb(request)
```
currycoder: I think it would be handy to rebase here as things have gotten a bit more complex in this view.  I don't think it stops you doing what you are doing with a utility though.

datahub/dnb_api/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+api_client = APIClient(
+    settings.DNB_SERVICE_BASE_URL,
+    TokenAuth(settings.DNB_SERVICE_TOKEN),
+    raise_for_status=False,
+)
+
+
+def search_dnb(request):
```
currycoder: I feel like changing away from using `request` here will help in a few ways.  
- It makes the method more reusable e.g. if we had a management command that needed to search DNB for whatever reason.
- It avoids the "read the stream once" problem that you described in our conversation earlier.

I would be tempted to make it `def search_dnb(search_criteria):` or similar.  I realise that this change would water down the "search as a proxy" approach that we currently have, but I think it's OK for DH API to expose a DNB search endpoint which is a little opinionated about how search is called on `dnb-service`. e.g. forcing a specific request method, content-type etc.

What do you think?

datahub/dnb_api/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        request.method,
+        'companies/search/',
+        data=request.body,
+        headers={
+            'Content-Type': request.content_type,
+        },
+    )
+
+
+def format_dnb_response(dnb_response):
```
currycoder: Small thing, but `dnb_response` could be `dnb_company_result` or similar

datahub/dnb_api/views.py
```diff
@@ -28,29 +30,62 @@ def post(self, request):
...
+            return JsonResponse(
+                {'error': error_message},
+                status=400,
+                content_type='application/json',
+            )
+        # If the dnb-service returns multiple companies for a duns_number
+        # we are going to create a company from the first one. This should
+        # not happen.
+        if not dnb_response['results']:
+            warning_message = 'Multiple companies returned for duns_number:{duns_number}'
```
currycoder: I would be tempted to just fail early here with a 400 and meaningful error message rather than assuming that the user would want the first result.  Since there's currently no checking on the request body, somebody could call this with `{"search_term": "sainsbury", "address_country": "GB"}` and it would create a DH company for the first (of many) results.

currycoder: My opinion might change if we had some input validation before this to assert that the user is only searching with `duns_number`.

datahub/dnb_api/views.py
```diff
@@ -28,29 +30,62 @@ def post(self, request):
...
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
+
+    @method_decorator(feature_flagged_view(FEATURE_FLAG_DNB_COMPANY_SEARCH))
+    @method_decorator(enforce_request_content_type('application/json'))
+    def post(self, request):
+        """
+        Proxy for POST requests.
+        """
+        dnb_response = search_dnb(request).json()
```
currycoder: I think in general, my feeling is that the "Create DH company from DNB DUNS number" API endpoint is not a proxy in the same way that our search endpoint was.

My mental model is that this endpoint should be a Data Hub endpoint which happens to call the DNB API as a source of data - in the same way that we might interrogate the DB for data in other Data Hub APIs.  

I think the "proxy-first" approach here is a little problematic.

alixedi:
TY for the early feedback @currycoder 

I would do the following:

1. Rebase.
2. Refactor `search_dnb` to take the query parameters instead of a request.
3. Re-introduce `duns_number` validation in the company-create view.
4. Fail when/if `dnb-service` returns multiple results for a `duns_number`.

datahub/dnb_api/views.py
```diff
@@ -123,22 +125,83 @@ def _format_and_hydrate(self, dnb_results):
...
+
+        if serializer.is_valid():
+            datahub_company = serializer.save()
+            return JsonResponse(
+                serializer.to_representation(datahub_company),
+                status=200,
+                content_type='application/json',
+            )
+
+        return JsonResponse(
```
reupen: This doesn't seem to be covered according to the coverage report. Are we able to add a test to cover it?

datahub/dnb_api/utils.py
```diff
@@ -0,0 +1,95 @@
...
+    except KeyError as exc:
+        raise InvalidDNBCompany(exc)
+
+    # Replace country code with ID
+    country = Country.objects.filter(
+        iso_alpha2_code=company['address']['country'],
+    ).first()
+
+    if country is None:
+        raise InvalidDNBCompany(
```
reupen: This doesn't seem to be covered according to the coverage report. Are we able to add a test to cover it?

datahub/company/serializers.py
```diff
@@ -572,3 +572,49 @@ class OneListCoreTeamMemberSerializer(serializers.Serializer):
...
+
+class DNBCompanySerializer(CompanySerializer):
+    """
+    For creating a company from DNB data.
+
+    We would like to keep the validators defined in the CompanySerializer
+    but we want to make the DNB fields writable.
+    """
+
+    class Meta:
```
reupen: If we don't inherit from the Meta class of the base class, won't we be missing the validators etc.?

alixedi: Initially, I thought it was a good idea to inherit the validators from `CompanySerializer.Meta` but there are a few e.g. one that makes `uk_region` required that are not going to work here.

datahub/dnb_api/utils.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+
+def search_dnb(query_params):
+    """
+    Queries the dnb-service with the given query_params
+    """
+    if not settings.DNB_SERVICE_BASE_URL:
+        raise ImproperlyConfigured('The setting DNB_SERVICE_BASE_URL has not been set')
+    return api_client.request(
```
currycoder: Neat!

datahub/dnb_api/utils.py
```diff
@@ -0,0 +1,95 @@
...
+from django.core.exceptions import ImproperlyConfigured
+
+from datahub.core.api_client import APIClient, TokenAuth
+from datahub.metadata.models import Country
+
+
+api_client = APIClient(
+    settings.DNB_SERVICE_BASE_URL,
+    TokenAuth(settings.DNB_SERVICE_TOKEN),
+    raise_for_status=False,
```
reupen: We probably do want `raise_for_status` on for the requests the new endpoint is making though (as it's not reverse proxying).

I'm not sure how it's handled in APIClient is ideal. I think `raise_for_status` could be moved from `APIClient.__init__()` to `APIClient.request()`, or it could follow the same pattern as timeout (with `default_timeout` and `timeout`).

Also, we should be setting `default_timeout` (otherwise it will not be using a timeout and could wait for a very long time).

datahub/dnb_api/utils.py
```diff
@@ -0,0 +1,95 @@
...
+class InvalidDNBCompany(Exception):
+    """
+    Exception raised when format_dnb_company runs into
+    issues.
+    """
+
+
+def search_dnb(query_params):
+    """
+    Queries the dnb-service with the given query_params
```
currycoder: Could it be worth noting some common query parameters for this purpose?  So that future callers have some hint about how this work or know where to go to find out more.

changelog/dnb-company-create.feature.rst
```diff
@@ -0,0 +1 @@
+An endpoint was added for creating companies through dnb-service (https://github.com/uktrade/dnb-service/) given a ``duns_number``.
```
currycoder: Could there also be an `dnb-company-create.api.rst` to explain the contract for the API endpoint?

datahub/dnb_api/test/test_views.py
```diff
@@ -333,3 +331,220 @@ def test_post_permissions(
...
+            data={
+                'duns_number': 123456,
+            },
+        )
+
+        assert response.status_code == 200
+
+        company = response.json()
+        dnb_company = dnb_response_non_uk['results'].pop()
+        self._assert_companies_same(company, dnb_company)
```
currycoder: Could it be worth checking the company saved in the DB also..?

datahub/dnb_api/views.py
currycoder: Thought for a future iteration: it might be nice to chuck in the number of companies created through DNB as a statsd metric?

datahub/company/serializers.py
reupen: OK, the docstring for the class above does need updating then (and a TODO adding). It definitely should be rectified before the feature flag is removed so it would be good to explicitly call that out.

datahub/dnb_api/test/test_views.py
alixedi: I am checking whether the record was created on the line below. I can also check individual fields but I am assuming that this would be duplicating the tests for the payload that is returned.

currycoder: I think in practice that's probably fine.  My thought was to catch the case where we respond with e.g. data from an instantiated serializer which has not been persisted to the DB.

alixedi: I have set `raise_exception=True` on `is_valid` so it should be good now.

datahub/dnb_api/utils.py
alixedi: Yes, moving `raise_for_status` to request is a good idea. 

ATM, I am dealing with non-200 responses explicitly in the view.

datahub/company/serializers.py
alixedi: I figured that none of the validators in `CompanySerializer` apply here at least in the context of this PR b/c we are not unpacking things like `uk_region`, `business_type` etc. 

That said, I have tried to move all the validation logic that is needed into the serialiser by e.g. declaring `name` field as `required=True`.

config/settings/common.py
```diff
@@ -590,6 +590,7 @@ def _add_hawk_credentials(id_env_name, key_env_name, scope):
 
 DNB_SERVICE_BASE_URL = env('DNB_SERVICE_BASE_URL', default=None)
 DNB_SERVICE_TOKEN = env('DNB_SERVICE_TOKEN', default=None)
+DNB_SERVICE_TIMEOUT = 15
```
elcct: Should we use an environment variable?

alixedi: I did then while putting it in README it occurred to me that I will probably not change this.

Also, 2/3 `_TIMEOUT`s in here don't use environment variables. Probably for the same reason.

datahub/dnb_api/views.py
```diff
@@ -123,22 +126,62 @@ def _format_and_hydrate(self, dnb_results):
...
         )
-        return api_client.request(
-            request.method,
-            '',
-            data=request.body,
-            headers={
-                'Content-Type': request.content_type,
-            },
+
+        if serializer.is_valid(raise_exception=True):
```
reupen: The `if` shouldn't be needed here as it'll raise an error if it fails.

datahub/dnb_api/views.py
```diff
@@ -123,22 +125,75 @@ def _format_and_hydrate(self, dnb_results):
...
+    @method_decorator(feature_flagged_view(FEATURE_FLAG_DNB_COMPANY_SEARCH))
+    @method_decorator(enforce_request_content_type('application/json'))
+    def post(self, request):
+        """
+        Given a duns_number, get the data for the company from dnb-service
+        and create a record in DataHub.
+        """
+        data = request.data
+        try:
+            duns_number = data['duns_number']
```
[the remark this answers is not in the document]
reupen: This comment is still relevant.

config/settings/common.py
elcct: Fair enough. It seems like if that was really need we could do a separate PR and expose all of those parameters.

datahub/company/serializers.py
```diff
@@ -578,43 +579,69 @@ class DNBCompanySerializer(CompanySerializer):
...
-            'company_number',
-
-            'number_of_employees',
-            'is_number_of_employees_estimated',
-
-            'turnover',
-            'is_turnover_estimated',
-
-            'website',
+    The validators would be put back in when we have done the work for
```
currycoder: Maybe prefix this particular sentence with `TODO:` ?

datahub/company/serializers.py
```diff
@@ -578,43 +579,69 @@ class DNBCompanySerializer(CompanySerializer):
...
-            'uk_region',
-            'sector',
-            'vat_number',
-            'global_headquarters',
-            'headquarter_type',
+    class Meta(CompanySerializer.Meta):
+        read_only_fields = []
+        dnb_read_only_fields = []
+        validators = (
+            RulesBasedValidator(
```
currycoder: As a super simple fix to avoid the repetition of this good logic, could this just be broken in to a `common_validators` attribute on the parent class or similar?

datahub/company/serializers.py
```diff
@@ -578,43 +579,69 @@ class DNBCompanySerializer(CompanySerializer):
...
+    """
+    Parses duns_number from request body and validates format.
+    """
+
+    duns_number = serializers.CharField(
+        required=True,
+        write_only=True,
+        max_length=9,
+        min_length=9,
+        validators=(integer_validator,),
```
currycoder: Can duns numbers have leading 0's?  Does integer_validator handle this case..?

datahub/company/serializers.py
alixedi: For the record, I do not like how we have shoehorned multiple contexts in `InteractionSerializer` but I do not have an elegant solution that uses inheritance in this case. 

Could have done something clever but decided to go non-DRY b/c we will turn all the validators on in due time.

datahub/dnb_api/views.py
```diff
@@ -135,73 +137,48 @@ class DNBCompanyCreateView(APIView):
...
+        dnb_companies = dnb_response.json().get('results', [])
 
         if not dnb_companies:
             error_message = f'Cannot find a company with duns_number: {duns_number}'
             logger.error(error_message)
-            return JsonResponse(
-                {'error': error_message},
-                status=400,
-            )
+            raise APIUpstreamException(error_message)
```
currycoder: Should this be a 502 or a 400..?  I think returning a 502 assumes that DNB (or dnb-service) is at fault, whereas for this case I would wager that most cases are because the DNB number provided by the caller is wrong.

datahub/company/serializers.py
reupen: I agree.

datahub/company/serializers.py
```diff
@@ -572,3 +573,75 @@ class OneListCoreTeamMemberSerializer(serializers.Serializer):
...
+        }
+
+
+class DUNSNumberSerializer(serializers.Serializer):
+    """
+    Parses duns_number from request body and validates format.
+    """
+
+    duns_number = serializers.CharField(
+        required=True,
```
reupen: `required=True` is the default, so we could omit that here.

datahub/dnb_api/test/conftest.py
```diff
@@ -0,0 +1,178 @@
...
+
+@pytest.fixture()
+def dnb_company_search_feature_flag(db):
+    """
+    Creates the dnb company search feature flag.
+    """
+    yield FeatureFlagFactory(code=FEATURE_FLAG_DNB_COMPANY_SEARCH)
+
+
+@pytest.fixture
```
reupen: It would be good to be consistent about `@pytest.fixture` vs `@pytest.fixture()` (at least in the same module) :slightly_smiling_face: .

datahub/dnb_api/test/conftest.py
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+
+@pytest.fixture()
+def dnb_company_search_datahub_companies():
+    """
+    Creates Data Hub companies for hydrating DNB search results with.
+    """
+    # Company with no interactions
+    CompanyFactory(duns_number='1234567', id='6083b732-b07a-42d6-ada4-c8082293285b')
```
reupen: Is there a reason for specifying this ID (and the ones below) explicitly?

datahub/dnb_api/test/test_views.py
```diff
@@ -1,66 +1,138 @@
...
-def dnb_company_search_feature_flag():
+class TestDNBAPICommon(APITestMixin):
     """
-    Creates the dnb company search feature flag.
+    Test common functionality in company-search as well
+    as company-create endpoints.
     """
-    yield FeatureFlagFactory(code=FEATURE_FLAG_DNB_COMPANY_SEARCH)
 
+    @pytest.mark.parametrize(
```
reupen: Could apply the parametrisation to the class here (to avoid repeating it for each method).

datahub/dnb_api/test/test_views.py
```diff
@@ -333,3 +323,322 @@ def test_post_permissions(
...
+        self,
+        requests_mock,
+        dnb_company_search_feature_flag,
+        dnb_response_non_uk,
+    ):
+        """
+        Test create-company endpoint for a non-uk company.
+        """
+        requests_mock.post(
+            settings.DNB_SERVICE_BASE_URL + 'companies/search/',
```
reupen: There is a bit of repetition with `settings.DNB_SERVICE_BASE_URL + 'companies/search/'` and the `reverse()` calls – maybe some named constants would help avoid that?

datahub/company/serializers.py
alixedi: I thought about it. Here is why I didn't do it:

In following PRs for this thread of work, we will start unpacking fields such as `business_type` & `uk_region` from DNB data. As things stand currently, I will have to simply add the necessary validation & remove the override when its the same as the parent class.

On the other hand, if I put in a `common_validators` attribute in the parent class, bringing in fields such as `business_type` and `uk_region` would mean I will be modifying the parent class.

IMO we can an argument for both approaches. Do let me know if you are easy on this, if not, lets discuss IRL.

datahub/dnb_api/views.py
```diff
@@ -123,22 +127,58 @@ def _format_and_hydrate(self, dnb_results):
...
+    def post(self, request):
+        """
+        Given a duns_number, get the data for the company from dnb-service
+        and create a record in DataHub.
+        """
+        duns_serializer = DUNSNumberSerializer(data=request.data)
+        duns_serializer.is_valid(raise_exception=True)
+        duns_number = duns_serializer.validated_data['duns_number']
+
+        if Company.objects.filter(duns_number=duns_number).exists():
```
reupen: Could consider adding a `validate_duns_number()` method to `DUNSNumberSerializer` for this.

datahub/company/serializers.py
alixedi: If I understand your Q correctly, the answer is that the validator does not mutate values.

datahub/dnb_api/views.py
alixedi: Good point. 400 it is 👍

datahub/dnb_api/views.py
```diff
@@ -123,22 +127,58 @@ def _format_and_hydrate(self, dnb_results):
...
+            error_message = f'Cannot find a company with duns_number: {duns_number}'
+            logger.error(error_message)
+            raise APIUpstreamException(error_message)
+
+        if len(dnb_companies) > 1:
+            error_message = f'Multiple companies found with duns_number: {duns_number}'
+            logger.error(error_message)
+            raise APIUpstreamException(error_message)
+
+        serializer = DNBCompanySerializer(
```
reupen: Since there is another serialiser above, it would be good to add something to the variable name to explicitly disambiguate them.

datahub/dnb_api/views.py
```diff
@@ -123,22 +127,58 @@ def _format_and_hydrate(self, dnb_results):
...
+            data=format_dnb_company(
+                dnb_companies.pop(),
```
reupen: Would consider `dnb_companies[0]` here since we know the length of `dnb_companies` and it will avoid mutating it (better for debugging, for example).

datahub/dnb_api/views.py
```diff
@@ -123,22 +127,58 @@ def _format_and_hydrate(self, dnb_results):
...
-            search_endpoint,
-            shared_key_auth,
-            raise_for_status=False,
+class DNBCompanyCreateView(APIView):
+    """
+    View for creating datahub company from DNB data.
+    """
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (IsAuthenticatedOrTokenHasScope,)
```
reupen: It should also check that the user has the permission required to create companies (since it creates companies).

(There is a `HasPermissions` permission class that you could use for that.)

datahub/company/serializers.py
currycoder: I think it's cool if we agree that the repetition will be tackled in some way in a subsequent PR.

datahub/dnb_api/test/conftest.py
alixedi: I will change it but this only shows up in the diff because I moved it to `conftest`.

datahub/dnb_api/test/conftest.py
alixedi: Again, this only shows up in the diff because it has been moved around.

datahub/dnb_api/views.py
alixedi: I don't get why it will help debugging b/c we are not referencing `dab_companies` after this line.

datahub/dnb_api/views.py
alixedi: Added the permission as well as a test.

datahub/dnb_api/views.py
reupen: That was just an example, but if you have a breakpoint on the next line you won't be able to look at `dnb_companies` and `company_serializer` or `datahub_company` at the same time.

I'm not seeing a particular reason to prefer mutation here, perhaps you could explain?

datahub/dnb_api/test/test_views.py
```diff
@@ -1,66 +1,121 @@
...
+from datahub.company.models import Company, CompanyPermission
 from datahub.company.test.factories import CompanyFactory
 from datahub.core.test_utils import APITestMixin, create_test_user
-from datahub.dnb_api.constants import FEATURE_FLAG_DNB_COMPANY_SEARCH
-from datahub.feature_flag.test.factories import FeatureFlagFactory
 from datahub.interaction.models import InteractionPermission
-from datahub.interaction.test.factories import CompanyInteractionFactory
+from datahub.metadata.models import Country
 
+DNB_SEARCH_URL = urljoin(settings.DNB_SERVICE_BASE_URL, 'companies/search/')
```
reupen: Would be good if we did this join in a consistent way across the code base – the way here seems to be the safest way: https://github.com/uktrade/data-hub-leeloo/blob/69a942d2de336b22c0e49ebe00eb893625f619c8/datahub/dnb_api/views.py#L130

reupen:
@alixedi did you see this comment?

> Also can't see any logic around created_by and modified_by – would be good to add tests for those and update the logic as needed

datahub/dnb_api/test/test_views.py
alixedi: I have fixed it but again the subject of this PR is not company search.

alixedi:
> @alixedi did you see this comment?
> 
> > Also can't see any logic around created_by and modified_by – would be good to add tests for those and update the logic as needed

Missed it. I've put something in now.

datahub/dnb_api/views.py
alixedi: It's easier for me to change it tbh but I do not buy the debugging argument.

PR #1945, opened 2019-08-15
Update company list models in preparation for multiple lists functionality

reupen:
### Description of change

This:

- adds a `CompanyList` model (representing a container of list items)
- adds a nullable `list` foreign key to the `CompanyListItem` model

This is in preparation for there being multiple company lists per user.

The intention is to make the existing company list endpoints operate on a specific list for each user and so these changes have been made with that in mind.

The next steps will be:

1. Update `PUT /v4/user/company-list/<company ID>` to set additionally the list attribute. (It will need to create a default legacy list for the user if one doesn't already exist).
2. Release that.
3. Add a data migration to populate `CompanyListItem.list` for the remaining objects that don't have it set and make `CompanyListItem.list` non-nullable.
4. Change the logic in the existing company list endpoints to use `CompanyListItem.list.adviser` instead of `CompanyListItem.adviser`
5. Another release.
6. Remove `CompanyListItem.adviser` (using the usual process)

### Checklist


changelog/adviser/add-company-list-model.db.rst
```diff
@@ -0,0 +1,12 @@
+A ``company_list_companylist`` table with the following columns was created:
+
+- ``"created_on" timestamp with time zone NULL``
+- ``"modified_on" timestamp with time zone NULL``
+- ``"id" uuid NOT NULL PRIMARY KEY``
+- ``"name" varchar(255) NOT NULL``
+- ``"legacy_default" boolean NOT NULL``
```
mcbhenwood: Migration code suggests this is called `is_legacy_default` (unless the "is_" gets dropped by some ORM process I don't know about)

datahub/user/company_list/models.py
```diff
@@ -1,11 +1,46 @@
...
+    )
+
+    def __str__(self):
+        """Human-friendly representation."""
+        return f'{self.name} – {self.adviser}'
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=('adviser', 'is_legacy_default'),
```
mcbhenwood: Won't this limit the number of lists an adviser can have to two? One with `True`, one with `False`?

changelog/adviser/add-company-list-model.db.rst
reupen: Thanks, will update the news fragment :+1:

datahub/user/company_list/models.py
reupen: Yes, good point. I'll update it to use the condition instead.

reupen: Actually, it would probably be a good idea to disallow `is_legacy_default=False` until we've filled in all the lists on the `CompanyListItem`s. I'll add another constraint for that for now.

PR #1946, opened 2019-08-16
Associate items added to company lists with a default list

reupen:
### Description of change

This makes sure that any new items added to a company list are associated with a default list for each user. This list is created if it does not already exist.

This is in preparation for there being multiple company lists per user.

The next steps will be:

1. Do a release.
2. Add a data migration to populate `CompanyListItem.list` for the remaining objects that don't have it set and make `CompanyListItem.list` non-nullable.
3. Change the logic in the existing company list endpoints to use `CompanyListItem.list`, and to not use `CompanyListItem.adviser` when filtering
4. Another release.
5. Tidy up.

### Checklist

### DIFF
--- a/datahub/user/company_list/tests/test_views.py
+++ b/datahub/user/company_list/tests/test_views.py
@@ -15,9 +15,12 @@ from datahub.company.test.factories import ArchivedCompanyFactory, CompanyFactor
 from datahub.core.test_utils import APITestMixin, create_test_user, format_date_or_datetime
 from datahub.interaction.test.factories import CompanyInteractionFactory
 from datahub.metadata.test.factories import TeamFactory
-from datahub.user.company_list.models import CompanyListItem
-from datahub.user.company_list.tests.factories import CompanyListItemFactory
-from datahub.user.company_list.views import CANT_ADD_ARCHIVED_COMPANY_MESSAGE
+from datahub.user.company_list.models import CompanyList, CompanyListItem
+from datahub.user.company_list.tests.factories import CompanyListFactory, CompanyListItemFactory
+from datahub.user.company_list.views import (
+    CANT_ADD_ARCHIVED_COMPANY_MESSAGE,
+    DEFAULT_LEGACY_LIST_NAME,
+)
 
 
 def company_with_interactions_factory(num_interactions):
@@ -256,7 +259,7 @@ class TestHeadCompanyListItemView(APITestMixin):
 class TestCreateOrUpdateCompanyListItemView(APITestMixin):
     """Tests for the PUT method in CompanyListItemView."""
 
-    def test_with_new_item(self):
+    def test_creates_new_items(self):
         """Test that a company can be added to the authenticated user's list."""
         company = CompanyFactory()
         url = reverse('api-v4:company-list:item', kwargs={'company_pk': company.pk})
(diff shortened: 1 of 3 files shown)


datahub/user/company_list/tests/test_views.py
```diff
@@ -256,15 +259,63 @@ def test_with_non_existent_company(self):
...
         url = reverse('api-v4:company-list:item', kwargs={'company_pk': company.pk})
         response = self.api_client.put(url)
 
         assert response.status_code == status.HTTP_204_NO_CONTENT
         assert response.content == b''
-        assert CompanyListItem.objects.filter(adviser=self.user, company=company).exists()
+        list_item = CompanyListItem.objects.filter(adviser=self.user, company=company).first()
+
+        assert list_item
+        assert list_item.created_by == self.user
```
elcct: I know we don't particularly care for `created_on` and `modified_on` at this moment, but since it is there maybe would be good to check it works? (not necessarily here)

datahub/user/company_list/views.py
```diff
@@ -124,9 +133,20 @@ def put(self, request, format=None, company_pk=None):
             }
             raise serializers.ValidationError(errors)
 
-        # update_or_create() is used to avoid an error if there is an existing
+        adviser = request.user
+        default_legacy_list = _get_default_list_for_user(adviser)
```
elcct: Maybe it would be good to mark it with a `# TODO: ` so we remember this will be gone in the future?

datahub/user/company_list/tests/test_views.py
reupen: I think it's OK; those are handled by Django and if we do want to test it we should do it on BaseModel rather than every subclass.

datahub/user/company_list/views.py
reupen: This will stay, but these endpoints will be removed eventually :)

datahub/user/company_list/tests/test_views.py
```diff
@@ -256,15 +259,63 @@ def test_with_non_existent_company(self):
...
+        assert list_item.modified_by == self.user
+
+    def test_creates_a_legacy_list_when_adding_a_new_item(self):
+        """
+        Test that when a company is added, a legacy default list is created when it didn't
```
mcbhenwood: "Test when a company is added _to a list_" - for added clarity?

datahub/user/company_list/tests/test_views.py
```diff
@@ -256,15 +259,63 @@ def test_with_non_existent_company(self):
...
+        assert list_.is_legacy_default
+        assert list_.name == DEFAULT_LEGACY_LIST_NAME
+        assert list_.adviser == self.user
+        assert list_.created_by == self.user
+        assert list_.modified_by == self.user
+
+        assert CompanyList.objects.count() == len(unrelated_lists) + 1
+
+    def test_reuses_an_existing_legacy_list_when_adding_a_new_item(self):
+        """Test that when a company is added, an existing legacy default list is reused."""
```
mcbhenwood: again maybe add "is added _to a list_" I say this bc if someone is grepping the code for `/company.*add/` which they might conceivably do, they might assume that these tests are doing something else.

datahub/user/company_list/tests/test_views.py
reupen: I've reworded it :+1:

PR #1948, opened 2019-08-16
Adapt notification app to be able to use multiple notify API credentials

currycoder:
### Description of change

This lays the initial groundwork to allow us to refactor the `datahub.omis` app to use our new `datahub.notification` app for sending GOVUK notify notifications.

Specifically, this makes the `datahub.notification` app able to work with multiple GOVUK notify API keys.  This will allow us to use the functions defined in `datahub/notification/notify.py` with multiple different notify credentials which are identified by a `notify_service_name` keyword argument.

For now the different notify service names/API keys are collated in `datahub/notification/constants.py` under the `NOTIFY_KEYS` variable.  The ambition is to move this in to a django setting rather than using the current `DATAHUB_NOTIFICATION_API_KEY` and `OMIS_NOTIFICATION_API_KEY` settings.

I didn't want this groundwork to change any settings for the project as I want to keep this light touch until we have proven that the new notification app works for omis with functionality switched between the new app and omis' notification package with a feature flag.

### Rollout plan

- This work to allow the notification app to work with multiple notify API keys.
- Deploy.
- Switch omis to optionally use `datahub.notification` under the hood for sending notifications - functionality determined by feature flag.
- Deploy. Activate feature flag progressively across environments and test.
- Refactor omis' notification integration to exclusively use `datahub.notification` including cleanup and decoupling.  Remove feature flag.
- Deploy.

### Checklist

### DIFF
--- a/datahub/notification/constants.py
+++ b/datahub/notification/constants.py
@@ -11,7 +11,7 @@ class NotifyServiceName(StrEnum):
 DEFAULT_SERVICE_NAME = NotifyServiceName.datahub
 
 # TODO: This module should not know or care about the specifics of our notify
-# service instances.  This should be moved up in to a setting.  To ensure easy
+# service instances. This should be moved up in to a setting. To ensure easy
 # rollback, we will turn this on once OMIS integration no longer uses a feature
 # flag.
 NOTIFY_KEYS = {
--- a/datahub/notification/constants.py
+++ b/datahub/notification/constants.py
@@ -1,13 +1,20 @@
-DATAHUB_SERVICE_NAME = 'datahub'
-OMIS_SERVICE_NAME = 'omis'
+from datahub.core.utils import StrEnum
 
-DEFAULT_SERVICE_NAME = DATAHUB_SERVICE_NAME
+
+class NotifyServiceName(StrEnum):
+    """Notify service name constants."""
+
+    datahub = 'datahub'
+    omis = 'omis'
+
+
+DEFAULT_SERVICE_NAME = NotifyServiceName.datahub
 
 # TODO: This module should not know or care about the specifics of our notify
 # service instances.  This should be moved up in to a setting.  To ensure easy
 # rollback, we will turn this on once OMIS integration no longer uses a feature
 # flag.
 NOTIFY_KEYS = {
-    DATAHUB_SERVICE_NAME: 'DATAHUB_NOTIFICATION_API_KEY',
-    OMIS_SERVICE_NAME: 'OMIS_NOTIFICATION_API_KEY',
+    NotifyServiceName.datahub: 'DATAHUB_NOTIFICATION_API_KEY',
+    NotifyServiceName.omis: 'OMIS_NOTIFICATION_API_KEY',
 }
--- a/datahub/notification/constants.py
+++ b/datahub/notification/constants.py
@@ -1,0 +1,13 @@
+DATAHUB_SERVICE_NAME = 'datahub'
+OMIS_SERVICE_NAME = 'omis'
+
+DEFAULT_SERVICE_NAME = DATAHUB_SERVICE_NAME
+
+# TODO: This module should not know or care about the specifics of our notify
+# service instances.  This should be moved up in to a setting.  To ensure easy
+# rollback, we will turn this on once OMIS integration no longer uses a feature
+# flag.
+NOTIFY_KEYS = {
+    DATAHUB_SERVICE_NAME: 'DATAHUB_NOTIFICATION_API_KEY',
+    OMIS_SERVICE_NAME: 'OMIS_NOTIFICATION_API_KEY',
+}
--- a/datahub/notification/core.py
+++ b/datahub/notification/core.py
@@ -25,6 +25,13 @@ class NotifyGateway:
             if api_key:
                 clients[service_name] = NotificationsAPIClient(api_key)
             else:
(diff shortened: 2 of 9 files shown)


datahub/notification/core.py
```diff
@@ -4,46 +4,63 @@
...
+        for service_name, api_key_setting_name in NOTIFY_KEYS.items():
+            api_key = getattr(settings, api_key_setting_name, None)
+            if api_key:
+                clients[service_name] = NotificationsAPIClient(api_key)
+            else:
+                client = mock.Mock(spec_set=NotificationsAPIClient)
+                client.send_email_notification.return_value = {'id': 'abc123'}
+                clients[service_name] = client
+                warnings.warn(
+                    f'`settings.{api_key_setting_name}` not specified therefore '
```
reupen: Would be good to review the line wrapping here.

datahub/notification/constants.py
```diff
@@ -0,0 +1,13 @@
+DATAHUB_SERVICE_NAME = 'datahub'
```
reupen: Could be a candidate for an enum? We have a `StrEnum` which might work well.

datahub/notification/core.py
```diff
@@ -4,46 +4,63 @@
...
-            )
+        """
+        Init underlying notification client(s).
+        """
+        self._initialise_clients()
+
+    def _initialise_clients(self):
+        clients = {}
+        for service_name, api_key_setting_name in NOTIFY_KEYS.items():
+            api_key = getattr(settings, api_key_setting_name, None)
```
reupen: I think we can avoid the default of `None` since the setting should at least always be present (and it will help avoid typos).

datahub/notification/notify.py
```diff
@@ -1,24 +1,29 @@
...
+def notify_by_email(email_address, template_identifier, context, notify_service_name=None):
+    """
+    Notify an email address, using a GOVUK notify template and some template
+    context.
+    """
     send_email_notification.apply_async(
-        args=(contact.email, template_identifier),
-        kwargs={'context': context},
+        args=(email_address, template_identifier),
+        kwargs={'context': context, 'notify_service_name': notify_service_name},
```
reupen: If we pass `notify_service_name` now, there is a small chance that there could be an error during deployment, because there is a small window in which an old Celery worker could pick up tasks sent by the new app.

So, if we want to be 100% safe, we should first add optional support for `notify_service_name` to the task, release and deploy that, and then start using the new keyword argument.

I think we can do that by just not passing `notify_service_name` here if it's `None`.

datahub/notification/constants.py
```diff
@@ -0,0 +1,13 @@
...
+OMIS_SERVICE_NAME = 'omis'
+
+DEFAULT_SERVICE_NAME = DATAHUB_SERVICE_NAME
+
+# TODO: This module should not know or care about the specifics of our notify
+# service instances.  This should be moved up in to a setting.  To ensure easy
```
elcct: The formatting looks odd here.

datahub/notification/core.py
```diff
@@ -4,46 +4,63 @@
...
+            if api_key:
+                clients[service_name] = NotificationsAPIClient(api_key)
+            else:
+                client = mock.Mock(spec_set=NotificationsAPIClient)
```
elcct: Should this belong in tests?

datahub/notification/core.py
```diff
@@ -4,46 +4,63 @@
...
         recipient_email,
         template_identifier,
         context=None,
+        notify_service_name=None,
     ):
         """
         Send an email notification using the GOVUK notification service.
         """
+        # TODO: the default notify service name should be in a setting, not a constant.
+        # This will be fixed when we fully move over omis notifications from its
```
mcbhenwood: "omis" as an acronym ought to be in uppercase

datahub/notification/notify.py
currycoder: Great observation Reupen - thanks.

datahub/notification/constants.py
currycoder: Can you explain how I can change it?

datahub/notification/core.py
currycoder: This is something I borrowed from the existing OMIS notify implementation.  I was a little sceptical of it at first as well, but I've discovered that it's quite a light-touch way to avoid some pain.

The mock allows us to do the following:
1) We do not need to set notify API keys on dev copies, as the mock will gracefully handle this situation for us.
2) We do not need to build any conditional code in to the notification core, to prevent calling a non-existent client object when a notification call is made.
3) The majority of our notification code can operate on a single path of execution that assumes a client object can be called to make a notification - we avoid adding extra complexity to the app to handle the case where it doesn't exist for dev.

Happy to discuss further of course!

elcct: I'd think of creating a dummy client to avoid an issue with the keys, but I guess the Mock does exactly that then. In that case I think it would be great to add a comment there to explain why mock is being used, so in the future someone working on it could get the context of why it is getting done this way much easier.

currycoder: Good idea yeah!

datahub/notification/constants.py
elcct: Sorry I should have explained - it has two spaces after a period.

currycoder: Ah :)

PR #2000, opened 2019-08-22
Fix minor invite notify issues

currycoder:
### Description of change

This PR does the following:
- Stops the meeting invite processing logic from sending error notifications when a meeting is cancelled.  This was some unexpected behaviour as most clients send cancellations in iCalendar format and the processing logic was handling this as if it was a new meeting request.
This was achieved by adding a new exception `UnconfirmedCalendarInviteError` - since this is not present in `NOTIFY_EXCEPTIONS`, advisers will not be notified about it.
- Prevents the notification celery task from retrying API calls that came back with a 400/403 error - these are raised as exceptions immediately.

### Checklist

### DIFF
--- a/datahub/notification/tasks.py
+++ b/datahub/notification/tasks.py
@@ -1,16 +1,16 @@
 from celery import shared_task
+from notifications_python_client.errors import HTTPError
 
 from datahub.notification.core import notify_gateway
 
 
 @shared_task(
+    bind=True,
     acks_late=True,
     priority=9,
-    max_retries=5,
-    autoretry_for=(Exception,),
-    retry_backoff=60,
 )
 def send_email_notification(
+    self,
     recipient_email,
     template_identifier,
     context=None,
@@ -20,10 +20,16 @@ def send_email_notification(
     Celery task to call the notify API to send a templated email notification
     to an email address.
     """
-    response = notify_gateway.send_email_notification(
-        recipient_email,
-        template_identifier,
-        context,
-        notify_service_name,
-    )
+    try:
+        response = notify_gateway.send_email_notification(
+            recipient_email,
+            template_identifier,
+            context,
+            notify_service_name,
+        )
+    except HTTPError as exc:
+        # Raise 400/403 responses without retry
+        if exc.status_code in (400, 403):
+            raise exc
+        raise self.retry(exc=exc, max_retries=5, countdown=60)
     return response['id']
--- a/datahub/notification/tasks.py
+++ b/datahub/notification/tasks.py
@@ -8,6 +8,7 @@ from datahub.notification.core import notify_gateway
     bind=True,
     acks_late=True,
     priority=9,
+    max_retries=5,
 )
 def send_email_notification(
     self,
@@ -28,8 +29,10 @@ def send_email_notification(
             notify_service_name,
         )
     except HTTPError as exc:
-        # Raise 400/403 responses without retry
+        # Raise 400/403 responses without retry - these are problems with the
+        # way we are calling the notify service and retries will not result in
+        # a successful outcome.
         if exc.status_code in (400, 403):
-            raise exc
-        raise self.retry(exc=exc, max_retries=5, countdown=60)
+            raise
+        raise self.retry(exc=exc, countdown=60)
     return response['id']
--- a/datahub/notification/test/test_tasks.py
+++ b/datahub/notification/test/test_tasks.py
@@ -1,11 +1,14 @@
+from unittest import mock
+
 import pytest
+from celery.exceptions import Retry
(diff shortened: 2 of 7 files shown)


datahub/notification/tasks.py
```diff
@@ -19,5 +19,11 @@ def send_email_notification(
     Celery task to call the notify API to send a templated email notification
     to an email address.
     """
-    response = client.send_email_notification(recipient_email, template_identifier, context)
+    try:
+        response = client.send_email_notification(recipient_email, template_identifier, context)
+    except HTTPError as exc:
+        # Raise 400/403 responses without retry
+        if exc.status_code in (400, 403):
+            raise exc
```
reupen: Just `raise` would be good here (to maintain the original context).

datahub/notification/tasks.py
```diff
@@ -19,5 +19,11 @@ def send_email_notification(
...
+        raise self.retry(exc=exc, max_retries=5, countdown=60)
```
reupen: It's still possible to specify `default_retry_delay` and `max_retries` on the task, so I guess there's a question of which style we want to follow. I think at least `max_retries` is more logical on the task.

datahub/notification/test/test_tasks.py
```diff
@@ -1,10 +1,15 @@
...
+
 import pytest
+from notifications_python_client.errors import HTTPError
 
 from datahub.notification import client
 from datahub.notification.tasks import send_email_notification
 
 
-@pytest.mark.django_db
+pytest.mark.django_db
```
reupen: Can this be removed?

datahub/notification/tasks.py
```diff
@@ -19,5 +19,11 @@ def send_email_notification(
     Celery task to call the notify API to send a templated email notification
     to an email address.
     """
-    response = client.send_email_notification(recipient_email, template_identifier, context)
+    try:
+        response = client.send_email_notification(recipient_email, template_identifier, context)
+    except HTTPError as exc:
+        # Raise 400/403 responses without retry
```
reupen: If we have a comment here it would be good to explain the intent.

datahub/notification/test/test_tasks.py
```diff
@@ -18,3 +20,41 @@ def test_send_email_notification():
...
+        (500, True),
+        (403, False),
+        (400, False),
+    ),
+)
+def test_send_email_notification_retries_errors(monkeypatch, error_status_code, expect_retry):
+    """
+    Test the send_email_notification utility.
+    """
+    notification_api_client = client.client
```
alixedi: This maybe something for a future PR but `client.client` doesn't read well.

datahub/notification/tasks.py
```diff
@@ -19,5 +20,13 @@ def send_email_notification(
     Celery task to call the notify API to send a templated email notification
     to an email address.
     """
-    response = client.send_email_notification(recipient_email, template_identifier, context)
+    try:
+        response = client.send_email_notification(recipient_email, template_identifier, context)
+    except HTTPError as exc:
+        # Raise 400/403 responses without retry - these are problems with the
+        # way we are calling the notify service and retryies will not result in
```
alixedi: Typo: `s/retryies/retrying/` perhaps

datahub/notification/test/test_tasks.py
alixedi: This maybe something for a future PR but `client.client` doesn't read well.

datahub/notification/test/test_tasks.py
```diff
@@ -18,3 +20,41 @@ def test_send_email_notification():
...
+    mock_response.status_code = error_status_code
+    mock_response.json.return_value = {}
+    error = HTTPError(mock_response)
+    notification_api_client.send_email_notification.side_effect = error
+
+    # Mock the task's retry method
+    retry_mock = mock.Mock(side_effect=MockRetryException())
+    monkeypatch.setattr('datahub.notification.tasks.send_email_notification.retry', retry_mock)
+
+    with pytest.raises((MockRetryException, HTTPError)):
```
alixedi: Might be worth giving [this](https://docs.celeryproject.org/en/latest/userguide/testing.html#tasks-and-unit-tests) a quick read.

E.g. 

1. It seems we can use `celery.exceptions.Retry` without having to `MockRetryException`.
2. I think its cleaner to use the patch decorator or context manager and set the side-effect to be `celery.exceptions.Retry` instead of `retry_mock = mock.Mock(side_effect=..)`.

currycoder: Thanks Ali.  I've improved this a bit according to your guidance here - I've still used the `monkeypatch` fixture as opposed to the patch decorator though.

PR #1991, opened 2019-08-20
Api endpoint for OMIS dataset

yusufertekin:
### Description of change
Adds new app datasets
Adds an api endpoint for OMIS dataset under datasets app (/v4/datasets/omis-dataset)
Adds new HAWK credentials for authenticating with data-flow-api
 Adds new HAWK scope for data-flow-api
### Checklist

* [ ] Has a new newsfragment been created? Check [changelog/README.rst](https://github.com/uktrade/data-hub-leeloo/blob/master/changelog/README.rst) for instructions
* [x ] Do any added or updated endpoints appear in the API documentation? See [docs/Maintaining the API documentation.md](https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Maintaining&#32;the&#32;API&#32;documentation.md) for more details
* [ ] Have any relevant search models been updated?
* [ ] Have any relevant fixtures (`fixtures/test_data.yaml`) been updated?
* [ ] Have any relevant select-/prefetch-related field lists in the views and search apps been updated?
* [ ] Has the admin site been updated (for new models, fields etc.)?
* [ ] Has the README been updated (if needed)?

PR #2001, opened 2019-08-23
Add API (V4) endpoint for fetching OMIS Dataset

yusufertekin:
### Description of change
New PR which addresses comments on and honors branch name convention, rebased develop
https://github.com/uktrade/data-hub-leeloo/pull/1991

Adds new app dataset
Adds an api endpoint for OMIS dataset under datasets app (/v4/datasets/omis-dataset)
Adds new HAWK credentials for authenticating with data-flow-api
Adds new HAWK scope for data-flow-api

### Checklist

### DIFF
--- a/changelog/omis/omis-dataset-api-endpoint.api.rst
+++ b/changelog/omis/omis-dataset-api-endpoint.api.rst
@@ -1,0 +1,2 @@
+The following endpoint was added:
+- ``GET /v4/dataset/omis-dataset``: Present required fields data of all orders to be consumed by data-flow and used in data-workspace for reporting and analyst access.
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -92,6 +92,7 @@ LOCAL_APPS = [
     'datahub.activity_stream.apps.ActivityStreamConfig',
     'datahub.user_event_log',
     'datahub.activity_feed',
+    'datahub.dataset',
 
     # TODO: delete after the whole data cleansing piece of work is complete
     'datahub.dnb_match',
@@ -526,6 +527,12 @@ _add_hawk_credentials(
     'MARKET_ACCESS_ACCESS_KEY_ID',
     'MARKET_ACCESS_SECRET_ACCESS_KEY',
     HawkScope.public_company,
+)
+
+_add_hawk_credentials(
+    'DATA_FLOW_API_ACCESS_KEY_ID',
+    'DATA_FLOW_API_SECRET_ACCESS_KEY',
+    HawkScope.data_flow_api,
 )
 
 # To read data from Activity Stream
--- a/datahub/dataset/pagination.py
+++ b/datahub/dataset/pagination.py
@@ -1,0 +1,9 @@
+from rest_framework.pagination import CursorPagination
+
+
+class OMISDatasetViewCursorPagination(CursorPagination):
+    """
+    Cursor Pagination for OMISDatasetView
+    """
+
+    ordering = ('created_on', 'pk')
--- a/datahub/dataset/test/test_views.py
+++ b/datahub/dataset/test/test_views.py
@@ -1,0 +1,168 @@
+import pytest
+from django.conf import settings
+from freezegun import freeze_time
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.core.test_utils import (
+    format_date_or_datetime,
+    get_attr_or_none,
+    HawkAPITestClient,
+    join_attr_values,
+)
+from datahub.omis.order.test.factories import (
+    OrderCancelledFactory,
+    OrderCompleteFactory,
+    OrderFactory,
+    OrderPaidFactory,
+    OrderWithAcceptedQuoteFactory,
+    OrderWithCancelledQuoteFactory,
+    OrderWithOpenQuoteFactory,
+    OrderWithoutAssigneesFactory,
+    OrderWithoutLeadAssigneeFactory,
+)
+
+
+@pytest.fixture
+def hawk_api_client():
+    """Hawk API client fixture."""
+    yield HawkAPITestClient()
+
+
+@pytest.fixture
+def data_flow_api_client(hawk_api_client):
+    """Hawk API client fixture configured to use credentials with the data_flow_api scope."""
+    hawk_api_client.set_credentials(
+        'data-flow-api-id',
+        'data-flow-api-key',
+    )
+    yield hawk_api_client
+
+
+@pytest.fixture
+def omis_dataset_view_url():
+    """Returns OMISDatasetView url"""
+    yield reverse('api-v4:dataset:omis-dataset')
+
+
+def get_expected_data_from_order(order):
+    """Returns expected dictionary based on given order"""
+    return {
+        'delivery_date': str(order.delivery_date),
+        'services': join_attr_values(order.service_types.order_by('name')),
+        'status': order.status,
+        'reference': order.reference,
+        'subtotal_cost': order.subtotal_cost,
+        'invoice__subtotal_cost': get_attr_or_none(order, 'invoice.subtotal_cost'),
+        'company__name': order.company.name,
+        'created_by__dit_team__name': order.created_by.dit_team.name,
+        'primary_market__name': order.primary_market.name,
+        'created_on': (format_date_or_datetime(order.created_on)
+                       if order.created_on else None),
+        'cancelled_on': (format_date_or_datetime(order.cancelled_on)
+                         if order.cancelled_on else None),
+        'cancellation_reason__name': get_attr_or_none(order, 'cancellation_reason.name'),
+        'uk_region__name': order.uk_region.name,
+        'sector__segment': order.sector.segment,
+        'paid_on': (format_date_or_datetime(order.paid_on)
+                    if order.paid_on else None),
+        'completed_on': (format_date_or_datetime(order.completed_on)
+                         if order.completed_on else None),
+        'contact__first_name': order.contact.first_name,
+        'contact__last_name': order.contact.last_name,
+        'contact__telephone_number': order.contact.telephone_number,
+        'contact__email': order.contact.email,
+        'company__address_1': order.company.address_1,
+        'company__address_2': order.company.address_2,
+        'company__address_town': order.company.address_town,
+        'company__address_county': order.company.address_county,
+        'company__address_country__name': order.company.address_country.name,
+        'company__address_postcode': order.company.address_postcode,
+        'company__registered_address_1': order.company.registered_address_1,
+        'company__registered_address_2': order.company.registered_address_2,
+        'company__registered_address_town': order.company.registered_address_town,
+        'company__registered_address_county': order.company.registered_address_county,
+        'company__registered_address_country__name': order.company.registered_address_country.name,
+        'company__registered_address_postcode': order.company.registered_address_postcode,
+    }
+
+
+@pytest.mark.django_db
+class TestOMISDatasetViewSet:
+    """
+    Tests for OMISDatasetView
+    """
+
+    @pytest.mark.parametrize('method', ('delete', 'patch', 'post', 'put'))
+    def test_other_methods_not_allowed(self, data_flow_api_client, method, omis_dataset_view_url):
+        """Test that various HTTP methods are not allowed."""
+        response = data_flow_api_client.request(method, omis_dataset_view_url)
+        assert response.status_code == status.HTTP_405_METHOD_NOT_ALLOWED
+
+    def test_without_scope(self, hawk_api_client, omis_dataset_view_url):
+        """Test that making a request without the correct Hawk scope returns an error."""
+        hawk_api_client.set_credentials(
+            'test-id-without-scope',
+            'test-key-without-scope',
+        )
+        response = hawk_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_without_credentials(self, api_client, omis_dataset_view_url):
+        """Test that making a request without credentials returns an error."""
+        response = api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    @pytest.mark.parametrize(
+        'order_factory', (
+            OrderFactory,
+            OrderCompleteFactory,
+            OrderCancelledFactory,
+            OrderPaidFactory,
+            OrderWithAcceptedQuoteFactory,
+            OrderWithCancelledQuoteFactory,
+            OrderWithOpenQuoteFactory,
+            OrderWithoutAssigneesFactory,
+            OrderWithoutLeadAssigneeFactory,
+        ))
+    def test_success(self, data_flow_api_client, omis_dataset_view_url, order_factory):
+        """Test that endpoint returns with expected data for a single order"""
+        order = order_factory()
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 1
+        result = response_results[0]
+        expected_result = get_expected_data_from_order(order)
+        assert result == expected_result
+
+    def test_with_multiple_orders(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that endpoint returns correct number of record in expected order"""
+        with freeze_time('2019-01-01 12:30:00'):
+            order_1 = OrderFactory()
+        with freeze_time('2019-01-03 12:00:00'):
+            order_2 = OrderFactory()
+        with freeze_time('2019-01-01 12:00:00'):
+            order_3 = OrderFactory()
+            order_4 = OrderFactory()
+
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 4
+        expected_order_list = sorted([order_3, order_4],
+                                     key=lambda item: item.pk) + [order_1, order_2]
+        for index, order in enumerate(expected_order_list):
+            assert order.reference == response_results[index]['reference']
+
+    def test_pagination(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that when page size higher than threshold response returns with next page url"""
+        OrderFactory.create_batch(settings.REST_FRAMEWORK['PAGE_SIZE'] + 1)
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        assert response.json()['next'] is not None
+
+    def test_no_data(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that without any data available, endpoint completes the request successfully"""
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
--- a/datahub/dataset/views.py
+++ b/datahub/dataset/views.py
@@ -1,0 +1,72 @@
+from rest_framework.views import APIView
+
+from config.settings.types import HawkScope
+from datahub.core.hawk_receiver import (
+    HawkAuthentication,
+    HawkResponseSigningMixin,
+    HawkScopePermission,
+)
+from datahub.core.query_utils import get_string_agg_subquery
+from datahub.dataset.pagination import OMISDatasetViewCursorPagination
+from datahub.omis.order.models import Order
+
+
+class OMISDatasetView(HawkResponseSigningMixin, APIView):
+    """
+    An APIView that provides 'get' action which queries and returns desired fields for OMIS Dataset
+    to be consumed by Data-flow periodically. Data-flow uses response result to insert data into
+    Dataworkspace through its defined API endpoints. The goal is presenting various reports to the
+    users out of flattened table and let analyst to work on denormalized table to get
+    more meaningful insight.
+    """
+
+    authentication_classes = (HawkAuthentication, )
+    permission_classes = (HawkScopePermission, )
+    required_hawk_scope = HawkScope.data_flow_api
+    pagination_class = OMISDatasetViewCursorPagination
+
+    def get(self, request):
+        """Endpoint which serves all records for OMIS Dataset"""
+        dataset = self.get_dataset()
+        paginator = self.pagination_class()
+        page = paginator.paginate_queryset(dataset, request, view=self)
+        return paginator.get_paginated_response(page)
+
+    def get_dataset(self):
+        """Returns list of OMIS Dataset records"""
+        return Order.objects.annotate(
+            services=get_string_agg_subquery(Order, 'service_types__name'),
+        ).values(
+            'cancelled_on',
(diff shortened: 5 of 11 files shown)


yusufertekin:
Can you help me to understand how can I improve test coverage to get green light?

reupen:
@yusufertekin Looks like the added tests aren't being picked up to me – I would try structuring them in the same way as the other apps. So add a `tests` sub-package containing `test_views.py` etc.

datahub/dataset/tests.py
```diff
@@ -0,0 +1,131 @@
...
+    """Returns OMISDatasetView url"""
+    yield reverse('api-v4:dataset:omis-dataset')
+
+
+def omis_dataset_expected_data_generator(order_list, next_page=None, previous_page=None):
+    """Generator that returns expected dictionary based on given list of orders"""
+    results = []
+    for order in order_list:
+        results.append({
+            'delivery_date': str(order.delivery_date),
```
mcbhenwood: It feels like an anti pattern, to have this scale of dictionary building going in terms of facilitating a post-condition check. It seem possible that a copy / paste error or typo in this section, supported by a corresponding one in the actual code, could lead to incorrect behaviour but with passing tests.

datahub/dataset/tests.py
```diff
@@ -0,0 +1,131 @@
...
+            'omis_order_reference': order.reference,
+            'net_price': order.subtotal_cost,
+            'subtotal': get_attr_or_none(order, 'invoice.subtotal_cost'),
+            'company_name': order.company.name,
+            'DIT_team': order.created_by.dit_team.name,
+            'market': order.primary_market.name,
+            'created_date': (format_date_or_datetime(order.created_on)
+                             if order.created_on else None),
+            'cancelled_date': (format_date_or_datetime(order.cancelled_on)
+                               if order.cancelled_on else None),
```
mcbhenwood: This seems like quite tightly-sprung, powerful logic inside what really ought to be just a dumb factory / object templating function.

datahub/dataset/tests.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    Tests for OMISDatasetView;
+    """
+
+    def test_success(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that endpoint returns with expected data for a single order"""
+        order = OrderFactory()
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        assert response.json() == omis_dataset_expected_data_generator([order])
```
mcbhenwood: See my comments on the generation of the expected dictionary in the `omis_dataset_expected_data_generator` function.

In the tests, instead of asserting that the JSON matches the output of this function called directly, would it be possible to break the equivalence comparisons into sections? e.g. something like (not tested)
```
response_json = response.json()
...
assert _response_trading_address_items_match(response_json, order)
assert _response_registered_address_items_match(response_json, order)
...
```
and then in those functions be like:
```
def _response_trading_address_items_match(json, order):
    assert json['company_trading_address_line_1'] == order.company.address_1
```
It's way more verbose this way but I feel leaves less room for errors. There is probably a better way than I suggest.

datahub/dataset/tests.py
yusufertekin: Thanks. What's your suggestion?

datahub/dataset/tests.py
yusufertekin: Sorry, I didn't understand what you mean by 'dumb factory' / 'object templating function'. There is no model for this dataset, it includes fields from various models and this is trying to create expected result json to validate response results based on dynamically generated variable mock orders. Also, this test will be the only place it's used.

datahub/dataset/tests.py
reupen: Generally, we will have at least _a_ test looking at the whole format of a response _item_ (not typically the whole response though), because:

- it will fail if extra fields are added to the response or fields are otherwise missing in the test
- the single assertion will report multiple failures at the same time
- it is clearer what the format of the response is

However, there is definitely a lot going on e.g. looking at pagination and multiple items at the same time.

yusufertekin: This tests a data critic operation. That's why, I wanted to test all fields against expected data. Having each field comparison in different functions might lead to maintainability problems. I agree that on the time of error, debugging comparing json as whole could lead to unreadability problems. As you suggested, I've re-structured it to test each field in separate assert statements. So that we will get more meaningful error like below
```
               AssertionError: assert 'wrong town' == 'Woodside'
                 - wrong town
                 + Woodside
```
Thanks !

yusufertekin: Separated pagination from expected data generator.

yusufertekin: Improved assertion message:
```
AssertionError: ("assert 'wrong town' == 'Woodside'\n  - wrong town\n  + Woodside", 'Order ref: VTW495/19', 'Field name company_trading_address_town')
```

config/settings/common.py
```diff
@@ -95,6 +95,7 @@
 
     # TODO: delete after the whole data cleansing piece of work is complete
     'datahub.dnb_match',
+    'datahub.dataset',
```
reupen: Small thing, but it would be nice to put this with the other apps before the TODO.

datahub/dataset/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class DatasetConfig(AppConfig):
```
reupen: The app config class doesn't seem to be used, can we remove it?

datahub/dataset/pagination.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.pagination import CursorPagination
+
+
+class OMISDatasetViewCursorPagination(CursorPagination):
+    """
+    Cursor Pagination for OMISDatasetView
+    """
+
+    ordering = ('created_date', 'omis_order_reference')
```
reupen:
```suggestion
    ordering = ('created_on', 'pk')
```

It'd be good to stick with fields on the model itself (since those are easier to index) and also use `pk` rather than the order reference (as the latter is a user-facing thing, it would be better to not rely on it for cases like these).

(There is also an existing index on `('created_on', 'pk')` that it will be able to use.)

However, I can also see that that wouldn't work with how we are using `CursorPagination` in this case. I think this needs a bit more thought.

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,146 @@
+import pytest
+from django.conf import settings
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.core.test_utils import format_date_or_datetime, get_attr_or_none, HawkAPITestClient
+from datahub.omis.order.test.factories import (
+    OrderCancelledFactory, OrderCompleteFactory, OrderFactory, OrderPaidFactory,
```
reupen: Small thing, but these factories should be one per line in line with the [coding style](https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Coding%20style.md#indentation).

datahub/dataset/pagination.py
yusufertekin: As far as I know, db indexes is still being used on the query doesn't matter which alias we are using.

Here is the end SQL.
```
SELECT "order_order"."delivery_date", 
(SELECT STRING_AGG(U2."name", ', 'ORDER BY U2."name") AS "_annotated_value" FROM "order_order" U0 LEFT OUTER JOIN "order_order_service_types" U1 ON (U0."id" = U1."order_id") LEFT OUTER JOIN "order_servicetype" U2 ON (U1."servicetype_id" = U2."id") WHERE U0."id" = ("order_order"."id") GROUP BY U0."id") AS "services", 
"order_order"."status" AS "order_status", "order_order"."reference" AS "omis_order_reference", "order_order"."subtotal_cost" AS "net_price", "omis-invoice_invoice"."subtotal_cost" AS "subtotal", "company_company"."name" AS "company_name", "metadata_team"."name" AS "DIT_team", "metadata_country"."name" AS "market", "order_order"."created_on" AS "created_date", "order_order"."cancelled_on" AS "cancelled_date", "order_cancellationreason"."name" AS "cancellation_reason_text", "metadata_ukregion"."name" AS "UK_region", "metadata_sector"."segment" AS "sector_name", "order_order"."paid_on" AS "payment_received_date", "order_order"."completed_on" AS "completion_date", "company_contact"."first_name" AS "contact_first_name", "company_contact"."last_name" AS "contact_last_name", "company_contact"."telephone_number" AS "contact_phone_number", "company_contact"."email" AS "contact_email_address", "company_company"."address_1" AS "company_trading_address_line_1", "company_company"."address_2" AS "company_trading_address_line_2", "company_company"."address_town" AS "company_trading_address_town", "company_company"."address_county" AS "company_trading_address_county", T11."name" AS "company_trading_address_country", "company_company"."address_postcode" AS "company_trading_address_postcode", "company_company"."registered_address_1" AS "company_registered_address_line_1", "company_company"."registered_address_2" AS "company_registered_address_line_2", "company_company"."registered_address_town" AS "company_registered_address_town", "company_company"."registered_address_county" AS "company_registered_address_county", T12."name" AS "company_registered_address_country", "company_company"."registered_address_postcode" AS "company_registered_address_postcode" 

FROM "order_order" 
LEFT OUTER JOIN "omis-invoice_invoice" ON ("order_order"."invoice_id" = "omis-invoice_invoice"."id") 
INNER JOIN "company_company" ON ("order_order"."company_id" = "company_company"."id") 
LEFT OUTER JOIN "company_advisor" ON ("order_order"."created_by_id" = "company_advisor"."id") 
LEFT OUTER JOIN "metadata_team" ON ("company_advisor"."dit_team_id" = "metadata_team"."id") 
LEFT OUTER JOIN "metadata_country" ON ("order_order"."primary_market_id" = "metadata_country"."id") 
LEFT OUTER JOIN "order_cancellationreason" ON ("order_order"."cancellation_reason_id" = "order_cancellationreason"."id") 
LEFT OUTER JOIN "metadata_ukregion" ON ("order_order"."uk_region_id" = "metadata_ukregion"."id") 
LEFT OUTER JOIN "metadata_sector" ON ("order_order"."sector_id" = "metadata_sector"."id") 
INNER JOIN "company_contact" ON ("order_order"."contact_id" = "company_contact"."id") 
LEFT OUTER JOIN "metadata_country" T11 ON ("company_company"."address_country_id" = T11."id")
LEFT OUTER JOIN "metadata_country" T12 ON ("company_company"."registered_address_country_id" = T12."id") 
ORDER BY "created_date" ASC, "omis_order_reference" ASC
```

yusufertekin: Used aliases are not user facing. This endpoint is designed to be machine-to-machine. It's consumed by data-flow-api, which uses the response data to insert in dataworkspace. We could do this name mapping in dataworkspace but I felt like it'd be better if it's done on source data.

datahub/dataset/apps.py
yusufertekin: I've added it back to be consistent with other apps which defines it without a need. Thought maybe reserved for future use. Removed it.

datahub/dataset/pagination.py
reupen: I missed that `created_date` was just an alias. However, the existing index is on `('created_on', 'pk')` and so the second ordering field should be `pk`.

Regarding the aliases, I think it's confusing to have them in this repository, and it creates unnecessary coupling. If the consumer of the API wants to store the fields under different names, that's more of a decision for it (rather than a concern for this project). So I'd stick as close to the original names here as possible.

datahub/dataset/views.py
```diff
@@ -0,0 +1,69 @@
...
+    HawkAuthentication,
+    HawkResponseSigningMixin,
+    HawkScopePermission,
+)
+from datahub.core.query_utils import get_string_agg_subquery
+from datahub.dataset.pagination import OMISDatasetViewCursorPagination
+from datahub.omis.order.models import Order
+
+
+class OMISDatasetView(HawkResponseSigningMixin, GenericAPIView):
```
reupen: This should inherit from `APIView` rather than `GenericAPIView` (as it's not really a generic view since it doesn't use a serialiser or the standard list etc. actions).

Changing it should just mean using the pagination class directly – something like:

```python
    def get(self, request):
        """Endpoint which serves all records for OMIS Dataset"""
        dataset = self.get_dataset()
        paginator = OMISDatasetViewCursorPagination()
        page = paginator.paginate_queryset(dataset, request, view=self)
        return paginator.get_paginated_response(page)
```

datahub/dataset/views.py
```diff
@@ -0,0 +1,69 @@
...
+    pagination_class = OMISDatasetViewCursorPagination
+
+    def get(self, request):
+        """Endpoint which serves all records for OMIS Dataset"""
+        dataset = self.get_dataset()
+        page = self.paginate_queryset(dataset)
+        return self.get_paginated_response(page)
+
+    def get_dataset(self):
+        """Returns list of OMIS Dataset revords"""
```
reupen: Small typo:

```suggestion
        """Returns list of OMIS Dataset records"""
```

datahub/dataset/pagination.py
yusufertekin: We have to do this mapping either in datahub or dataworkspace because you don't want this mapping in datahub, I had to create a serializer just to match field names on deserialization. I don't call it coupling but now dataworkspace knows about datahub datastructure with ugly names including relations (i.e. created_by__dit_team__name).  FYI, with these changes, if one of the fields in this api endpoint is changed somehow, code change on dataworkspace is required.

datahub/dataset/views.py
```diff
@@ -0,0 +1,69 @@
...
+        return Order.objects.prefetch_related('service_types').annotate(
```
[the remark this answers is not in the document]
reupen: This comment is still relevant.

datahub/dataset/views.py
```diff
@@ -0,0 +1,69 @@
...
+        paginator = self.pagination_class()
+        page = paginator.paginate_queryset(dataset, request, view=self)
+        return paginator.get_paginated_response(page)
+
+    def get_dataset(self):
+        """Returns list of OMIS Dataset records"""
+        return Order.objects.prefetch_related('service_types').annotate(
+            services=get_string_agg_subquery(Order, 'service_types__name'),
+        ).values(
+            'services',
```
reupen: It would be good to sort the field names alphabetically to make the list of fields easier to scan through.

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,152 @@
...
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_without_credentials(self, api_client, omis_dataset_view_url):
+        """Test that making a request without credentials returns an error."""
+        response = api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    @pytest.mark.parametrize(
+        'order_factory',
+        (OrderFactory, OrderCompleteFactory, OrderCancelledFactory, OrderPaidFactory,
```
reupen: The factories here should be one per line (as per earlier comment about the coding style).

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,152 @@
...
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 1
+        result = response_results[0]
+        expected_result = omis_dataset_expected_data_generator(order)
+        assert result == expected_result
+
+    def test_with_multiple_orders(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that endpoint returns with expected data for multiple orders"""
+        orders = OrderFactory.create_batch(3)
```
reupen: Would be good to use `freeze_time()` (from `freezegun`) to control the `created_on` date for the orders we are creating, so the ordering can be fully tested.

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,152 @@
...
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 3
+        expected_sorted_orders = sorted(orders, key=lambda x: (x.created_on, x.reference))
```
reupen: Would be nice to use a more meaningful variable name than `x` e.g. `order` or `obj`.

Also, shouldn't the sorting be using `pk` instead of `reference` now?

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,152 @@
...
+def omis_dataset_view_url():
+    """Returns OMISDatasetView url"""
+    yield reverse('api-v4:dataset:omis-dataset')
+
+
+def omis_dataset_expected_data_generator(order):
+    """Generator that returns expected dictionary based on given list of orders"""
+    return {
+        'delivery_date': str(order.delivery_date),
+        'services': ', '.join((st_name
```
reupen: There is a `join_attr_values()` in `datahub.core.test_utils` that could be used here.

Also the sorting should be done by using calling `order_by()` on the query set, as the database sorts strings differently from how Python does in some cases so mixing the two could lead to occasional test failures.

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,152 @@
...
+    def test_with_multiple_orders(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that endpoint returns with expected data for multiple orders"""
+        orders = OrderFactory.create_batch(3)
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 3
+        expected_sorted_orders = sorted(orders, key=lambda x: (x.created_on, x.reference))
+        for index in range(3):
+            assert format_date_or_datetime(expected_sorted_orders[index].created_on) == \
```
reupen: `\` should be avoided for line continuation (as per PEP 8). So, it would be good to avoid them here – either by using shorter expressions, or wrapping them in round brackets.

datahub/dataset/pagination.py
reupen: Happy to discuss further in person, but I'd add:

- in all cases API compatibility needs to be considered and fields can't just be removed
- there's a deprecation process for removing fields if it does become a necessity
- the structure is basically the same as the endpoint the Data Hub front end uses (except that it's flat rather than nested)
- we can still use aliases if we need to, the point was that they should make sense in the context of this repository

datahub/dataset/test/test_views.py
yusufertekin: something broken with diff, this is fixed.

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,166 @@
...
+            order_2 = OrderFactory()
+        with freeze_time('2019-01-01 12:00:00'):
+            order_3 = OrderFactory()
+        with freeze_time('2019-01-01 12:00:00'):
+            order_4 = OrderFactory()
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 4
+        for index, order in enumerate([order_3, order_4, order_1, order_2]):
```
reupen: The `pk` is a UUID, so the orders will still need to be sorted here

datahub/dataset/views.py
```diff
@@ -0,0 +1,69 @@
...
+)
+from datahub.core.query_utils import get_string_agg_subquery
+from datahub.dataset.pagination import OMISDatasetViewCursorPagination
+from datahub.omis.order.models import Order
+
+
+class OMISDatasetView(HawkResponseSigningMixin, APIView):
+    """
+    An APIView that provides 'get' action which queries and returns desired
+    fields for OMIS Dataset to be consumed by data-flow-api
```
reupen: It would be good to elaborate a bit on how it is used to help anyone else that needs to work on it in future i.e. I assume it is used to periodically copy all the data for all orders from Data Hub to XXX.

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,166 @@
...
+        response = data_flow_api_client.get(omis_dataset_view_url)
+        assert response.status_code == status.HTTP_200_OK
+        response_results = response.json()['results']
+        assert len(response_results) == 1
+        result = response_results[0]
+        expected_result = omis_dataset_expected_data_generator(order)
+        assert result == expected_result
+
+    def test_with_multiple_orders(self, data_flow_api_client, omis_dataset_view_url):
+        """Test that endpoint returns with expected data for multiple orders"""
```
reupen: Mainly this is testing the ordering with multiple orders, so it would be good to clarify the docstring a little bit.

changelog/dataset/omis-dataset-api-endpoint.api.rst
```diff
@@ -0,0 +1 @@
+New API Endpoint OMISDatasetView presents required fields data of all orders to be used in data-workspace reporting
```
reupen: This won't be picked up in the dataset directory as there is a predefined list of sections in `pyproject.toml`. I think it can go under `omis` as that's the general area of functionality (for example search changes for OMIS in particular would also be put under OMIS).

changelog/omis/omis-dataset-api-endpoint.api.rst
```diff
@@ -0,0 +1 @@
+New API Endpoint OMISDatasetView presents required fields data of all orders to be consumed by data-flow and used in data-workspace reporting and analyst access.
```
currycoder: It could be helpful to explain the expected signature of the new API endpoint in this changelog entry, e.g. https://raw.githubusercontent.com/uktrade/data-hub-leeloo/develop/changelog/adviser/update-company-list.api.rst

datahub/dataset/test/test_views.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+@pytest.fixture
+def omis_dataset_view_url():
+    """Returns OMISDatasetView url"""
+    yield reverse('api-v4:dataset:omis-dataset')
+
+
+def omis_dataset_expected_data_generator(order):
+    """Generator that returns expected dictionary based on given list of orders"""
```
currycoder: "Generator" gives me some python-specific connotations - e.g. https://wiki.python.org/moin/Generators

Maybe "transformer" could be better?

sample.env
```diff
@@ -35,3 +35,5 @@ COMPOSE_PROJECT_NAME=data-hub
 # Some extra ENV variables to make superuser creation easier on docker copies
 #SUPERUSER_USERNAME=brendan.smith
 #SUPERUSER_PASSWORD=foobarbaz
+DATA_FLOW_API_ACCESS_KEY_ID=data-flow-api-id
```
currycoder: Could these be added to the README?

changelog/omis/omis-dataset-api-endpoint.api.rst
elcct: I agree with above.

datahub/dataset/test/test_views.py
yusufertekin: It was a generator at some point :) Changed the name.

PR #2012, opened 2019-08-27
Add a configurable blacklist to prohibit certain email addresses using email ingestion

currycoder:
### Description of change

It was agreed in a meeting about securing the email ingestion feature that we should have a configurable blacklist of emails that are prohibited from using email ingestion.  This has been implemented as a new setting `DIT_EMAIL_INGEST_BLACKLIST`.  

It's possible that we may want to add some logging in the future, but for right now a simple blacklist lookup is good enough to satisfy the business requirements.

### Checklist

### DIFF
--- a/datahub/email_ingestion/validation.py
+++ b/datahub/email_ingestion/validation.py
@@ -55,6 +55,9 @@ def was_email_sent_by_dit(message):
     if from_email.lower() not in settings.DIT_EMAIL_INGEST_WHITELIST:
         return False
 
+    if from_email.lower() in settings.DIT_EMAIL_INGEST_BLACKLIST:
+        return False
+
     try:
         domain_auth_methods = settings.DIT_EMAIL_DOMAINS[from_domain]
     except KeyError:
--- a/README.md
+++ b/README.md
@@ -260,6 +260,8 @@ Leeloo can run on any Heroku-style platform. Configuration is performed via the 
 | `DATAHUB_SUPPORT_EMAIL_ADDRESS` | No | Email address for DataHub support team. |
 | `DEBUG`  | Yes | Whether Django's debug mode should be enabled. |
 | `DIT_EMAIL_DOMAIN_*` | No | An allowable DIT email domain for email ingestion along with it's allowed email authentication methods. Django-environ dict format e.g. example.com=dmarc:pass\|spf:pass\|dkim:pass |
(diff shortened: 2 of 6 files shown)


reupen:
@currycoder The changes aren't covered by the tests according to the coverage report – presumably as the blacklisted email isn't on the whitelist?

currycoder:
Good point @reupen thanks!  I should probably also share a bit more context about why this PR has both a whitelist and a blacklist..  We are hoping to remove the whitelist shortly as meeting invite ingestion gets out of the pilot stage, whereas the blacklist will live on as a safety measure that we can use in the future.

datahub/email_ingestion/validation.py
```diff
@@ -55,6 +55,9 @@ def was_email_sent_by_dit(message):
     if from_email.lower() not in settings.DIT_EMAIL_INGEST_WHITELIST:
         return False
 
+    if from_email.lower() in settings.DIT_EMAIL_INGEST_BLACKLIST:
```
reupen: Should we check the domain before checking the blacklist?

reupen:
@currycoder So should there be a feature flag to switch to the new behaviour of using a blacklist instead of a whitelist? As blacklisting things on the whitelist doesn't seem too useful.

Alternatively, keep this on hold and remove the whitelist at the same time?

datahub/email_ingestion/validation.py
currycoder: I think we could do it either way really.  Doing it high up in the validation function made the most sense to me as we immediately check if the user has been blacklisted before performing any further validation logic.

currycoder:
> @currycoder So should there be a feature flag to switch to the new behaviour of using a blacklist instead of a whitelist? As blacklisting things on the whitelist doesn't seem too useful.
> 
> Alternatively, keep this on hold and remove the whitelist at the same time?

We could do that.  I saw this as a way that we could get some logic merged now ahead of switching off the whitelist.  I don't mind moving the commit over to the (blocked) PR which removes the whitelist if that's preferable.

datahub/email_ingestion/validation.py
reupen: Yes, there shouldn't be any difference in the end result, but I think logically it doesn't make sense to blacklist things not on the domain whitelist so ordering it that way might make that clearer.

Not too fussed though.

currycoder: Yeah no worries!  I'll close this and pick the commit over to the other PR :)

PR #2013, opened 2019-08-27
Adjust email ingestion to delete ingested emails

currycoder:
### Description of change

Email ingestion was adjusted so that emails are deleted after they are ingested. Previously, email ingestion would mark the emails as "seen" but now that we are out of the pilot for meeting invite ingestion we have switched to deletes as this is safer for data retention/protection reasons.

**NOTE** This should not be merged until we have agreed that the invite ingestion pilot is finished.

### Checklist

### DIFF
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -200,7 +200,7 @@ class Mailbox:
                     )
                     logger.exception(error_message)
                     # Just set the message to None so that we still mark it as
-                    # read later
+                    # deleted later
                     message = None
                 except EmailRetrievalError:
                     # We should fail and exit immediately in this case, as it's
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -172,9 +172,13 @@ class Mailbox:
 
     def get_new_mail(self):
         """
-        Generator method which gets new messages from the email inbox. After a
-        message has been yielded (or an error has been logged while parsing it),
-        it is flagged as DELETED on the inbox so that it will not be ingested again later.
+        Generator method which gets new messages from the email inbox.
+
+        After a message has been yielded (or an error has been logged while parsing it), it is
+        flagged as DELETED on the inbox. Finally, the function will call `expunge()` on the mailbox
+        (which will delete all messages marked for deletion) - this will be called after all unread
+        messages have been yielded by the generator.
+
         We only consider messages that have not been seen for ingestion.
 
         :yields: A mailparser.Message object for each parsed message.
--- a/datahub/email_ingestion/mailbox.py
+++ b/datahub/email_ingestion/mailbox.py
@@ -28,7 +28,7 @@ class Mailbox:
     them using associated processing classes.
 
     When an email is retrieved from the inbox successfully, it is marked as
-    SEEN on the upstream IMAP inbox.  This is the mechanism we use to ensure
+    DELETED on the upstream IMAP inbox.  This is the mechanism we use to ensure
     that emails are only ever processed once.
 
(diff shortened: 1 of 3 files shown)


datahub/email_ingestion/mailbox.py
```diff
@@ -174,7 +174,7 @@ def get_new_mail(self):
         """
         Generator method which gets new messages from the email inbox. After a
         message has been yielded (or an error has been logged while parsing it),
-        it is flagged as SEEN on the inbox so that it will not be ingested again later.
+        it is flagged as DELETED on the inbox so that it will not be ingested again later.
```
reupen: This says they'll be flagged as deleted, but the expunge call will actually delete them, right? It would be good to make that a bit clearer in the docstring here (and the one above).

mcbhenwood: Agreed. Even though the message is marked as deleted initially, it's expunged when the generator ends and so the comment should probably reflect the end position rather than the intermediate state.

I note also that the tests check for this marking as "Deleted" - that makes sense but I still think that the comment on the code itself should be clear that deletion takes place. (Either that or expand the comment to explain the whole process but that feels a bit verbose for a docstring).

datahub/email_ingestion/mailbox.py
```diff
@@ -209,8 +209,10 @@ def get_new_mail(self):
                 if message:
                     yield message
 
-                # Mark the email as read in the inbox
-                connection.uid('store', uid, '+FLAGS', '(\\SEEN)')
+                # Mark the email for deletion in the inbox
+                connection.uid('store', uid, '+FLAGS', '(\\Deleted)')
+            # Delete the emails which were marked for deletion
+            connection.expunge()
```
mcbhenwood: Reading the `expunge()` [docs](https://docs.python.org/3/library/imaplib.html#imaplib.IMAP4.expunge), it sounds like the `expunge` method works on the mailbox as a whole, removing all messages marked for Deletion, rather than just messages marked _during this connection_. So if for example there were two consumers processing this mailbox at the same time, it sounds possible that an `expunge()` in one could remove messages marked in the other, still-running consumer.

I don't see why this would cause us a problem under the current processing model but it may be worth a moment's thought for sanity checking.

datahub/email_ingestion/mailbox.py
currycoder: Nice observation.  I'll go with the verbose route - thanks!

datahub/email_ingestion/mailbox.py
currycoder: I appreciate the diligence here.  Thinking it through, I agree that it's not a problem for us.  Even if two ingestion tasks were running in parallel, the worst case would be that the expunge from task 1 deletes some emails marked for deletion by task 2 a little earlier than expected.

Worth noting also that right now, the email ingestion is run on a celery beat and the celery task uses a lock to ensure that it runs exclusively.

datahub/email_ingestion/mailbox.py
```diff
@@ -172,9 +172,13 @@ def _get_message(self, uid, connection):
...
+        Generator method which gets new messages from the email inbox.
+
+        After a message has been yielded (or an error has been logged while parsing it), it is
+        flagged as DELETED on the inbox. Finally, the function will call `expunge()` on the mailbox
+        (which will delete all messages marked for deletion) - this will be called after all unread
+        messages have been yielded by the generator.
+
         We only consider messages that have not been seen for ingestion.
 
         :yields: A mailparser.Message object for each parsed message.
```
reupen: Just spotted that there's a comment on line 203 (further down) that looks like it needs updating.

currycoder: Nice spot - thanks!

PR #2020, opened 2019-08-28
Add pending_dnb_investigation flag to Company model

alixedi:
### Description of change

Adds a boolean field called `pending_dnb_investigation` to the Company model. The purpose of this field is to record whether a company is under DNB investigation.

Initially, I set `default=False` for `pending_dnb_investigation` and timed the migration on staging. Here is what I did:

1. Add the field with `default=False` in the Company model:

```
    pending_dnb_investigation = models.BooleanField(
        null=True,
        blank=True,
        default=False,
        help_text='Whether this company is to be investigated by DNB.',
    )
```

2. Generate the migration:

```
$ python manage.py makemigrations company
```

3. Get the SQL for the migration:

```
$ python manage.py sqlmigrate company 0087_company_pending_dnb_investigation
```

4. Run it on staging:

```
$ cg ssh datahub-api-staging
$ ./manage.py shell_plus
```

```
>>> from django.db import connection
>>> with connection.cursor() as cursor:
>>>     cursor.execute(f"BEGIN; %s; ROLLBACK;",  [sql_from_step_3])
```

It took 20 seconds which is a little too much in my opinion. As a result, I am planning to do the following:

1. Create a column in the DB without a default value - See the migration.
2. Set the field in the Django model to a default value. This would ensure that new companies will have `pending_dnb_investigation` default to `False`. 
3. Create a migration for the default value.
4. Release.
5. Use `datahub.dbmaintenance.tasks.replace_null_with_default` to set the default for existing companies in batches.
6. Create the migration that removes `null=True, blank=True` for the field.
7. Release.

### Checklist


changelog/company/add-failed-dnb-investigation-flag.db.rst
```diff
@@ -0,0 +1 @@
+The ``company_company`` table now has a boolean field called ``pending_dnb_investigation``. This is to record whether a company is under DNB investigation.
```
currycoder: Don't know if the filename matters in practice, but it's not using the updated field name.

datahub/company/test/models/test_company.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from datahub.company.models import Company
+from datahub.company.test.factories import CompanyFactory
+
+
+@pytest.mark.django_db
+def test_pending_dnb_investigation():
+    """
+    Check if a newly create company has `pending_dnb_investigation`
```
elcct: I think it should be `created`

datahub/company/test/models/test_company.py
```diff
@@ -0,0 +1,16 @@
...
+    set to False.
+    """
+    company = CompanyFactory()
```
elcct: I am not sure if this test checks that a newly created company has `pending_dnb_investigation` set to False. Here it checks if company created by a factory has that field set to False, but is it the case when company actually gets created in the system?

alixedi: Cool, I have added a test [here](https://github.com/uktrade/data-hub-leeloo/pull/2020/commits/e81c0606fd43a7795d5f1674ed297aca58fcbaba)

PR #2021, opened 2019-08-28
Add endpoint to add company to user company lists

elcct:
### Description of change

This adds an endpoint for adding a company to user's multiple company lists. 
The work is based on a legacy implementation done by @reupen for a single list per user.

 ``PUT /v4/company-list/<company list ID>/item/<company ID>``

  This adds a company to the user's own selected list of companies.

  If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or company with the specified company ID, a 404 will be returned.

  If an archived company is specified, a 400 status code will be returned and response body will contain::

      {
          "non_field_errors": "An archived company can't be added to a company list."
      }

  Otherwise, the response body will be empty.

### Checklist

### DIFF
--- a/changelog/adviser/company-list-add-item.api.rst
+++ b/changelog/adviser/company-list-add-item.api.rst
@@ -1,0 +1,15 @@
+The following endpoint was added:
+
+  - ``PUT /v4/company-list/<company list ID>/item/<company ID>``
+
+  This adds a company to the user's own selected list of companies.
+
+  If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or company with the specified company ID, a 404 will be returned.
+
+  If an archived company is specified, a 400 status code will be returned and response body will contain::
+
+      {
+          "non_field_errors": "An archived company can't be added to a company list."
+      }
+
+  Otherwise, the response body will be empty.
--- a/datahub/user/company_list/urls.py
+++ b/datahub/user/company_list/urls.py
@@ -4,7 +4,7 @@ from datahub.user.company_list.legacy_views import (
     LegacyCompanyListItemView,
     LegacyCompanyListViewSet,
 )
-from datahub.user.company_list.views import CompanyListViewSet
+from datahub.user.company_list.views import CompanyListItemAPIView, CompanyListViewSet
 
 urlpatterns = [
     path(
@@ -26,6 +26,11 @@ urlpatterns = [
             },
         ),
         name='list-detail',
+    ),
+    path(
+        'company-list/<uuid:company_list_pk>/item/<uuid:company_pk>',
+        CompanyListItemAPIView.as_view(),
+        name='item-detail',
     ),
     path(
         'user/company-list',
--- a/datahub/user/company_list/views.py
+++ b/datahub/user/company_list/views.py
@@ -1,13 +1,33 @@
+from django.db import transaction
 from django.db.models import Count
+from django.shortcuts import get_object_or_404
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext_lazy
 from django_filters.rest_framework import DjangoFilterBackend
+from oauth2_provider.contrib.rest_framework import IsAuthenticatedOrTokenHasScope
+from rest_framework import serializers, status
 from rest_framework.filters import OrderingFilter
 from rest_framework.mixins import DestroyModelMixin
+from rest_framework.permissions import DjangoModelPermissions
+from rest_framework.response import Response
+from rest_framework.settings import api_settings
+from rest_framework.views import APIView
 
+from datahub.company.models import Company, CompanyPermission
 from datahub.core.query_utils import get_aggregate_subquery
 from datahub.core.viewsets import CoreViewSet
 from datahub.oauth.scopes import Scope
-from datahub.user.company_list.models import CompanyList
+from datahub.user.company_list.models import (
+    CompanyList,
+    CompanyListItem,
+    CompanyListItemPermissionCode,
+)
+from datahub.user.company_list.serializers import CompanyListItemSerializer
 from datahub.user.company_list.serializers import CompanyListSerializer
+
+CANT_ADD_ARCHIVED_COMPANY_MESSAGE = gettext_lazy(
+    "An archived company can't be added to a company list.",
+)
 
 
 class CompanyListViewSet(CoreViewSet, DestroyModelMixin):
@@ -48,3 +68,65 @@ class CompanyListViewSet(CoreViewSet, DestroyModelMixin):
             **additional_data,
             'adviser': self.request.user,
         }
+
+
+class CompanyListItemAPIPermissions(DjangoModelPermissions):
+    """DRF permissions class for the company list item view."""
+
+    perms_map = {
+        'PUT': [
+            f'company.{CompanyPermission.view_company}',
+            f'company_list.{CompanyListItemPermissionCode.add_company_list_item}',
+        ],
+    }
+
+
+class CompanyListItemAPIView(APIView):
+    """
+    A view for adding a company to a selected list of companies that belongs to a user.
+    """
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (
+        IsAuthenticatedOrTokenHasScope,
+        CompanyListItemAPIPermissions,
+    )
+    # Note: A query set is required for CompanyListItemPermissions
+    queryset = CompanyListItem.objects.all()
+    serializer_class = CompanyListItemSerializer
+
+    @method_decorator(transaction.non_atomic_requests)
+    def put(self, request, company_list_pk, company_pk, format=None):
+        """Add company to a list."""
+        company_list = self._get_company_list_or_404(request, company_list_pk)
+
+        company = get_object_or_404(Company, pk=company_pk)
+        if company.archived:
+            errors = {
+                api_settings.NON_FIELD_ERRORS_KEY: CANT_ADD_ARCHIVED_COMPANY_MESSAGE,
+            }
+            raise serializers.ValidationError(errors)
+
+        adviser = request.user
+
+        # get_or_create() is used to avoid an error if there is an existing
+        # CompanyListItem for this adviser and company
+        self.queryset.get_or_create(
+            company=company,
+            list=company_list,
+            defaults={
+                'created_by': adviser,
+                'modified_by': adviser,
+            },
+        )
+
+        return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+    def _get_company_list_or_404(self, request, company_list_pk):
+        obj = get_object_or_404(
+            CompanyList,
+            adviser=request.user,
+            pk=company_list_pk,
+        )
+        self.check_object_permissions(request, obj)
+        return obj
--- a/datahub/user/company_list/legacy_views.py
+++ b/datahub/user/company_list/legacy_views.py
@@ -17,11 +17,10 @@ from datahub.oauth.scopes import Scope
 from datahub.user.company_list.models import (
     CompanyList,
     CompanyListItem,
-    CompanyListItemPermissionCode,
 )
+from datahub.user.company_list.models import CompanyListItemPermissionCode
 from datahub.user.company_list.queryset import get_company_list_item_queryset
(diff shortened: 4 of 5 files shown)


datahub/user/company_list/views.py
```diff
@@ -1,10 +1,53 @@
...
+)
+from datahub.user.company_list.serializers import CompanyListItemSerializer
 from datahub.user.company_list.serializers import CompanyListSerializer
 
+CANT_ADD_ARCHIVED_COMPANY_MESSAGE = gettext_lazy(
+    "An archived company can't be added to a company list.",
+)
+
+
+class CompanyListItemPermissions(DjangoModelPermissions):
```
reupen: I think it would be a bit cleaner to define a new permissions class for this, since we won't be using all of the same methods.

datahub/user/company_list/views.py
```diff
@@ -22,3 +65,53 @@ class CompanyListViewSet(CoreViewSet):
...
+        """Add company to a list."""
+        company = get_object_or_404(Company, pk=company_pk)
+        if company.archived:
+            errors = {
+                api_settings.NON_FIELD_ERRORS_KEY: CANT_ADD_ARCHIVED_COMPANY_MESSAGE,
+            }
+            raise serializers.ValidationError(errors)
+
+        adviser = request.user
+        company_list = self._get_company_list(request, company_list_pk)
```
reupen: It would be good for this to be the first thing that the function does so that it also 404s if an archived company is specified for a non-existent list.

datahub/user/company_list/views.py
```diff
@@ -22,3 +65,53 @@ class CompanyListViewSet(CoreViewSet):
...
+            list=company_list,
+            defaults={
+                'created_by': adviser,
+                'modified_by': adviser,
+            },
+        )
+
+        return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+    def _get_company_list(self, request, company_list_pk):
```
reupen: It would be good to add `or_404` to the name of this function so that it follows the same pattern as `get_object_or_404()`.

datahub/user/company_list/urls.py
```diff
@@ -16,6 +16,11 @@
         ),
         name='list-collection',
     ),
+    path(
+        'company-list/<uuid:company_list_pk>/<uuid:company_pk>',
+        CompanyListItemAPIView.as_view(),
+        name='list-item',
```
reupen: Maybe we should call this `item-detail` so that we can have `item-collection` as well (also to make it clearer that it's not a list object itself)? (So the full name would end up being `api-v4:company-list:item-detail` etc.)

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,252 @@ def test_lists_are_sorted_by_name(self):
...
+                'company_list_pk': company_list.pk,
+                'company_pk': company.pk,
+            },
+        )
+
+        response = self.api_client.put(url)
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert response.content == b''
+
+        list_item = _get_queryset_for_list_adviser_and_company(
```
reupen: Shouldn't we just be passing the list and company into this now (and not the adviser)?

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,252 @@ def test_lists_are_sorted_by_name(self):
...
+            company_list,
+            self.user,
+            company,
+        ).first()
+
+        assert list_item
+        assert list_item.list.adviser == self.user
```
reupen: I think we'd want to check that `list_item.list == company_list` here.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,252 @@ def test_lists_are_sorted_by_name(self):
...
+                'company_list_pk': company_list.pk,
+                'company_pk': company_to_add.pk,
+            },
+        )
+        response = self.api_client.put(url)
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+
+        list_item_queryset = CompanyListItem.objects.filter(
+            list=company_list,
+            list__adviser=self.user,
```
reupen: Shouldn't need to filter by adviser here (as the list only has one adviser).

(Same comment for the other tests.)

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,252 @@ def test_lists_are_sorted_by_name(self):
...
+
+class TestCreateOrUpdateCompanyListItemAPIView(APITestMixin):
+    """Tests for the PUT method in CompanyListItemAPIView."""
+
+    def test_creates_new_items(self):
+        """Test that a company can be added to the authenticated user's list."""
+        company = CompanyFactory()
+        company_list = CompanyListFactory(adviser=self.user)
+
+        url = list_item_url(
```
reupen: We could define a `_get_list_item_url()` that takes a company and list ID to simplify this a little bit.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,252 @@ def test_lists_are_sorted_by_name(self):
...
+            company_list,
+            self.user,
+            company,
+        ).exists()
+
+    def test_adviser_can_have_the_same_company_on_multiple_lists(self):
+        """Tests that adviser can have the same company on multiple lists."""
+        company = CompanyFactory()
+        lists = CompanyListFactory.create_batch(5, adviser=self.user)
+        for company_list in lists:
```
reupen: This test looks a little bit complicated – it would be good to restrict it the scope of it, and do the set-up using the factories.

So something like create two lists (using `CompanyListFactory`), create a list item for one of the lists and the adviser (using `CompanyListItemFactory`) and then make an API request to add the company to the second list.

datahub/user/company_list/views.py
```diff
@@ -22,3 +65,53 @@ class CompanyListViewSet(CoreViewSet):
     def get_queryset(self):
         """Get a query set filtered to the authenticated user's lists."""
         return super().get_queryset().filter(adviser=self.request.user)
+
+
+class CompanyListItemAPIView(APIView):
+    """
+    A view for adding a company to a selected list of companies that belongs to a user.
```
currycoder: Question: Why aren't we inheriting from the `CoreViewSet` class here?  I'm sure there's a reason, but it might be nice to highlight in the docstring?

datahub/user/company_list/views.py
elcct: I created a new permission class that includes only what is being used.

datahub/user/company_list/views.py
elcct: Very good point! Updated. Thank you.

datahub/user/company_list/views.py
elcct: True! I added that. Thank you.

datahub/user/company_list/urls.py
elcct: This is a good idea. I changed it so it is now called `item-detail`.

datahub/user/company_list/tests/test_views.py
elcct: I wanted to be extra sure that the list belongs to that user, but I agree this is not necessary. Removed.

datahub/user/company_list/tests/test_views.py
elcct: That's true. Updated. Thank you.

datahub/user/company_list/tests/test_views.py
elcct: Good point! Removed. Thank you.

datahub/user/company_list/tests/test_views.py
elcct: Great suggestion. I added a function that creates an url and it looks much simpler. Thank you!

datahub/user/company_list/tests/test_views.py
elcct: I agree this is a bit overdone. I have simplified it following your suggestions. Thank you!

changelog/adviser/company-list-add-item.api.rst
```diff
@@ -0,0 +1,15 @@
+The following endpoint was added:
+
+  - ``PUT /v4/company-list/<company list ID>/<company ID>``
+
+  This adds a company to the user's own selected list of companies.
+
+  If the operation is successful, a 2xx status code will be returned. If there is no company list with specified company list ID or company with the specified company ID, a 404 wil lbe returned.
```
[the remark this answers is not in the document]
reupen: Although it's a 204 at the moment, if we add a response body later on then it will be a 200 or 201. so it might be good to say it's subject to change to another 2xx code.

changelog/adviser/company-list-add-item.api.rst
```diff
@@ -0,0 +1,15 @@
+The following endpoint was added:
+
+  - ``PUT /v4/company-list/<company list ID>/<company ID>``
+
+  This adds a company to the user's own selected list of companies.
+
+  If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or company with the specified company ID, a 404 wil lbe returned.
```
reupen: Small typo (originally mine) :slightly_smiling_face: 
```suggestion
  If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or company with the specified company ID, a 404 will be returned.
```

changelog/adviser/company-list-add-item.api.rst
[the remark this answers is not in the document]
elcct: I have updated it now. Thank you!

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,187 @@ def test_lists_are_sorted_by_name(self):
...
+        response = self.api_client.put(url)
+
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+        assert not CompanyListItem.objects.filter(
+            list__adviser=self.user,
+            company=company,
+        ).exists()
+
+
+def _get_queryset_for_list_adviser_and_company(company_list, company):
```
reupen: I think we can drop 'adviser from the function name now:

```suggestion
def _get_queryset_for_list_and_company(company_list, company):
```

datahub/user/company_list/views.py
```diff
@@ -22,3 +65,53 @@ class CompanyListViewSet(CoreViewSet):
...
+        self.queryset.get_or_create(
+            company=company,
+            list=company_list,
+            defaults={
+                'created_by': adviser,
+                'modified_by': adviser,
+            },
+        )
+
+        return Response(None, status=status.HTTP_204_NO_CONTENT)
```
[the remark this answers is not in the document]
elcct: This is not an usual pattern, but it exists in the legacy list code. My reason is that I think the serialised object that is being created will not be useful as the caller already would have all the information.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +101,187 @@ def test_lists_are_sorted_by_name(self):
...
+            company=company,
+        ).count() == 2
+
+    def test_with_existing_item(self):
+        """
+        Test that no error is returned if the specified company is already on the
+        authenticated user's list.
+        """
+        creation_date = datetime(2018, 1, 2, tzinfo=utc)
+        modified_date = datetime(2018, 1, 2, tzinfo=utc)
```
reupen: Small thing (again from the legacy view tests) – `modified_date` is the same as `creation_date` date here when they probably should be different.

The endpoint doesn't actually change `modified_on` in this case, which I think is fine since nothing has really been changed, so the last assertion should need updating as well.

datahub/user/company_list/views.py
elcct: The reason for not using `CoreViewSet` is to keep this simple. I could definitely write some info in the docstring.

datahub/user/company_list/tests/test_views.py
elcct: Good point! Updated.

changelog/adviser/company-list-add-item.api.rst
elcct: :blush: Fixed!

PR #2022, opened 2019-08-28
Add endpoint for listing company list contents

elcct:
### Description of change

This adds an endpoint for getting contents of user's multiple company lists. The work is based on a legacy implementation done by @reupen for a single list per user.

``GET /v4/company-list/<company list ID>/item``

It lists all the companies on the authenticated user's selected list, with responses in the following format::

    {
        "count": <int>,
        "previous": <url>,
        "next": <url>,
        "results": [
            {
                "company": {
                    "id": <string>,
                    "archived": <boolean>,
                    "name": <string>,
                    "trading_names": [<string>, <string>, ...]
                },
                "created_on": <ISO timestamp>,
                "latest_interaction": {
                    "id": <string>,
                    "created_on": <ISO timestamp>,
                    "date": <ISO date>,
                    "subject": <string>
                }
            },
            ...
        ]
    }

``latest_interaction`` may be ``null`` if the company has no interactions.

Results are sorted by ``latest_interaction.date`` in reverse chronological order, with ``null`` values last.

The endpoint has pagination in line with other endpoints; to retrieve all results pass a large value for the ``limit`` query parameter (e.g. ``?limit=1000``).

If selected list does not exist, the endpoint will return 404 status code.

### Checklist

### DIFF
--- a/datahub/user/company_list/views.py
+++ b/datahub/user/company_list/views.py
@@ -1,5 +1,6 @@
 from django.db import transaction
 from django.db.models import Count
+from django.http import Http404
 from django.shortcuts import get_object_or_404
 from django.utils.decorators import method_decorator
 from django.utils.translation import gettext_lazy
@@ -22,6 +23,7 @@ from datahub.user.company_list.models import (
     CompanyListItem,
     CompanyListItemPermissionCode,
 )
+from datahub.user.company_list.queryset import get_company_list_item_queryset
 from datahub.user.company_list.serializers import CompanyListItemSerializer
 from datahub.user.company_list.serializers import CompanyListSerializer
 
@@ -74,11 +76,54 @@ class CompanyListItemAPIPermissions(DjangoModelPermissions):
     """DRF permissions class for the company list item view."""
 
     perms_map = {
+        'GET': [
+            f'company.{CompanyPermission.view_company}',
+            f'company_list.{CompanyListItemPermissionCode.view_company_list_item}',
+        ],
         'PUT': [
             f'company.{CompanyPermission.view_company}',
             f'company_list.{CompanyListItemPermissionCode.add_company_list_item}',
         ],
     }
+
+
+class CompanyListItemViewSet(CoreViewSet):
+    """A view set for returning the contents of a company list."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (
+        IsAuthenticatedOrTokenHasScope,
+        CompanyListItemAPIPermissions,
+    )
+    serializer_class = CompanyListItemSerializer
+    filter_backends = (OrderingFilter,)
+    # Note that we want null to be treated as the oldest value when sorting by
+    # how long ago the interaction happened. This happens automatically when sorting by
+    # latest_interaction_time_ago (as opposed to sorting by latest_interaction_date in
+    # descending order)
+    ordering = (
+        'latest_interaction_time_ago',
+        '-latest_interaction_created_on',
+        'latest_interaction_id',
+    )
+    queryset = get_company_list_item_queryset()
+
+    def initial(self, request, *args, **kwargs):
+        """
+        Raise an Http404 if user's company list specified in the URL path does not exist.
+        """
+        super().initial(request, *args, **kwargs)
+
+        if not CompanyList.objects.filter(
+            pk=self.kwargs['company_list_pk'],
+            adviser=self.request.user,
+        ).exists():
+            raise Http404()
+
+    def filter_queryset(self, queryset):
+        """Filter the query set to the items relating to the authenticated users."""
+        queryset = super().filter_queryset(queryset)
+        return queryset.filter(list__pk=self.kwargs['company_list_pk'])
 
 
 class CompanyListItemAPIView(APIView):
--- a/changelog/adviser/company-list-list-item.api.rst
+++ b/changelog/adviser/company-list-list-item.api.rst
@@ -1,0 +1,38 @@
+The following endpoint was added:
+
+``GET /v4/company-list/<company list ID>/item``
+
+It lists all the companies on the authenticated user's selected list, with responses in the following format::
+
+    {
+        "count": <int>,
+        "previous": <url>,
+        "next": <url>,
+        "results": [
+            {
+                "company": {
+                    "id": <string>,
+                    "archived": <boolean>,
+                    "name": <string>,
+                    "trading_names": [<string>, <string>, ...]
+                },
+                "created_on": <ISO timestamp>,
+                "latest_interaction": {
+                    "id": <string>,
+                    "created_on": <ISO timestamp>,
+                    "date": <ISO date>,
+                    "subject": <string>
+                }
+            },
+            ...
+        ]
+    }
+
+
+``latest_interaction`` may be ``null`` if the company has no interactions.
+
+Results are sorted by ``latest_interaction.date`` in reverse chronological order, with ``null`` values last.
+
+The endpoint has pagination in line with other endpoints; to retrieve all results pass a large value for the ``limit`` query parameter (e.g. ``?limit=1000``).
+
+If selected list does not exist, the endpoint will return 404 status code.
--- a/datahub/user/company_list/queryset.py
+++ b/datahub/user/company_list/queryset.py
@@ -8,7 +8,7 @@ from datahub.user.company_list.models import CompanyListItem
 
 def get_company_list_item_queryset():
     """
-    Returns an annotated query set used by LegacyCompanyListViewSet.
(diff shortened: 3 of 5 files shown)


datahub/user/company_list/tests/test_views.py
```diff
@@ -79,9 +97,227 @@ def test_lists_are_sorted_by_name(self):
...
+        interaction with the company.
+
+        Note that we want companies without any interactions to be sorted last.
+        """
+        # These dates are in the order we expect them to be returned
+        interaction_dates = [
+            datetime(2019, 10, 8, tzinfo=utc),
+            datetime(2016, 9, 7, tzinfo=utc),
+            datetime(2009, 5, 6, tzinfo=utc),
+            None,
```
currycoder: I'm having difficulty understanding why `None` is in this list?  It seems to just complicate some of the subsequent code.  Could you explain the benefit?

elcct: When we sort by a date of latest interaction, if a company doesn't have any interactions will have last interaction date as Null. We want these companies to be at the bottom of the list. This test checks if that indeed is happening.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+def _get_queryset_for_list_adviser_and_company(company_list, adviser, company):
+    return CompanyListItem.objects.filter(
+        list=company_list,
+        list__adviser=adviser,
+        company=company,
+    )
+
+
+def _get_list_item_collection_url(company_list_pk):
+    return list_item_collection_url(kwargs={'company_list_pk': company_list_pk})
```
reupen: Do we need `list_item_collection_url`? If we keep it, I think we should give it a better name.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+            for result in results
+        ]
+        expected_interaction_dates = [
+            format_date_or_datetime(date_.date()) if date_ else None
+            for date_ in interaction_dates
+        ]
+        assert actual_interaction_dates == expected_interaction_dates
+
+
+def _get_queryset_for_list_adviser_and_company(company_list, adviser, company):
```
reupen: This doesn't seem to be used.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+        api_client = self.create_api_client(user=user)
+
+        url = _get_list_item_collection_url(company_list.pk)
+        response = api_client.get(url)
+
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_with_no_list_items(self):
+        """
+        Test that an empty list is returned if the user does not have any companies on their
```
reupen: It would be good to rephrase this docstring a bit since users can have multiple lists now.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+        url = _get_list_item_collection_url(company_list.pk)
+        response = self.api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['results'] == []
+
+    def test_with_list_that_does_not_exist(self):
+        """
+        Test that an empty list is returned if the user does not have any companies on their
```
reupen: The docstring needs updating here.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+        list.
+        """
+        url = _get_list_item_collection_url(uuid4())
+        response = self.api_client.get(url)
+
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    def test_with_only_items_on_other_users_list(self):
+        """
+        Test that an empty list is returned if the user has no companies on their list,
```
reupen: Also would be good to update this docstring, since it sounds like each user has one list.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+
+        url = _get_list_item_collection_url(company_list.pk)
+        response = self.api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['results'] == []
+
+    def test_with_multiple_user_lists(self):
+        """Test that user can list their own lists."""
```
reupen: Would be good to expand this docstring a bit.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +104,197 @@ def test_lists_are_sorted_by_name(self):
...
+        creation_date = datetime(2018, 1, 2, tzinfo=utc)
+
+        # other items on other users lists, that shouldn't appear in results
+        CompanyListItemFactory.create_batch(10)
```
reupen: We have `test_with_only_items_on_other_users_list()` above, do we need to consider other users in this test?

I think this test could be simplified in general – something like create two lists for the user, create some items for both lists, make the request for one of  the lists and check we get the right items back.

datahub/user/company_list/views.py
```diff
@@ -22,3 +42,42 @@ class CompanyListViewSet(CoreViewSet):
...
+    )
+    queryset = get_company_list_item_queryset()
+
+    def initial(self, request, *args, **kwargs):
+        """
+        Raise an Http404 if company list specified in the URL path does not exist.
+        """
+        super().initial(request, *args, **kwargs)
+
+        if not CompanyList.objects.filter(pk=self.kwargs['company_list_pk']).exists():
```
reupen: We should filter by `adviser` here (rather than in `filter_queryset()`) to be consistent with the other views (so a 404 is returned rather than an empty list).

Also, I think we need to add a test for trying to list another user's list.

datahub/user/company_list/views.py
```diff
@@ -22,3 +42,42 @@ class CompanyListViewSet(CoreViewSet):
...
+        """
+        Raise an Http404 if user's company list specified in the URL path does not exist.
+        """
+        super().initial(request, *args, **kwargs)
+
+        if not CompanyList.objects.filter(
+            pk=self.kwargs['company_list_pk'],
+            adviser=self.request.user,
+        ).exists():
+            raise Http404
```
reupen:
```suggestion
            raise Http404()
```

datahub/user/company_list/views.py
```diff
@@ -22,3 +42,42 @@ class CompanyListViewSet(CoreViewSet):
...
+
+    def filter_queryset(self, queryset):
+        """Filter the query set to the items relating to the authenticated users."""
+        queryset = super().filter_queryset(queryset)
+        return queryset.filter(list__pk=self.request.parser_context['kwargs']['company_list_pk'])
```
reupen: This should work:

```suggestion
        return queryset.filter(list__pk=self.kwargs['company_list_pk'])
```

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +102,192 @@ def test_lists_are_sorted_by_name(self):
...
+
+        url = _get_list_item_collection_url(company_list.pk)
+        response = self.api_client.get(url)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['results'] == []
+
+    def test_with_multiple_user_lists(self):
+        """Test that user who owns multiple lists can list all their contents."""
```
reupen: I still think this test is a bit too complicated – did you see my earlier suggestion:

> I think this test could be simplified in general – something like create two lists for the user, create some items for both lists, make the request for one of the lists and check we get the right items back.

elcct: I thought it would be better to check the contents of all lists in case the content of the other list goes into vacuum? But sure I can look into simplifying that.

datahub/user/company_list/views.py
elcct: Thanks!

datahub/user/company_list/views.py
elcct: Thank you!

datahub/user/company_list/tests/test_views.py
elcct: So I removed the code that creates lists for other users but left checking all lists as I think it will be better to see that multiple lists' content is accessible.

PR #2023, opened 2019-08-28
Add endpoint for deleting company list contents

elcct:
### Description of change

This add the following endpoint:

  - ``DELETE /v4/company-list/<company list ID>/item/<company ID>``

It removes a company from the user's own selected list of companies.

If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or a list doesn't belong to the user, a 404 will be returned.

### Checklist

### DIFF
--- a/changelog/adviser/company-list-remove-item.api.rst
+++ b/changelog/adviser/company-list-remove-item.api.rst
@@ -1,0 +1,7 @@
+The following endpoint was added:
+
+  - ``DELETE /v4/company-list/<company list ID>/item/<company ID>``
+
+This removes a company from the user's own selected list of companies.
+
+If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or a list doesn't belong to the user, a 404 will be returned.
--- a/datahub/user/company_list/views.py
+++ b/datahub/user/company_list/views.py
@@ -84,6 +84,10 @@ class CompanyListItemAPIPermissions(DjangoModelPermissions):
             f'company.{CompanyPermission.view_company}',
             f'company_list.{CompanyListItemPermissionCode.add_company_list_item}',
         ],
+        'DELETE': [
+            f'company.{CompanyPermission.view_company}',
+            f'company_list.{CompanyListItemPermissionCode.delete_company_list_item}',
+        ],
     }
 
 
@@ -164,6 +168,19 @@ class CompanyListItemAPIView(APIView):
                 'modified_by': adviser,
             },
         )
+
+        return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+    @method_decorator(transaction.non_atomic_requests)
+    def delete(self, request, company_list_pk, company_pk=None, format=None):
+        """
+        Delete a CompanyListItem for the selected list of authenticated user and
+        specified company if it exists.
+        """
+        company_list = get_object_or_404(CompanyList, pk=company_list_pk, adviser=request.user)
+        company = get_object_or_404(Company, pk=company_pk)
+
+        self.queryset.filter(list=company_list, company=company).delete()
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
 
--- a/datahub/user/company_list/test/test_views.py
+++ b/datahub/user/company_list/test/test_views.py
@@ -481,20 +481,21 @@ class TestDeleteCompanyListView(APITestMixin):
         assert response.status_code == status.HTTP_404_NOT_FOUND
 
 
+@pytest.mark.parametrize('http_method', ('delete', 'put'))
 class TestCompanyListItemAuth(APITestMixin):
     """Tests authentication and authorisation for the company list item views."""
 
-    def test_returns_401_if_unauthenticated(self, api_client):
+    def test_returns_401_if_unauthenticated(self, api_client, http_method):
         """Test that a 401 is returned for an unauthenticated user."""
         company = CompanyFactory()
         company_list = CompanyListFactory()
 
         url = _get_list_item_url(company_list.pk, company.pk)
-        response = api_client.put(url)
+        response = api_client.generic(http_method, url)
 
         assert response.status_code == status.HTTP_401_UNAUTHORIZED
 
-    def test_returns_403_if_without_permissions(self, api_client):
+    def test_returns_403_if_without_permissions(self, api_client, http_method):
         """Test that a 403 is returned for a user with no permissions."""
         company = CompanyFactory()
         user = create_test_user(dit_team=TeamFactory())
@@ -503,7 +504,7 @@ class TestCompanyListItemAuth(APITestMixin):
         url = _get_list_item_url(company_list.pk, company.pk)
         api_client = self.create_api_client(user=user)
 
-        response = api_client.put(url)
+        response = api_client.generic(http_method, url)
 
         assert response.status_code == status.HTTP_403_FORBIDDEN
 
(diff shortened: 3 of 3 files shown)


datahub/user/company_list/urls.py
```diff
@@ -16,6 +16,11 @@
         ),
         name='list-collection',
     ),
+    path(
+        'company-list/<uuid:company_list_pk>/item/<uuid:company_pk>',
```
mcbhenwood: No biggie, but I thought we'd discussed the idea that we didn't need the `/item/` element in the path. Was there a reason we kept it in?

changelog/adviser/company-list-remove-item.api.rst
```diff
@@ -0,0 +1,7 @@
+The following endpoint was added:
+
+  - ``DELETE /v4/company-list/<company list ID>/item/<company ID>``
+
+This removes a company from the user's own selected list of companies.
+
+If the operation is successful, a 204 status code will be returned. If there is no company list with specified company list ID or a company with the specified company ID or a list doesn't belong to the user, a 404 will be returned.
```
mcbhenwood: I can see a test for the case where the company exists but is not on a list which belongs to the user. However I cannot see tests for:

1. Company does not exist in DH
2. Specified company list does not exist
3. List does not belong to user

datahub/user/company_list/views.py
```diff
@@ -22,3 +47,40 @@ class CompanyListViewSet(CoreViewSet):
...
+        """
+        company = get_object_or_404(Company, pk=company_pk)
+
+        to_delete_qs = self.queryset.filter(
+            list_id=company_list_pk,
+            list__adviser=request.user,
+            company=company,
+        )
+
+        if not to_delete_qs.exists():
```
mcbhenwood: As per above, missing these tests:

1. Specified company list does not exist
2. List does not belong to user

datahub/user/company_list/views.py
```diff
@@ -22,3 +47,40 @@ class CompanyListViewSet(CoreViewSet):
...
+    queryset = CompanyListItem.objects.all()
+    serializer_class = CompanyListItemSerializer
+
+    @method_decorator(transaction.non_atomic_requests)
+    def delete(self, request, company_list_pk, company_pk=None, format=None):
+        """
+        Delete a CompanyListItem for the selected list of authenticated user and
+        specified company if it exists.
+        """
+        company = get_object_or_404(Company, pk=company_pk)
```
mcbhenwood: As per above, missing test for when company does not exist in DH

datahub/user/company_list/urls.py
elcct: It is kept in becuase otherwise it is clashing with the list api endpoints.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +92,100 @@ def test_lists_are_sorted_by_name(self):
...
+
+        url = _get_list_item_url(company_list.pk, company.pk)
+        response = self.api_client.delete(url)
+
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert response.content == b''
+
+        assert not CompanyListItem.objects.filter(list=company_list, company=company).exists()
+
+    def test_with_new_item(self):
```
mcbhenwood: This test could do with a clearer name I think.

mcbhenwood: Something like `test_existing_company_not_on_list`

datahub/user/company_list/views.py
```diff
@@ -22,3 +47,40 @@ class CompanyListViewSet(CoreViewSet):
...
+
+        to_delete_qs = self.queryset.filter(
+            list_id=company_list_pk,
+            list__adviser=request.user,
+            company=company,
+        )
+
+        if not to_delete_qs.exists():
+            raise Http404
```
reupen: If we're following the same model as the legacy views (and also adding a company to the list) then it should behave the same regardless of whether the company was on the list or not, no? (Checking that the list belongs to the user could be done separately.)

elcct: This is good suggestion. I think the way I added this is incorrect and indeed the delete should return the same result regardless whether the company is on the list or not. This is now fixed. Thank you!

changelog/adviser/company-list-remove-item.api.rst
elcct: Great suggestion! I have added those tests now.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +93,133 @@ def test_lists_are_sorted_by_name(self):
...
+        response = self.api_client.delete(url)
+
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+        assert response.content == b'{"detail":"Not found."}'
+
+        # company has not been deleted from another user list
+        assert CompanyListItem.objects.filter(list=company_list, company=company).exists()
+
+    def test_with_list_that_does_not_exist(self):
+        """Test that a company cannnot be removed from another user's list."""
```
reupen: The docstring needs updating here.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +93,133 @@ def test_lists_are_sorted_by_name(self):
         response_data = response.json()
         actual_list_names = [result['name'] for result in response_data['results']]
         assert actual_list_names == expected_list_names
+
+
+class TestDeleteCompanyListItemAPIView(APITestMixin):
+    """Tests for the DELETE method in CompanyListItemAPIView."""
+
+    def test_with_existing_item(self):
+        """Test that a company can be removed from the authenticated user's list."""
```
reupen: It would be good to rephrase the docstrings slightly as they sound like the user has one list, when it's the list that was specified in the URL path.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +93,133 @@ def test_lists_are_sorted_by_name(self):
...
+        assert response.content == b''
+
+        assert not CompanyListItem.objects.filter(list=company_lists[0], company=company).exists()
+
+        # The company exists on 4 other lists
+        assert CompanyListItem.objects.filter(list__in=company_lists, company=company).count() == 4
+
+
+def _get_list_item_url(company_list_pk, company_pk):
+    return list_item_url(
```
reupen: We removed `list_item_url` in one of the other PRs, should we do the same here?

elcct: Yes, definitely I am going to remove it.

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +91,131 @@ def test_lists_are_sorted_by_name(self):
...
+        response = self.api_client.delete(url)
+
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+        assert response.content == b'{"detail":"Not found."}'
+
+        # company has not been deleted from another user list
+        assert CompanyListItem.objects.filter(list=company_list, company=company).exists()
+
+    def test_with_list_that_does_not_exist(self):
+        """Test that a company cannot be removed from another user's lists."""
```
mcbhenwood: This doc string isn't accurate; looks like a paste from the test function above :)

datahub/user/company_list/tests/test_views.py
```diff
@@ -85,3 +91,131 @@ def test_lists_are_sorted_by_name(self):
...
+        response = self.api_client.delete(url)
+
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert response.content == b''
+
+        assert not CompanyListItem.objects.filter(list=company_list, company=company).exists()
+
+    def test_with_company_not_on_the_list(self):
+        """
+        Test that 204 status is returned regardless if company is on the authenticated user's list.
```
mcbhenwood: "...if company is NOT on the...."

datahub/user/company_list/tests/test_views.py
elcct: :blush:

PR #2026, opened 2019-08-29
Adjust OMIS to use notification app based on a feature flag

currycoder:
### Description of change

This allows us to switch OMIS notifications to be sent through the new `datahub.notification` app by activating a feature flag.  It allows us to switch over to the new app in a way that is low risk - as we can easily switch back to the existing implementation if there are any problems.

Due to the nature of making this switch achievable with a feature flag, the logic which triggers omis notifications has gotten a little ugly.  Once we have successfully switched to the `datahub.notification` app, a PR will follow which cleans up `datahub.omis.notification`.  There's some easy refactors that can be made to make this much cleaner.

This follows on from previous work to allow `datahub.notification` to work with multiple API keys/client instances. https://github.com/uktrade/data-hub-leeloo/pull/1948

### Checklist

### DIFF
--- a/datahub/omis/notification/client.py
+++ b/datahub/omis/notification/client.py
@@ -7,8 +7,14 @@ from django.conf import settings
 from notifications_python_client.notifications import NotificationsAPIClient
 
 from datahub.core.thread_pool import submit_to_thread_pool
+from datahub.feature_flag.utils import is_feature_flag_active
+from datahub.notification.constants import NotifyServiceName
+from datahub.notification.notify import notify_by_email
 from datahub.omis.market.models import Market
-from datahub.omis.notification.constants import Template
+from datahub.omis.notification.constants import (
+    OMIS_USE_NOTIFICATION_APP_FEATURE_FLAG_NAME,
+    Template,
+)
 from datahub.omis.region.models import UKRegionalSettings
 
 
@@ -17,13 +23,7 @@ logger = getLogger(__name__)
 
 def send_email(client, **kwargs):
     """Send email and catch potential errors."""
-    data = dict(kwargs)
-
-    # override recipient if needed
-    if settings.OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL:
-        data['email_address'] = settings.OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL
-
-    client.send_email_notification(**data)
+    client.send_email_notification(**kwargs)
 
 
 class Notify:
@@ -55,9 +55,22 @@ class Notify:
                 stacklevel=2,
             )
 
-    def _send_email(self, **kwargs):
+    def _send_email(self, **data):
         """Send email in a separate thread."""
-        submit_to_thread_pool(send_email, self.client, **kwargs)
+        # override recipient if needed
+        if settings.OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL:
+            data['email_address'] = settings.OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL
+
+        use_notification_app = is_feature_flag_active(OMIS_USE_NOTIFICATION_APP_FEATURE_FLAG_NAME)
+        if use_notification_app:
+            notify_by_email(
+                data['email_address'],
+                data['template_id'],
+                data.get('personalisation'),
+                NotifyServiceName.omis,
+            )
+        else:
+            submit_to_thread_pool(send_email, self.client, **data)
 
     def _prepare_personalisation(self, order, data=None):
         """Prepare the personalisation data with common values."""
--- a/datahub/omis/notification/test/test_templates.py
+++ b/datahub/omis/notification/test/test_templates.py
@@ -1,11 +1,16 @@
 import pytest
 from dateutil.parser import parse as dateutil_parse
 from django.conf import settings
+from django.test.utils import override_settings
 
 from datahub.company.test.factories import AdviserFactory
 from datahub.core.constants import UKRegion
+from datahub.feature_flag.test.factories import FeatureFlagFactory
+from datahub.notification.core import NotifyGateway
+from datahub.notification.tasks import send_email_notification
 from datahub.omis.market.models import Market
 from datahub.omis.notification.client import Notify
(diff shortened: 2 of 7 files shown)


datahub/omis/notification/test/test_templates.py
```diff
@@ -33,7 +33,7 @@ class TestTemplates:
     not strictly mandatory.
     """
 
-    def test_order_info(self, settings):
+    def test_order_info(self, settings, notify_client):
```
reupen: Do we need to check the result of the task in these tests for when they are using Celery? (`task.apply_async()` returns a task result and e.g. calling `task_result.get()` will propagate any exceptions and there is also `task_result.successful()`.)

currycoder: Thanks Reupen, nice observation here.  It switched me on to the fact that this was only really testing half of the functionality - as the clients within the `datahub.notification` app were still mocked, regardless of instantiating a new omis notification client in these tests.

I've reworked things a little bit to use a new `spy_notify_client` fixture for these tests.  This fixture will ensure that the mocked notify clients are replaced with spy implementations (according to whether the feature flag is set or not).  This then allows us to interrogate whether the spies were called.

As for your actual comment here about interrogating the task result, I'm finding it a little difficult to conceptualise how we might get at the task result from these test cases.  Unfortunately we can't yet change any of the contracts within `datahub.omis.notify.client` to expose these task results because we are still allowing the functionality to be optional behind a feature flag.  Do you have any ideas about how I might achieve it?

reupen: I think as these tests are essentially integration tests to check that the templates and code agree on things (e.g. personalisation parameters, template IDs) then it's probably not very interesting to test both the Celery and thread pool cases here.

So I think we could restrict these particular tests to just test the Celery case to simplify things. Then we could modify the functions that send emails to return the task result (and hence the tests would then have access to it). Alternatively, you may be able to use the [`track_return_values`](https://github.com/uktrade/data-hub-leeloo/blob/22ceaa936a62471e330e52c1b21c73605623034b/conftest.py#L88) fixture to get the task results.

datahub/omis/notification/test/test_templates.py
```diff
@@ -33,25 +34,24 @@ class TestTemplates:
     not strictly mandatory.
     """
 
-    def test_order_info(self, settings):
+    @override_settings(OMIS_NOTIFICATION_API_KEY=settings.OMIS_NOTIFICATION_TEST_API_KEY)
```
reupen: Does `@override_settings()` work on the class here (rather than each method)?

currycoder: It didn't as the class didn't inherit from django test case.. I've reworked things here to make things more DRY in this respect by making this part of a test fixture instead.

datahub/omis/notification/test/test_templates.py
currycoder: Thanks for these thoughts Reupen.

I've decided to rename the `TemplateTest` class to `TestTemplatesLegacyOMISNotification` and keep these tests identical to how they were before.  I appreciate your point about the value of having full integration tests for both implementations, but right now I see it as easiest to leave these unchanged with the view that we will remove them in the cleanup PR that will follow successfully switching over to the new notification app.

I've added a new `TemplateTest` class which sets up it's own `Notify` object (and patches the `datahub.notification` client to call through to the web service instead of using a mock).  This largely apes the test cases present in the original class, but verifies that the tasks were successful with `track_return_values` as you suggested.

datahub/omis/notification/test/test_templates.py
```diff
@@ -19,6 +22,34 @@
...
+def end_to_end_notify(monkeypatch, settings):
+    """
+    A fixture for a notify client which uses the new datahub.notification app
+    under the hood and calls through to the GOVUK notify service (with
+    settings.OMIS_NOTIFICATION_TEST_API_KEY).
+
+    By contrast, our other test cases will use mocked clients so that no actual
+    web requests are made to GOVUK notify.
+    """
+    settings.OMIS_NOTIFICATION_API_KEY = settings.OMIS_NOTIFICATION_TEST_API_KEY
```
reupen: It should be possible to do this in a safer way similar here:

https://github.com/uktrade/data-hub-leeloo/blob/05caa3527b5c2f9b386226003b556d07e40c3f26/datahub/documents/test/test_views.py#L19-L20

datahub/omis/notification/test/test_templates.py
```diff
@@ -28,7 +61,162 @@ class TestTemplates:
...
     and it's meant to check that the templates in GOV.UK notifications have not been
     changed.
+    If `OMIS_NOTIFICATION_TEST_API_KEY` is not set they will not run as they are
+    not strictly mandatory.
+    """
+
+    def _assert_tasks_successful(self, task_count, return_value_tracker):
+        task_results = return_value_tracker.return_values
+        assert len(task_results) == task_count
+        assert all(task_result.successful() for task_result in task_results)
```
reupen: I don't believe `.successful()` propagates the exception. If it fails, will the reason be reported? 

We could call `.get()` or similar on the task result instead (and possibly wrap it with a `try`/`except` and `pytest.fail()`).

datahub/omis/notification/client.py
```diff
@@ -57,7 +57,22 @@ def __init__(self):
 
     def _send_email(self, **kwargs):
         """Send email in a separate thread."""
-        submit_to_thread_pool(send_email, self.client, **kwargs)
+        data = dict(kwargs)
```
alixedi: This can probably be replaced by:

```
def _send_email(self, **data):
    """...""
    ...
```

datahub/omis/notification/client.py
```diff
@@ -57,7 +57,22 @@ def __init__(self):
...
+        # override recipient if needed
+        if settings.OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL:
+            data['email_address'] = settings.OMIS_NOTIFICATION_OVERRIDE_RECIPIENT_EMAIL
+
+        use_notification_app = is_feature_flag_active(OMIS_USE_NOTIFICATION_APP_FEATURE_FLAG_NAME)
+        if use_notification_app:
+            notify_by_email(
+                data['email_address'],
+                data['template_id'],
+                data.get('personalisation', None),
```
alixedi: `None` here is redundant.

PR #2027, opened 2019-08-29
Add a validator to restrict unarchiving companies with certain archive reasons

currycoder:
# Description of change

This adds an `unarchive_validator` to the Company viewset which prevents unarchiving companies that were archived with a specific reason.

The business case here is to ensure that a company record which was determined to be invalid (and so archived by DH support) after a DNB investigation cannot be subsequently unarchived by Data Hub users.

I believe we would add to the canned archive reasons shown on the frontend here:
archive

### Things of note
- It would be good to have some discussion around this approach - I'm a little uncomfortable about the approach of checking whether `archived_reason in RESTRICTED_REASONS`.  Especially since `RESTRICTED_REASONS` just consists of raw strings and that a reason is a user friendly string which could be liable to change in future.  This is complicated in that the canned reasons available on the frontend do not appear to be driven by backend data.
- The string for the restricted reason is not final.

### Checklist

### DIFF
--- a/datahub/company/test/test_company_views_archiving.py
+++ b/datahub/company/test/test_company_views_archiving.py
@@ -104,3 +104,45 @@ class TestUnarchiveCompany(APITestMixin):
                     'record for the following reason: Duplicate record.',
                 ],
         }
+
+    def test_cannot_unarchive_transferred_company(self):
+        """Test that a transferred company cannot be unarchived."""
+        transfer_reason = 'duplicate'
+        transfer_company = CompanyFactory()
+        company = CompanyFactory(
+            archived=True,
+            transferred_to=transfer_company,
+            transfer_reason=transfer_reason,
+        )
+        url = reverse('api-v4:company:unarchive', kwargs={'pk': company.id})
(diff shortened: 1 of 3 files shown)


reupen:
> It would be good to have some discussion around this approach - I'm a little uncomfortable about the approach of checking whether archived_reason in RESTRICTED_REASONS. Especially since RESTRICTED_REASONS just consists of raw strings and that a reason is a user friendly string which could be liable to change in future. This is complicated in that the canned reasons available on the frontend do not appear to be driven by backend data.

Yes, I agree. Presumably we will be adding a field to store that a company has failed the investigation, so are we able to use that instead (similar to `NotATransferredCompanyValidator`)?

currycoder:
We are going to experiment with turning off company unarchiving in the frontend (behind a feature flag).  Users being able to easily unarchive doesn't make the best of sense and it's hard to understand how this feature was intended to be used.  It also appears from data that the archive feature is not well utilised by users.

The user feedback (if any) should inform the work that follows this.

PR #2032, opened 2019-08-30
Deprecate unused contact search sorting options

reupen:
### Description of change

This deprecates various sorting options for contact search that aren't being used by the front end.

This is so that we can simplify the code base and make future refactoring easier. It will also allow us to make the contact search model leaner.

Many of these sorting options were also untested.

Logging has been added to make sure we don't break things inadvertently.

The approach used is the same as in #1450.

[The options that are in use in the front end can be seen here.](https://github.com/uktrade/data-hub-frontend/blob/ebccec4a08ea01b50957a357c7b5d709f25f0583/src/apps/contacts/macros.js#L26-L34)

### Checklist

### DIFF
--- a/datahub/search/contact/serializers.py
+++ b/datahub/search/contact/serializers.py
@@ -1,3 +1,5 @@
+from logging import getLogger
+
 from rest_framework import serializers
 
 from datahub.search.serializers import (
@@ -5,6 +7,9 @@ from datahub.search.serializers import (
     SingleOrListField,
     StringUUIDField,
 )
+
+
+logger = getLogger(__name__)
 
 
 class SearchContactQuerySerializer(EntitySearchQuerySerializer):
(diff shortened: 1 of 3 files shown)


datahub/search/contact/serializers.py
```diff
@@ -21,30 +26,48 @@ class SearchContactQuerySerializer(EntitySearchQuerySerializer):
     created_by = SingleOrListField(child=StringUUIDField(), required=False)
     created_on_exists = serializers.BooleanField(required=False)
 
-    # TODO: Deprecate unused sort by values, and add logging to double-check that they aren't
-    #  being used
-    SORT_BY_FIELDS = (
-        'address_country.name',
+    # Deprecated sorting options
+    # TODO: Remove following deprecation period.
```
elcct: I am thinking if it will be helpful to add a date in the comment, when this is deprecated.

reupen: I was thinking that we might want to start logging them centrally, but probably something for another time.

PR #2048, opened 2019-09-03
Change sync_es management command to use Celery

reupen:
### Description of change

This makes some changes to the `sync_es` management command. It:

- makes the task run asynchronously by default (using Celery)
- removes the `--batch_size` argument as it's not really used
- adds a `--foreground` argument for running it synchronously (mainly for convenience)

The reason for this change is to make things more consistent, and also as syncing (asynchronously) using Celery is generally faster due to parallelism. (This change will also benefit the staging refresh job.)

`--foreground` is similar to running `CELERY_TASK_ALWAYS_EAGER=True ./manage.py sync_es` so we could potentially not bother adding `--sync`.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Dependencies:
     ```shell
     ./manage.py loaddata fixtures/test_data.yaml
 
-    ./manage.py sync_es
+    ./manage.py sync_es --foreground
     ```
 
 13. Create a superuser:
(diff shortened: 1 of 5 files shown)


mcbhenwood:
To the untrained eye, doing `./manage.py sync_es --sync` looks like the `--sync` switch is actually telling the command to do the sync for real, which possibly suggests that omitting the switch would result in a dry run. This is I think a result of the command in that form making use of the word "sync" in two places; the first meaning "to synchronise ES" and the second meaning "synchronously". That differentiation in meaning isn't clear since both forms shorten to just "sync".

How about something like `--foreground` instead?

elcct:
I agree that `--sync` switch may be confusing. Maybe we could just remove it?

reupen:
I've renamed `--sync` to `--foreground`.

PR #2049, opened 2019-09-03
WIP Feature/add imputed countries of interest fields

benpbenp:
### Checklist


mcbhenwood:
I appreciate it's a bit late in the process to ask this questions, but.... Is "imputed" really the term we want?

> impute | ɪmˈpjuːt |
> verb [with object]
>  1 represent (something, especially something undesirable) as being done or possessed by someone; attribute: the crimes imputed to Richard.
>
> 2 _Finance_ assign (a value) to something by inference from the value of the products or processes to which it contributes: (as adjective imputed) : recovering the initial outlay plus imputed interest.

datahub/company/models/company.py
```diff
@@ -1,4 +1,6 @@
 """Company models."""
+from logging import getLogger
```
mcbhenwood: Usually the `forms would come after the `import`s in each group

datahub/company/models/company.py
```diff
@@ -241,6 +259,68 @@ def uk_based(self):
         united_kingdom_id = uuid.UUID(constants.Country.united_kingdom.value.id)
         return self.address_country.id == united_kingdom_id
 
+    @property
+    def combined_future_interest_countries(self):
```
mcbhenwood: Really I'd have wanted to see a whole bunch of tests and code changes in separate commits showing how this logic had been put together. i.e. a series of several atomic commits each of which contains one iteration of "Fail; Fix; Refactor" - each adding a new level of complexity to the operations. (i.e. pure Kent Beck TDD)

This logic is really complicated and adding in tests after the code is written will make it hard to manage refactoring etc later on.

I would go so far as to say that in its current state, this code is no higher than prototype and it would need to be removed and rebuilt, test by test, commit by commit, before we could merge it in.

datahub/company/models/company.py
```diff
@@ -241,6 +259,68 @@ def uk_based(self):
...
+        future interest countries.
+        1. future_interest_countries contains manually entered list of countries
+        2. imputed_future_interest_countries comes from external sources and
+            is periodically refreshed.
+        3. disregarded_future_interest_countries are those countries in 2. which
+            a user has manually chosen to disregard.
+
+        The combined view we return is 1 + (2 - 3)
+        """
+        return sorted(list(
```
mcbhenwood: I feel there ought to be a way to do this within a single ORM query which returns just those countries which are appropriate. Doing it this way, with 3 separate queries and a bunch of set operations seems more expensive.

benpbenp: You are right, but in the context of where this property is used, all of those many-to-many relations have been prefetched, so there is no additional query doing it this way, where as a single custom query would have to be an additional query.

benpbenp:
@mcbhenwood No it's not too late... maybe "reputed" is a better fit?

Also, yes this is indeed basically a prototype. We can break it into smaller commits -- I don't think you'll be disappointed with the test coverage!

datahub/company/models/company.py
```diff
@@ -241,6 +259,68 @@ def uk_based(self):
...
+            set(self.future_interest_countries.all()) |
+            (set(self.imputed_future_interest_countries.all()) -
+                set(self.disregarded_future_interest_countries.all()))
+        ), key=lambda c: c.name)
+
+    @combined_future_interest_countries.setter
+    def combined_future_interest_countries(self, input_countries):
```
reupen: Typically this repo prefers methods to properties (especially for things containing significant logic) since it is more explicit. For example, this setter submits changes to the database which won't be obvious in the code that assigns to the attribute.

So, it would be good to replace the property with a couple of methods.

(It should be possible to use `source` on the serialiser field to make it call the `get_*` method.)

Also, it would be good to use the `@transaction.atomic` decorator on this so partial changes aren't committed after an error part way through.

datahub/company/serializers.py
```diff
@@ -274,6 +274,12 @@ class CompanySerializer(PermittedFieldsModelSerializer):
     future_interest_countries = NestedRelatedField(
         meta_models.Country, many=True, required=False,
     )
+    # This accesses a Company @property.
+    # If future_interest_countries is also supplied, its changes will
+    # be applied after the changes to combined_future_interest_countries.
+    combined_future_interest_countries = NestedRelatedField(
+        meta_models.Country, many=True, required=False
```
reupen: Shouldn't this be `read_only=True` (rather than `required=False`)?

datahub/company/models/company.py
```diff
@@ -241,6 +259,68 @@ def uk_based(self):
...
+        user = set(self.future_interest_countries.all())
+        imputed  = set(self.imputed_future_interest_countries.all())
+        disregarded = set(self.disregarded_future_interest_countries.all())
+        # If we are removing a country that is user supplied, then we
+        # remove it from self.future_interest_countries (these are the manually
+        # entered countries).
+        # However, to remove a country which is in imputed_future_interest_countries,
+        # we have to add it to disregarded_future_interest_countries,
+        # because imputed will be refreshed periodically by external sources.
+        for dc in deleted:
```
reupen: It looks like this logic could be simplified:

```python
to_disregard = user & deleted
self.future_interest_countries.remove(*to_disregard)  # same as self.future_interest_countries.remove(*deleted) 
self.disregarded_future_interest_countries.add(*to_disregard)
```

(Similar comment for some of the other logic.)

datahub/company/models/company.py
```diff
@@ -241,6 +259,68 @@ def uk_based(self):
...
+        1. future_interest_countries contains manually entered list of countries
+        2. imputed_future_interest_countries comes from external sources and
+            is periodically refreshed.
+        3. disregarded_future_interest_countries are those countries in 2. which
+            a user has manually chosen to disregard.
+
+        The combined view we return is 1 + (2 - 3)
+        """
+        return sorted(list(
+            set(self.future_interest_countries.all()) |
```
reupen: It looks like from the logic below that `future_interest_countries` will always contain all countries. Is that right?

datahub/company/admin/company.py
```diff
@@ -157,6 +157,8 @@ class CompanyAdmin(BaseModelAdminMixin, VersionAdmin):
                     'export_experience_category',
                     'export_to_countries',
                     'future_interest_countries',
+                    'imputed_future_interest_countries',
```
reupen: Should these be read-only as well?

datahub/company/models/company.py
benpbenp: All the ones that the user manually enters via the company endpoint, yes. However, there is an additional change in the works whereby the imputed/reputed future interest countries will be updated in bulk by a management command operating on an export of data-science-data. When that data comes in, there will be countries in imputed/reputed that are not in future_interest_countries.

alixedi:
I would like to review this but I have no context of why we are making this change. Can you please put something in the PR description? Thanks 🙏

PR #2057, opened 2019-09-06
Interaction search one list group tier filter

currycoder:
### Description of change
This PR adds `company_one_list_group_tier` to the interaction search document as well as a filter `company__one_list_group_tier` to the interaction search endpoint.

This will enable Data Hub users to filter interaction searches by the one list group tier of companies.  There will be some frontend work to follow in order to hook this up. 

### Checklist

### DIFF
--- a/changelog/interaction/interaction-search-one-list-group-tier.feature.rst
+++ b/changelog/interaction/interaction-search-one-list-group-tier.feature.rst
@@ -1,0 +1,20 @@
+The ``/v3/search/interaction/`` endpoint was modified to return 
+``company_one_list_group_tier`` in search results. This will be in the following
+format:
+
+
+```
+...
+"company_one_list_group_tier": {
+    "id": "b91bf800-8d53-e311-aef3-441ea13961e2",
+    "name": "Tier A - Strategic Account"
+}
+...
+```
+
+The value could alternatively be null (if the interaction's company does not
+have a one list group tier).
+
+A filter was added to ``/v3/search/interaction/`` - ``company__one_list_group_tier`` -
+which allows callers to filter interaction searches to companies attributed to a
+particular one list group tier.
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -312,6 +312,13 @@ class Company(ArchivableModel, BaseModel):
             return self.global_headquarters
         return self
 
+    @cached_property
+    def one_list_group_tier(self):
+        """
+        :returns: the One List Tier of the group this company is part of.
+        """
+        return self.get_one_list_group_tier()
+
     def get_one_list_group_tier(self):
         """
         :returns: the One List Tier of the group this company is part of.
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -312,13 +312,6 @@ class Company(ArchivableModel, BaseModel):
             return self.global_headquarters
         return self
 
-    @cached_property
-    def one_list_group_tier(self):
-        """
-        :returns: the One List Tier of the group this company is part of.
-        """
-        return self.get_one_list_group_tier()
-
     def get_one_list_group_tier(self):
         """
         :returns: the One List Tier of the group this company is part of.
--- a/datahub/search/interaction/apps.py
+++ b/datahub/search/interaction/apps.py
@@ -21,6 +21,7 @@ class InteractionSearchApp(SearchApp):
         'company__sector',
         'company__sector__parent',
         'company__sector__parent__parent',
+        'company__one_list_tier',
         'communication_channel',
         'investment_project',
         'investment_project__sector',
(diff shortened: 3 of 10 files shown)


datahub/search/interaction/apps.py
```diff
@@ -21,6 +21,7 @@ class InteractionSearchApp(SearchApp):
         'company__sector',
         'company__sector__parent',
         'company__sector__parent__parent',
+        'company__one_list_tier',
```
reupen: Do we need to add `company__global_headquarters__one_list_tier` as well?

datahub/company/models/company.py
```diff
@@ -286,6 +286,13 @@ def get_group_global_headquarters(self):
             return self.global_headquarters
         return self
 
+    @cached_property
+    def one_list_group_tier(self):
```
reupen: Is there a reason for adding this? Are we able to use `get_one_list_group_tier()` instead?

changelog/interaction/interaction-search-one-list-group-tier.feature.rst
```diff
@@ -0,0 +1,20 @@
+The ``/v3/search/interaction/`` endpoint was modified to return
```
reupen: This looks like an `api` news fragment rather a `feature` one (could have a feature one as well, but usually those are a bit higher-level and not go into API details).

Also, there are some extra trailing slashes on the URL paths in this news fragment.

datahub/search/interaction/test/test_views.py
```diff
@@ -111,6 +130,36 @@ def interactions(setup_es):
...
+                ],
+                dit_participants__adviser__first_name='Trevor',
+                dit_participants__adviser__last_name='Saleman',
+            ),
+            CompanyInteractionFactory(
+                subject='Regional Subsidiary chat',
+                date=dateutil_parse('2016-09-02T00:00:00Z'),
+                company=company_4,
+                contacts=[
+                    ContactFactory(company=company_3, first_name='Bob', last_name='Smith'),
```
reupen: Typo here – should be `company_4`.

datahub/search/interaction/test/test_views.py
```diff
@@ -50,6 +51,12 @@
 
 pytestmark = pytest.mark.django_db
 
+USED_ONE_LIST_TIER_IDS = [
+    'b91bf800-8d53-e311-aef3-441ea13961e2',
```
reupen: These values create a bit of a non-obvious dependency with the fixtures – would suggest not using the `interactions` fixture in the new tests and instead creating a few relevant interactions in those tests (e.g. `CompanyInteractionFactory(company__one_list_tier=...)`. (That keeps all the dependencies clear and the tests a bit more manageable.)

For the tier objects themselves, rather than using the IDs, we could either add a new factory to create some on demand or just pick some (at random) from the ones that already exist in the database.

datahub/company/models/company.py
currycoder: I added it so that I could use it here: https://github.com/uktrade/data-hub-leeloo/pull/2057/files#diff-cfbade75c2889f708945dbe195299436R92

My understanding is that I would need to add an extra utility function for this otherwise, as `dict_utils.computed_nested_id_name_dict` does not accept a callable.  Happy to discuss this.

changelog/interaction/interaction-search-one-list-group-tier.feature.rst
currycoder: Good spot!

datahub/search/interaction/test/test_views.py
currycoder: Agreed - thanks for suggesting a fix here.

datahub/company/models/company.py
elcct: Maybe it will be worth adding a comment why this is this way?

reupen: We have control of both so I don't think adding a property here that just calls the method is desirable – there shouldn't be anything to cache either, since it just accesses foreign keys (which will be cached automatically).

Ideally I think it would be a property of `company` in the interaction search model, but if it's easier to make another top-level field then OK. Something like:

```python
        'company_one_list_group_tier': lambda obj: dict_utils.id_name_dict(
            obj.company.get_one_list_group_tier() if obj.company else None,
        ),
```

could be used if we don't want to make `computed_nested_id_name_dict` accept the names of callables.

datahub/search/interaction/test/test_views.py
```diff
@@ -467,6 +472,123 @@ def test_filter_by_company_name(self, setup_es, name_term, matched_company_name)
...
+        company_1 = CompanyFactory(
+            name='Global HQ Ltd',
+            one_list_tier=one_list_tier,
+        )
+        company_2 = CompanyFactory(
+            name='Regional Subsidiary Ltd',
+            global_headquarters=company_1,
+        )
+        CompanyInteractionFactory(
+            subject='Global HQ chat',
```
reupen: It would be good to only override the fields needed to make these tests a bit more concise.

datahub/company/models/company.py
currycoder: Neat idea - thanks

PR #2065, opened 2019-09-09
Add serializer to create stub companies for DNB investigation

alixedi:
### Description of change

This PR adds a new serialiser: DNBCompanyInvestigation. This will be used in a new endpoint that lets users create a stub company to be investigated by DNB.

### Checklist

### DIFF
--- a/datahub/dnb_api/serializers.py
+++ b/datahub/dnb_api/serializers.py
@@ -124,10 +124,68 @@ class DUNSNumberSerializer(serializers.Serializer):
 
     def validate_duns_number(self, duns_number):
         """
-        Check if the duns
+        Check if the duns_number is valid i.e. isn't already assigned
+        to another company.
         """
         if Company.objects.filter(duns_number=duns_number).exists():
             raise serializers.ValidationError(
                 f'Company with duns_number: {duns_number} already exists in DataHub.',
             )
         return duns_number
+
+
+class DNBInvestigationDataSerializer(serializers.Serializer):
+    """
+    Serializer for DNBInvestigationData - a JSON field that contains
+    auxuliary data needed for submitting to DNB for investigation.
+    """
+
+    telephone_number = serializers.CharField(
+        required=False,
+        allow_null=True,
+        allow_blank=True,
+    )
+
+
+class DNBCompanyInvestigationSerializer(CompanySerializer):
+    """
+    For creating Company record to be investigated by DNB.
+
+    Sets `dnb_investigation_data`.
+    """
+
+    dnb_investigation_data = serializers.JSONField(
+        required=False,
+        allow_null=True,
+        write_only=True,
+    )
+
+    def validate_dnb_investigation_data(self, dnb_investigation_data):
+        """
+        Check if dnb_investigation_data is valid.
+        """
+        if dnb_investigation_data in (None, ''):
+            return None
+        serializer = DNBInvestigationDataSerializer(data=dnb_investigation_data)
+        serializer.is_valid(raise_exception=True)
+        return serializer.validated_data
+
+    def validate(self, data):
+        """
+        Validate if either website or telephone_number is present.
+        """
+        data = super().validate(data)
+        investigation_data = data.get('dnb_investigation_data') or {}
+
+        if (
+            data.get('website') in (None, '')
+            and investigation_data.get('telephone_number') in (None, '')
+        ):
+            raise serializers.ValidationError(
+                f'Either website or telephone_number must be provided.',
+            )
+
+        return data
+
+    class Meta(CompanySerializer.Meta):
+        fields = CompanySerializer.Meta.fields + ('dnb_investigation_data', )
--- a/datahub/dnb_api/test/test_serializers.py
+++ b/datahub/dnb_api/test/test_serializers.py
@@ -1,0 +1,237 @@
+from urllib.parse import urlparse
+
+import pytest
+
+from datahub.dnb_api.serializers import DNBCompanyInvestigationSerializer
+from datahub.dnb_api.utils import format_dnb_company_investigation
+
+
+override_functions = (
+    pytest.param(
+        lambda data, null_keys: {k: v for k, v in data.items() if k not in null_keys},
+        id='no-keys-for-null-values',
+    ),
+    pytest.param(
+        lambda data, null_keys: {**data, **{k: None for k in null_keys}},
+        id='null-values-for-null-values',
+    ),
+    pytest.param(
+        lambda data, null_keys: {**data, **{k: '' for k in null_keys}},
+        id='blank-for-null-values',
+    ),
+)
+
+
+def assert_company_data(company, data):
+    """
+    Check if each field of the given company has the same value as given data.
+    """
+    assert company.name == data['name']
+    assert company.dnb_investigation_data == data['dnb_investigation_data']
+    assert company.address_1 == data['address']['line_1']
+    assert company.address_2 == data['address']['line_2']
+    assert company.address_town == data['address']['town']
+    assert company.address_county == data['address']['county']
+    assert company.address_postcode == data['address']['postcode']
+    assert str(company.address_country.id) == data['address']['country']['id']
+    assert str(company.business_type.id) == data['business_type']
+    assert str(company.sector.id) == data['sector']
+    assert str(company.uk_region.id) == data['uk_region']
+
+    website = data.get('website')
+    if website not in (None, ''):
+        url = urlparse(website)
+        website = f'{url.scheme or "http"}://{url.path or url.netloc}'
+    assert company.website == website
+
+
+@pytest.mark.parametrize(
+    'investigation_payload_override',
+    (
+        # No override
+        {},
+        # URL without scheme
+        {'website': 'www.test.com'},
+        # URL with https scheme
+        {'website': 'https://www.test.com'},
+        # Telephone number with non-numeric characters
+        {'telephone_number': '123ABC'},
+        # Telephone number with non-alphanumeric characters
+        {'telephone_number': '+(44)123#456'},
+    ),
+)
+def test_investigation_serializer_valid(
+        db,
+        investigation_payload,
+        investigation_payload_override,
+):
+    """
+    Test if DNBCompanyInvestigationSerializer saves a Company record given
+    a valid payload.
+    """
+    data = format_dnb_company_investigation(
+        {
+            **investigation_payload,
+            **investigation_payload_override,
+        },
+    )
+    serializer = DNBCompanyInvestigationSerializer(data=data)
+    assert serializer.is_valid()
+
+    company = serializer.save()
+    company.refresh_from_db()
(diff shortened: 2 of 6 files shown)


datahub/company/models/company.py
```diff
@@ -249,8 +249,8 @@ def get_dnb_investigation_context(self):
                 'postcode': self.address_postcode,
             },
             'website': self.website,
-            'telephone_number': dnb_investigation_data.get(
-                'telephone_number',
+            'telephone': dnb_investigation_data.get(
```
alixedi: The old `telephone_number` is now `telephone: {'country_code': .., 'number': .. }`

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,64 @@ class DUNSNumberSerializer(serializers.Serializer):
 
     def validate_duns_number(self, duns_number):
         """
-        Check if the duns
+        Check if the duns_number is valid i.e. isn't already assigned
```
alixedi: This seemed incomplete so I fixed it.

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,64 @@ class DUNSNumberSerializer(serializers.Serializer):
...
+
+    Sets `dnb_investigation_data`.
+    """
+
+    dnb_investigation_data = serializers.JSONField(
+        required=False,
+        allow_null=True,
+        write_only=True,
+    )
+    website = serializers.CharField(
```
reupen: Overriding the field in this way will lose the `max_length` value. You could use [`Meta.extra_kwargs`](https://www.django-rest-framework.org/api-guide/serializers/#additional-keyword-arguments) to avoid that.

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,64 @@ class DUNSNumberSerializer(serializers.Serializer):
...
         if Company.objects.filter(duns_number=duns_number).exists():
             raise serializers.ValidationError(
                 f'Company with duns_number: {duns_number} already exists in DataHub.',
             )
         return duns_number
+
+
+class DNBCompanyInvestigationSerializer(CompanySerializer):
+    """
+    For creating Comapany record to be investigated by DNB.
```
currycoder: `s/Comapany/Company/g`

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,64 @@ class DUNSNumberSerializer(serializers.Serializer):
...
+        allow_null=True,
+        validators=(URLValidator(), ),
+    )
+
+    def validate_dnb_investigation_data(self, dnb_investigation_data):
+        """
+        Check if dnb_investigation_data is valid i.e. if a non-null telephone
+        field is present, it should have integer number and country_code.
+        """
+        if dnb_investigation_data is not None:
```
currycoder: Is `telephone` required?  If so, are we going to enforce the requirement or is frontend?

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,64 @@ class DUNSNumberSerializer(serializers.Serializer):
...
+            telephone = dnb_investigation_data.get('telephone')
+            if telephone is not None:
+                for field in ('country_code', 'number'):
+                    integer_validator(
+                        telephone.get(field),
+                    )
+
+        return dnb_investigation_data
```
currycoder: My understanding is that although this validates that the `dnb_investigation_data` contains a valid `telephone` item, it does not clear away any other cruft that might appear in a free-form JSON field.  I think it could be good if this method explicitly extracted the data that was needed instead of allowing anything.

currycoder: I guess I don't know the context of how this will be used in the view though!

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,64 @@ class DUNSNumberSerializer(serializers.Serializer):
...
+    def validate_dnb_investigation_data(self, dnb_investigation_data):
+        """
+        Check if dnb_investigation_data is valid i.e. if a non-null telephone
+        field is present, it should have integer number and country_code.
+        """
+        if dnb_investigation_data is not None:
+            telephone = dnb_investigation_data.get('telephone')
+            if telephone is not None:
+                for field in ('country_code', 'number'):
+                    integer_validator(
```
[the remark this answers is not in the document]
alixedi: No. I have added a test case for this as well.

datahub/dnb_api/serializers.py
reupen: It would be nice to try to use a serialiser for this validation (as it's typically more robust and easier to understand data structure).

Possibly, something like:

```python
class TelephoneNumberSerializer(serializers.Serializer):
    country_code = serializers.CharField(validators=[integer_validator])
    # but see other comment about spaces
    number = serializers.CharField(validators=[integer_validator])


class DNBInvestigationData(serializers.Serializer):
    telephone = TelephoneNumberSerializer(required=False, allow_null=True)

...

    def validate_dnb_investigation_data(self, dnb_investigation_data):
        """
        Check if dnb_investigation_data is valid i.e. if a non-null telephone
        field is present, it should have integer number and country_code.
        """
        if dnb_investigation_data is None:
            return None

        serializer = DNBInvestigationData(data=dnb_investigation_data)
        serializer.is_valid(raise_exception=True)

        return serializer.validated_data
```

(With https://github.com/encode/django-rest-framework/pull/6916 it should be possible to use the serialiser directly as a field in this case.)

datahub/dnb_api/test/test_serializers.py
```diff
@@ -0,0 +1,140 @@
...
+):
+    """
+    Test if DNBCompanyInvestigationSerializer fails given an invalid payload.
+    """
+    serializer = DNBCompanyInvestigationSerializer(
+        data=format_dnb_company_investigation(
+            {**investigation_payload, **investigation_payload_override},
+        ),
+    )
+    assert not serializer.is_valid()
```
currycoder: Could it be good to ensure that the errors are as expected here too?

reupen:
The Django convention is [to use an empty string as the blank value for CharFields](https://docs.djangoproject.com/en/2.2/ref/models/fields/#null), and we've tried to stick with that in new code. It would be good to add more test cases for blank strings.

datahub/dnb_api/serializers.py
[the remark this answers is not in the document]
reupen: Is that the desired behaviour though?

datahub/dnb_api/test/test_serializers.py
alixedi: Good idea!

datahub/dnb_api/test/test_serializers.py
```diff
@@ -0,0 +1,139 @@
...
+    assert str(db_company.business_type.id) == data['business_type']
+    assert str(db_company.sector.id) == data['sector']
+    assert str(db_company.uk_region.id) == data['uk_region']
+
+
+@pytest.mark.parametrize(
+    'investigation_payload_override',
+    (
+        # Website and telephone_number cannot be both None
+        {'website': None, 'telephone': None},
```
reupen: Since these are more complex expressions, it would be good to use `pytest.param()` for these and pass a value for `id` so it is easier to find the parametrisation for each the test result and avoid things like this:

image

(There are some examples here: https://github.com/uktrade/data-hub-leeloo/blob/56a900c8b7b08ed877eba671dfea82761cfed995/datahub/interaction/test/admin_csv_import/test_row_form.py#L89-L96.)

datahub/dnb_api/serializers.py
alixedi: Good point. I will make [format_dnb_company_investigation](https://github.com/uktrade/data-hub-leeloo/pull/2065/files#diff-f3cf320eaf9fd1b399380f387051a25aR81) explicitly use the keys that we expect.

datahub/dnb_api/serializers.py
alixedi: Great idea! 👍

alixedi: Either a `telephone` or a valid `website` is required. This is enforced in the `validate` method.

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,68 @@ class DUNSNumberSerializer(serializers.Serializer):
...
+            raise serializers.ValidationError(
+                f'Either website or telephone_number must be provided.',
+            )
+
+        return data
+
+    class Meta(CompanySerializer.Meta):
+        fields = CompanySerializer.Meta.fields + ('dnb_investigation_data', )
+        extra_kwargs = {
+            'website': {
```
reupen: Isn't `website` already a `URLField`?

datahub/dnb_api/serializers.py
```diff
@@ -124,10 +124,67 @@ class DUNSNumberSerializer(serializers.Serializer):
...
         return duns_number
+
+
+class DNBInvestigationDataSerializer(serializers.Serializer):
+    """
+    Serializer for DNBInvestigationData - a JSON field that contains
+    auxuliary data needed for submitting to DNB for investigation.
+    """
+
+    telephone_number = serializers.CharField(
```
currycoder: It's probably good to call out here that this data will only ever be used to help DNB with investigation and so a simple charfield will suffice.  i.e. We don't need to worry about complex phone number/area code/string vs integer validation.

datahub/dnb_api/utils.py
```diff
@@ -76,3 +76,14 @@ def format_dnb_company(dnb_company):
         # 'global_headquarters': None,
         # 'headquarter_type': None,
     }
+
+
+def format_dnb_company_investigation(data):
```
reupen: Is this function just for the tests?

alixedi: No. This will be used in the views. I am using the same pattern as: https://github.com/uktrade/data-hub-leeloo/pull/1943/files#diff-f3cf320eaf9fd1b399380f387051a25aR32-R78

datahub/dnb_api/test/test_serializers.py
```diff
@@ -0,0 +1,189 @@
...
+        # Website can be null if we have telephone_number
+        {'website'},
+    ),
+)
+@pytest.mark.parametrize(
+    'override_function',
+    override_functions,
+
+)
+def test_investigation_serializer_valid(
```
reupen: I find this test a bit indirect and therefore not easy to follow what is going on. Things like https://github.com/uktrade/data-hub-leeloo/blob/2a250a20682b62c0302b80df71497723167fb747/datahub/interaction/test/views/test_interaction.py#L42 and https://github.com/uktrade/data-hub-leeloo/blob/56a900c8b7b08ed877eba671dfea82761cfed995/datahub/interaction/test/admin_csv_import/test_row_form.py#L89 seem a bit more explicit about what the inputs are and therefore easier to follow.

But I'll leave it up to you if you want to change it.

PR #2099, opened 2019-09-23
Add an admin tool for unarchiving companies

currycoder:
### Description of change

We are turning off the company unarchive feature in the frontend.  This change will give admins an easy tool to unarchive companies through django admin.

### Checklist 

### DIFF
--- a/datahub/company/test/admin/archiving/test_unarchive.py
+++ b/datahub/company/test/admin/archiving/test_unarchive.py
@@ -4,7 +4,11 @@ from django.urls import reverse
 from rest_framework import status
 
 from datahub.company.models import Company, CompanyPermission
-from datahub.company.test.factories import ArchivedCompanyFactory, CompanyFactory
+from datahub.company.test.factories import (
+    ArchivedCompanyFactory,
+    CompanyFactory,
+    DuplicateCompanyFactory,
+)
 from datahub.core.test_utils import AdminTestMixin, create_test_user
 from datahub.core.utils import reverse_with_query_string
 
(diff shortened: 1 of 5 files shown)


datahub/company/test/admin/archiving/test_unarchive.py
```diff
@@ -0,0 +1,172 @@
...
+                status.HTTP_400_BAD_REQUEST,
+                False,
+            ),
+        ),
+    )
+    def test_unarchive_view(
+        self,
+        company_callable,
+        expected_status_code,
+        expected_archived_value,
```
reupen: Looks like `expected_archived_value` is `False` for both cases – are there any cases where it should be `True`, or can it be removed from the parametrisation?

datahub/company/test/admin/archiving/test_unarchive.py
```diff
@@ -0,0 +1,172 @@
...
+        unarchive_url = reverse_with_query_string(
+            unarchive_route_name,
+            unarchive_query_args,
+        )
+
+        response = self.client.get(unarchive_url)
+        assert response.status_code == expected_status_code
+        company.refresh_from_db()
+        assert company.archived == expected_archived_value
+        if expected_archived_value is False:
```
elcct: As Reupen mentioned above, this seems to be always False, but if a case was added where this value is True, then it would be good to assert that these values exist.

datahub/company/test/admin/archiving/test_unarchive.py
currycoder: I removed it and then realised that this was handy for testing that transferred companies cannot be un-archived.

datahub/company/test/admin/archiving/test_unarchive.py
currycoder: I removed it and then realised that this was handy for testing that transferred companies cannot be un-archived.

PR #2102, opened 2019-09-25
Add countries of interest as a field in the file download

yozzo:
### Checklist


yozzo:
waiting for @benpbenp changes to the countries of interest

.gitignore
```diff
@@ -79,3 +79,9 @@ docker-volumes/
 
 # celery
 celerybeat-schedule
+
+# Redis
+dump.rdb
+
+#IDE
+.idea/
```
currycoder: We generally try to make sure that the gitignore does not contain references to developer-specific patterns. In this case, `.git/info/exclude` is probably a better place for IDE ignores as it won't pollute the project-wide file.
See: https://git-scm.com/docs/gitignore

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -581,10 +582,9 @@ def test_export(
             is_turnover_estimated=True,
             number_of_employees=95,
             is_number_of_employees_estimated=True,
+            export_to_countries=[c.id for c in Country.objects.all()[:2]],
         )
 
-        setup_es.indices.refresh()
```
reupen: Did you mean to remove this? It should be needed here, otherwise the documents might not be returned by the search query.

datahub/search/company/views.py
```diff
@@ -169,13 +172,25 @@ class SearchCompanyExportAPIView(SearchCompanyAPIViewMixin, SearchExportAPIView)
             default='employee_range__name',
             output_field=CharField(),
         ),
+
+        export_to_countries_list=get_string_agg_subquery(
+            DBCompany,
+            Cast('export_to_countries__id', CharField()),
```
reupen: Is there are reason we're returning the IDs rather than the country names?

yozzo: I noticed other fields are using id's also (e.g. uk_region) so in the sense of having a uniform return from the queryset I used `id`, but personally I feel it's a tad superfluous. Suggestions?

datahub/search/company/test/test_entity_search_views.py
yozzo: whoops!

datahub/search/company/views.py
reupen: UK region is a name here (see further down) – are you looking at the MI dashboard downloads maybe? I don't think there are any raw IDs in the search CSV exports from Data Hub itself (on the collection pages).

yozzo:
Because of a the long tail of commits not related to this feature, the PR has been replicated here: https://github.com/uktrade/data-hub-leeloo/pull/2144 thus closing this PR. All change requests have been addressed

PR #2118, opened 2019-10-01
Add script to automate creation of changelog

reupen:
### Description of change

This adds a script that automates the tasks associated with creating a changelog.

Python was used for the script because:

- we're all familiar with it
- it's portable
- it handles arguments well

Most of the tasks associated with creating a changelog can now be performed by running `scripts/create_changelog.py major|minor|patch`.

The script will increment the version automatically, create a branch with the updated changelog, push the new branch, and open your web browser ready to create a PR.

The release instructions have been updated to reflect these changes.

### Checklist

### DIFF
--- a/scripts/create_changelog.py
+++ b/scripts/create_changelog.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python
+
+"""
+This is a script that does the following:
+
+- takes a release type (major, minor or patch) as input and determines the new version number
+- runs `git fetch`
+- creates a local branch based on origin/develop
+- generates a changelog for the release
+- commits and pushes the changelog
+- opens your web browser to the create PR page for the pushed branch
+
+The script will abort if:
+
+- there are uncommitted changes
+- there are no news fragments on origin/develop
+- the changelog branch for the new version already exists
+- there are news fragments left behind after the changelog is generated
+"""
+
+import argparse
+import glob
+import re
+import subprocess
+import webbrowser
+from pathlib import Path
+from urllib.parse import quote
+
+from script_utils.version import Version
+
+parser = argparse.ArgumentParser(description='Create and push a changelog for a new version.')
+parser.add_argument('release_type', choices=Version._fields)
+
+
+class CommandError(Exception):
+    """A fatal error when running the script."""
+
+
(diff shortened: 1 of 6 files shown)


scripts/create_changelog.py
```diff
@@ -0,0 +1,168 @@
...
+
+    return True
+
+
+def list_news_fragments():
+    """Return a list of files that are probable news fragments."""
+    path = Path(__file__)
+    root_path = path.resolve().parents[1]
+    excluded_files = {'.gitignore', '_template.md.jinja2', 'README.md'}
+    changelog_files = glob.iglob(f'{root_path}/changelog/**/*')
```
currycoder: Could this be preferable? `glob.iglob(f'{root_path}/changelog/**/*.md')`

scripts/create_changelog.py
```diff
@@ -0,0 +1,168 @@
...
+    return set(changelog_files) - excluded_files
+
+
+def get_current_version():
+    """Gets the current version number from the changelog."""
+    with open('CHANGELOG.md', 'r') as file:
```
currycoder: In theory, the latest release tag should be the most bullet proof place to get this (since there's an obvious programmatic mechanism)..  In practice, updating this is forgotten at the end of the release process - so I see why you went with the approach of parsing the changelog.

Might be something to strive to later with the extra iterations of release automation?  Maybe a TODO would be handy.

scripts/script_utils/version.py
```diff
@@ -0,0 +1,40 @@
+from typing import NamedTuple
+
+
+class Version(NamedTuple):
```
currycoder: This class: :ok_hand:

scripts/create_changelog.py
reupen: Technically the `.md` is optional – originally we were just naming new news fragments as `.internal`, `.feature` etc. so I didn't require `.md` here.

scripts/create_changelog.py
reupen: Yes, I was similarly wondering if we should be writing the version to e.g. module so the code also has access to it. I'll add a note to the docstring about this.

PR #2121, opened 2019-10-01
Add interactions dataset endpoint

niross:
### Description of change

Added Interactions dataset api endpoint under the datasets app (/v4/datasets/interactions-dataset)

### Checklist


datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,101 @@
...
+                InteractionDITParticipant.interaction.field,
+                'adviser__telephone_number',
+                ('-id',),
+            ),
+            adviser_team=get_top_related_expression_subquery(
+                InteractionDITParticipant.interaction.field,
+                'adviser__dit_team__name',
+                ('-id',),
+            ),
+            contact_id=Subquery(first_contact_queryset.values('contact__id')),
```
reupen: Should be able to use `contact_id` here to avoid a join.

datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,101 @@
...
+            interaction_link=get_front_end_url_expression('interaction', 'pk'),
+            sector=get_sector_name_subquery('company__sector'),
+            service_delivery=get_service_name_subquery('service'),
+        ).values(
+            'adviser_email',
+            'adviser_name',
+            'adviser_phone',
+            'adviser_team',
+            'communication_channel__name',
+            'company__id',
```
reupen: Should be able to use `company_id` here to avoid a join.

datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,101 @@
...
+            'event__address_country__name',
+            'event__address_town',
+            'event__event_type__name',
+            'event__name',
+            'event__start_date',
+            'event__uk_region__name',
+            'event_service',
+            'grant_amount_offered',
+            'interaction_link',
+            'investment_project__cdms_project_code',
```
reupen: Since investment project is going to be a data set, shouldn't we just return `investment_project_id` here?

datahub/interaction/migrations/0069_auto_20191002_1315.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('interaction', '0068_remove_interaction_location_from_database'),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name='interaction',
+            index=models.Index(fields=['created_on', 'id'], name='interaction_created_9a2917_idx'),
```
reupen: Just an FYI, this migration will take 10 seconds or so (and lock the table in the process) – that shouldn't be a problem, but we should make sure we do a small release when we release it (without other non-quick migrations).

datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,101 @@
...
+                ('-id',),
+            ),
+            adviser_team=get_top_related_expression_subquery(
+                InteractionDITParticipant.interaction.field,
+                'adviser__dit_team__name',
+                ('-id',),
+            ),
+            contact_id=Subquery(first_contact_queryset.values('contact__id')),
+            event_service=get_service_name_subquery('event__service'),
+            interaction_link=get_front_end_url_expression('interaction', 'pk'),
```
reupen: I haven't seen  front-end URLs in the other data sets – just wondering if there's a specific reason for including it in this one?

datahub/dataset/interactions/views.py
reupen: At some point I think returning an array of IDs was mooted – is there a reason we aren't going for that here? (I think it could be done with `ArrayAgg`.)

datahub/dataset/interactions/views.py
niross: I added it as it's in the service deliveries & interactions report currently. Not sure how necessary it is though... I will check.

datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,101 @@
...
+        first_contact_queryset = contacts_m2m_model.objects.filter(
+            interaction_id=OuterRef('pk'),
+        ).order_by(
+            'pk',
+        )[:1]
+
+        return get_base_interaction_queryset().annotate(
+            adviser_email=get_top_related_expression_subquery(
+                InteractionDITParticipant.interaction.field,
+                'adviser__email',
```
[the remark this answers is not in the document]
niross: Gotcha thanks, do you think it's worth bringing both over or `contact_email` should be sufficient?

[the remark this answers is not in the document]
reupen: Good question. `contact_email` seems pretty thoroughly populated for active advisers in production, so it should be enough I reckon (if what we want is a real email address).

datahub/dataset/interactions/views.py
niross: Cool, wasn't aware of that conversation, have switch to returning all contact ids

datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,93 @@
...
+                InteractionDITParticipant.interaction.field,
+                'adviser__telephone_number',
+                ('-id',),
+            ),
+            adviser_team=get_top_related_expression_subquery(
+                InteractionDITParticipant.interaction.field,
+                'adviser__dit_team__name',
+                ('-id',),
+            ),
+            contact_ids=ArrayAgg('contacts__id', ordering='id'),
```
reupen: Would suggest doing this via a subquery (similar to how `get_string_agg_subquery()` works). Reason being, it avoids the join and `group by` in the main query (which can end up being expensive).

reupen:
Just spotted that the new sub-package is named `interactions` – would be good to make it `interaction` to make it consistent with the other ones.

datahub/dataset/interactions/pagination.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.pagination import CursorPagination
+
+
+class InteractionsDatasetViewCursorPagination(CursorPagination):
+    """
+    Cursor Pagination for InteractionsDatasetView
```
elcct: Just cosmetic, but this should fit in one line.

datahub/dataset/interactions/views.py
```diff
@@ -0,0 +1,93 @@
...
+                InteractionDITParticipant.interaction.field,
+                get_full_name_expression('adviser'),
+                ('-id',),
+            ),
+            adviser_phone=get_top_related_expression_subquery(
+                InteractionDITParticipant.interaction.field,
+                'adviser__telephone_number',
+                ('-id',),
+            ),
+            adviser_team=get_top_related_expression_subquery(
```
reupen: FYI the team(s) associated with the interaction are stored in `InteractionDITParticipant.team`. (`InteractionDITParticipant.team` is usually consistent with `InteractionDITParticipant.adviser`, but not always e.g. when an adviser moves team, or for old data.)

datahub/dataset/interactions/test/test_views.py
```diff
@@ -0,0 +1,176 @@
...
+    def test_without_scope(self, hawk_api_client):
+        """Test that making a request without the correct Hawk scope returns an error."""
+        hawk_api_client.set_credentials(
+            'test-id-without-scope',
+            'test-key-without-scope',
+        )
+        response = hawk_api_client.get(self.view_url)
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_without_credentials(self, api_client):
```
elcct: After IP check was separated from the Hawk we also check if the endpoint can be accessed from the IP address that is not whitelisted. E.g. https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/dataset/test/test_views.py#L123

datahub/dataset/interactions/views.py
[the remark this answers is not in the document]
niross: Cheers, so it turns out we are going to start syncing advisers across as a separate dataset so I've switched to ArrayAgg on adviser_id. I'll make a note of the above in the advisers ticket though :+1:

datahub/dataset/interactions/views.py
niross: Done, made the same change to adviser also, cheers.

datahub/dataset/interactions/views.py
niross: Noted for the Advisers dataset :+1:

datahub/dataset/interactions/test/test_views.py
niross: Ah good catch. I've added IP authentication to the view (and tests)

niross:
> Looks good to me (subject to @elcct 's comments).
> 
> But, can we hold off on merging until the company data set is released (due to the indexes being created)?

Will do. Thanks for your help.

datahub/dataset/interaction/test/test_views.py
```diff
@@ -0,0 +1,181 @@
...
+    """Hawk API client fixture configured to use credentials with the data_flow_api scope."""
+    hawk_api_client.set_credentials(
+        'data-flow-api-id',
+        'data-flow-api-key',
+    )
+    yield hawk_api_client
+
+
+def get_expected_data_from_interaction(interaction):
+    """Returns expected API response dictionary for the an interaction"""
```
currycoder: Minor typo:

```
"""Returns expected API response dictionary for an interaction"""
```

datahub/interaction/migrations/0069_auto_20191002_1315.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.5 on 2019-10-02 13:15
```
reupen: Small thing, but it would be good to give this migration a name (usually done by passing `--name <name>` to `./manage.py makemigrations`).

PR #2122, opened 2019-10-01
Add export_potential field to company model

sekharpanja:
### Description of change
A new read-only field was added to company model `export_potential`. Values for this field are imported from Data Science platform, as a seperate exercise. It can have one of the constant values out of `Very High`, `High`, `Medium`, `Low` and `Very Low`.

### Checklist


datahub/company/test/test_company_views.py
```diff
@@ -1090,6 +1094,30 @@ def test_update_headquarter_type(
...
 
+    @pytest.mark.parametrize(
+        'input_score,expected_score',
+        (
+            ('very_high', 'very_high'),
+            ('high', 'high'),
+            ('medium', 'medium'),
+            ('low', 'low'),
+            ('very_low', 'very_low'),
+            (None, None),
```
benpbenp: 1. Since the expected is the same as the input in every case, can you just use one parameter here?
2. I think it is unnecessary to test every single option... None and two others should be plenty.

datahub/company/models/company.py
```diff
@@ -59,6 +59,14 @@ class Company(ArchivableModel, BaseModel):
         ('duplicate', 'Duplicate record'),
     )
 
+    EXPORT_POTENTIAL_SCORE = Choices(
```
reupen: This should probably be a plural to be consistent with other Choices objects.

changelog/company/export-potential-score.md
```diff
@@ -0,0 +1 @@
+`GET /v4/search/company`: A new read-only field was added to company model `export_potential`. Values for this field are imported from Data Science platform, as a seperate exercise. It can have one of the constant values out of `Very High`, `High`, `Medium`, `Low` and `Very Low`.
```
reupen: The values will be `very_high`, `high` etc. right? Also would be good to clarify that it can be `null`.

Also, this PR seems to be changing the responses of `/v4/company` and `/v4/company/<id>` (rather than the search endpoint).

datahub/company/models/company.py
```diff
@@ -223,6 +231,13 @@ class Company(ArchivableModel, BaseModel):
         null=True,
         blank=True,
     )
+    export_potential = models.CharField(
+        max_length=MAX_LENGTH,
+        null=True,
+        blank=True,
+        choices=EXPORT_POTENTIAL_SCORE,
+        help_text='Score that signifies export potential, imported from DS',
```
reupen: Would be good to write out 'data science' in full because it doesn't seem obvious what DS is.

datahub/company/test/test_company_views.py
```diff
@@ -687,6 +688,7 @@ def test_cannot_update_read_only_fields(self):
             number_of_employees=95,
             is_number_of_employees_estimated=False,
             pending_dnb_investigation=True,
+            export_potential='very_high',
```
reupen: Would be good to use `Company.EXPORT_POTENTIAL_SCORE.very_high` here (and in similar places elsewhere) so the values are only defined in one place.

changelog/company/export-potential-score.md
reupen: The file name of this news fragment is missing a category – it should be e.g. `.api.md`.

datahub/company/test/test_company_views.py
reupen: I agree.

datahub/company/test/test_company_views.py
```diff
@@ -1095,28 +1095,26 @@ def test_update_headquarter_type(
...
-        'input_score,expected_score',
+        'score',
         (
-            ('very_high', 'very_high'),
-            ('high', 'high'),
-            ('medium', 'medium'),
-            ('low', 'low'),
-            ('very_low', 'very_low'),
-            (None, None),
+            ('very_high'),
```
reupen: Small thing – the inner brackets around each value aren't needed here (and can be removed).

datahub/company/migrations/0092_company_export_potential.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
```
reupen: We'll also want to rename this migration to `0093_company_export_potential.py` (0093 instead of 0092) as well (if I haven't missed something).

PR #2126, opened 2019-10-03
Add function to create new StatsClient.

alixedi:
### Description of change

From discussion with PaaS devs the other day:

> One thing that we haven’t communicated very well to tenants is that internal networking is meshed, and load balancing is done client side.
When you make a request to x.cloudapps.digital it goes through our routers, which are fronted by ELBs (who’s IPs do not change often), or by Cloudfront (if you are using a custom domain)
When you make a request to x.apps.internal it goes directly between apps, with no routing component other than a sidecar proxy. This means that applications which do any sort of DNS caching without retries run into networking issues when applications are redeployed, or cells are rolled. Ideally applications would not cache DNS at all for internal apps.

[Python StatsD module](https://github.com/jsocol/pystatsd) creates a `StatsClient` singleton at the [start of the process](https://github.com/jsocol/pystatsd/blob/master/statsd/defaults/django.py). 

This combined with UDP means that e.g. if the IP of `StatsD` changes, the packets will not make their way to StatsD and there is no error and therefore no way to create a new `StatsClient` instance.

This PR has the simplest solution to this which is to create a new `StatsClient` for every metric update. In future, we might want to cache the `StatsClient` with a suitable TTL.

If you want to see this working end-to-end:

```
$ docker run -d -p 9102:9102 -p 9125:9125/udp prom/statsd-exporter
$ pytest datahub/dnb_api/test/test_views.py
$ curl localhost:9102 | grep dnb
```

### Checklist

### DIFF
--- a/datahub/core/statsd.py
+++ b/datahub/core/statsd.py
@@ -1,0 +1,37 @@
+from django.conf import settings
+from statsd import defaults
+from statsd.client import StatsClient
+
+HOST = settings.STATSD_HOST
+PORT = settings.STATSD_PORT
+PREFIX = settings.STATSD_PREFIX
+MAXUDPSIZE = defaults.MAXUDPSIZE
+IPV6 = defaults.IPV6
+
+
+def statsd():
+    """
+    Returns a new StatsD client.
+
+    Use this instead of:
+        statsd.defaults.django.statsd
+
+    because statsd.defaults.django.statsd is a singleton
+    initialised at the start of the process and therefore
+    does not deal with chnages to the IP of the statsd instance.
+    """
+    return StatsClient(
+        host=HOST,
+        port=PORT,
+        prefix=PREFIX,
+        maxudpsize=MAXUDPSIZE,
+        ipv6=IPV6,
+    )
+
+
+def incr(*args, **kwargs):
+    """
+    Increments the given stat by `count` after
+    creating a new `StatsClient`.
+    """
+    statsd().incr(*args, **kwargs)
--- a/datahub/dnb_api/test/conftest.py
+++ b/datahub/dnb_api/test/conftest.py
@@ -1,5 +1,4 @@
 import datetime
-from unittest.mock import Mock
 
 import pytest
 from django.utils.timezone import utc
@@ -204,16 +203,3 @@ def investigation_payload():
         'sector': Sector.renewable_energy_wind.value.id,
         'uk_region': UKRegion.east_midlands.value.id,
     }
-
-
-@pytest.fixture
-def statsd_mock(monkeypatch):
-    """
-    Returns a mock statsd client instance.
-    """
-    mock_statsd = Mock()
-    monkeypatch.setattr(
-        'datahub.dnb_api.views.statsd',
-        mock_statsd,
-    )
-    return mock_statsd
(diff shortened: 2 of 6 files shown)


datahub/core/statsd.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def statsd():
+    """
+    Returns a new StatsD client.
+
+    Use this instead of:
+        statsd.defaults.django.statsd
+
+    because this does not stop working when the IP of the statsd
```
mcbhenwood: Lexically I'm binding the word "this" to the last mentioned thing (statsd.defaults.django.statsd) in which case I am parsing the comment to mean what I imagine is its opposite. Perhaps something explicit like "because statsd.defaults.django.statsd stops working when the IP of the statsd instance changes"

datahub/core/statsd.py
```diff
@@ -0,0 +1,36 @@
...
+    instance changes
+    """
+    return StatsClient(
+        host=HOST,
+        port=PORT,
+        prefix=PREFIX,
+        maxudpsize=MAXUDPSIZE,
+        ipv6=IPV6,
+    )
```
mcbhenwood: Do you think that we might be in danger of memory-leaking with loads of dead StatsClient objects being retained, holding open "black hole" sockets to missing endpoitns? This feels like the sort of thing which would side-step Python's usual garbage collection. Maybe there should be a singleton (new) StatsClient which explicitly holds a "private"  statsd.defaults.django.statsd.StatsClient and sets that to None just before opening a new one each time?

This may be overkill. Feel free to discard. Checked the pystatsd StatsClientBase code and there's no clear way to "end" a client so maybe I am worrying about nothing.

datahub/core/statsd.py
```diff
@@ -0,0 +1,36 @@
+from django.conf import settings
+from statsd import defaults
+from statsd.client import StatsClient
+
+HOST = getattr(settings, 'STATSD_HOST', defaults.HOST)
+PORT = getattr(settings, 'STATSD_PORT', defaults.PORT)
+PREFIX = getattr(settings, 'STATSD_PREFIX', defaults.PREFIX)
+MAXUDPSIZE = getattr(settings, 'STATSD_MAXUDPSIZE', defaults.MAXUDPSIZE)
+IPV6 = getattr(settings, 'STATSD_IPV6', defaults.IPV6)
```
reupen: `STATSD_HOST`, `STATSD_PORT` and `STATSD_PREFIX` should always be defined in settings, so shouldn't need to use `getattr()` for those.

Also, `STATSD_MAXUDPSIZE` and `STATSD_IPV6` aren't not defined in our settings, so `getattr()` is will always pick the defaults.

datahub/core/statsd.py
```diff
@@ -0,0 +1,36 @@
...
+    return StatsClient(
+        host=HOST,
+        port=PORT,
+        prefix=PREFIX,
+        maxudpsize=MAXUDPSIZE,
+        ipv6=IPV6,
+    )
+
+
+def incr(*args, **kwargs):
```
reupen: Maybe it would be simpler to just let clients call `statsd().incr()`?

alixedi: I wanted to make minimal changes in this PR. 

In future, if we end up using functions other than `incr`, I would agree `statsd().incr()` would be simpler.

datahub/core/statsd.py
currycoder: I think that once the `incr()` function has finished execution, the `StatsClient` index will be eligible for garbage collection as it's essentially a local variable to the `incr()` function.  That's my understanding anyway.

PR #2131, opened 2019-10-04
Factor out `get_company`

alixedi:
### Description of change
This factors out code that searches the company with the given `duns_number` and handles exceptions into a utility function. 

I will be reusing this new function in the upcoming unhappy-path-admin PR.

### Checklist

### DIFF
--- a/datahub/dnb_api/utils.py
+++ b/datahub/dnb_api/utils.py
@@ -1,10 +1,15 @@
+import logging
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from rest_framework import status
 
 from datahub.core import statsd
 from datahub.core.api_client import APIClient, TokenAuth
 from datahub.metadata.models import Country
 
+
+logger = logging.getLogger(__name__)
 
 api_client = APIClient(
     settings.DNB_SERVICE_BASE_URL,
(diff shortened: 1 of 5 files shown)


datahub/dnb_api/utils.py
```diff
@@ -32,6 +38,47 @@ def search_dnb(query_params):
     return response
 
 
+def get_company(duns_number):
```
currycoder: Is it worth making some unit tests for this considering that we are going to use it in a few different locations now?

datahub/dnb_api/utils.py
```diff
@@ -32,6 +38,47 @@ def search_dnb(query_params):
...
+    """
+    Pull data for the company with the given duns_number from DNB and
```
reupen: I'm not sure about raising `APIException`s in a utility function that's not just for API views – `APIException`s are just for DRF really.

alixedi: I am planning to call this in an admin view. What are you suggesting?

reupen: Most probably raising a couple of custom exceptions here (that don't inherit from APIException) e.g. `DNBServiceInternalError`, `DNBServiceNotFoundError`.

Then the API view can translate those to `APIUpstreamException` and `APIBadRequestException`, and the admin view can display the error back to the user.

datahub/dnb_api/utils.py
```diff
@@ -20,6 +19,27 @@
 )
 
 
+class DNBServiceError(Exception):
+    """
+    Exception for when DNB service doesn't return
```
currycoder: Minor: Looks like the column width is a bit short here..?

alixedi: I get anxious 😂

datahub/dnb_api/views.py
```diff
@@ -163,7 +167,15 @@ def post(self, request):
         duns_serializer.is_valid(raise_exception=True)
         duns_number = duns_serializer.validated_data['duns_number']
 
-        dnb_company = get_company(duns_number)
+        try:
+            dnb_company = get_company(duns_number)
+
+        except (DNBServiceError, DNBServiceInvalidResponse) as exc:
+            raise APIUpstreamException(exc.args[0])
```
reupen: Would suggest `str(exc)` rather than `exc.args[0]` (same result, but avoids digging into `exc.args` directly).

PR #2146, opened 2019-10-07
POC for a way to speed up search tests.

benpbenp:
There are a good number of tests that have expensive setup which is not
mutated, and which can be re-used across a number of tests.

This PR shows how this might be done with one particularly slow set of tests. Setup time on my machine for each of the 6 tests in TestOrderExportView was 21.5 seconds. Therefore by reusing the setup we can potentially save around 105 seconds. However, as noted below, this gain is reduced by parallelism.

Some things that need more consideration:
1. ~How can we reuse the boiler plate at the beginning of the one_time_setup function?~
*I have now put this boilerplate in a django_db_for_class fixture in conftest.*
2. The gains are reduced by parallelism as the expensive setup needs to be performed by each worker. If there were some way of making a whole class of tests remain together in a single worker that would be ideal. *Suggestion: Mark these tests as "non-parallelized". Run the non-parallelised tests as a separate step in the build pipeline.*
3.~It would be ideal if we didn't need separate fixtures for the class scope for each of monkeypatch, setup_es_class, and synchronous_on_commmit, but I'm not sure if there is any way of avoiding that.~ *I am no longer concerned with this*
4. ~The test_user_without_permission_cannot_export test in particular only uses the setup_es fixture and its setup time is 9 seconds. If there are many other tests like this then we could potentially have a session-scoped setup_es fixture for massive savings.~ Turns out this 9 seconds is not primarily a result of the setup_es fixture

### Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -146,6 +146,21 @@ def synchronous_on_commit(monkeypatch):
     monkeypatch.setattr('django.db.transaction.on_commit', _synchronous_on_commit)
 
 
+@pytest.fixture(scope='class')
+def monkeyclass():
+    """Monkeypatch fixture which works in the class scope"""
+    from _pytest.monkeypatch import MonkeyPatch
+    mpatch = MonkeyPatch()
+    yield mpatch
+    mpatch.undo()
+
+
+@pytest.fixture(scope='class')
+def synchronous_on_commit_for_class(monkeyclass):
+    """synchronous_on_commit fixture which works in the class scope"""
+    monkeyclass.setattr('django.db.transaction.on_commit', _synchronous_on_commit)
+
+
 def _synchronous_submit_to_thread_pool(fn, *args, **kwargs):
     fn(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


reupen:
Could we use the `class` fixture scope to do this?

https://docs.pytest.org/en/latest/fixture.html#scope-sharing-a-fixture-instance-across-tests-in-a-class-module-or-session

reupen:
It's a good observation about that test.

Just a bit apprehensive about using internals for this.

I had a look locally and it seems like a lot of time is wasted syncing objects to ES individually.

For example, this yields a similar improvement (and can still be parallelised):

```python
        with disable_search_signal_receivers(Order), \
                disable_search_signal_receivers(Company), \
                disable_search_signal_receivers(Contact):
            order_with_multiple_refunds = OrderPaidFactory()
            ApprovedRefundFactory(
                order=order_with_multiple_refunds,
                requested_amount=order_with_multiple_refunds.total_cost / 5,
            )
            ApprovedRefundFactory(
                order=order_with_multiple_refunds,
                requested_amount=order_with_multiple_refunds.total_cost / 4,
            )

            for factory_ in factories:
                factory_.create_batch(2)

        sync_all_models()

        setup_es.indices.refresh()
```

I think we could come up with a smarter fixture for tests like these that syncs in bulk (and hence works more generically).

benpbenp:
@reupen What do you mean internals? We could of course combine your optimisation with mine for even faster tests.

reupen:
> What do you mean internals? 

Anything prefixed with `_` is usually considered internal. For example, the entire `_pytest` package is internal.

benpbenp:
Yeah, I don't think I'm using any of those am I? The private methods are all ours.

benpbenp:
Oh I see, the monkeypatch one is. Perhaps that part of it could be done another way, but the main point is the class-scoped data fixture.

benpbenp:
@reupen see the alternative way of implementing synchronous_on_commit_for_class that I just pushed, I believe that should work the same as using the monkeypatch fixture.

currycoder:
Some of the ideas here have made it in to other PRs which improve speed of the test suite.  Closing this for now.

PR #2157, opened 2019-10-10
Add CompanyExportCountry model

benpbenp:
This replaces the previous PR https://github.com/uktrade/data-hub-leeloo/pull/2085

* Add data migration to migrate future_interest_countries to
CompanyExportCountry.
* Change API future_interest_countries field to use this new model
* Ignore these objects during company merge (replicates current
behaviour)
* Add set_user_edited_export_countries method to perform logic when user
sets these countries
* Add get_active_future_export_countries method to query the new model
and return a list of active countries of all sources.

### Description of change

Add a new model which can record future interest countries from a variety of sources.
This model is populated from the old future_interest_country m2m using a INSERT INTO... SELECT.

The two main sources of information are 'user' and 'external'. User represents those countries manually entered by the user on the front-end. External represents those countries that only came in from an import from the data science pipeline.

The key difference is that we want the user to be able to delete countries that came in from data science, and have those deletions stick when the import is refreshed. On the other hand, data science imports should only be able to add new countries, not delete user-entered countries or override any that users have deleted.

### Checklist


reupen:
The general approach looks OK to me, but the PR is a bit large at just over 900 lines added ([the guideline is up to 400](https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Code%20review%20guidelines.md#pr-hygiene)).

Can I suggest splitting it to make reviewing easier? For example, first add the new model and the underlying logic, and then do the data migration and plug it into the API etc. (Feature flags can also be used if needed.)

benpbenp:
@reupen only if it is 100% necessary. I understand the line limit as an aspirational goal, but this has been dragging on for so long already.

benpbenp:
Split into https://github.com/uktrade/data-hub-leeloo/pull/2177/files and https://github.com/uktrade/data-hub-leeloo/pull/2178

PR #2173, opened 2019-10-14
Remove legacy metadata endpoints

elcct:
### Description of change

This removes legacy `/metadata/*` endpoints. It should be merged on or after 17th October.

It also fixes tests for ordered metadata views and updates a comment.

### Checklist


datahub/metadata/test/test_views.py
```diff
@@ -236,7 +236,7 @@ def test_team_view(metadata_client):
 
 
 class TestServiceView:
-    """Tests for the /metadata/service/ view."""
+    """Tests for the /v4/metadata/service/ view."""
```
reupen: Does the trailing slash need removing too?

```suggestion
    """Tests for the /v4/metadata/service view."""
```

datahub/metadata/views.py
```diff
@@ -25,12 +24,14 @@ def _create_metadata_view(mapping, view_set_types=None, auth_attributes=None):
         'queryset': mapping.queryset,
         'serializer_class': mapping.serializer,
         '__doc__': f'List all {model._meta.verbose_name_plural}.',
+        'authentication_classes': (PaaSIPAuthentication, HawkAuthentication),
```
reupen: Should we put these new auth-related entries first (and keep `__doc__` last)?

rafenden:
Could we please double check if no apps are using them? 

Looks like the market access are still pointing to the old endpoints:
https://github.com/uktrade/market-access-frontend/blob/develop/src/app/lib/metadata.js

elcct:
Thank you for checking this out. The metadata you are referring to calls Market Access backend, not Data Hub. Market Access backend fetches Data Hub metadata and then serves it to front end. The change has been made so that Market Access backend uses new /v4 endpoints with Hawk, here: https://github.com/uktrade/market-access-api/commit/83e47c9a1b018650f43f7566b8fe8ad46f6d7ef6

elcct:
@rafenden just confirmed that indeed Market Access is no longer using legacy metadata endpoints.

rafenden:
> @rafenden just confirmed that indeed Market Access is no longer using legacy metadata endpoints.

Thank you for clarification 👍

PR #2179, opened 2019-10-15
Rename and refactor Elasticsearch pytest fixtures

reupen:
### Description of change

This:

- renames the `_es_client` pytest fixture to `_es_session`
- splits the `setup_es` pytest fixture into `es` and `es_with_signals`

I'd recommend reviewing the three commits individually. (Most of the changes are just changing references to `setup_es` to `es_with_signals`.)

### Context

It was observed that a fair bit of time is wasted in some search tests indexing search documents individually rather than in bulk.

We're also currently implicitly testing search app signal receivers in every search test, which is not really necessary (they have their own dedicated tests).

This refactor is a precursor to adding a new fixture that can be used to index documents in bulk (rather than relying on the signal receivers), and hence reduce the running time of search tests that create a lot of model objects.

### Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -221,7 +221,26 @@ def _es_session(_es_client):
 
 
 @pytest.fixture
-def es_with_signals(_es_session, synchronous_on_commit):
+def es(_es_session):
+    """
+    Function-scoped pytest fixture that:
+
+    - ensures Elasticsearch is available for the test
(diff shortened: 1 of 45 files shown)


reupen:
>  Could it be worth an internal changelog entry to help communicate that the approach for ES tests has changed subtly

I omitted it in this case since it's not a releasable change as such, but happy to add one.

PR #2180, opened 2019-10-15
import data science export potential data into company model

sekharpanja:
### Description of change
Adding a django management command to import export propensity data in csv format into Data Hub's company model.

### Checklist

### DIFF
--- a/datahub/dbmaintenance/management/commands/update_company_export_potential.py
+++ b/datahub/dbmaintenance/management/commands/update_company_export_potential.py
@@ -1,0 +1,35 @@
+from logging import getLogger
+
+import reversion
+
+from datahub.company.models import Company
+from datahub.dbmaintenance.management.base import CSVBaseCommand
+from datahub.dbmaintenance.utils import parse_limited_string, parse_uuid
+
+
+logger = getLogger(__name__)
+
+
+class Command(CSVBaseCommand):
+    """Command to update Company.export_potential."""
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        score_dict = {value.lower(): key for key, value in Company.EXPORT_POTENTIAL_SCORES}
+
+        pk = parse_uuid(row['datahub_company_id'])
+        company = Company.objects.get(pk=pk)
+        raw_potential = parse_limited_string(row['export_propensity'])
+        export_potential = score_dict[raw_potential.lower()]
+
+        if company.export_potential == export_potential:
+            return
+
+        company.export_potential = export_potential
+
+        if simulate:
+            return
+
+        with reversion.create_revision():
+            company.save(update_fields=('export_potential',))
+            reversion.set_comment('Export potential updated.')
--- a/datahub/dbmaintenance/test/commands/test_update_company_export_potential.py
+++ b/datahub/dbmaintenance/test/commands/test_update_company_export_potential.py
@@ -1,0 +1,164 @@
+from datetime import datetime, timezone
+from io import BytesIO
+
+import factory
+import pytest
+from django.core.management import call_command
+from freezegun import freeze_time
+from reversion.models import Version
+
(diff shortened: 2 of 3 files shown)


reupen:
@sekharpanja It looks like this would fit in the `dbmaintenance` framework. Is there a reason it wasn't used here?

datahub/dbmaintenance/test/commands/test_update_company_export_potential.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def test_run(s3_stubber, caplog):
+    """Test that the command updates the specified records (ignoring ones with errors)."""
+    caplog.set_level('ERROR')
+
+    original_datetime = datetime(2017, 1, 1, tzinfo=timezone.utc)
+
+    with freeze_time(original_datetime):
+        exp_potential_scores = [
```
reupen: I think 'exp' could be confused with 'expected' – would be good to write out export in full to avoid the ambiguity.

datahub/dbmaintenance/management/commands/update_company_export_potential.py
```diff
@@ -0,0 +1,35 @@
...
+
+logger = getLogger(__name__)
+
+
+class Command(CSVBaseCommand):
+    """Command to update Company.export_potential."""
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        score_dict = {value: key for key, value in Company.EXPORT_POTENTIAL_SCORES}
```
reupen: Minor point, but I think doing the mapping based on the titles isn't too obvious. Is there a possibility of data science instead outputting the scores with our identifiers (to avoid having to map them)?

If not, I think an explicit mapping from their identifiers to ours would be better.

datahub/dbmaintenance/management/commands/update_company_export_potential.py
```diff
@@ -0,0 +1,35 @@
...
+
+        pk = parse_uuid(row['datahub_company_id'])
+        company = Company.objects.get(pk=pk)
+        raw_potential = parse_limited_string(row['export_propensity'])
+        export_potential = score_dict[raw_potential.title()]
```
reupen: If we normalise the case (to make the comparison case insensitive), I think using lower case (on both sides) would be more conventional.

changelog/company/import-csv-export-potential-score.md
```diff
@@ -0,0 +1 @@
+A new dbmaintenance Django command was added to import data from csv format, with two fields `datahub_company_id` and `export_propensity` provided by Data Science platform, into Data Hub comapany model.
```
reupen: Looks like a type was missed in the news fragment file name – it should end in e.g. `.feature.md`.

datahub/dbmaintenance/management/commands/update_company_export_potential.py
sekharpanja: Done it using normalised lower case comparison instead.

reupen: I'm not sure this one was resolved – have approved though as it was a small point.

datahub/dbmaintenance/management/commands/update_company_export_potential.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Command(CSVBaseCommand):
+    """Command to update Company.export_potential."""
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        score_dict = {value.lower(): key for key, value in Company.EXPORT_POTENTIAL_SCORES}
+
+        pk = parse_uuid(row['datahub_company_id'])
```
elcct: We would just use `id` column for the object that is being changed, so `datahub_company_id` looks a bit redundant, but I'll leave it to you.

PR #2181, opened 2019-10-15
Add CompanyExportCountry model - A

benpbenp:
* Ignore these objects during company merge (replicates current
behaviour)
* Add get_active_future_export_countries method to query the new model
and return a list of active countries of all sources.

### Checklist


datahub/company/migrations/0095_add_company_export_country_model.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0094_add_change_regional_account_manager_perm'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
currycoder: Not too sure,. but I believe that the change to the model file for this is missing. Maybe as part of the branch splitting..?

datahub/company/models/company.py
```diff
@@ -284,6 +285,33 @@ def companies_house_data(self):
...
+            return sorted([
+                cec.country
+                for cec
+                in self.unfiltered_export_countries.all()
+                if not cec.disabled
+            ], key=lambda c: c.id)
+        else:
+            # An optimised query to get the countries in the case where
+            # the necessary prefetches have not been made.
+            return list(metadata_models.Country.objects.filter(
```
currycoder: I'm a little concerned about the two parallel paths of execution here depending on whether the queryset had the necessary prefetches or not. I feel like there is room for the two approaches to diverge without us knowing.
I'm also wary about the use of `_prefectched_objects_cache` as the contract that it provides may change unless it is explicitly documented by Django?

What's the drawback of always going with the "optimised query" approach?

datahub/company/models/company.py
```diff
@@ -459,3 +487,138 @@ def business_type(self):
...
+        Return boolean indicating whether country is disabled.
+        """
+        return bool(self.disabled_on)
+
+    def disable(self, user=None):
+        """
+        Disable by setting disabled_on to current time and disabled_by to `user`.
+        `user` should be None if this is not the result of user action.
+        """
+        self._dirty = True
```
currycoder: Can you help me understand what this is for?

datahub/company/models/company.py
```diff
@@ -459,3 +487,138 @@ def business_type(self):
...
+        self.disabled_on = now()
+        self.disabled_by = user
```
currycoder: Should this also set `modified_by`?

datahub/company/models/company.py
```diff
@@ -459,3 +487,138 @@ def business_type(self):
...
+        ):
+            # User source removal should always disable.
+            # Other sources should only disable if there are no other sources left.
+            self.disable(user)
+
+    def save_if_dirty(self):
+        """
+        If the attribute _dirty is set and truthy,
+        save the instance and then remove the _dirty attr.
+        """
```
currycoder: I'm a little unconvinced by this. Perhaps I haven't gotten there yet but I'm struggling to understand exactly where this will be used.

Perhaps add_source/remove_source could be adapted to save if a modification was made? This would prevent us needing to carry around this temporary state and make things a bit more explicit.

pytest.ini
```diff
@@ -3,3 +3,6 @@ addopts = --ds=config.settings.test
 norecursedirs = env
 filterwarnings =
     ignore:`settings.OMIS_NOTIFICATION_API_KEY`
+markers =
```
currycoder: I'm unfamiliar with this - what does it do? :)

datahub/company/models/company.py
```diff
@@ -284,6 +285,33 @@ def companies_house_data(self):
...
+            # that unfiltered_export_countries__country has been as well.
+            # If it hasn't then the below will cause N queries to get each country.
+            return sorted([
+                cec.country
+                for cec
+                in self.unfiltered_export_countries.all()
+                if not cec.disabled
+            ], key=lambda c: c.id)
+        else:
+            # An optimised query to get the countries in the case where
```
elcct: Is this comment necessary? I don't think it adds anything.

datahub/company/models/company.py
benpbenp: The "optimised query" would still be slow in the case of a list of a companies e.g. in the company search api, because it would have to make the query for every company in the list. If we wanted to just keep one of the branches it should probably be the first one, because it is probably acceptable to be a little bit slow when you are just dealing with one company.

pytest.ini
benpbenp: As I developed my feature I added tests in a number of different modules, so as a convenient way of running all of them I decorated them with pytest markers. However pytest raises a warning if the marker is not registered, to guard against people just mistyping a markers.

datahub/company/models/company.py
benpbenp: Yeah maybe you are right, I'll see if I can remove.

datahub/company/models/company.py
benpbenp: I don't know... not setting modified_by retains the information about the last person to touch the record before it was deleted. However, looking at it another way, the deletion is itself a modification. What do you think?

datahub/company/migrations/0095_add_company_export_country_model.py
benpbenp: yeah you are right. I will add the help text back into the model.

datahub/company/models/company.py
benpbenp: removed

datahub/company/models/company.py
benpbenp: Went with setting modified_by

datahub/company/models/company.py
benpbenp: ok removed

datahub/company/models/company.py
currycoder: That sounds good to me, especially if we can strip out the conditional which checks django internals.

sekharpanja: removed alternative path of execution

mcbhenwood:
Closed pending restructure of solution.

PR #2187, opened 2019-10-16
Add an API endpoint to export advisers

allait:
### Description of change

Exports adviser objects ordered by creation time and primary key.

The new approach for data-flow API endpoints is to export individual
objects as is (including their foreign key relationship IDs) and join
different tables/datasets within data-workspace instead of flattening
fields from the related models into the API response.

### Checklist

* [X] Do any added or updated endpoints appear in the API documentation? See [docs/Maintaining the API documentation.md](https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Maintaining&#32;the&#32;API&#32;documentation.md) for more details


datahub/dataset/adviser/views.py
```diff
@@ -0,0 +1,45 @@
...
+        paginator = self.pagination_class()
+        page = paginator.paginate_queryset(dataset, request, view=self)
+        return paginator.get_paginated_response(page)
+
+    def get_dataset(self):
+        """Returns list of Advisers Dataset records"""
+        return Adviser.objects.values(
+            'id',
+            'date_joined',
+            'email',
```
reupen: For info (and if not already aware), the `email` is unfortunately named and is really a username.

allait: Good shout. I've checked with Nick and we probably don't need it, so I've removed it for now.

datahub/dataset/adviser/views.py
```diff
@@ -0,0 +1,43 @@
...
+    def get_dataset(self):
+        """Returns list of Advisers Dataset records"""
+        return Adviser.objects.values(
+            'id',
+            'date_joined',
+            'first_name',
+            'last_name',
+            'telephone_number',
+            'contact_email',
+            'dit_team__id',
```
reupen: Just `dit_team_id` should work – I think we've restricted `__` to when we explicitly want a join.

allait: Oh yeah, you're right. Updated.

datahub/dataset/adviser/test/test_views.py
```diff
@@ -0,0 +1,108 @@
...
+        """Test that endpoint returns correct order of records"""
+        adviser_1 = AdviserFactory(date_joined=datetime(2019, 1, 2))
+        adviser_2 = AdviserFactory(date_joined=datetime(2019, 1, 3))
+        adviser_3 = AdviserFactory(date_joined=datetime(2019, 1, 1))
+
+        response = data_flow_api_client.get(self.view_url)
+        assert response.status_code == status.HTTP_200_OK
+
+        assert response.json()['results'] == [
+            get_expected_data_from_adviser(adviser_3),
```
reupen: In the equivalent test for the other data sets it only checks the IDs (since it's only testing the order here while the previous test checks the full adviser object). 

Also, they also test the scenario of duplicate `date_joined` values.

It would be good to keep this consistent.

datahub/dataset/adviser/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class AdvisersDatasetView(HawkResponseSigningMixin, APIView):
+    """
+    An APIView that provides 'get' action which queries and returns desired fields for
+    Advisers Dataset to be consumed by Data-flow periodically. Data-flow uses response result
+    to insert data into Dataworkspace through its defined API endpoints.
+    """
+
+    authentication_classes = (HawkAuthentication, )
```
reupen: I think we're adding `PaaSIPAuthentication` as the first authentication class – see https://github.com/uktrade/data-hub-leeloo/pull/2189.

changelog/adviser/dataset-api-endpoint.md
```diff
@@ -0,0 +1 @@
+A new dataset endpoint (`GET /v4/dataset/advisers-dataset`) was added to be consumed by data-flow and used in data-workspace.
```
reupen: The news fragment will need a type as well in the file name (so e.g. `dataset-api-endpoint.api.md`).

reupen:
@allait Could you check what's happened with the commits?

Also, I think this will need updating following https://github.com/uktrade/data-hub-leeloo/pull/2191. 

(With the introduction of `BaseDatasetView` and `DatasetCursorPagination`, we might benefit from an easier way of overriding the ordering on individual views. I think we could do that by adding `OrderingFilter` to `BaseDatasetView` and defining the ordering in the `ordering` attribute on the view.)

datahub/dataset/adviser/test/test_views.py
```diff
@@ -0,0 +1,115 @@
...
+import pytest
+from django.conf import settings
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.company.test.factories import AdviserFactory
+from datahub.core.test_utils import format_date_or_datetime, HawkAPITestClient
+
+
+@pytest.fixture
```
currycoder: Similar to the refactoring involved with `BaseDatasetView`, it could be good to reduce down some of the repetition in these test cases.

I see that e.g. `datahub/dataset/interaction/test/test_views.py` (and probably others) share a lot of test fixtures (which could be factored out in to a `conftest.py` file) and test methods (`test_other_methods_not_allowed`, `test_without_scope`, etc) which could be factored out in to a common base class.

This isn't critical for this PR, but it could be good to note this for future work.

reupen:
@allait Yep that works, this should be another option open to us:

> I think we could do that by adding OrderingFilter to BaseDatasetView and defining the ordering in the ordering attribute on the view.

But using a custom pagination class works as well.

datahub/dataset/adviser/test/test_views.py
currycoder: @allait Did this manage to go on to the backlog at all?

allait: @currycoder I don't think I have access to Data Hub boards, sorry.

PR #2193, opened 2019-10-18
Add self-assign company account manager endpoint

reupen:
### Description of change

This adds a new endpoint `POST /v4/company/<ID>/self-assign-account-manager` that:

- sets the authenticated user as the One List account manager
- sets the One List tier of the company to 'Tier D - Interaction Trade Adviser Accounts'

The operation is not allowed if:

- the company is a subsidiary of a One List company (on any tier)
- the company is already a One List company on a different tier (i.e. not 'Tier D - Interaction Trade Adviser Accounts')

The `company.change_company` and `company.change_regional_account_manager` permissions are required to use the endpoint.

I followed a similar pattern to the  OMIS action views (e.g. cancel order) as those have worked well.

### Checklist

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -7,7 +7,7 @@ from django.db import models
 from django.utils.translation import gettext_lazy
 from rest_framework import serializers
 
-from datahub.company.constants import BusinessTypeConstant
+from datahub.company.constants import BusinessTypeConstant, OneListTierID
 from datahub.company.models import (
     Advisor,
     CompaniesHouseCompany,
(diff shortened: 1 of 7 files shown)


datahub/company/serializers.py
```diff
@@ -522,6 +522,48 @@ class Meta:
...
+class SelfAssignAccountManagerSerializer(serializers.Serializer):
+    """
+    Serialiser for assigning an interaction trade adviser as the account manager of a
+    company.
+    """
+
+    target_one_list_tier_id = OneListTierID.tier_d_international_trade_advisers.value
+    default_error_messages = {
+        'cannot_change_account_manager_of_one_list_subsidiary':
+            "A lead adviser can't be set on a subsidiary of a One List company.",
```
elcct: Are we still using the translation functions?

reupen: Good point! Have updated it to use them.

datahub/company/views.py
```diff
@@ -12,6 +13,7 @@
     Advisor,
     CompaniesHouseCompany,
     Company,
+    CompanyPermission,
```
currycoder: Out of scope for this particular review, but can I take from this that the `add_company` and `change_company` permissions are not currently needed for a user adding/changing a company via the API?  I don't see any reference to those permissions on `CompanyVierSet`

reupen: They should be – it's handled by `DjangoCrudPermission`.

datahub/company/serializers.py
```diff
@@ -522,6 +522,48 @@ class Meta:
         }
 
 
+class SelfAssignAccountManagerSerializer(serializers.Serializer):
```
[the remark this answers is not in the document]
reupen: Yes, I followed a similar pattern to OMIS:

> I followed a similar pattern to the OMIS action views (e.g. cancel order) as those have worked well.

The various OMIS views and the archive and unarchive views are the main action views we have and they follow a similar pattern. The serialiser handles validation for us.

> the fields accepted are sort of a composite with CompanySerializer - right

No, the fields set by this view are read-only in the main view.

> `omis.order.serializers.CancelOrderSerializer` inherit from the overall serializer for the model - `OrderSerializer`

It looks to me like `CancelOrderSerializer` inherits from `OrderSerializer` for the class-level validators (since it doesn't use the same fields). So, we could do the same here – but to be honest there isn't really anything currently relevant in `CompanySerializer` (rather the opposite, the validators there shouldn't affect this operation). We can always review that later on, though.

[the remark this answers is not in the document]
currycoder: Thanks for the explainer - that's helpful.

PR #2195, opened 2019-10-18
Add `global_ultimate_duns_number` field to Company

alixedi:
### Description of change

This field will be used to store the `duns_number` for the global ultimate. We get this from D&B. Code for ingesting it in following PR.

### Checklist


datahub/company/models/company.py
```diff
@@ -239,6 +239,18 @@ class Company(ArchivableModel, BaseModel):
         choices=EXPORT_POTENTIAL_SCORES,
         help_text='Score that signifies export potential, imported from Data Science',
     )
+    global_ultimate_duns_number = models.CharField(
+        blank=True,
+        null=True,
+        help_text='Dun & Bradstreet unique identifier for global ultimate.',
+        max_length=9,
+        unique=True,
```
reupen: Is this meant to have `unique=True`? Doesn't that mean a global ultimate can only have one subsidiary?

alixedi: Good spot! Fixing now.

datahub/company/migrations/0095_company_global_ultimate_duns_number.py
```diff
@@ -20,7 +20,6 @@ class Migration(migrations.Migration):
                 help_text='Dun & Bradstreet unique identifier for global ultimate.',
                 max_length=9,
                 null=True,
```
reupen: FYI, we haven't generally been using `null=True` on `CharField`s (when `unique=False`) in line with this: https://docs.djangoproject.com/en/2.2/ref/models/fields/#null

datahub/company/models/company.py
currycoder: Whoops - I missed this too. Might be good to have an index instead still..

reupen: Yep, good point, assume we're planning to filter by it so `db_index=True` would indeed be good.

PR #2204, opened 2019-10-21
Add is_global_ultimate property to Company

currycoder:
### Description of change

Add `is_global_ultimate` property to Company model.  This will later be exposed on the CompanySerializer class for company related API endpoints.

### Checklist

### DIFF
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -308,6 +308,15 @@ class Company(ArchivableModel, BaseModel):
             except CompaniesHouseCompany.DoesNotExist:
                 return None
 
+    @property
+    def is_global_ultimate(self):
+        """
+        Whether this company is the global ultimate or not.
+        """
+        if not self.duns_number:
+            return False
+        return self.duns_number == self.global_ultimate_duns_number
+
     def mark_as_transferred(self, to, reason, user):
         """
         Marks a company record as having been transferred to another company record.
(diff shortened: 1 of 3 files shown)


datahub/company/models/company.py
```diff
@@ -308,6 +308,15 @@ def companies_house_data(self):
...
                 return None
 
+    @property
+    def is_global_ultimate(self):
+        """
+        Whether this company is the global ultimate or not.
+        """
+        if not self.duns_number:
+            return False
+        return self.duns_number == self.global_ultimate_duns_number
```
alixedi: 1. `self.duns_number` is `null=True, blank=True`, this might mean that it can be `None` and therefore we should check explicitly i.e. `if self.duns_number is not None`. 

2. Does this mean for companies that have `self.duns_number == self.global_ultimate_duns_number == None`, we return True? 

3. Might be good to add tests for these.

currycoder: > 1. `self.duns_number` is `null=True, blank=True`, this might mean that it can be `None` and therefore we should check explicitly i.e. `if self.duns_number is not None`.

My understanding is that the `if not self.duns_number:` block will be sufficient here - it will trigger if `duns_number==None` or `duns_number==''`

> 2. Does this mean for companies that have `self.duns_number == self.global_ultimate_duns_number == None`, we return True?

Nope because we return early in this case.

> 3. Might be good to add tests for these.

Will do!

PR #2207, opened 2019-10-21
Add an API endpoint to export Teams

allait:
### Description of change

This is similar to Advisors #2187 and other dataset endpoints,
but since teams don't have a creation date we're ordering the
table by ID only.

This has a small chance of potentially breaking pagination, but
(based on a discussion with @reupen) adding a creation date to the
model is out of scope for this and the table data doesn't change
that often so the risk is low.

Since Teams are pre-populated by a session-wide fixture the test
endpoint returns a lot of existing results. So instead of creating
new ones we check serialization for one of the existing teams and
test that items are sorted by ID.

### Checklist

* [X] Do any added or updated endpoints appear in the API documentation? See [docs/Maintaining the API documentation.md](https://github.com/uktrade/data-hub-leeloo/blob/develop/docs/Maintaining&#32;the&#32;API&#32;documentation.md) for more details

### DIFF
--- a/datahub/dataset/team/test/test_views.py
+++ b/datahub/dataset/team/test/test_views.py
@@ -1,0 +1,107 @@
+import pytest
+from django.conf import settings
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.core.test_utils import HawkAPITestClient
+from datahub.metadata.models import Team
+from datahub.metadata.test.factories import TeamFactory
+
+
+@pytest.fixture
+def hawk_api_client():
+    """Hawk API client fixture."""
+    yield HawkAPITestClient()
+
+
+@pytest.fixture
+def data_flow_api_client(hawk_api_client):
+    """Hawk API client fixture configured to use credentials with the data_flow_api scope."""
+    hawk_api_client.set_credentials(
+        'data-flow-api-id',
+        'data-flow-api-key',
+    )
+    yield hawk_api_client
+
+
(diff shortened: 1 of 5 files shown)


datahub/dataset/team/test/test_views.py
```diff
@@ -0,0 +1,107 @@
...
+        'id': str(team.id),
+        'name': team.name,
+        'role__name': team.role.name,
+        'uk_region__name': team.uk_region.name,
+        'country__name': team.country.name,
+    }
+
+
+@pytest.mark.django_db
+class TestTeamDatasetViewSet:
```
currycoder: My thoughts here still apply: https://github.com/uktrade/data-hub-leeloo/pull/2187#discussion_r336916412

Would be good to know if this is on the radar for some future refactor work.

allait: Yeah, the tests are largely the same across the endpoints and it makes sense to group them either into a base test class or as tests for the base view.

I'm not really sure how the Data Hub team tracks/prioritizes work like this though, so I'm probably not the best person to make sure it's captured somewhere. If you feel like it's worth doing now I can probably pick it up as part of this PR, as long as we're sure it's not conflicting with any other work or plans you have for these endpoints.

currycoder: Sounds good @allait thanks.  I'm not saying that this particular PR is blocked on this - just wanted to make sure that it's kept in mind for the data workspace team for some future work.

PR #2208, opened 2019-10-22
Speed up search tests

reupen:
### Description of change

This adds a new pytest fixture, `es_with_collector`, that collects saved objects and syncs them to Elasticsearch in bulk.

Tests that do not specifically test the search signal receivers have been updated to use this fixture. (Tests that do specifically test the signal receivers are still using `es_with_signals`.)

Locally, the running time of `pytest -n2 --reuse-db datahub/search/` is reduced from about 6.2 to 2.2 minutes. (Running _all_ tests locally now takes 7 minutes.)

There are two reasons that the new fixture helps:

- the standard signal-based syncing path re-fetches objects from the database individually (as it goes via a Celery task)
- the standard signal-based syncing path indexes objects individually

The new fixture re-fetches and indexes objects in bulk (resulting in the speed-up).

Another bit of slowness on CircleCI seems to be the database migrations (it spends about two minutes applying them) – we should be able to speed that up by squashing and optimising the migrations.

### Checklist

### DIFF
--- a/datahub/core/utils.py
+++ b/datahub/core/utils.py
@@ -1,6 +1,7 @@
 from enum import Enum
 from itertools import islice
 from logging import getLogger
+from uuid import UUID
 
 import requests
 from django.conf import settings
@@ -44,6 +45,25 @@ class EchoUTF8:
         if isinstance(value, str):
             return value.encode('utf-8')
         return value
+
+
+def force_uuid(value):
+    """
+    Convert value to a UUID if it isn't already and isn't None.
+
+    Useful if you have a value that could be a UUID object or a UUID as a string, but
+    want it to be a UUID in all cases.
+    """
+    if value is None:
+        return None
+
+    if isinstance(value, UUID):
+        return value
+
+    if isinstance(value, str):
+        return UUID(value)
+
+    raise TypeError('The value must be None or an instance of str or UUID.')
 
 
 def join_truthy_strings(*args, sep=' '):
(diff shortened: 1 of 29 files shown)


reupen:
Just pushed a rebase as I had a commit rewording locally that I forgot to push (the first two commits are otherwise unchanged).

I also pushed a small fix-up as I realised that it is better to use the `disable()` and `enable()` methods on `SignalReceiver` to temporarily disable them rather than `disconnect()` and `connect()`. (It shouldn't make a difference for the tests, but `disable()` and `enable()` are safe to use in this manner with `gevent` while `disconnect()` and `connect()` aren't.)

(Going to do a separate PR to make the same correction in `datahub.search.deletion.Collector`.)

reupen:
Rebasing to resolve the coverage complaint.

datahub/core/utils.py
```diff
@@ -46,6 +47,22 @@ def write(self, value):
...
+    Convert value to a UUID if it isn't already.
+
+    Useful if you have a value that could be a UUID object or a UUID as a string, but
+    want it to be a UUID in all cases.
+    """
+    if isinstance(value, UUID):
+        return value
+
+    if not value:
+        return None
```
elcct: I am not sure about this - the comment says we want UUID in all cases, but it seems there is a possibility to receive `None`. Maybe a comment could be updated?

reupen: I've tightened the logic and updated the docstring.

reupen:
@elcct @currycoder speaking to @sekharpanja , we are probably going to use an annotation for a new field being added to the company search model. 

This PR wasn't compatible with that, but I just pushed a fairly small change that will make it still work if annotations are used. The speed-up is still about the same. Could you have a look at the new commit and reapprove if happy?

PR #2214, opened 2019-10-23
Replace duplicate dataset tests with a base test class

allait:
### Add a dataset base view test class and shared fixtures

We have some largely identical tests that are repeated for each view. They tests authentication and basic responses and the only differences are the view url and the object factory used to set them up, which can be easily configured with class attributes.

This moves common tests to a base class in dataset.core.test. Since the tests depend on pytest fixtures we need to move them to a shared conftest to avoid unused imports in other test modules.

The base class could be in the conftest as well, but I think it makes more sense alongside the base view itself.

### Replace duplicate dataset tests with a base test class

This removes repeated tests and fixtures and replaces them with the conftest fixtures and test methods on the BaseDatasetViewTest class.

### Checklist

### DIFF
--- a/datahub/dataset/core/test.py
+++ b/datahub/dataset/core/test.py
@@ -1,5 +1,6 @@
+from unittest import mock
+
 import pytest
-from django.conf import settings
 from rest_framework import status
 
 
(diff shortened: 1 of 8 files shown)


datahub/dataset/core/test.py
```diff
@@ -0,0 +1,58 @@
...
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    def test_no_data(self, data_flow_api_client):
+        """Test that without any data available, endpoint completes the request successfully"""
+        response = data_flow_api_client.get(self.view_url)
+        assert response.status_code == status.HTTP_200_OK
+
+    def test_pagination(self, data_flow_api_client):
+        """Test that when page size higher than threshold response returns with next page url"""
+        self.factory.create_batch(settings.REST_FRAMEWORK['PAGE_SIZE'] + 1)
```
reupen: I realise this was like this before, but usually DRF settings are accessed via `api_settings` in `rest_framework.settings` – do you know if there is reason for it being this way?

Also, it looks like it's using the default of 100 items per page which seems a bit excessive for the tests, so it might be good to override it to a smaller value.

reupen:
Personally, I think I'd find parametrising the generic tests e.g. using `urlpatterns` a bit more direct... But, this is an improvement, so I'm cool with it.

datahub/dataset/core/test.py
currycoder: Agreed - would be good to override the setting.

allait: Updated. CursorPagination stores page size in an attribute on import, so this mocks the page size on the pagination class just for the test.

PR #2224, opened 2019-10-25
Add one list account manager filter to company search.

elcct:
### Description of change

`POST /v3/search/company/`: new filter `one_list_group_global_account_manager` was added.

This enables filtering by computed field `one_list_group_global_account_manager` that contains an id of one list account manager. If company is a subsidiary, this value will be inherited from the parent company.

### Checklist

### DIFF
--- a/changelog/company/add-one-list-group-account-manager-filter.api.md
+++ b/changelog/company/add-one-list-group-account-manager-filter.api.md
@@ -1,0 +1,1 @@
+`POST /v4/search/company`: new filter `one_list_group_global_account_manager` was added.
--- a/datahub/search/company/signals.py
+++ b/datahub/search/company/signals.py
@@ -4,7 +4,7 @@ from django.db.models.signals import post_save
 from datahub.company.models import Company as DBCompany
 from datahub.search.company import CompanySearchApp
 from datahub.search.signals import SignalReceiver
-from datahub.search.sync_object import sync_object_async
+from datahub.search.sync_object import sync_object_async, sync_related_objects_async
 
 
 def company_sync_es(instance):
@@ -14,4 +14,14 @@ def company_sync_es(instance):
     )
 
 
-receivers = (SignalReceiver(post_save, DBCompany, company_sync_es),)
+def company_subsidiaries_sync_es(instance):
+    """Sync company subsidiaries to the Elasticsearch."""
+    transaction.on_commit(
+        lambda: sync_related_objects_async(instance, 'subsidiaries'),
+    )
+
+
+receivers = (
+    SignalReceiver(post_save, DBCompany, company_sync_es),
+    SignalReceiver(post_save, DBCompany, company_subsidiaries_sync_es),
+)
--- a/datahub/search/company/test/test_elasticsearch.py
+++ b/datahub/search/company/test/test_elasticsearch.py
@@ -218,6 +218,26 @@ def test_mapping(es):
                         },
                     },
                 },
+                'one_list_group_global_account_manager': {
+                    'properties': {
+                        'first_name': {
+                            'index': False,
+                            'type': 'text',
+                        },
+                        'id': {
+                            'type': 'keyword',
+                        },
+                        'last_name': {
+                            'index': False,
+                            'type': 'text',
+                        },
+                        'name': {
+                            'index': False,
+                            'type': 'text',
+                        },
+                    },
+                    'type': 'object',
+                },
                 'sector': {
                     'properties': {
                         'ancestors': {
@@ -531,6 +551,7 @@ def test_indexed_doc(es):
         'reference_code',
         'address',
         'registered_address',
+        'one_list_group_global_account_manager',
         'sector',
         'suggest',
         'trading_names',
--- a/datahub/search/company/test/test_entity_search_views.py
+++ b/datahub/search/company/test/test_entity_search_views.py
@@ -1,3 +1,4 @@
+import random
 import uuid
 from cgi import parse_header
 from csv import DictReader
@@ -13,7 +14,7 @@ from rest_framework import status
 from rest_framework.reverse import reverse
 
 from datahub.company.models import Company, CompanyPermission
-from datahub.company.test.factories import CompanyFactory
+from datahub.company.test.factories import AdviserFactory, CompanyFactory
 from datahub.core import constants
 from datahub.core.exceptions import DataHubException
 from datahub.core.test_utils import (
@@ -117,12 +118,13 @@ class TestSearch(APITestMixin):
 
     def test_response_body(self, es_with_collector):
         """Tests the response body of a search query."""
+        one_list_account_owner = AdviserFactory()
         company = CompanyFactory(
             company_number='123',
             trading_names=['Xyz trading', 'Abc trading'],
             global_headquarters=None,
             one_list_tier=None,
-            one_list_account_owner=None,
+            one_list_account_owner=one_list_account_owner,
         )
         es_with_collector.flush_and_refresh()
 
@@ -164,6 +166,12 @@ class TestSearch(APITestMixin):
                             'id': str(company.registered_address_country.id),
                             'name': company.registered_address_country.name,
                         },
+                    },
+                    'one_list_group_global_account_manager': {
+                        'id': str(one_list_account_owner.id),
+                        'first_name': one_list_account_owner.first_name,
+                        'last_name': one_list_account_owner.last_name,
+                        'name': one_list_account_owner.name,
                     },
                     'uk_based': (
                         company.address_country.id == uuid.UUID(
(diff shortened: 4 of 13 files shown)


datahub/search/dict_utils.py
```diff
@@ -138,11 +138,30 @@ def get_dict(obj):
...
+
+    return get_dict
+
+
 def computed_nested_id_name_dict(nested_field):
     """Creates a dictionary with id and name from a nested field."""
     return _computed_nested_dict(nested_field, id_name_dict)
 
 
+def computed_nested_contact_or_adviser_dict(nested_field):
```
reupen: Is this used anywhere?

datahub/search/company/test/test_elasticsearch.py
```diff
@@ -218,6 +218,35 @@ def test_mapping(es):
                         },
                     },
                 },
+                'one_list_group_global_account_manager': {
```
reupen: We're indexing quite a lot here, when it seems like we only need `id` to be indexed. Would be good to only index the field being filtered by, and set `index=False` on anything else.

datahub/search/dict_utils.py
elcct: Oops, I forgot to remove. Thank you!

datahub/search/company/test/test_elasticsearch.py
elcct: That's a very good point. I replaced this field with one that only indexes the id.

elcct:
The `one_list_account_owner` was already in the select-related fields list, but `global_headquarters__one_list_account_owner` wasn't, so I added it. Thank you!

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -361,6 +368,97 @@ def test_headquarter_type_filter(self, setup_headquarters_data, query, results):
...
+            (0, 1, 2),
+            (0, 1),
+            (0, 2),
+            (2, ),
+            (1, 0),
+        ),
+    )
+    def test_one_list_account_manager_with_global_headquarters_filter(
+        self,
+        selected_account_manager,
```
reupen: Doesn't seem like `selected_account_manager` is used here.

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -361,6 +368,97 @@ def test_headquarter_type_filter(self, setup_headquarters_data, query, results):
...
+        )
+
+        assert response.status_code == status.HTTP_200_OK
+
+        search_results = {company['id'] for company in response.data['results']}
+        expected_results = {
+            str(global_headquarters.id),
+            *{str(target_company.id) for target_company in target_companies},
+        }
+        assert search_results == expected_results
```
reupen: Would be good to add another assertion for the number of results as well (to make sure there are no duplicates).

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -361,6 +368,97 @@ def test_headquarter_type_filter(self, setup_headquarters_data, query, results):
...
         assert search_results == results
 
+    @pytest.mark.parametrize(
+        'selected_account_manager',
+        (
+            (0, 1, 2),
+            (0, 1),
+            (0, 2),
+            (2, ),
+            (1, 0),
```
reupen: I'm not sure on the practical difference between `(0, 1)`, `(0, 2)`, `(1, 0)`. If we want to test with 1, 2 and 3 advisers maybe we could use `random.sample(account_managers, n)` and parametrise `n`?

datahub/search/company/test/test_elasticsearch.py
```diff
@@ -18,7 +18,6 @@ def test_mapping(es):
         CompanySearchApp.es_model.get_write_index(),
         CompanySearchApp.name,
     )
-
```
reupen: Is this change intended?

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -361,6 +368,97 @@ def test_headquarter_type_filter(self, setup_headquarters_data, query, results):
...
+        ),
+    )
+    def test_one_list_account_manager_filter(self, selected_account_manager, es_with_signals):
```
reupen: Should update these to `es_with_collector` (instead of `es_with_signal`) when we rebase.

datahub/search/company/test/test_entity_search_views.py
elcct: Oh. Yes, I should have removed that. Thank you!

datahub/search/company/test/test_entity_search_views.py
elcct: Good point! I added additional assertions now. Thank you!

datahub/search/company/test/test_entity_search_views.py
elcct: I changed this test to follow your suggestion. I think it is better now because it also checks the contents of `one_list_group_global_account_manager` field. Thank you!

datahub/search/company/test/test_elasticsearch.py
elcct: It is not. Fixed now. Thank you.

datahub/search/company/test/test_entity_search_views.py
elcct: I'll keep in mind!

elcct:
This is a good point, I missed that! I have changed the signal to also sync subsidiaries if a company is a headquarter and added test to check that `one_list_group_global_account_manager` is populated when  `one_list_account_owner` of global hq is changed.

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -361,6 +369,132 @@ def test_headquarter_type_filter(self, setup_headquarters_data, query, results):
...
+        search_results = {company['id'] for company in response.data['results']}
+        expected_results = {
+            str(global_headquarters.id),
+            *{str(target_company.id) for target_company in target_companies},
+        }
+        assert response.data['count'] == 3
+        assert len(response.data['results']) == 3
+        assert search_results == expected_results
+
+    def test_subsidiary_is_synced_to_es(self, es_with_signals):
```
reupen: I think we should add a specific test for the signal receiver for this (in `test_signals`) instead of testing this via the view.

datahub/search/company/signals.py
```diff
@@ -8,10 +8,15 @@
...
-    """Sync company to the Elasticsearch."""
-    transaction.on_commit(
-        lambda: sync_object_async(CompanySearchApp, instance.pk),
-    )
+    """Sync company and subsidiaries to the Elasticsearch."""
+    def sync_es_wrapper():
+        sync_object_async(CompanySearchApp, instance.pk)
 
+        if instance.headquarter_type_id:
+            for company in instance.subsidiaries.all():
```
reupen: It would be good to use `sync_related_objects_async` for this (similar to other search apps):

https://github.com/uktrade/data-hub-leeloo/blob/470e8ad003057b3af655fea7ea4b7d23e8d1d6d8/datahub/search/contact/signals.py#L17-L21

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -361,6 +369,132 @@ def test_headquarter_type_filter(self, setup_headquarters_data, query, results):
         search_results = {company['name'] for company in response.data['results']}
         assert search_results == results
 
+    @pytest.mark.parametrize(
+        'n',
```
reupen: Sorry, I just meant `n` as example – would be good to use something more descriptive e.g. `num_account_managers`.

datahub/search/company/test/test_entity_search_views.py
elcct: This is a good idea. I removed this test and created a new one in `test_signals`. Thanks!

datahub/search/company/signals.py
elcct: I did it this way first but it didn't seem to be working. Maybe I got something else wrong, because I redo it with `sync_related_objects_async` and it now works. Thank you!

datahub/search/company/test/test_entity_search_views.py
elcct: Haha yes I was thinking the use of `n` was odd, but perhaps didn't have enough coffee to transform it into something like `num_account_managers`. It is fixed now. Thank you!

changelog/company/add-one-list-group-account-manager-filter.api.md
```diff
@@ -0,0 +1 @@
+`POST /v3/search/company/`: new filter `one_list_group_global_account_manager` was added.
```
currycoder: Should this be `POST /v4/search/company`?

elcct: Oh good catch! Updated now. Thank you.

datahub/search/company/test/test_signals.py
```diff
@@ -35,3 +35,48 @@ def test_company_auto_updates_to_es(es_with_signals):
...
+    )
+    same_name = 'very_hard_to_find_company_international'
+    subsidiaries = CompanyFactory.create_batch(
+        2,
+        name=same_name,
+        global_headquarters=global_headquarters,
+    )
+    es_with_signals.indices.refresh()
+
+    result = get_basic_search_query(Company, same_name).execute()
```
reupen: I know the other tests in this module work this way, but rather than doing a search (which seems irrelevant for testing if the signal receiver works), I think it would be better if we fetched the document directly using something like:

```python
es_with_signals.get(
    index=CompanySearchApp.es_model.get_read_alias(),
    doc_type=CompanySearchApp.name,
    id=company.pk,
)
```

(Could add a helper function to `datahub.search.test.utils` for this as well.)

elcct: I see. I would think that `get_basic_search_query` plays a role of such helper, but I can imagine how this is not the best way to do it. I'll try to modify the test though. Thank you!

datahub/search/test/utils.py
```diff
@@ -62,3 +62,14 @@ def db_objects_to_es_documents(db_objects, index=None):
         get_write_alias=Mock(return_value='test-write-alias'),
         get_target_index_name=Mock(return_value=f'test-index-{target_mapping_hash}'),
     )
+
+
+def get_documents_by_id(es_with_signals, app, ids):
```
reupen: `get_documents_by_ids` might be clearer (since it takes multiple IDs).

elcct:
I'll fix the typos whenI rebase.

PR #2272, opened 2019-11-07
Sync company from dnb celery task

currycoder:
### Description of change

This PR adds a celery task to sync a company's data with DNB.  The task can optionally be called with a list of fields to sync.

Based on Reupen's suggestion in the previous PR, I've ensured that `Company.modified_on` is not changed by this task as the aspiration is for this to be initiated by a batch/automated process.

The next PR to follow will be a management command which calls this task for a batch of company IDs.

### Checklist

### DIFF
--- a/datahub/dnb_api/serializers.py
+++ b/datahub/dnb_api/serializers.py
@@ -14,6 +14,33 @@ from datahub.core.validators import EqualsRule, OperatorRule, RulesBasedValidato
 from datahub.interaction.models import InteractionPermission
 
 
+# TODO: Use this as a DRY reference for all code relating to updated DNB fields in the future
+ALL_DNB_UPDATED_FIELDS = [
+    'name',
+    'trading_names',
+    'address_1',
+    'address_2',
+    'address_town',
+    'address_county',
+    'address_country',
+    'address_postcode',
+    'registered_address_1',
+    'registered_address_2',
+    'registered_address_town',
+    'registered_address_county',
+    'registered_address_country',
+    'registered_address_postcode',
+    'website',
+    'number_of_employees',
+    'is_number_of_employees_estimated',
+    'turnover',
+    'is_turnover_estimated',
+    'website',
+    'global_ultimate_duns_number',
+    'company_number',
+]
+
+
 class DNBMatchedCompanySerializer(PermittedFieldsModelSerializer):
     """
     Serialiser for data hub companies matched with a DNB entry.
@@ -115,6 +142,23 @@ class DNBCompanySerializer(CompanySerializer):
                 ),
             ),
         )
+
+    def save_dnb_fields(self, **kwargs):
+        """
+        Method to save the instance - by writing only the fields updated by DNB.
+        Takes kwargs to override the values of specific fields for the model
+        on save - these fields are not subject to the DNB-only restriction.
+
+        Note: modified_on will not be updated by this method.
+        """
+        instance = self.instance
+        validated_data = dict(
+            list(self.validated_data.items()) + list(kwargs.items()),
+        )
+        for field, value in validated_data.items():
+            setattr(instance, field, value)
+        update_fields = ALL_DNB_UPDATED_FIELDS + list(kwargs.keys())
+        instance.save(update_fields=update_fields)
 
 
 class DUNSNumberSerializer(serializers.Serializer):
--- a/datahub/dnb_api/serializers.py
+++ b/datahub/dnb_api/serializers.py
@@ -14,33 +14,6 @@ from datahub.core.validators import EqualsRule, OperatorRule, RulesBasedValidato
 from datahub.interaction.models import InteractionPermission
 
 
-# TODO: Use this as a DRY reference for all code relating to updated DNB fields in the future
-ALL_DNB_UPDATED_FIELDS = [
-    'name',
-    'trading_names',
-    'address_1',
-    'address_2',
-    'address_town',
-    'address_county',
-    'address_country',
-    'address_postcode',
-    'registered_address_1',
-    'registered_address_2',
-    'registered_address_town',
-    'registered_address_county',
-    'registered_address_country',
-    'registered_address_postcode',
-    'website',
-    'number_of_employees',
-    'is_number_of_employees_estimated',
-    'turnover',
-    'is_turnover_estimated',
-    'website',
-    'global_ultimate_duns_number',
-    'company_number',
-]
-
-
 class DNBMatchedCompanySerializer(PermittedFieldsModelSerializer):
     """
     Serialiser for data hub companies matched with a DNB entry.
@@ -143,21 +116,20 @@ class DNBCompanySerializer(CompanySerializer):
             ),
         )
 
-    def save_dnb_fields(self, **kwargs):
+    def partial_save(self, **kwargs):
         """
-        Method to save the instance - by writing only the fields updated by DNB.
-        Takes kwargs to override the values of specific fields for the model
-        on save - these fields are not subject to the DNB-only restriction.
+        Method to save the instance - by writing only the fields updated by the serializer.
+        Takes kwargs to override the values of specific fields for the model on save.
 
         Note: modified_on will not be updated by this method.
         """
+        if not self.partial:
+            raise Exception('partial_save() called, but serializer is not set as partial.')
         instance = self.instance
-        validated_data = dict(
-            list(self.validated_data.items()) + list(kwargs.items()),
-        )
+        validated_data = {**self.validated_data, **kwargs}
         for field, value in validated_data.items():
             setattr(instance, field, value)
-        update_fields = ALL_DNB_UPDATED_FIELDS + list(kwargs.keys())
+        update_fields = validated_data.keys()
         instance.save(update_fields=update_fields)
 
 
--- a/datahub/dnb_api/serializers.py
+++ b/datahub/dnb_api/serializers.py
@@ -14,6 +14,13 @@ from datahub.core.validators import EqualsRule, OperatorRule, RulesBasedValidato
 from datahub.interaction.models import InteractionPermission
 
 
+class SerializerNotPartial(Exception):
+    """
+    Exception for when some logic was called which expects a serializer object
+    to be self.partial=True.
+    """
+
+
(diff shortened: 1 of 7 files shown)


datahub/dnb_api/utils.py
```diff
@@ -213,9 +226,24 @@ def update_company_from_dnb(dh_company, dnb_company, user, fields_to_update=None
...
+
+        company_serializer.save(**company_kwargs)
+
+        # When an update is made by an automatic process, keep the modified_on field
+        # unchanged as this field should only be when a change was initiated by a user
+        # Unfortunately, we have to do this as a second update call to avoid django's
+        # `auto_now` machinery from triggering - there's no way to avoid this when
+        # saving with a serializer
+        if not user:
+            Company.objects.filter(id=dh_company.id).update(modified_on=original_modified_on)
```
currycoder: Note: I tried using `update_fields` with the above serializer `save()` method to avoid this approach - but having looked through the DRF source I'm pretty confident that the `serializer.save()` method does not pass this kwarg through to the django model `save()` method.  It looks to me like this is unavoidable while we use `serializer.save()`

elcct: Is this going to bypass signals so that the record e.g. won't get updated in Elasticsearch?

reupen: What if we added another method to `DNBCompanySerializer` to do the saving for this case? It would only need to handle updating, so should be fairly simple I think...

datahub/dnb_api/test/test_tasks.py
```diff
@@ -0,0 +1,211 @@
...
+    original_company = Company.objects.get(id=company.id)
+    sync_company_with_dnb.apply_async(
+        args=[company.id],
+        kwargs={'fields_to_update': ['global_ultimate_duns_number']},
+    )
+    company.refresh_from_db()
+    assert model_to_dict(company) == {
+        'address_1': original_company.address_1,
+        'address_2': original_company.address_2,
+        'address_country': original_company.address_country.id,
```
reupen: I think it would be good to use `original_company.address_country_id` (rather than `original_company.address_country.id`):

- it has better behaviour if the field is None
- avoids extra queries (if the related object isn't cached)

datahub/dnb_api/tasks.py
```diff
@@ -0,0 +1,59 @@
...
+logger = get_task_logger(__name__)
+
+
+def _sync_company_with_dnb(company_id, fields_to_update, task):
+    dh_company = Company.objects.get(id=company_id)
+
+    try:
+        dnb_company = get_company(dh_company.duns_number)
+    except DNBServiceError as exc:
+        server_error_statuses = (
```
reupen: All 5xx codes are server errors – could we just check the range?

datahub/dnb_api/tasks.py
```diff
@@ -0,0 +1,59 @@
...
+    company record to sync and `fields_to_update` defines an iterable of
+    company serializer fields that should be updated - if it is None, all fields
+    will be synced.
+    """
+    try:
+        _sync_company_with_dnb(company_id, fields_to_update, self)
+    except Retry:
+        raise
+    except Exception:
+        logger.error(f'Encountered an error when syncing Company "{company_id}" with DNB')
```
reupen: Perhaps we could rely on the default behaviour of an error (and Sentry alert) if it fails on the final attempt (and info messages on retries)?

That way, Sentry alerts will be reserved for when we need to do something/when it failed fatally.

datahub/dnb_api/utils.py
currycoder: Good point @elcct - thanks

@reupen I think this is a very good idea - I'll have a go at this and see what I can come up with.

currycoder: Ok please see the new `save_dnb_fields()` method.

datahub/dnb_api/tasks.py
```diff
@@ -52,8 +46,10 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
     """
     try:
         _sync_company_with_dnb(company_id, fields_to_update, self)
-    except Retry:
-        raise
-    except Exception:
-        logger.error(f'Encountered an error when syncing Company "{company_id}" with DNB')
+    except Retry as exc:
+        status_code = exc.exc.status_code
+        logger.info(
```
reupen: We do get a similar log message from Celery – I think we could just define a better `__repr__()` on `DNBServiceError` so the Celery message is more useful

datahub/dnb_api/tasks.py
```diff
@@ -19,13 +19,7 @@ def _sync_company_with_dnb(company_id, fields_to_update, task):
...
         dnb_company = get_company(dh_company.duns_number)
     except DNBServiceError as exc:
-        server_error_statuses = (
-            status.HTTP_500_INTERNAL_SERVER_ERROR,
-            status.HTTP_502_BAD_GATEWAY,
-            status.HTTP_503_SERVICE_UNAVAILABLE,
-            status.HTTP_504_GATEWAY_TIMEOUT,
-        )
-        if exc.status_code in server_error_statuses:
+        if is_server_error(exc.status_code):
```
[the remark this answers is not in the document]
currycoder: Thanks for pointing this out. I'd like to break this out in to a subsequent PR which will follow this.  I'd like to prevent the scope creeping too much in this PR.

datahub/dnb_api/tasks.py
currycoder: Agreed.

currycoder: Looks like the message passed in is probably good enough in this case - have adjusted it as:
`error_message = f'DNB service returned an error status: {dnb_response.status_code}'`

datahub/dnb_api/serializers.py
```diff
@@ -116,6 +116,22 @@ class Meta(CompanySerializer.Meta):
...
+        """
+        Method to save the instance - by writing only the fields updated by the serializer.
+        Takes kwargs to override the values of specific fields for the model on save.
+
+        Note: modified_on will not be updated by this method.
+        """
+        if not self.partial:
+            raise Exception('partial_save() called, but serializer is not set as partial.')
+        instance = self.instance
+        validated_data = {**self.validated_data, **kwargs}
```
alixedi: Does this mean that we are bypassing validation?

reupen: No, `self.validated_data` means it's been validated.

alixedi: Discussed IRL, I think the broader point I was trying to make is that we should only be using this method to not trigger setting `modified_on` and therefore it might be useful to rename this method to reflect that.

datahub/dnb_api/test/test_utils.py
```diff
@@ -168,20 +169,34 @@ class TestUpdateCompanyFromDNB:
     Test update_company_from_dnb utility function.
     """
 
+    @pytest.mark.parametrize(
```
alixedi: Should we nest the parametrisation here to test all the permutations?

```
@pytest.mark.parametrize(
    'adviser',
    (None, AdviserFactory()),
)
@pytest.mark.parametrize(
    'update_descriptor',
    (None, 'automatic'),
)
```

P.S. See next comment on `s/adviser_callable/adviser`

datahub/dnb_api/serializers.py
```diff
@@ -116,6 +116,22 @@ class Meta(CompanySerializer.Meta):
...
 
+    def partial_save(self, **kwargs):
+        """
+        Method to save the instance - by writing only the fields updated by the serializer.
+        Takes kwargs to override the values of specific fields for the model on save.
+
+        Note: modified_on will not be updated by this method.
+        """
+        if not self.partial:
+            raise Exception('partial_save() called, but serializer is not set as partial.')
```
alixedi: Might be a good idea to raise a more specific Exception here.

datahub/dnb_api/test/test_utils.py
```diff
@@ -272,18 +302,26 @@ def test_update_company_from_dnb_partial_fields_single(
         assert company.name == original_company.name
         assert company.number_of_employees == original_company.number_of_employees
 
+    @pytest.mark.parametrize(
+        'adviser_callable',
```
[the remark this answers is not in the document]
currycoder: You can't call DB operations at the file level.  It errors with message:
```
Failed: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it
```

datahub/dnb_api/test/test_utils.py
currycoder: Nice idea!

datahub/dnb_api/serializers.py
currycoder: Good point

datahub/dnb_api/test/test_tasks.py
```diff
@@ -0,0 +1,211 @@
...
+        'address_country': original_company.address_country_id,
+        'address_county': original_company.address_county,
+        'address_postcode': original_company.address_postcode,
+        'address_town': original_company.address_town,
+        'archived': original_company.archived,
+        'archived_by': original_company.archived_by,
+        'archived_documents_url_path': original_company.archived_documents_url_path,
+        'archived_on': original_company.archived_on,
+        'archived_reason': original_company.archived_reason,
+        'business_type': original_company.business_type.id,
```
reupen: Seems like we still have a few `.id`s which could be converted to `_id` – would be good to update those to avoid extra queries.

datahub/dnb_api/test/test_tasks.py
```diff
@@ -0,0 +1,211 @@
...
+        json=dnb_response_uk,
+    )
+    company = CompanyFactory(duns_number='123456789')
+    original_company = Company.objects.get(id=company.id)
+    sync_company_with_dnb.apply_async(
+        args=[company.id],
+        kwargs={'fields_to_update': ['global_ultimate_duns_number']},
+    )
+    company.refresh_from_db()
+    assert model_to_dict(company) == {
```
reupen: It would be good to check if the task succeeded before doing this assertion (should be some examples in other tests).

(Same comment for the test above.)

PR #2288, opened 2019-11-11
Optimise entity search tests by syncing only the app being tested

reupen:
### Description of change

This adds a simple optimisation to the entity search view tests so that only the search app being tested is synced to Elasticsearch.

This is done by making the `es_with_collector` fixture only sync the apps set in the `es_collector_apps` marker (if it is set).

(I went for this approach as it is easy to set a marker on a whole module.)

(This optimisation is possible as each entity search view only searches the index for a particular search app.)

This reduces the running time of `pytest -n2 --reuse-db datahub/search/` by around 13% for me (locally).

### Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -270,23 +270,24 @@ class SavedObjectCollector:
     """
     Collects the search apps of saved search objects and indexes those apps in bulk in
     Elasticsearch.
-
-    (A possible future optimisation would be to further restrict the synced apps to the app
-    being tested i.e. in tests for the interaction search app, only sync interaction objects
-    and not companies, contacts etc.)
     """
 
-    def __init__(self, es_client):
-        """Initialises the collector."""
-        self.apps_to_sync = set()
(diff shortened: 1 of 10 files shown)


conftest.py
```diff
@@ -332,11 +333,11 @@ def _collect(self, obj):
         """
         model = obj.__class__
         search_app = get_search_app_by_model(model)
-        self.apps_to_sync.add(search_app)
+        self.collected_apps.add(search_app)
 
 
 @pytest.fixture
-def es_with_collector(es, synchronous_on_commit):
+def es_with_collector(es, synchronous_on_commit, request):
```
currycoder: I don't quite understand where `request` comes from or what the significance is here..?

reupen: It's a built-in fixture: https://docs.pytest.org/en/latest/reference.html#request

It's used a few lines down to get the markers that have been applied to a test.

currycoder: I missed that somehow!  Looks good to me.

PR #2293, opened 2019-11-12
Sync multiple DNB companies tool

currycoder:
### Description of change

This PR brings the work for the `datahub.dnb_api.tasks.sync_company_from_dnb` task in to a manual tool that we can use to update a list of company IDs with the latest DNB data.  This will be used to progressively update DNB data for companies in Tier A of the one list.

The key logic in this management command relates to enforcing a maximum limit of API calls per second.  Right now, this is set to one call per second - this should keep us well within safeties for ensuring that we do not exceed our limit (of 5 calls per second) with DNB.  The command is long running and will effectively call the API synchronously.

As discussed in the Backend channel, we determined that making a long running synchronous command was fine for for the purposes of this manual tool.  Celery's `countdown` and `rate_limit` options were investigated but it was determined that they did not give enough control/confidence that we would avoid exceeding our desired rate limit for API calls.

### Checklist


datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if not company.duns_number:
+            # TODO: Log this?
+            logger.warning('Company does not have a duns_number.')
+            raise CompanyNotDunsLinked()
+
+        if simulate:
+            return
+
+        self._limit_call_rate()
```
alixedi: For the benefit of simplicity, can we not replace this method with `sleep(1000)` given that an approximate rate is probably fine in this case?

currycoder: So it can take between 300-500ms (anecdotal) to get a response from dnb-service. Hence why I was trying to be a little more clever here so that we can get through a long list of companies as quickly as possible. Happy to discuss this though.

datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
```diff
@@ -0,0 +1,93 @@
...
+class Command(CSVBaseCommand):
+    """
+    Command to update companies with the latest DNB data.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Set some initial state related to API rate limiting.
+        """
+        self.last_called_api_time = time.time()
```
reupen: Would suggest using [`time.perf_counter()`](https://docs.python.org/3.7/library/time.html#time.perf_counter) instead as it isn't subject to system clock changes etc.

datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
```diff
@@ -0,0 +1,93 @@
...
+        self.api_call_wait = 1 / API_CALLS_PER_SECOND
```
reupen: Seems like this is only written to once, can it be a module-level constant?

datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
```diff
@@ -0,0 +1,93 @@
...
+        pk = parse_uuid(row['id'])
+        try:
+            company = Company.objects.get(pk=pk)
+        except Company.DoesNotExist:
+            logger.warning('Company does not exist.')
+            raise
+
+        if not company.duns_number:
+            # TODO: Log this?
+            logger.warning('Company does not have a duns_number.')
```
reupen: Might be able to get rid of this – `CSVBaseCommand` logs failures using `logger.exception()` so the exception shoudl be logged.

datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
elcct: I also think that `interval` will describe it better.

datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
elcct: Is this limiting coming from the API requirements?

currycoder: Yes it is. We are limited to a maximum of 5 requests per second based on our plan with DNB and there is no separation between the API account used for these bulk calls and the calls on the "add a company" journey - so we need to be careful not to hit the limit.

There is also some parallel work that will happen in https://github.com/uktrade/dnb-service to add some extra mechanisms to ensure that we do not exceed the agreed rate limits.

datahub/dbmaintenance/test/commands/test_update_companies_dnb_data.py
```diff
@@ -0,0 +1,189 @@
...
+    command._limit_call_rate()
+    sleep_time = mock_sleep.call_args[0][0]
+    assert sleep_time == API_CALL_INTERVAL
+    assert command.last_called_api_time == time_at_freeze + API_CALL_INTERVAL
+
+    # Simulate some execution time outside of _limit_call_rate which takes
+    # (maximum wait time / 2) and ensure that sleep is called to maintain
+    # the API_CALLS_PER_SECOND limit
+    gap_seconds = API_CALL_INTERVAL / 2
+    mock_perf_counter.return_value += gap_seconds
```
reupen: It's possible to do e.g. `Mock(side_effect=[1.0, 2.0])` as well – might be cleaner than mutating `return_value`.

datahub/dbmaintenance/management/commands/update_companies_dnb_data.py
```diff
@@ -0,0 +1,91 @@
+import logging
```
reupen: The other company commands start with `update_company_` – would be nice to keep it consistent :slightly_smiling_face:

datahub/dbmaintenance/test/commands/test_update_companies_dnb_data.py
```diff
@@ -0,0 +1,189 @@
...
+    mock_sleep = mock_time['mock_sleep']
+    mock_perf_counter = mock_time['mock_perf_counter']
+    time_at_freeze = mock_perf_counter.return_value
+    command = Command()
+
+    # Call the _limit_call_rate() method and ensure that sleep is called with
+    # the maximum wait time
+    command._limit_call_rate()
+    sleep_time = mock_sleep.call_args[0][0]
+    assert sleep_time == API_CALL_INTERVAL
```
reupen: I think `assert mock_sleep.call_args == mock.call(API_CALL_INTERVAL)` would be better since it avoids indexing (it requires some mental effort to work out what `call_args[0][0]` is...)

datahub/dbmaintenance/test/commands/test_update_companies_dnb_data.py
currycoder: I built it this way in order to try to emulate freezegun's `tick()` method. Which effectively allows the caller to progress time by a certain value.

The solution here modifies the `return_value` in a couple of places (also see the side effect for `time.sleep()`).  This is a simplification that allows us to have a mocked time value that we can tick for the needs of testing - I agree that the underlying calls aren't as readable as freezegun's explicit API.  For now, could better commenting help?

PR #2295, opened 2019-11-12
Optimise cleanup management command tests

reupen:
### Description of change

This optimises tests for the `cleanup` app so that they sync objects to Elasticsearch in bulk during set-up.

These tests need to use the normal Elasticsearch signals when the command is run, so a new pytest fixture was added that can be used to temporarily activate the Elasticsearch collector (so that it can be used for test set-up only in this case).

Locally, this reduces the time to run `pytest --reuse-db datahub/cleanup/` from 49 seconds to 27 seconds.

(Some of the changes are white-space only, those can be hidden using the cog button on the diff view.)

### Checklist

### DIFF
--- a/datahub/cleanup/test/commands/test_delete_orphans.py
+++ b/datahub/cleanup/test/commands/test_delete_orphans.py
@@ -194,6 +194,7 @@ def test_run(
     dep_field_name,
     track_return_values,
     es_with_signals,
+    es_collector_context_manager,
 ):
     """
     Test that:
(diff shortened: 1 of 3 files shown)


datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -278,19 +288,21 @@ def test_simulate(cleanup_configs, track_return_values, es_with_signals):
...
 
     model = apps.get_model(model_name)
     search_app = get_search_app_by_model(model)
     read_alias = search_app.es_model.get_read_alias()
 
     assert model.objects.count() == 3
     assert es_with_signals.count(read_alias, doc_type=search_app.name)['count'] == 3
-    management.call_command(command, model_name, simulate=True)
 
+    # Run the command
```
elcct: Do we need this comment?

reupen: It doesn't add much by itself, but just thought it would be useful to add headings for each bit since there is a lot going on.

PR #2312, opened 2019-11-19
Add Staff SSO OAuth admin views

elcct:
### Description of change

This is the first PR.

This PR adds views to perform OAuth2 authentication with Staff SSO for accessing Data Hub Django admin. 

Following environment variables need to be set for OAuth2 authentication:

| Variable name | Required | Description |
| ------------- | ------------- | ------------- |
| `ADMIN_OAUTH2_ENABLED` | Yes | Enables Django Admin SSO login when is True. |
| `ADMIN_OAUTH2_BASE_URL` | If `ADMIN_OAUTH2_ENABLED` is set | A base URL of OAuth2 provider. |
| `ADMIN_OAUTH2_TOKEN_FETCH_PATH` | If `ADMIN_OAUTH2_ENABLED` is set | OAuth fetch token path for Django Admin SSO login. |
| `ADMIN_OAUTH2_USER_PROFILE_PATH` | If `ADMIN_OAUTH2_ENABLED` is set | OAuth user profile path for Django Admin SSO login. |
| `ADMIN_OAUTH2_AUTH_PATH` | If `ADMIN_OAUTH2_ENABLED` is set | OAuth auth path for Django Admin SSO login. |
| `ADMIN_OAUTH2_CLIENT_ID` | If `ADMIN_OAUTH2_ENABLED` is set | OAuth client ID for Django Admin SSO login. |
| `ADMIN_OAUTH2_CLIENT_SECRET` | If `ADMIN_OAUTH2_ENABLED` is set | OAuth client secret for Django Admin SSO login. |

There are following views.

`login` - it will replace Django admin login view. It redirects user to Staff SSO.
`callback` - it is being called by Staff SSO with appropriate parameters to enable authentication. User details are being stored in the session.
`logout` - clears the Django session and redirects user to Staff SSO logout.

The next PR will have integration of these views into Django admin.

I am not sure about the location of the app. Probably it shouldn't be `django.oauth.admin`.

### Checklist


datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    user = Advisor.objects.filter(
+        staff_sso_user_id=sso_user_profile['user_id'],
+        **active_admin_user_filter,
+    ).first()
+
+    if user is None:
+        # If sso_user_id is not set, try to find user by email login
+        # TODO: this should be removed once SSO user IDs have been migrated
+        user = Advisor.objects.filter(
+            email=sso_user_profile['email'],
```
currycoder: Are there situations where the SSO email could differ from the adviser email for a particular staff member?

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,109 @@
...
+        return error_response(request, 'Forbidden.')
+
+    request.session['oauth.access_token'] = access_token_data
+    request.session['user_id'] = str(user.pk)
+
+    if 'oauth.next_url' in request.session:
+        next_url = request.session.pop('oauth.next_url')
+        return redirect(next_url)
+
+    return redirect('/admin/')
```
currycoder: Could be better to look this URL up with `reverse()`

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    response = requests.post(settings.ADMIN_OAUTH2_TOKEN_FETCH_URL, data=oauth_params)
+
+    access_token_data = response.json()
+    if 'error' in access_token_data:
+        raise AuthenticationFailed(access_token_data['error'])
+
+    if 'access_token' not in access_token_data:
+        raise AuthenticationFailed('No access token.')
+
+    access_token_data['received_on'] = int(time())
```
currycoder: What will this be used for?

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,109 @@
...
+    if code is None:
+        return error_response(request, 'Forbidden.')
+
+    try:
+        access_token_data = get_access_token(request, code)
+        sso_user_profile = get_sso_user_profile(access_token_data['access_token'])
+
+        user = get_adviser_by_sso_user_profile(sso_user_profile)
+    except AuthenticationFailed:
+        return error_response(request, 'Forbidden.')
```
currycoder: Looking at all of the parts of the code that can raise `AuthenticationFailed` for this block, I feel like it would be good to log the error too. At least at first.

datahub/oauth/admin/utils.py
elcct: Ideally no, but I am not sure if there is any mechanism that would update the DH user if email changes on the SSO etc. I got request to only match by SSO user ID, so if we get to import these IDs, I could remove this part of code.

datahub/oauth/admin/utils.py
elcct: To logout the user after certain time passes. It will be checked in the middleware in next PR.

reupen:
I agree that the PR is a bit on the large side. Do we need to do the SSO user ID changes at the same time? That seems to be adding complexity and I'm not sure it needs to be mixed in with this.

config/settings/common.py
```diff
@@ -225,6 +226,16 @@
 ALLOWED_ADMIN_IPS = env.list('ALLOWED_ADMIN_IPS', default=[])
 ALLOWED_ADMIN_IP_RANGES = env.list('ALLOWED_ADMIN_IP_RANGES', default=[])
 
+# OAuth2 settings to authenticate Django admin users
+
+ADMIN_OAUTH2_TOKEN_FETCH_URL = env('ADMIN_OAUTH2_TOKEN_FETCH_URL')
```
reupen: Could we only require these if admin SSO authentication is enabled? (I'm not sure we would want to always use it locally, also we might want to turn it on in each environment progressively.)

datahub/oauth/admin/views.py
elcct: Good idea! I added logging now.

datahub/oauth/admin/views.py
elcct: Good idea! This is now fixed.

config/settings/common.py
elcct: I have added `ADMIN_OAUTH2_ENABLED` flag so that if it is False then these environment variables are optional.

elcct:
Good suggestion! I have removed `staff_sso_user_id` bits, so the PR is slightly smaller now.

config/settings/common.py
```diff
@@ -225,6 +226,29 @@
...
+
+ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED', True)
+ADMIN_OAUTH2_TOKEN_FETCH_URL = env('ADMIN_OAUTH2_TOKEN_FETCH_URL', default=None)
+ADMIN_OAUTH2_USER_PROFILE_URL = env('ADMIN_OAUTH2_USER_PROFILE_URL', default=None)
+ADMIN_OAUTH2_AUTH_URL = env('ADMIN_OAUTH2_AUTH_URL', default=None)
+ADMIN_OAUTH2_CLIENT_ID = env('ADMIN_OAUTH2_CLIENT_ID', default=None)
+ADMIN_OAUTH2_CLIENT_SECRET = env('ADMIN_OAUTH2_CLIENT_SECRET', default=None)
+ADMIN_OAUTH2_LOGOUT_URL = env('ADMIN_OAUTH2_LOGOUT_URL', default=None)
+
+_admin_oauth2_params = [
```
reupen: Maybe we could handle this in a similar way to Celery i.e. something like:

```python
ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED', True)

if ADMIN_OAUTH2_ENABLED:
    ADMIN_OAUTH2_TOKEN_FETCH_URL = env('ADMIN_OAUTH2_TOKEN_FETCH_URL')
    ...
```

config/settings/common.py
```diff
@@ -225,6 +226,29 @@
 ALLOWED_ADMIN_IPS = env.list('ALLOWED_ADMIN_IPS', default=[])
 ALLOWED_ADMIN_IP_RANGES = env.list('ALLOWED_ADMIN_IP_RANGES', default=[])
 
+# OAuth2 settings to authenticate Django admin users
+
+ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED', True)
+ADMIN_OAUTH2_TOKEN_FETCH_URL = env('ADMIN_OAUTH2_TOKEN_FETCH_URL', default=None)
```
reupen: With these four URLs, would we be repeating the scheme and hostname in each one? Is there a particular need to have four URLs in the config instead of just the base URL?

config/urls.py
```diff
@@ -13,6 +14,7 @@
     path('', include('datahub.admin_report.urls')),
     path('', include('datahub.investment.project.report.urls')),
     path('ping.xml', ping, name='ping'),
+    path('oauth/callback', oauth_callback, name='oauth_callback'),
```
reupen: It would be good to somehow make it clear that this is specifically about the admin site (and not the normal SSO flow through the front end). Maybe we could nest the path under `admin/` (if it's possible)?

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    """Fetch OAuth2 access token from remote server."""
+    oauth_params = {
+        'code': code,
+        'grant_type': 'authorization_code',
+        'client_id': settings.ADMIN_OAUTH2_CLIENT_ID,
+        'client_secret': settings.ADMIN_OAUTH2_CLIENT_SECRET,
+        'redirect_uri': request.build_absolute_uri(reverse('oauth_callback')),
+    }
+
+    response = requests.post(settings.ADMIN_OAUTH2_TOKEN_FETCH_URL, data=oauth_params)
```
reupen: It would be good to use `APIClient` with `TokenAuth` (from `datahub.core.api_client`) for the API requests (also would be good to set a timeout).

datahub/oauth/admin/utils.py
reupen: Yes, it would be good to understand a bit more. Is it something that we could use Redis for (with an expiry timeout instead of tracking the time manually)?

datahub/oauth/admin/utils.py
reupen: To elaborate, I think it would be nice to add a minimal staff SSO client on top of that. I.e. something like:

```python
class StaffSSOAPIClient:

    def __init__(self):
        auth = TokenAuth(...)
        self.api_client = APIClient(...)

    def get_user_profile(self, access_token):
        request = self.api_client.request(...)
        return request.json()

    ...
```

config/settings/common.py
elcct: I think it should be okay to set it only with particular paths plus a host name, although it will be less flexible in case we want to replace the auth server for one that uses different configuration e.g having profile endpoint on completely different URL.

reupen: I think eliminating the redundancy would be good. If we want to make the paths configurable we could also do that – however, at least `ADMIN_OAUTH2_USER_PROFILE_URL` seems to be specific to Staff SSO so I'm not sure swapping Staff SSO out for a different OAuth provider would be as simple as changing the paths.

config/settings/common.py
elcct: I have modified it so it is taking the environment variables this way. Thank you!

config/settings/common.py
elcct: Fair point. I have modified it so it is using relative paths now. Thank you!

config/urls.py
elcct: I have moved it under `admin/`, but here it is only used for `reverse`, so the actual usage would be addressed in the following PR.

datahub/oauth/admin/utils.py
elcct: I have not seen your comment before I started working on it, so I just used the `APIClient` inline. I think the use case would be too small to create separate client. The token is only used for fetch the SSO profile.

datahub/oauth/admin/utils.py
elcct: I think Redis could be a good way to manage that. I changed it so that the state is being stored in cache and the timeout is added, so the state will disappear after a certain period of time.

datahub/oauth/admin/utils.py
elcct: This part has been removed for now.

config/settings/common.py
```diff
@@ -225,6 +226,21 @@
 ALLOWED_ADMIN_IPS = env.list('ALLOWED_ADMIN_IPS', default=[])
 ALLOWED_ADMIN_IP_RANGES = env.list('ALLOWED_ADMIN_IP_RANGES', default=[])
 
+# OAuth2 settings to authenticate Django admin users
+
+ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED', True)
```
reupen: I think we could make this required like `SSO_ENABLED`. That way, anyone working locally will be told to configure this first instead of the other settings.

elcct: I agree this is a good idea. Changed it now so it is required. Thank you!

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,124 @@
...
+ERROR_PAGE_TEMPLATE = 'admin/oauth/error.html'
+
+
+def login(request):
+    """
+    View replacing Django admin login page.
+
+    Instead of using standard Django authentication, we use OAuth2 protocol to authenticate user.
+    """
+    next_url = request.GET.get('next') if request.path == reverse('admin:login') \
```
reupen: Since `next` is coming in externally, we'll want to validate its host name (to avoid a dodgy redirect).

Django has an `is_safe_url()` function which I think we can use like `is_safe_url(url=next, allowed_hosts={request.get_host()}, require_https=request.is_secure()))`.

(For info, the Flask client does something similar: https://github.com/uktrade/flask-authbroker-client/blob/master/authbroker_client.py)

reupen: As we discussed offline, is the condition here and else logic correct? Should it just not set `next_url` in the session if we don't have one (as `callback()` handles that)?

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,95 @@
...
+    return user
+
+
+def _get_api_client(token=None):
+    api_client_args = [
+        settings.ADMIN_OAUTH2_BASE_URL,
+    ]
+    if token:
+        api_client_args.append(TokenAuth(token))
+    return APIClient(
```
reupen: Looks like this could be simplified:

```python
token = TokenAuth(token) if token else None

return APIClient(
    settings.ADMIN_OAUTH2_BASE_URL,
    auth=auth,
    default_timeout=settings.ADMIN_OAUTH2_REQUEST_TIMEOUT,
)
```

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,124 @@
...
+
+    request.session['oauth.state'] = state_id
+
+    return redirect(authorization_redirect_url)
+
+
+def logout(request):
+    """Perform Django log out and redirect to SSO log out."""
+    django_logout(request)
+    logout_url = f'{settings.ADMIN_OAUTH2_BASE_URL}{settings.ADMIN_OAUTH2_LOGOUT_PATH}'
```
reupen: Would suggest we do the join similar to this (it behaves the same whether the first part has a trailing slash or not) :

https://github.com/uktrade/data-hub-api/blob/97b475104c01fd0baf4c4146ced887bc5749449c/datahub/company/test/admin/test_update_from_dnb.py#L16

(Or even better, define a utility function like `safe_urljoin()` utility function that does joins like that.)

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,124 @@
...
+        return error_response(request, 'Forbidden.')
+
+    oauth_state = fetch_oauth2_state(state_id)
+    oauth_state['access_token_data'] = access_token_data
+
+    next_url = oauth_state.pop('next_url', None)
+
+    store_oauth2_state(state_id, oauth_state)
+
+    request.session['user_id'] = str(user.pk)
```
reupen: Is there a reason we can't use `login()` from `django.auth.contrib` here? Seems to do quite a few things we are missing.

https://docs.djangoproject.com/en/2.2/topics/auth/default/#how-to-log-a-user-in

config/urls.py
```diff
@@ -15,6 +16,9 @@
     path('ping.xml', ping, name='ping'),
     path('token/', TokenView.as_view(), name='token'),
     path('whoami/', who_am_i, name='who_am_i'),
+
+    # This endpoint is used for Django Admin OAuth2 authentication
+    path('admin/oauth/callback', oauth_callback, name='oauth_callback'),
```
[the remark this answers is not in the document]
elcct: I missed that - I agree it will be better to have `admin_oauth_callback`. Changed. Thank you!

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,124 @@
...
+    2. Get access token using provided code.
+    3. Request SSO profile using access token.
+    4. Find an active user matching SSO profile and having `is_staff` flag.
+    5. Store user details in session.
+    """
+    request_state_id = request.GET.get('state')
+    if request_state_id is None:
+        return error_response(request, 'No state.')
+
+    state_id = request.session['oauth.state']
```
reupen: As discussed offline – the state could be legitimately _missing_ in some cases (this happens when clicking links in Excel etc.).

In this case, we should send the user through another login loop with a new state (this should be transparent to the user as they'll already be authenticated in SSO).

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,92 @@
...
+from requests import HTTPError
+from rest_framework.exceptions import AuthenticationFailed
+
+from datahub.company.models import Advisor
+from datahub.core.api_client import APIClient, TokenAuth
+
+logger = logging.getLogger(__name__)
+
+
+def store_oauth2_state(state_id, state):
```
reupen: Just wondering if we need a separate timeout for the state, or we can shorten the session timeout instead (either using [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-SESSION_COOKIE_AGE) or [the `set_expiry()` session method](https://docs.djangoproject.com/en/2.2/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase.set_expiry)?

datahub/oauth/admin/views.py
elcct: This is an excellent idea. I have added suggested check and a test. Thank you!

elcct: I removed this and now `next_url` is only added if we have one. Thank you!

datahub/oauth/admin/utils.py
elcct: Nice one, this is indeed simpler. I have updated it. Thank you!

datahub/oauth/admin/views.py
elcct: This is a very good point. I should have used Django's `login()` - this is now corrected. Thank you!

datahub/oauth/admin/utils.py
elcct: I think I should use `expires_in` from the `access_token_data` to set how long this is valid, so that would be set by SSO.

datahub/oauth/admin/views.py
elcct: I have changed it to use `urljoin`, but would take a look later on for having `safe_urljoin` utility function. Thank you!

datahub/oauth/admin/views.py
elcct: Now the login process is restarted when the state is missing. Thank you!

datahub/oauth/admin/utils.py
elcct: I changed it now so the timeout is set to equal `expires_in` when access token is received, but between login and callback the state is using the value from settings - which could be smaller now.
I will have a look into session timeout. Thank you!

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+    return access_token_data
+
+
+def get_sso_user_profile(access_token):
+    """Fetch a user profile using given access token from remote server."""
+    try:
+        api_client = _get_api_client(access_token)
+        response = api_client.request('GET', settings.ADMIN_OAUTH2_USER_PROFILE_PATH)
+    except HTTPError as excinfo:
```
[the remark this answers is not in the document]
elcct: That's true :) Fixed it now, thank you!

datahub/oauth/admin/test/test_utils.py
```diff
@@ -0,0 +1,183 @@
...
+    get_access_token,
+    get_adviser_by_sso_user_profile,
+    get_sso_user_profile,
+    store_oauth2_state,
+)
+
+pytestmark = pytest.mark.django_db
+
+
+@patch('datahub.oauth.admin.utils.APIClient')
```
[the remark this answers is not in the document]
elcct: There is no reason. I just forgot that we use it. I have replaced this with `requests_mock` now. Thank you for pointing this out!

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,134 @@
...
+    except AuthenticationFailed as excinfo:
+        logger.warning(f'Django Admin OAuth2 authentication failed: {excinfo}')
+        return error_response(request, 'Forbidden.')
+
+    oauth_state = fetch_oauth2_state(state_id)
+    oauth_state['access_token_data'] = access_token_data
+
+    next_url = oauth_state.pop('next_url', None)
+
+    store_oauth2_state(state_id, oauth_state, timeout=access_token_data['expires_in'])
```
reupen: Did we reach a conclusion on if we could just [set the session expiry](https://docs.djangoproject.com/en/2.2/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase.set_expiry) (after calling `django_login()` I guess)?

elcct: My concern is that if something outside of oauth stuff modifies the session that would extend the expiry time?

reupen: OK, looks like we can't pass an absolute time without using `PickleSerializer` anyway (which we wouldn't want to do).

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,134 @@
...
+    View replacing Django admin login page.
+
+    Instead of using standard Django authentication, we use OAuth2 protocol to authenticate user.
+    """
+    state_id = str(uuid4())
+
+    oauth_url_params = {
+        'response_type': 'code',
+        'client_id': settings.ADMIN_OAUTH2_CLIENT_ID,
+        'redirect_uri': request.build_absolute_uri(reverse('admin_oauth_callback')),
```
reupen: Did we try passing the next URL as a query parameter in `redirect_uri`? It should be more reliable (due to e.g. the earlier problem with Microsoft Office mentioned earlier).

datahub/oauth/admin/views.py
```diff
@@ -0,0 +1,134 @@
...
+        'state': state_id,
+        'idp': 'cirrus',
+    }
+
+    authorization_url = f'{settings.ADMIN_OAUTH2_BASE_URL}{settings.ADMIN_OAUTH2_AUTH_PATH}'
```
reupen: Would be good keep the URL joining consistent – we went for `urljoin(f'{base_url}/', path)` elsewhere since it gives the same result whether `base_url` ends in a `/` or not.

datahub/oauth/admin/views.py
elcct: I have not tested it yet, but will do soon.

config/urls.py
```diff
@@ -15,6 +16,9 @@
     path('ping.xml', ping, name='ping'),
     path('token/', TokenView.as_view(), name='token'),
     path('whoami/', who_am_i, name='who_am_i'),
+
+    # This endpoint is used for Django Admin OAuth2 authentication
+    path('admin/oauth/callback', admin_oauth_callback, name='admin_oauth_callback'),
```
reupen: Should we only add this if `settings.ADMIN_OAUTH2_ENABLED` is truthy?

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,92 @@
...
+from requests import HTTPError
+from rest_framework.exceptions import AuthenticationFailed
+
+from datahub.company.models import Advisor
+from datahub.core.api_client import APIClient, TokenAuth
+
+logger = logging.getLogger(__name__)
+
+
+def store_oauth2_state(state_id, state, timeout=settings.ADMIN_OAUTH2_STATE_TIMEOUT):
```
reupen: `settings.ADMIN_OAUTH2_STATE_TIMEOUT` is not set when `ADMIN_OAUTH2_ENABLED` is falsey – we'll probably need to change that, as otherwise there's an error here when the app starts.

datahub/oauth/admin/views.py
reupen: I tested doing:

```
    request.session.set_expiry(60)
```

After doing that, I am signed out in 60 seconds, seemingly no matter what I do in the admin site (even resaving the session).

If using the session expiry means that we don't need `OAuthSessionMiddleware` (or can simplify it) then I'd be in favour of using the session expiry.

config/urls.py
elcct: I've done that in second PR, but will add it here

datahub/oauth/admin/views.py
reupen: We discussed offline – seems like the default expiry behaviour isn't what we want after all :disappointed:

datahub/oauth/admin/utils.py
```diff
@@ -0,0 +1,118 @@
...
+    except Advisor.DoesNotExist:
+        raise AuthenticationFailed('User not found.')
+
+    return user
+
+
+def get_safe_next_url_from_request(request):
+    """Get safe next URL from request."""
+    next_url = request.GET.get('next')
+    if not is_safe_url(
```
reupen: Probably would be more logical to handle the true case first here.

elcct: I agree it will be clearer to return the true case first. Fixed. Thank you!

datahub/oauth/admin/utils.py
elcct: I should have removed this as we always set `timeout` from `expires_in` passed by Staff SSO. It's now fixed. Thank you!

datahub/oauth/admin/utils.py
elcct: This has been removed now.

datahub/oauth/admin/views.py
elcct: We are now passing the next URL in redirect_uri.

PR #2314, opened 2019-11-19
Handle DNB search API connection errors gracefully

currycoder:
### Description of change

This PR ensures that we handle DNB search connection errors gracefully from several callers: the DNB company create API endpoint, the "update from DNB" admin command and the `sync_company_with_dnb` celery task.

This change was requested in a PR here: https://github.com/uktrade/data-hub-api/pull/2272#discussion_r344178926 - we deferred resolving the issue in to a separate PR; this PR!

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -6,6 +6,7 @@ from datahub.company.models import Company
 from datahub.dnb_api.utils import (
     DNBServiceConnectionError,
     DNBServiceError,
+    DNBServiceTimeoutError,
     get_company,
     update_company_from_dnb,
 )
@@ -22,7 +23,7 @@ def _sync_company_with_dnb(company_id, fields_to_update, task):
         if is_server_error(exc.status_code):
             raise task.retry(exc=exc, countdown=60)
         raise
-    except DNBServiceConnectionError as exc:
+    except (DNBServiceConnectionError, DNBServiceTimeoutError) as exc:
         raise task.retry(exc=exc, countdown=60)
 
     update_company_from_dnb(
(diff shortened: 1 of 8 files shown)


datahub/dnb_api/utils.py
```diff
@@ -75,7 +82,12 @@ def get_company(duns_number):
     found or if the `duns_number` for the company is not the same as the one
     we searched for.
     """
-    dnb_response = search_dnb({'duns_number': duns_number})
+    try:
+        dnb_response = search_dnb({'duns_number': duns_number})
+    except ConnectionError:
```
reupen: Just looking though the requests exception hierarchy – we might want `Timeout` as well.

datahub/dnb_api/tasks.py
```diff
@@ -21,6 +22,8 @@ def _sync_company_with_dnb(company_id, fields_to_update, task):
         if is_server_error(exc.status_code):
             raise task.retry(exc=exc, countdown=60)
         raise
+    except DNBServiceConnectionError as exc:
```
reupen: I think we should also consider that we get 404s from Cloud Foundry when there are routing problems (route unmapped, app still starting etc.) but these are only transient as well.

currycoder: Just to talk this through...  In this case I believe that the block will trigger here: https://github.com/uktrade/data-hub-api/pull/2314/files/fbaaefeff61e9ea37a13dc4b6f9373cbff40097b#diff-f3cf320eaf9fd1b399380f387051a25aR93

The exception class/error is probably a little misleading (it's not strictly a service error) but I feel like it's OK to flag this as a logged exception even when the vast majority of 404s might be from a transient routing problem.  This will also flag up the possible situation where the endpoint URL has changed on dnb-service, etc.

My feeling is that the application is not the place to solve (or suppress errors for) an ops-specific oddity.  Are there other considerations that I'm missing here?

currycoder:
Re-requesting reviews as the code to also handle `Timeout` exceptions was not trivial.

PR #2321, opened 2019-11-21
Company export countries model

sekharpanja:
### Description of change
This is first of multiple PRs in an attempt to migrate company export countries that are currently stored within Company model as `export_to_countries` and `future_interest_countries`.

This adds new `CompanyExportCountry` model, to hold multiple countries and its export status for each company. Status being one of `not_interested`, `future_interest` or `currently_exporting`.

This model is unused and just a placeholder for now. Existing fields are still intact and still are primary. They will be eventually phased out in PRs that will follow.

### Checklist


datahub/company/migrations/0100_auto_20191121_1608.py
```diff
@@ -0,0 +1,91 @@
+# Generated by Django 2.2.7 on 2019-11-21 16:08
```
elcct: Would be good to rename this file so that it says what's inside e.g. `0100_add_companyexportcountry_model.py`.

datahub/company/models/company.py
```diff
@@ -466,3 +466,58 @@ class Meta:
...
+        - 'currently exporting to'
+        - 'future interest'
+        - 'not interested'
+
+    This will eventually replace company fields:
+        - export_to_countries
+        - future_interest_countries
+    """
+
+    EXPORT_INTEREST_CATEGORIES = Choices(
```
elcct: The field is called `status` but choices are referred to as categories, I think better name will be `EXPORT_INTEREST_STATUSES`, so that it hints what field uses it.

datahub/company/admin/company_export_country.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib import admin
+
+from datahub.company.models import CompanyExportCountry
+from datahub.core.admin import BaseModelAdminMixin
+
+
+class CompanyExportCountryAdmin(BaseModelAdminMixin, admin.ModelAdmin):
+    """CompanyExportCountry Admin"""
```
elcct: This is minor, but would be great to have a full stop at the end of each comment, so we know it is complete and not cut in the middle by accident.

datahub/company/test/factories.py
```diff
@@ -174,3 +184,17 @@ class ArchivedContactFactory(ContactFactory):
...
     archived_by = factory.LazyFunction(lambda: random_obj_for_model(Advisor))
     archived_reason = factory.Faker('sentence')
+
+
+class CompanyExportCountryFactory(factory.django.DjangoModelFactory):
+    """Factory for Company export country"""
+
+    id = factory.LazyFunction(uuid.uuid4)
+    company = factory.SubFactory(CompanyFactory)
+    country = factory.Iterator(Country.objects.all())
```
elcct: We probably don't need to create all countries - perhaps this could be replaced with something that picks just one random e.g. `factory.LazyFunction(lambda: random_obj_for_model(Country))`? Otherwise our tests may get unnecessarily slow.

datahub/company/test/factories.py
```diff
@@ -174,3 +184,17 @@ class ArchivedContactFactory(ContactFactory):
...
+    status = CompanyExportCountry.EXPORT_INTEREST_CATEGORIES.currently_exporting
+    created_on = now()
```
reupen: Should be able to remove this as `created_on` uses `auto_add_now` and so will be set automatically (it can't normally be overridden except by using `freeze_time()`).

datahub/company/models/company.py
```diff
@@ -466,3 +466,58 @@ class Meta:
...
+        ('not_interested', 'Not interested'),
+        ('currently_exporting', 'Currently exporting to'),
+        ('future_interest', 'Future country of interest'),
+    )
+
+    id = models.UUIDField(
```
currycoder: Out of interest, is there any plan to add fields to this model in the future?  I remember in the first stab at this, there was an extra dimension of information sources - e.g. whether a user had entered this data or whether it was from an automatic mechanism.

datahub/company/test/test_models.py
```diff
@@ -232,3 +235,32 @@ def test_str(self, first_name, last_name, company_factory, expected_output):
...
+        status = f'{export_country.company} {export_country.country} {export_country.status}'
+        assert str(export_country) == status
+
+    def test_unique_constraint(self):
+        """
+        Test unique constraint
+        a company and country combination can't be added more than once
+        """
+        company_1 = CompanyFactory()
+        canada = Country.objects.get(iso_alpha2_code='CA')
```
reupen: Small thing but would suggest

```suggestion
        country = random_obj_for_model(Country)
```

as the particular country isn't important.

datahub/company/migrations/0100_auto_20191121_1608.py
reupen: For info, this can be done when generating the migration by passing `--name <name>` to `./manage.py makemigrations`.

datahub/company/admin/company_export_country.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib import admin
+
+from datahub.company.models import CompanyExportCountry
+from datahub.core.admin import BaseModelAdminMixin
+
+
+class CompanyExportCountryAdmin(BaseModelAdminMixin, admin.ModelAdmin):
```
reupen: I'm not convinced that adding this to admin on its own will be too useful, since it will just be a very long list of company and country combinations. Could go for an inline view on the company form, similar to interaction participants:

https://github.com/uktrade/data-hub-api/blob/e292a29dcb1a0172afbe3f6169344cd5b2e6a531/datahub/interaction/admin.py#L32-L58

(Could deal with the admin site in a separate PR if that's easier.)

sekharpanja: Agreed Reupen, I was contemplating with that too. But wasn't sure DH standard.

datahub/company/models/company.py
sekharpanja: We may not ingest the data from external sources. If we end up ingesting, we might have to add that dimension to it. But I am trying to keep things as simple as possible.

PR #2322, opened 2019-11-22
Add Staff SSO OAuth admin views integration

elcct:
### Description of change

This is branched off #2312 and contains integration of OAuth2 views into Django Admin.

Django admin native login and logout views are replaced with the views performing OAuth2 authentication.

New middleware has been added that checks if the requested path is a protected admin path and checks if OAuth2 state exists. If OAuth2 state doesn't exist, user is logged out and redirected to login page.

OAuth2 state is created during login process and stored in cache for a certain period of time (using cache timeout). OAuth2 state would only exist if user has successfully logged in.

### Checklist

### DIFF
--- a/changelog/django-admin-staff-sso-integration.feature.md
+++ b/changelog/django-admin-staff-sso-integration.feature.md
@@ -1,0 +1,1 @@
+It is now possible to configure the application to use Staff SSO to log into the admin site.
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -40,10 +40,22 @@ ALLOWED_HOSTS = ['*']
 USE_TZ = True
 TIME_ZONE = 'Etc/UTC'
 
+ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED')
+
+# If Django Admin OAuth2 authentication is enabled, we swap stock Django admin
+# with our own. We can only have either stock Django admin app or our OAuth2 admin app
+# enabled at a time.
+if ADMIN_OAUTH2_ENABLED:
+    _ADMIN_DJANGO_APP = []
+    _ADMIN_OAUTH2_APP = ['datahub.oauth.admin.apps.OAuthAdminConfig']
+else:
+    _ADMIN_DJANGO_APP = ['django.contrib.admin']
+    _ADMIN_OAUTH2_APP = []
+
 # Application definition
 
 DJANGO_APPS = [
-    'django.contrib.admin',
+    *_ADMIN_DJANGO_APP,
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
@@ -78,7 +90,7 @@ LOCAL_APPS = [
     'datahub.investment.investor_profile',
     'datahub.metadata',
     'datahub.oauth',
-    'datahub.oauth.admin.apps.OAuthAdminConfig',
+    *_ADMIN_OAUTH2_APP,
     'datahub.admin_report',
     'datahub.search.apps.SearchConfig',
     'datahub.user',
@@ -221,9 +233,9 @@ AUTHENTICATION_BACKENDS = [
 
 # OAuth2 settings to authenticate Django admin users
 
-ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED')
 if ADMIN_OAUTH2_ENABLED:
     ADMIN_OAUTH2_REQUEST_TIMEOUT = 15
+    ADMIN_OAUTH2_TOKEN_BYTE_LENGTH = 64
     ADMIN_OAUTH2_BASE_URL = env('ADMIN_OAUTH2_BASE_URL')
     ADMIN_OAUTH2_TOKEN_FETCH_PATH = env('ADMIN_OAUTH2_TOKEN_FETCH_PATH')
     ADMIN_OAUTH2_USER_PROFILE_PATH = env('ADMIN_OAUTH2_USER_PROFILE_PATH')
@@ -231,6 +243,13 @@ if ADMIN_OAUTH2_ENABLED:
     ADMIN_OAUTH2_CLIENT_ID = env('ADMIN_OAUTH2_CLIENT_ID')
     ADMIN_OAUTH2_CLIENT_SECRET = env('ADMIN_OAUTH2_CLIENT_SECRET')
     ADMIN_OAUTH2_LOGOUT_PATH = env('ADMIN_OAUTH2_LOGOUT_PATH')
+
+    authentication_middleware_label = 'django.contrib.auth.middleware.AuthenticationMiddleware'
+    authentication_middleware_index = MIDDLEWARE.index(authentication_middleware_label)
+    MIDDLEWARE.insert(
+        authentication_middleware_index + 1,
+        'datahub.oauth.admin.middleware.OAuthSessionMiddleware',
+    )
 
 # django-oauth-toolkit settings
 
--- a/datahub/oauth/admin/admin.py
+++ b/datahub/oauth/admin/admin.py
@@ -1,0 +1,18 @@
+from django.contrib import admin
+from django.views.decorators.cache import never_cache
+
+from datahub.oauth.admin.views import login, logout
+
+
+class OAuth2AdminSite(admin.AdminSite):
+    """Replace login and logout views with corresponding OAuth2 views."""
+
+    @never_cache
+    def login(self, request, extra_context=None):
+        """Replace login view."""
+        return login(request)
+
+    @never_cache
+    def logout(self, request, extra_context=None):
+        """Replace logout view."""
+        return logout(request)
--- a/datahub/oauth/admin/middleware.py
+++ b/datahub/oauth/admin/middleware.py
@@ -1,0 +1,31 @@
+from time import time
+
+from django.contrib.auth import logout as django_logout
+from django.shortcuts import redirect
+
+from datahub.core.utils import reverse_with_query_string
+
+
+class OAuthSessionMiddleware:
+    """
+    OAuthSessionMiddleware checks if user has been logged in via OAuth.
+    """
+
+    def __init__(self, get_response):
+        """Get response."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """
+        Check if OAuth2 expiration time is present in the session.
+
+        If OAuth2 expires on has passed it means we have to logout the user.
+        """
+        if request.user.is_authenticated:
+            oauth_expires_on = request.session.get('oauth.expires_on')
+            if oauth_expires_on is not None and oauth_expires_on < time():
+                django_logout(request)
+
+                return redirect(reverse_with_query_string('admin:login', {'next': request.path}))
+
+        return self.get_response(request)
--- a/datahub/oauth/admin/test/test_middleware.py
+++ b/datahub/oauth/admin/test/test_middleware.py
@@ -1,0 +1,65 @@
+from datetime import datetime
+
+import pytest
+from django.contrib.auth import login as django_login
+from django.urls import reverse
+from django.utils.timezone import utc
+from freezegun import freeze_time
+from rest_framework import status
+
+from datahub.company.test.factories import AdviserFactory
+from datahub.core.utils import reverse_with_query_string
+from datahub.oauth.admin.middleware import OAuthSessionMiddleware
+from datahub.oauth.admin.test.utils import get_request_with_session
+
+pytestmark = pytest.mark.django_db
+
+
+FROZEN_TIME = datetime(2018, 6, 1, 2, tzinfo=utc)
+
+
+@freeze_time(FROZEN_TIME)
+def test_user_is_logged_out_from_admin_when_oauth2_session_has_expired():
+    """Tests that user is logged out if OAuth2 session is expired."""
+    adviser = AdviserFactory()
+
+    admin_url = reverse('admin:index')
+    request = get_request_with_session(admin_url)
+    request.session['oauth.expires_on'] = int(FROZEN_TIME.timestamp()) - 1
+    django_login(request, adviser)
+
+    assert request.user.is_authenticated
+
+    oauth_session_middleware = OAuthSessionMiddleware(_get_response)
+    response = oauth_session_middleware(request)
+
+    assert not request.user.is_authenticated
+
+    assert response.status_code == status.HTTP_302_FOUND
+    assert response.url == reverse_with_query_string('admin:login', {'next': request.path})
+    assert 'oauth.state' not in request.session
+
+
+@freeze_time(FROZEN_TIME)
+def test_user_is_not_logged_out_when_oauth2_session_is_not_expired():
+    """
+    Tests that user is not logged out if OAuth2 session is not expired.
+    """
+    adviser = AdviserFactory()
+
+    admin_url = reverse('admin:index')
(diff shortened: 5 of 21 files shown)


datahub/oauth/admin/admin.py
```diff
@@ -0,0 +1,9 @@
+from django.conf import settings
+from django.contrib import admin
+
+from datahub.oauth.admin.views import login, logout
+
+if settings.ADMIN_OAUTH2_ENABLED:
+    admin.autodiscover()
+    admin.site.login = login
```
reupen: Are we monkey-patching the default site here? It's possible to override the default site, so we should do that instead:

https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#overriding-default-admin-site

datahub/oauth/admin/middleware.py
```diff
@@ -0,0 +1,38 @@
...
+        """Get response."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """
+        Check if OAuth2 state is present in the session.
+
+        If OAuth2 state is not available it means we have to logout the user.
+        """
+        required_path = request.path.startswith(self.ADMIN_PATH_FRAGMENT) \
```
reupen: I'm not sure able restricting this to `/admin` since we have other views protected by the same authentication (e.g. the API docs and potentially the DRF browseable API). What would happen if we removed that check?

datahub/oauth/admin/middleware.py
```diff
@@ -0,0 +1,38 @@
...
+            and request.path != reverse('admin:login')
+
+        state_exists = hasattr(request, 'session') \
+            and 'oauth.state' in request.session \
+            and fetch_oauth2_state(request.session['oauth.state'])
+
+        if not required_path or (required_path and state_exists):
+            return self.get_response(request)
+
+        django_logout(request)
```
reupen: The Django `logout()` function uses `request.user` but at the moment `OAuthSessionMiddleware` is before `AuthenticationMiddleware`. Should it be after it? Also, should we be checking `request.user.is_authenticated` as well?

datahub/oauth/admin/middleware.py
```diff
@@ -0,0 +1,38 @@
...
+
+        If OAuth2 state is not available it means we have to logout the user.
+        """
+        required_path = request.path.startswith(self.ADMIN_PATH_FRAGMENT) \
+            and request.path != reverse('admin_oauth_callback') \
+            and request.path != reverse('admin:login')
+
+        state_exists = hasattr(request, 'session') \
+            and 'oauth.state' in request.session \
+            and fetch_oauth2_state(request.session['oauth.state'])
```
reupen: It seems a bit indirect to do this via the state parameter (which is to protect against CSRF attacks). Maybe it would be simpler to just put the expiry time in the session? (Also bearing in mind that we may at some point want to add the periodic introspection that happens with the normal SSO flow too.)

datahub/oauth/admin/test/my_sso/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path
```
reupen: Maybe we can find a name clearer than `my_sso` for this – maybe `mock_sso_server`, `fake_sso_server` or similar?

datahub/omis/order/test/test_admin.py
```diff
@@ -26,10 +27,11 @@ def test_403_if_not_logged_in(self):
         url = reverse('admin:order_order_cancel', args=(OrderFactory().pk,))
 
         client = Client()
-        response = client.post(url, data={}, follow=True)
+        with my_test_sso_server(None):
```
reupen: I'm not sure that the use of SSO should be leaking into other tests like this – maybe it would be simpler if we removed `follow=True` and just checked `response.status_code` and `response['Location']`?

reupen:
Sorry @elcct , I missed in the last PR that we were generating the value of the `state` parameter using `uuid4()`. I think we should use `token_urlsafe()` from `secrets` for that (since it was designed for things like that and should be more secure).

datahub/oauth/admin/admin.py
elcct: I changed it, so it is now subclassing the `AdminSite`. Thank you!

datahub/oauth/admin/middleware.py
elcct: I think I was overthinking this. This is now removed.

datahub/oauth/admin/middleware.py
elcct: I changed the middleware and its order, so now it is checking whether user is authenticated.  Thank you!

datahub/oauth/admin/middleware.py
elcct: I thought we agreed to do it this way, maybe I got confused. I have now added `oauth.expires_on` value in session and if that is present and the current time is past that, user gets logged out. That will enable us to refresh the token if there will be a need. I have not added periodic introspection because we don't seem to do it and thought it may not be necessary.

datahub/oauth/admin/test/my_sso/urls.py
elcct: I wasn't sure if just checking that request gets redirected to login page is enough, but after giving it second look I think it is okay, so I removed all this `my_sso` stuff. Thank you!

datahub/omis/order/test/test_admin.py
elcct: I changed the tests following your suggestion. Thank you!

config/settings/common.py
```diff
@@ -38,10 +38,21 @@
 USE_TZ = True
 TIME_ZONE = 'Etc/UTC'
 
+ADMIN_OAUTH2_ENABLED = env.bool('ADMIN_OAUTH2_ENABLED')
+
+# If Django Admin OAuth2 authentication is enabled, we swap stock Django admin
+# with our own
```
currycoder: Might be good to extend the comment with something to express that the ordering of apps matters.  I had a little difficulty understanding why there needed to be two mutually exclusive variables here.

config/settings/common.py
```diff
@@ -38,10 +38,21 @@
...
+    ADMIN_OAUTH2_APP = ['datahub.oauth.admin.apps.OAuthAdminConfig']
+else:
+    ADMIN_DJANGO_APP = ['django.contrib.admin']
+    ADMIN_OAUTH2_APP = []
+
 # Application definition
 
 DJANGO_APPS = [
-    'django.contrib.admin',
+    *ADMIN_DJANGO_APP,
```
reupen: Maybe we could inline the definitions of `ADMIN_DJANGO_APP` and `ADMIN_OAUTH2_APP` to keep the app lists in one place (i.e. `*(['django.contrib.admin'] if ADMIN_OAUTH2_ENABLED else [])`).

Or, at least prefix `ADMIN_DJANGO_APP` with an `_` (as it looks like a setting itself right now).

datahub/core/test_utils.py
```diff
@@ -150,6 +150,7 @@ def get_admin_user(password=None):
     return admin_user
 
 
+@pytest.mark.usefixtures('local_memory_cache')
 class AdminTestMixin:
     """All the tests using the DB and accessing admin endpoints should use this class."""
```
reupen: Just below this it uses the `pytestmark` class attribute to set `pytest.mark.django_db` – we should probably set both marks in the same way (if there isn't some reason not to).

config/settings/common.py
reupen: As an alternative, maybe it would be simpler to check `settings.ADMIN_OAUTH2_ENABLED` in `OAuth2AdminSite` instead?

datahub/oauth/admin/middleware.py
```diff
@@ -0,0 +1,31 @@
...
+        """Get response."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """
+        Check if OAuth2 expiration time is present in the session.
+
+        If OAuth2 expires on has passed it means we have to logout the user.
+        """
+        if hasattr(request, 'user') and request.user.is_authenticated:
```
reupen: It looks like `AuthenticationMiddleware` sets `request.user` unconditionally to me – I think we should avoid `hasattr(request, 'user')` since that would silently skip the check if the middleware are reordered.

datahub/oauth/admin/middleware.py
```diff
@@ -0,0 +1,31 @@
...
+            oauth_expires_on = request.session.get('oauth.expires_on')
+            if oauth_expires_on and oauth_expires_on < time():
+                django_logout(request)
+
+                login_url = reverse('admin:login')
+                return redirect(f'{login_url}?next={request.path}')
```
reupen: Does this need `urlencode()`? Could we use `reverse_with_query_string()`?

datahub/user/test/test_views.py
```diff
@@ -48,9 +47,9 @@ def test_who_am_i_authenticated(self):
             response_data['permissions'].sort()
 
         expected_permissions = [
-            'admin.add_cats',
-            'admin.view_ipsum',
-            'admin.view_lorem',
+            f'{content_type.app_label}.add_cats',
```
reupen: Was this just flakey (as opposed to the app label changing)?

datahub/oauth/admin/test/test_middleware.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+from django.contrib.auth import login as django_login
+from django.urls import reverse
+from freezegun import freeze_time
+from rest_framework import status
+
+from datahub.company.test.factories import AdviserFactory
+from datahub.oauth.admin.middleware import OAuthSessionMiddleware
+from datahub.oauth.admin.test.utils import (
```
reupen: Will this fit on one line?

datahub/oauth/admin/test/test_middleware.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@freeze_time('2018-01-01 00:00:00')
+def test_user_is_logged_out_from_admin_when_oauth2_session_has_expired():
+    """Tests that user is logged out if OAuth2 session is expired."""
+    adviser = AdviserFactory()
+
+    admin_url = reverse('admin:index')
+    request = get_request_with_session(admin_url)
+    request.session['oauth.expires_on'] = 1514764799
```
reupen: Maybe we could do:

```python
FROZEN_TIME = datetime(...)

@freeze_time(FROZEN_TIME)
```

above and then something like `expires_on = int(FROZEN_TIME.timestamp()) + 1` in the test (to make it clearer what the significance of `1514764799` is).

changelog/django-admin-staff-sso-integration.feature.md
```diff
@@ -0,0 +1 @@
+The login mechanism for Django admin has changed and now it is using Staff SSO.
```
reupen: Would be good to make this clearer since that's only true if it's enabled. So maybe something like 'It's now possible to configure the application to use Staff SSO to log into the admin site.'

reupen:
Some additional points:

- It would be good to expand on the 'Forbidden.' message if you don't have access on our side (e.g. add 'Please contact Data Hub support if you think this is an error.'
- Can the state be dropped from the session now in the callback view?

datahub/user/test/test_views.py
elcct: It was dependent on the order of apps in the settings.

config/settings/common.py
elcct: I think it will be better to leave this check in the config, so that the `OAuth2AdminSite` is not used when OAuth2 admin is disabled. I think leaving it in separate block would aid visibility that stock admin or oauth2 admin are mutually exclusive and I expanded the comment a little. I agree that we should add `_` prefix to those variables, so I did. Thank you!

datahub/core/test_utils.py
elcct: Since we are no longer using the cached state, I just removed it. Thank you!

datahub/oauth/admin/middleware.py
elcct: That's a very good point. I have removed this bit now. Thank you!

datahub/oauth/admin/middleware.py
elcct: Oh I forgot about this handy function. I replaced it now. Thank you!

datahub/oauth/admin/test/test_middleware.py
elcct: It will. Fixed. Thank you!

datahub/oauth/admin/test/test_middleware.py
elcct: Great idea - this way it is much clearer. Fixed. Thank you!

changelog/django-admin-staff-sso-integration.feature.md
elcct: I have replaced the newsfragment. Thank you!

elcct:
- I have added additional message in the template. Thank you!

- I am not sure about dropping the state from the session, as we are checking if the request doing callback has the same state as the one initiated the authentication. We, however, no longer seem to be using the state kept in cache, so this could potentially be dropped, but if we decide to periodically refresh the token in the future it would be a handy thing to have. I understand this may not happen. I would still prefer to leave it, but I am happy to remove it if you feel strongly about it.

reupen:
@elcct I don't see why we would need the state again after callback – it's only for CSRF protection. (The normal SSO flow does do introspection currently every `RESOURCE_SERVER_TOKEN_CACHING_SECONDS` seconds without needing the state.)

At the moment it opens up the possibility of making extra requests to the callback view after the initial one, which is a bit of a liability, so I think it would be better to get rid of the state once it has been used.

elcct:
@reupen fair enough. I have removed it where it is not necessary. Thank you!

datahub/oauth/admin/middleware.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __call__(self, request):
+        """
+        Check if OAuth2 expiration time is present in the session.
+
+        If OAuth2 expires on has passed it means we have to logout the user.
+        """
+        if request.user.is_authenticated:
+            oauth_expires_on = request.session.get('oauth.expires_on')
+            if oauth_expires_on and oauth_expires_on < time():
```
reupen: I think `if oauth_expires_on is not None` would be better (doesn't seem like we have any reason to skip this if `oauth_expires_on` is somehow 0).

elcct: That's a good point. I have updated it. Thank you!

sample.env
```diff
@@ -43,3 +43,13 @@ COMPOSE_PROJECT_NAME=data-hub
 # be the same as MOCK_SSO_USERNAME in the frontend's .env file
 #SUPERUSER_USERNAME=some.person@digital.trade.gov.uk
 #SUPERUSER_PASSWORD=foobarbaz
+
+# OAuth2 settings for Django Admin access
+
+ADMIN_OAUTH2_ENABLED=False
```
reupen: `ADMIN_OAUTH2_ENABLED` is repeated from further up now...

elcct: Nice! I have removed one. Thank you.

templates/admin/oauth/error.html
```diff
@@ -15,4 +15,6 @@
```
reupen: Was there any particular reason that this template moved?

elcct: If I remember correctly Django wouldn't find it in the previous location.

PR #2330, opened 2019-11-25
Record export countries with the CompanyExportCountry model

yozzo:
### Description of change
This is part of a larger feature our team is building where we aim to handle export information separately from the Company model. 

Following the protocol described here: https://github.com/uktrade/data-hub-api/blob/develop/docs/How%20to%20change%20the%20type%20of%20a%20field.md, this is the step where we're adding the functionality to record the export information with both models - the new one **CompanyExportCountry** and the historical **Company** model from which we intend to retire the export information.

Use the initial company API to record export country data to the newly created CompanyExportCountry model

### Checklist


datahub/company/serializers.py
```diff
@@ -286,6 +287,65 @@ def __init__(self, *args, **kwargs):
...
+    def update(self, instance, validated_data):
+        """
+        Using writable nested representations to copy export country elements
+        from the `Company` model into the `CompanyExportCountry`.
+
+        N.B. Currently 'export_to_countries' and 'future_interest_countries'
+        are the only m2m fields, and due to the how the ManyRelatedManager operates
+        setattr() does not work hence instance.<field>.set(<value>) was used.
+        """
+        for key, value in validated_data.items():
```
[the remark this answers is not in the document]
yozzo: went for the second option

datahub/company/serializers.py
```diff
@@ -286,6 +287,65 @@ def __init__(self, *args, **kwargs):
...
+        countries_to_keep = []
+        for country in data:
+            obj, created = CompanyExportCountry.objects.update_or_create(
+                country=country,
+                company=instance,
+                status=status,
+            )
+
+            obj.save()
+            countries_to_keep.append(country)
```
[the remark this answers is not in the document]
yozzo: nice one

datahub/company/serializers.py
```diff
@@ -286,6 +287,65 @@ def __init__(self, *args, **kwargs):
...
+                export_countries=future_interest_countries,
+                status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.future_interest,
+            )
+
+        return company
+
+    @staticmethod
+    def _save_to_company_export_country_model(*, company, adviser_pk, export_countries, status):
+        for country in export_countries:
+            country_query_string = CompanyExportCountry.objects.filter(
```
reupen: Not sure about the logic here – shouldn't it be:

- try to get the `CompanyExportCountry` with the same country and company
- if successful and it has a different status then update the status
- if there is no existing CompanyExportCountry with the same country and company, create one

datahub/company/serializers.py
```diff
@@ -286,6 +287,65 @@ def __init__(self, *args, **kwargs):
...
+                country=country,
+                company=company,
+                status=status)
+
+            if country_query_string.exists():
+                country_query_string.update(
```
reupen: `QuerySet.update()` skips emitting signals, updating `modified_on` etc. so it tends to only be used in specific circumstances. I think you could probably use `CompanyExportCountry.objects.get_or_create()` here.

datahub/company/serializers.py
```diff
@@ -286,6 +287,65 @@ def __init__(self, *args, **kwargs):
...
+
+        return company
+
+    @staticmethod
+    def _save_to_company_export_country_model(*, company, adviser_pk, export_countries, status):
+        for country in export_countries:
+            country_query_string = CompanyExportCountry.objects.filter(
+                country=country,
+                company=company,
+                status=status)
```
reupen: The closing bracket should be dropped down on to the next line here (see https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md#indentation).

datahub/company/serializers.py
yozzo: luckily the new business rule is to not have a country with two statuses, but yes - removing the status from the initial filtering simplified the logic.

sekharpanja: I agree with Reupen. There is a chance that same country could appear in both 'currently_exporting_to' and 'future_interest' fields. In that we need to choose one. With the current logic, 'future_interest' takes precedence over 'currently_exporting_to'. Please speak to Gus and apply it here. Same business rule will have to applied in the migration script as well.

datahub/company/serializers.py
```diff
@@ -286,6 +287,88 @@ def __init__(self, *args, **kwargs):
             for field in self.Meta.dnb_read_only_fields:
                 self.fields[field].read_only = True
 
+    def update(self, instance, validated_data):
+        """
+        Using writable nested representations to copy export country elements
+        from the `Company` model into the `CompanyExportCountry`.
+        """
+        export_to_countries = validated_data.get('export_to_countries', None)
```
reupen: `None` is the default, so doesn't need to be passed explicitly:

```suggestion
        export_to_countries = validated_data.get('export_to_countries')
```

datahub/company/serializers.py
```diff
@@ -286,6 +287,88 @@ def __init__(self, *args, **kwargs):
...
+                status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting,
+            )
+
+        CompanyExportCountry.objects.filter(
+            company=instance,
+        ).exclude(
+            country__in=self._get_countries_to_exclude(
+                export_to_countries,
+                future_interest_countries,
+            ),
```
reupen: I don't think `self._get_countries_to_exclude()` is needed. Should be able to do something like:

```python
        all_countries = {
            *(export_to_countries or []),
            *(future_interest_countries or []),
        }

        CompanyExportCountry.objects.filter(
            company=instance,
        ).exclude(
            country__in=all_countries,
        ).delete()
```

datahub/company/serializers.py
```diff
@@ -286,6 +287,88 @@ def __init__(self, *args, **kwargs):
...
                 self.fields[field].read_only = True
 
+    def update(self, instance, validated_data):
+        """
+        Using writable nested representations to copy export country elements
+        from the `Company` model into the `CompanyExportCountry`.
+        """
+        export_to_countries = validated_data.get('export_to_countries', None)
+        future_interest_countries = validated_data.get('future_interest_countries', None)
+        adviser_pk = self._get_adviser_primary_key(validated_data.get('modified_by'))
```
reupen: Doesn't look like `self._get_adviser_primary_key()` is needed. Should be able to do:

```python
adviser = validated_data.get('modified_by')

...

obj.modified_by = adviser
```
etc.

datahub/company/serializers.py
```diff
@@ -286,6 +287,88 @@ def __init__(self, *args, **kwargs):
...
+        return countries
+
+    @staticmethod
+    def _get_adviser_primary_key(adviser=None):
+        if adviser is not None and hasattr(adviser, 'pk'):
+            return adviser.pk
+
+    @staticmethod
+    def _save_to_company_export_country_model(*, company, adviser_pk, export_countries, status):
+        for country in export_countries:
```
reupen: I don't think resaving the export country if the status hasn't changed is desirable, because then the `modified_on` and `modified_by` values will change (and all the company's export countries will have the same values for those fields).

Also, the logic can be simplified. So I would suggest something like:

```python
            export_country, created = CompanyExportCountry.objects.get_or_create(
                country=country,
                company=company,
                defaults={
                    'created_by_id': adviser_pk,
                    'modified_by_id': adviser_pk,
                    'status': status,
                }
            )

            if not created and export_country.status != status:
                export_country.status = status
                export_country.modified_by_id = adviser_pk
                export_country.save()
```

(keeping in mind the earlier comment about replacing `adviser_pk` with just `adviser`).

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1290,214 @@ def test_update_headquarter_type(
...
+        """
+        Utility fn aimed at hiding the list comprehension from the assertion logic.
+        """
+        benchmark_list = sorted(initial_list)
+        current_ids = sorted([_['id'] for _ in current_model_list])
+        new_ids = sorted([str(_.country.id) for _ in new_model_list])
+
+        if (benchmark_list == current_ids) and (benchmark_list == new_ids):
+            return True
+
```
reupen: It would be good to write these tests in the same style as other ones. Some specific points:

- other tests try not to reference countries using constants and instead pick random ones for tests
- other tests don't use `before` functions (pytest fixtures can be used if needed)
- assertions tend to compare two values or data structures rather than the result of a function (which won't give useful output if the assertion fails)

Could you have a go at simplifying them and making them more in line with the other tests? Let me know if you need any help.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,144 @@ def test_update_headquarter_type(
...
+        assert sorted(export_to_countries) == sorted(
+            [_['id'] for _ in response_data['export_to_countries']],
+        )
+        assert sorted(export_to_countries) == sorted([
+            str(_.country.id) for _ in company_export_country_currently_exporting_data
+        ])
+
+        if not duplicate_data_entry:
+            assert sorted(future_interest_countries) == sorted(
+                [_['id'] for _ in response_data['future_interest_countries']])
```
reupen: `_` implies the variable is unused (which isn't the case). So, would be good to use a proper name e.g. `country`.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,144 @@ def test_update_headquarter_type(
...
+            assert sorted(future_interest_countries) == sorted([
+                str(_.country.id) for _ in company_export_country_future_interest_data
+            ])
+        else:
+            duplicates = []
+            for i in export_to_countries:
+                for j in future_interest_countries:
+                    if i == j:
```
reupen: `i` and `j` makes it seem like they are indices which isn't the case. But it looks like this can be simplified using set logic instead e.g. `duplicates = set(export_to_countries) & set(future_interest_countries)`.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,144 @@ def test_update_headquarter_type(
             error = ['Subsidiaries have to be unlinked before changing headquarter type.']
             assert response_data['headquarter_type'] == error
 
+    @pytest.mark.parametrize(
+        'export_to_countries,future_interest_countries,duplicate_data_entry',
+        (
+            ([Country.cayman_islands.value.id], [Country.canada.value.id], False),
```
reupen: There are still country constants being used here so it would be good to discuss how we could avoid that.

I think there are also more cases to consider. For example, the cases might be:

1. adding exporting-to countries to a company without any
1. adding countries of interest to a company without any
1. changing exporting-to countries to completely different ones
1. changing countries of interest to completely different ones
1. adding an additional exporting-to country to a company that has existing ones
1. adding an additional country of interest to a company that has existing ones
1. adding overlapping exporting-to countries and countries of interest

(These don't need to all be done via parametrisation – it would probably be several test functions.)

config/settings/sample.env
```diff
@@ -36,3 +36,4 @@ ADMIN_OAUTH2_USER_PROFILE_PATH=/api/v1/user/me/
 ADMIN_OAUTH2_AUTH_PATH=/o/authorize/
 ADMIN_OAUTH2_CLIENT_ID=oauth2-client-id
 ADMIN_OAUTH2_CLIENT_SECRET=oauth2-client-secret
+
```
reupen: Was this change accidental?

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,335 @@ def test_update_headquarter_type(
...
+                [country['id'] for country in response_data['future_interest_countries']])
+            assert sorted(future_interest_countries) == sorted([
+                str(_.country.id) for _ in company_export_country_future_interest_data
+            ])
+        else:
+            duplicates = set(export_to_countries) & set(future_interest_countries)
+
+            new_ids = [str(_.country.id) for _ in company_export_country_future_interest_data]
+
+            assert len(new_ids) == len(future_interest_countries) - len(duplicates)
```
currycoder: Could sets be used a little more to make the assertions more readable here?  e.g. set up variables `expected_export_to_countries`, `expected_future_interest_countries` which are sets and then assert that the API responses/saved data is as expected.

I think this could help get rid of the `duplicate_data_entry` flag which IMO is a little difficult to understand.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,335 @@ def test_update_headquarter_type(
...
+
+    @staticmethod
+    def update_company_export_country_model(*, self, new_countries, field, company, model_status):
+        """
+        Standard action for updating the model with
```
currycoder: I think explaining the args/return signature would be helpful for readers to understand what this does.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,335 @@ def test_update_headquarter_type(
...
+        ]
+
+        actual_export_countries = CompanyExportCountry.objects.filter(
+            company=company,
+            status=model_status,
+        ).order_by(
+            'country__pk',
+        )
+
+        return response, actual_export_countries, actual_response_country_ids
```
currycoder: A dict might be a better/more readable container here for this multi-value response.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,335 @@ def test_update_headquarter_type(
...
+
+    @pytest.mark.parametrize(
+        'field,model_status',
+        (
+            ('export_to_countries',
```
currycoder: The usual convention in the API codebase for multi-line iterables is e.g.
```
(
    'export_to_countries',
    CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting,
)
```
As per https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md#indentation

datahub/company/test/test_company_views.py
currycoder: Could the variable names change to something like:
- `actual_export_countries` to `saved_export_countries`
- `actual_response_country_ids` to `response_country_ids`

I find the use of "actual" a little confusing here.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,335 @@ def test_update_headquarter_type(
...
+             CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting),
+            ('future_interest_countries',
+             CompanyExportCountry.EXPORT_INTEREST_STATUSES.future_interest),
+        ),
+    )
+    def test_adding_to_empty_company_export_to_country_model(
```
currycoder: Looking at all of the test cases that use the `self.update_company_export_country_model()` helper, I'm wondering why these test cases couldn't be rolled in to the `test_update_company_export_country_model()` method above.

I appreciate that they each test the update API based on different state for the existing company model, but I wonder if this could be captured as extra parameters on the `test_update_company_export_country_model()` method?

e.g. That test method could be built with the following signature:
```
    @pytest.mark.parametrize(
        'existing_export_to_countries,existing_future_interest_countries,export_to_countries,future_interest_countries',
        (
            ...
        ),
    )
    def test_update_company_export_country_model(
            self,
            existing_export_to_countries,
            existing_future_interest_countries,
            export_to_countries,
            future_interest_countries,
    ):
        """Test updating export countries to CompanyExportCountry model"""
```

This would then allow blocks of parameters which can set the existing state and set the variables for the API request.  The body of the test would need to do some logic to assert that the end values saved in the DB and in the API response are as expected.

These other test cases and the (fairly tricky) helper method could all be removed in this case, I believe.

What do you think?

datahub/company/test/test_company_views.py
currycoder: Might be superseded by the large comment below.

datahub/company/test/test_company_views.py
currycoder: Might be superseded by the large comment below.

currycoder: Might be superseded by the large comment below.

datahub/company/test/test_company_views.py
reupen: I agree with trying to get rid of `update_company_export_country_model()`. 

But just to note that we were trying to avoid using the constants (based on previous discussions a while ago – can explain if needed) and get rid of `test_update_company_export_country_model()`. So, I think things should look simpler after that.

I think, if we have a few methods in the end, grouping these tests in a dedicated class would also help.

yozzo: as of now `test_update_company_export_country_model()` doesn't exist anymore.

currycoder: Cool - this approach sounds good!

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,335 @@ def test_update_headquarter_type(
...
+        future_interest_countries = set(future_interest_countries)
+
+        # initialise the models in scope
+        company = CompanyFactory(
+            export_to_countries=export_ids_fixture,
+            future_interest_countries=future_ids_fixture,
+        )
+
+        for country in export_countries_fixture:
+            CompanyExportCountryFactory(
```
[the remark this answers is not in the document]
yozzo: this test was removed

datahub/company/serializers.py
```diff
@@ -286,6 +287,68 @@ def __init__(self, *args, **kwargs):
...
+        CompanyExportCountry.objects.filter(
+            company=instance,
+        ).exclude(
+            country__in=all_countries,
+        ).delete()
+
+        return company
+
+    @staticmethod
+    def _get_adviser_primary_key(adviser=None):
```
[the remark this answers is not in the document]
yozzo: Got left out from a previous refactoring, I removed it now.

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,283 @@ def test_update_headquarter_type(
...
+        ]
+        assert [
+            export_country.country for export_country in actual_export_to_countries
+        ] == new_export_to_countries
+
+        duplicates = set(new_export_to_countries) & set(new_future_interest_countries)
+
+        assert len(
+            list(actual_future_interest_countries),
+        ) == len(new_future_interest_countries) - len(duplicates)
```
reupen: It would be good to check the values here instead e.g. `assert set(actual_future_interest_countries) == new_future_interest_countries.keys() - new_export_to_countries.keys()`

datahub/company/test/test_company_views.py
```diff
@@ -1283,6 +1292,283 @@ def test_update_headquarter_type(
             error = ['Subsidiaries have to be unlinked before changing headquarter type.']
             assert response_data['headquarter_type'] == error
 
+    @staticmethod
+    def update_company_export_country_model(*, self, new_countries, field, company, model_status):
```
reupen: It might be nice to move these tests to a separate class to keep them grouped together.

changelog/company/company-export-country.view.md
```diff
@@ -0,0 +1,3 @@
+We are now recording `future_interest_countries`
```
reupen: The extension is currently `.view.md` but there is no `view` news fragment type – the valid types are listed in https://github.com/uktrade/data-hub-api/blob/develop/changelog/README.md. (I think this could go in `internal` or `feature`, I'd lean towards `internal` as there aren't any externally visible changes.)

PR #2338, opened 2019-11-28
Add celery task to apply company updates

currycoder:
### Description of change

This PR adds a `update_company` celery task which will be used in an automatic process for regularly updating DNB company details.  The producer task which pages through an API endpoint of updates will emit `update_company` tasks for each company to update.

This task is a thin wrapper around existing DNB utility functions that we are already using. 

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -48,10 +48,11 @@ def _sync_company_with_dnb(company_id, fields_to_update, task):
 )
 def sync_company_with_dnb(self, company_id, fields_to_update=None):
     """
-    Sync a company record with data sourced from DNB. `company_id` identifies the
-    company record to sync and `fields_to_update` defines an iterable of
-    company serializer fields that should be updated - if it is None, all fields
-    will be synced.
+    Sync a company record with data sourced from DNB. This task will interact with dnb-service to
+    get the latest data for the company.
+
+    `company_id` identifies the company record to sync and `fields_to_update` defines an iterable
+    of company serializer fields that should be updated - if it is None, all fields will be synced.
     """
     _sync_company_with_dnb(company_id, fields_to_update, self)
 
@@ -117,11 +118,13 @@ def get_company_updates(self, last_updated_after=None, fields_to_update=None):
     acks_late=True,
     priority=9,
 )
-def update_company(company_data, fields_to_update=None):
+def update_company_from_dnb_data(dnb_company_data, fields_to_update=None):
     """
-    Update the company from latest data from dnb-service.
+    Update the company with the latest data from dnb-service. This task should be called
+    when some other logic interacts with dnb-service to get the company data as the task itself
+    will not interact with dnb-service.
     """
-    dnb_company = format_dnb_company(company_data)
+    dnb_company = format_dnb_company(dnb_company_data)
     duns_number = dnb_company['duns_number']
 
     try:
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -11,6 +11,7 @@ from datahub.dnb_api.utils import (
     DNBServiceConnectionError,
     DNBServiceError,
     DNBServiceTimeoutError,
+    format_dnb_company,
     get_company,
     get_company_update_page,
     update_company_from_dnb,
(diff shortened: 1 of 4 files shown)


datahub/dnb_api/tasks.py
```diff
@@ -48,3 +49,21 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
+@shared_task(
+    acks_late=True,
+    priority=9,
+)
+def update_company(company_data, fields_to_update=None):
+    """
+    Update the company from latest data from dnb-service.
+    """
+    dnb_company = format_dnb_company(company_data)
+    dh_company = Company.objects.get(duns_number=dnb_company['duns_number'])
```
alixedi: What happens if a company with the given `duns_number` doesn't exist in Data Hub?

datahub/dnb_api/tasks.py
```diff
@@ -48,3 +49,21 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
+    update_company_from_dnb(
```
alixedi: What happens if there is a `ValidationError`?

datahub/dnb_api/tasks.py
currycoder: We just get a `DoesNotExist` exception and the task quits early.  I did consider wrapping and raising a custom exception, but I didn't really see much value as I don't believe that there's much extra context that can be added to that exception.  In writing this out though, it would probably be handy to know the `duns_number` - so I will add some handling/logging :+1:

datahub/dnb_api/tasks.py
currycoder: I'll take a similar approach here to ensure that the sentry error that we end up with has enough context for us to investigate, without saturating the error log with many different unique messages.

currycoder: So it turns out that the `update_company_from_dnb` utility already does a good job at logging this, so I think it's safe to just allow the ValidationError to bubble up and for our task to quit early.

datahub/dnb_api/tasks.py
```diff
@@ -48,3 +49,34 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
     will be synced.
     """
     _sync_company_with_dnb(company_id, fields_to_update, self)
+
+
+@shared_task(
+    acks_late=True,
+    priority=9,
+)
+def update_company(company_data, fields_to_update=None):
```
reupen: I assume `company_data` is D&B company data – maybe `dnb_company_data` will be more self explanatory?

```suggestion
def update_company(dnb_company_data, fields_to_update=None):
```

datahub/dnb_api/tasks.py
```diff
@@ -48,3 +49,34 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
+    """
+    Update the company from latest data from dnb-service.
+    """
+    dnb_company = format_dnb_company(company_data)
+    duns_number = dnb_company['duns_number']
+
+    try:
+        dh_company = Company.objects.get(duns_number=duns_number)
+    except Company.DoesNotExist:
+        logger.error(
```
reupen: Does this being an error mean such companies will filtered out before the task is called?

datahub/dnb_api/tasks.py
```diff
@@ -48,3 +49,34 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
     _sync_company_with_dnb(company_id, fields_to_update, self)
+
+
+@shared_task(
+    acks_late=True,
+    priority=9,
+)
+def update_company(company_data, fields_to_update=None):
+    """
+    Update the company from latest data from dnb-service.
```
reupen: It would be good to make the difference between this and `sync_company_with_dnb` a bit clearer (in the task names and docstrings).

datahub/dnb_api/tasks.py
currycoder: According to the contract we have agreed for the dnb-service, the endpoint will only return companies that are in Data Hub - so in this sense it would be an exceptional situation that we would want to know about.

datahub/dnb_api/tasks.py
currycoder: You're right - will do.  Thanks!

PR #2349, opened 2019-12-02
Add integration tests for get_company_updates task

currycoder:
### Description of change
Integration tests were added for the `datahub.dnb_api.tasks.get_company_updates` task. These were not added as part of the original development as the task and it's dependent task (`datahub.dnb_api.tasks.update_company_from_dnb_data`) were developed in parallel.

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -81,8 +81,8 @@ def _get_company_updates(task, last_updated_after, fields_to_update):
         # Spawn tasks that updates Data Hub companies
         for data in response.get('results', []):
             update_company_from_dnb_data.apply_async(
-                data,
-                fields_to_update=fields_to_update,
+                args=(data,),
+                kwargs={'fields_to_update': fields_to_update},
             )
 
         cursor = response.get('next')
--- a/datahub/dnb_api/test/test_tasks.py
+++ b/datahub/dnb_api/test/test_tasks.py
@@ -231,8 +231,7 @@ def test_sync_company_with_dnb_retries_errors(monkeypatch, error, expect_retry):
 
 class TestGetCompanyUpdates:
     """
-    Tests for the get_company_updates task and the
-    associated _get_company_updates function.
+    Tests for the get_company_updates task and the associated _get_company_updates function.
     """
 
     @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


datahub/dnb_api/test/test_tasks.py
```diff
@@ -414,6 +413,55 @@ def test_lock(self, monkeypatch, lock_acquired, call_count):
...
         assert mock_get_company_updates.call_count == call_count
 
+    @freeze_time('2019-01-02T2:00:00')
+    def test_updates_with_update_company_from_dnb_data(self, monkeypatch, dnb_response_uk):
+        """
+        Test full integration for the `get_company_updates` task with the
+        `update_company_from_dnb_data` task when all fields are updated.
+        """
+        company = CompanyFactory(duns_number='123456789')
+        update_response_body = dnb_response_uk
```
reupen: Is there a particular reason for reassigning `dnb_response_uk` to a different variable?

currycoder: Just to make the variable name a bit more explicit/meaningful for the way it is being used in this test i.e. we are adding a pagination attribute to make it a valid response from the company updates endpoint.

currycoder: Note that the OG fixture should probably be called `dnb_search_response_uk` to make things more explicit in this case.

reupen: I think aliasing something and then modifying the object via alias could seem a bit confusing. Maybe a copy would make things clearer e.g.:

```python
update_response_body = {
    **dnb_response_uk,
    'next': None,
}
```

Or, alternatively, a new fixture could be defined...

datahub/dnb_api/tasks.py
```diff
@@ -81,8 +81,8 @@ def _get_company_updates(task, last_updated_after, fields_to_update):
         # Spawn tasks that updates Data Hub companies
         for data in response.get('results', []):
             update_company_from_dnb_data.apply_async(
-                data,
-                fields_to_update=fields_to_update,
+                args=(data,),
```
reupen: Doesn't this mean that we logic in the task was fixed (i.e. the change is more than adding tests)? It would be good to update the news fragment if so.

currycoder: True!

PR #2351, opened 2019-12-02
Limit automatic dnb company updates

currycoder:
### Description of change

A setting `DNB_AUTOMATIC_UPDATE_LIMIT` was added which can be used to limit the number of companies updated by the `datahub.dnb_api.tasks.get_company_updates` task.

### Checklist


datahub/dnb_api/test/test_tasks.py
```diff
@@ -413,6 +414,57 @@ def test_lock(self, monkeypatch, lock_acquired, call_count):
 
         assert mock_get_company_updates.call_count == call_count
 
+    @pytest.mark.parametrize(
+        'data',
+        (
+            {
+                'next': None,
```
reupen: Would be good to have a test with multiple pages (to test that the limit is working correctly if it falls on a subsequent page).

currycoder: Great spot, turns out that this didn't work (pass by value for ints - whoops!).  I've now fixed it up.

datahub/dnb_api/tasks.py
```diff
@@ -57,11 +58,32 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
 
 
+class UpdateLimitReachedException(Exception):
+    """
+    Logic-flow exception which is raised when settings.DNB_AUTOMATIC_UPDATE_LIMIT
+    is exceeded.
+    """
+
+
+def _spawn_update_tasks(dnb_updates, fields_to_update, update_count, max_updates):
```
reupen: I think it would be a bit cleaner if the caller decreased `max_updates` for each page so this function doesn't need to maintain the state of `update_count` (it can just just return how many items were processed).

But will leave that with you to consider.

currycoder: That is much cleaner, agreed.  Thanks.

datahub/dnb_api/tasks.py
```diff
@@ -57,34 +58,46 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
-            response = get_company_update_page(last_updated_after, cursor)
+        response = _get_company_updates_from_api(last_updated_after, cursor, task)
+        dnb_company_updates = response.get('results', [])
 
-        except DNBServiceError as exc:
-            if is_server_error(exc.status_code):
-                raise task.retry(exc=exc, countdown=60)
-            raise
+        if updates_remaining and len(dnb_company_updates) > updates_remaining:
+            dnb_company_updates = dnb_company_updates[:updates_remaining]
```
reupen: I think you can drop the `if` statement here and just do `dnb_company_updates[:updates_remaining]` in all cases (should work for `None` too).

PR #2353, opened 2019-12-02
add interaction export country model

sekharpanja:
### Description of change
Adding new model `InteractionExportCountry`, recording `Interaction`'s exporting status to a `Country`.
Status is expressed as:
- 'currently exporting to'
- 'future interest'
- 'not interested'

Interaction API changes, to consume and expose export countries will follow this PR.

### Checklist


datahub/interaction/models.py
```diff
@@ -293,6 +295,8 @@ class Interaction(ArchivableModel, BaseModel):
         related_name='interactions',
     )
     policy_feedback_notes = models.TextField(blank=True, default='')
+    # export countries
+    were_countries_discussed = models.BooleanField(default=False)
```
elcct: Should this just say `were_export_countries_discussed`?

datahub/interaction/models.py
```diff
@@ -345,3 +349,52 @@ class Meta:
             'delete',
             'view_all',
         )
+
+
+@reversion.register_base_model()
+class InteractionExportCountry(BaseModel):
+    """
+    Record `Interaction`'s exporting status to a `Country`.
+    Status is expressed as:
```
elcct: Not sure if it is worth listing statuses in a comment - maybe point which constant keeps them? Otherwise we will have to maintain the comment as well if statuses change.

datahub/interaction/test/factories.py
```diff
@@ -181,3 +183,17 @@ class InteractionDITParticipantFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = 'interaction.InteractionDITParticipant'
+
+
+class InteractionExportCountryFactory(factory.django.DjangoModelFactory):
+    """Factory for Interaction export country"""
```
elcct: Would be good to end all comments with full stop (optional).

datahub/interaction/test/test_models.py
```diff
@@ -11,3 +16,34 @@ def test_interaction_get_absolute_url():
...
+        assert str(export_country) == status
+
+    @pytest.mark.django_db
+    def test_unique_constraint(self):
+        """
+        Test unique constraint
+        a interaction and country combination can't be added more than once
+        """
+        interaction_1 = CompanyInteractionFactory()
+        canada = Country.objects.get(iso_alpha2_code='CA')
```
elcct: Maybe we could pick random country instead (using `random_obj_for_model`)?

datahub/interaction/models.py
sekharpanja: Fair enough. Will change the field name.

datahub/cleanup/management/commands/delete_old_records.py
```diff
@@ -97,6 +97,7 @@ class Command(BaseCleanupCommand):
             # We want to delete the relations below along with any expired interactions
             excluded_relations=(
                 Interaction._meta.get_field('dit_participants'),
+                Interaction._meta.get_field('export_countries'),
```
mcbhenwood: Is it definite, defined business logic that an Interaction's export_countries should live on even if the Interaction is removed?

datahub/interaction/models.py
mcbhenwood: Actually, not sure that 'export' is wanted here because they may have been discussed like "We will never export to Belgium" and that doesn't feel like a discussion about _export_ countries (sorry @elcct )

reupen: I agree with @elcct as the context is export (as opposed to investment or something else).

datahub/cleanup/management/commands/delete_old_records.py
reupen: I think the name `excluded_relations` might have given the wrong impression – the comment above tries to explain it, but it means 'don't worry if these exist' (so they will be deleted if `on_delete=CASCADE`).

From memory the policy is to delete interactions ten years after they happened, so I think this should be fine.

sekharpanja: Apparently, entry in excluded_relations would mean they are removed irrespective.

datahub/interaction/models.py
sekharpanja: Left it the same in the end.

elcct: In this context I find the purpose of `export countries` comment confusing. Maybe you could add a little bit more information about this field or remove the comment?

sekharpanja: I am split between having `export` in the field or not. In the end left it alone. If you guys feel strongly, let me know.

datahub/interaction/models.py
```diff
@@ -294,6 +296,8 @@ class Interaction(ArchivableModel, BaseModel):
     )
     policy_feedback_notes = models.TextField(blank=True, default='')
 
+    were_countries_discussed = models.BooleanField(default=False)
```
reupen: Have a look at https://github.com/uktrade/data-hub-api/blob/develop/docs/How%20to%20add%20a%20non-nullable%20field.md to add a non-nullable field – we'll need to follow this process to add this field if we want a default or it to be non-nullable.

So, if we want it non-nullable, it should be first added as nullable without a default. The default can be added in a separate migration in this PR (but removing `null=True` has to be done separately once `replace_null_with_default` has been run).

(Another option would be to leave it as null for existing rows.)

sekharpanja: Thinking about it, leaving the field to allow null is indeed a good option.

reupen: :+1: 

(Will still need to split the migration I believe (first without `default` and then with it) – `sqlmigrate` can always be used to check what the migration is doing.)

PR #2354, opened 2019-12-04
Fix dnb company updates ingest pagination

currycoder:
### Description of change
A bug was fixed to ensure that DNB company updates can be ingested over multiple pages from dnb-service.  Previously, the cursor value was not being extracted from the URL for the next page correctly.

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -1,4 +1,6 @@
+import urllib.parse as urlparse
 from datetime import datetime, time, timedelta
+from urllib.parse import parse_qs
 
 from celery import shared_task
 from celery.utils.log import get_task_logger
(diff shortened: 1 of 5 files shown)


datahub/dnb_api/tasks.py
```diff
@@ -97,7 +109,8 @@ def _get_company_updates(task, last_updated_after, fields_to_update):
             if updates_remaining <= 0:
                 break
 
-        cursor = response.get('next')
+        next_page = response.get('next')
+        cursor = _get_cursor(next_page)
```
reupen: Are we able to request `next` as it is rather than try to parse the URL? Seems a bit brittle this way.

datahub/dnb_api/tasks.py
```diff
@@ -71,6 +73,16 @@ def _get_company_updates_from_api(last_updated_after, cursor, task):
         raise task.retry(exc=exc, countdown=60)
 
 
+def _get_cursor(url):
+    if not url:
+        return None
+    parsed = urlparse.urlparse(url)
+    try:
+        return parse_qs(parsed.query).get('cursor')[0]
+    except IndexError:
```
reupen: Not sure what case `IndexError` is handling here – when would it be raised? (`None[0]` is a `TypeError`.)

datahub/dnb_api/tasks.py
currycoder: Good point, I've binned this and am respecting `next` as a whole.

datahub/dnb_api/tasks.py
currycoder: This extraction code was removed.

PR #2355, opened 2019-12-04
Add dnb-company-updates feature flag

currycoder:
### Description of change
A feature flag was added `"dnb-company-updates"` which governs whether or not to run the logic within the `datahub.dnb_api.tasks.get_company_updates` celery task.
This affords us the ability to easily switch on/off DNB company updates as needed  during the rollout of this feature.

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -8,6 +8,7 @@ from django_pglocks import advisory_lock
 from rest_framework.status import is_server_error
 
 from datahub.company.models import Company
+from datahub.dnb_api.constants import FEATURE_FLAG_DNB_COMPANY_UPDATES
 from datahub.dnb_api.utils import (
     DNBServiceConnectionError,
     DNBServiceError,
@@ -17,6 +18,7 @@ from datahub.dnb_api.utils import (
     get_company_update_page,
     update_company_from_dnb,
 )
+from datahub.feature_flag.utils import is_feature_flag_active
 
 logger = get_task_logger(__name__)
 
@@ -117,6 +119,14 @@ def get_company_updates(self, last_updated_after=None, fields_to_update=None):
     task goes through the pages and spawns tasks that update the records in
     Data Hub.
     """
+    # TODO: remove this feature flag after a reasonable period after going live
+    # with unlimited company updates
+    if not is_feature_flag_active(FEATURE_FLAG_DNB_COMPANY_UPDATES):
+        logger.info(
+            f'Feature flag "{FEATURE_FLAG_DNB_COMPANY_UPDATES}" is not active, exiting.',
+        )
+        return
+
     with advisory_lock('get_company_updates', wait=False) as acquired:
 
         if not acquired:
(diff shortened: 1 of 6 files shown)


datahub/email_ingestion/tasks.py
```diff
@@ -15,8 +15,8 @@ def ingest_emails():
     Ingest and process new emails for all mailboxes in the application - i.e.
     those in the MAILBOXES django setting.
     """
-    # TODO: remove feature flag check once we are happy with meeting invite
-    # email processing
+    # NOTE: This is a long-lived feature flag which allows us to quickly switch off email
```
currycoder: I noticed that this TODO was outdated, so I've made it in to a NOTE instead.

datahub/dnb_api/tasks.py
```diff
@@ -105,6 +107,14 @@ def get_company_updates(self, last_updated_after=None, fields_to_update=None):
     task goes through the pages and spawns tasks that update the records in
     Data Hub.
     """
+    # TODO: remove this feature flag after a reasonable period after going live
+    # with unlimited company updates
+    if not is_feature_flag_active(FEATURE_FLAG_DNB_COMPANY_UPDATES):
```
elcct: Why feature flag isn't around the point where this task would be called? Wouldn't this potentially generate unnecessary calls and pollute logs?

datahub/dnb_api/test/test_tasks.py
```diff
@@ -293,7 +293,7 @@ class TestGetCompanyUpdates:
             ),
         ),
     )
-    def test_errors(self, monkeypatch, error, expect_retry):
+    def test_errors(self, monkeypatch, error, expect_retry, dnb_company_updates_feature_flag):
```
reupen: I think it would be good to use `@pytest.mark.usefixtures('dnb_company_updates_feature_flag')` if the fixture is unreferenced (to make it clear that that's deliberate).

(Could apply it to the entire class if the case without the feature flag is moved to a separate class.)

datahub/dnb_api/tasks.py
currycoder: The aspiration for this task is that it goes in to a celery beat config - so there's no other place to put the feature flag check in this case, as far as I know.

datahub/dnb_api/test/test_tasks.py
currycoder: Nice idea - didn't know about that.

datahub/dnb_api/tasks.py
elcct: We do use feature flags to enable / disable task e.g. https://github.com/uktrade/data-hub-api/blob/develop/config/settings/common.py#L406

currycoder: I believe because they are in the settings file, those are environment variable flags rather than DB-backed feature flags. I used the same approach for the email ingestion feature, with both an environment variable and a feature flag e.g.

https://github.com/uktrade/data-hub-api/blob/develop/config/settings/common.py#L418

and

https://github.com/uktrade/data-hub-api/blob/develop/datahub/email_ingestion/tasks.py#L20

elcct: Yes, we can toggle features either way. I understand the need is to toggle it without having to re-deploy the application?

currycoder: Yes that's it.  I agree that it feels odd to now have **both** mechanisms, but I think the environment variable approach will end up being long lived and the DB feature flag will be removed after some period of confidence building.  Does that sound ok?

PR #2368, opened 2019-12-09
DNB company updates logging

currycoder:
### Description of change

Info log messages and a sentry-based audit log were added to the DNB company updates tasks to help provide better visibility for task runs.

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -1,5 +1,6 @@
 from datetime import datetime, time, timedelta
 
+import sentry_sdk
 from celery import shared_task
 from celery.utils.log import get_task_logger
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


datahub/dnb_api/tasks.py
```diff
@@ -60,6 +61,39 @@ def sync_company_with_dnb(self, company_id, fields_to_update=None):
...
+    """
+    audit = {
+        'success_count': 0,
+        'failure_count': 0,
+        'updated_company_ids': [],
+        'producer_task_id': producer_task.request.id,
+        'start_time': start_time.isoformat(),
+        'end_time': now().isoformat(),
+    }
+    for result in update_results:
```
reupen: I might've missed it, but `get_company_updates()` is using asynchronous sub-tasks so it looks like there's no guarantee that the result is available at this point (the tests run everything synchronously...)

If so, perhaps a group and a chain would do the trick?

https://docs.celeryproject.org/en/latest/getting-started/next-steps.html#groups
https://docs.celeryproject.org/en/latest/userguide/tasks.html#task-synchronous-subtasks

currycoder: My bad, I made an assumption about the contract of `.successful()` - I had thought that it would block until success/failure could be ascertained.  Will look in to those - thanks!

currycoder: I've fixed this without having to change too much of the existing code: https://github.com/uktrade/data-hub-api/pull/2368/files#diff-c940877bcad229a174569b04a6b01b7cR146

FYI, it will get more complicated later though.  We have an upcoming card (before we raise the update limit to 1000) which states that we should run these updates in fixed sized batches.  e.g. to ensure that no more than 250 companies are updated a time, so we don't accidentally run in to queue saturation problems (from knock on search index tasks etc).  I'm going to defer refactoring this function until we take on that card though.

datahub/dnb_api/tasks.py
```diff
@@ -103,6 +142,11 @@ def _get_company_updates(task, last_updated_after, fields_to_update):
         if next_page is None:
             break
 
+    # Wait for all update tasks to finish...
+    GroupResult(results=update_results).join(propagate=False)
```
reupen: Maybe `ResultSet` would be more appropriate?

PR #2369, opened 2019-12-09
Add CSRF header to SwaggerUI requests

elcct:
### Description of change

This updates API docs template, so the CSRF token is included in the POST request. This enables us to try POST requests in the API browser. 
Currently the API would return an error about missing CSRF token when such request is made.

### Checklist

### DIFF
--- a/datahub/core/templates/core/docs/swagger-ui.html
+++ b/datahub/core/templates/core/docs/swagger-ui.html
@@ -27,7 +27,11 @@
           SwaggerUIBundle.presets.apis,
           SwaggerUIBundle.SwaggerUIStandalonePreset
         ],
-        layout: "BaseLayout"
+        layout: "BaseLayout",
+        requestInterceptor: (req) => {
+            req.headers['X-CSRFToken'] = "{{csrf_token}}";
+            return req;
+        }
       })
     </script>
   </body>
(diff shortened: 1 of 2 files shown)


datahub/core/templates/core/docs/swagger-ui.html
```diff
@@ -27,7 +27,11 @@
           SwaggerUIBundle.presets.apis,
           SwaggerUIBundle.SwaggerUIStandalonePreset
         ],
-        layout: "BaseLayout"
+        layout: "BaseLayout",
+        requestInterceptor: (req) => {
+            req.headers['X-CSRFToken'] = "{{csrf_token}}"
```
reupen: Is there a semicolon missing?

```suggestion
            req.headers['X-CSRFToken'] = "{{csrf_token}};"
```

changelog/add-csrf-token-to-api-docs.internal.md
```diff
@@ -0,0 +1 @@
+The CSRF token is now being added to API Docs request header. It is now possible to try POST requests.
```
reupen: I think this should be under `feature` rather than `internal` since it's visible outside of the code base.

datahub/core/templates/core/docs/swagger-ui.html
elcct: Yes! Updated. Thank you.

changelog/add-csrf-token-to-api-docs.internal.md
elcct: Makes sense. Updated. Thank you!

PR #2370, opened 2019-12-09
Add countries to interaction journey - add validations

sekharpanja:
### Description of change
This is second of three PR set, adding export countries to interaction journey.
First one is: https://github.com/uktrade/data-hub-api/pull/2358

This PR handles all business rule validations:
* export countries can only be added to non-investment themed interactions. Hence `were_countries_discussed` is mandatory field for both `export` and `other` themed interactions.
* At least one country/status combination is mandatory if `were_countries_discussed` is set to True
* Same country is not allowed to be added more than once, be it with different status

### Checklist


datahub/core/validators/rules_based.py
```diff
@@ -36,6 +37,26 @@ def field(self):
...
 
 
+class IsFeatureFlagActive(BaseRule):
+    """Rule to check if feature flag is active."""
+
+    def __call__(self, combiner) -> bool:
+        """
+        Returns True if the feature flag is active.
+        """
+        return is_feature_flag_active(self.field)
```
reupen: Using `field` to store the name of the feature flag doesn't seem ideal. Maybe we could override `__init__` and store it in a dedicated attribute?

datahub/interaction/serializers.py
```diff
@@ -466,6 +472,35 @@ class Meta:
...
+                            'theme',
+                            [Interaction.THEMES.export, Interaction.THEMES.other],
+                        ),
+                    ),
+                ),
+                ValidationRule(
+                    'required',
+                    OperatorRule('export_countries', is_not_blank),
+                    when=AndRule(
+                        EqualsRule('were_countries_discussed', is_not_blank),
```
reupen: Is this rule needed? Seems like the next one would cover it (if it's True then it isn't blank).

datahub/interaction/serializers.py
```diff
@@ -466,6 +472,35 @@ class Meta:
...
+                        EqualsRule('were_countries_discussed', True),
```
reupen: Small thing, but `OperatorRule('were_countries_discussed', bool)` would keep this consistent with the other rule.

datahub/interaction/test/views/test_interaction.py
```diff
@@ -336,15 +347,157 @@ def test_add(self, extra_data, permissions):
...
+                    'service': Service.inbound_referral.value.id,
+                    'was_policy_feedback_provided': False,
+                    'communication_channel': partial(random_obj_for_model, CommunicationChannel),
+                    'were_countries_discussed': None,
+                },
+                {
+                    'were_countries_discussed': ['This field is required.'],
+                },
+            ),
+            (
```
reupen: I know the comment above is meant to cover this set of parameters, but could we add a comment above each one to make it clearer?

datahub/interaction/validators.py
```diff
@@ -190,3 +190,35 @@ def __call__(self, data):
...
+
+        if not export_countries:
+            return
+
+        deduped = set()
+        for item in export_countries:
+            country = item['country']
+            if country in deduped:
+                raise serializers.ValidationError(
+                    "Same country can't be added more than once to export_countries.",
```
reupen: Looks like the logic can be simplified e.g.:

```python
countries = [item['country'] for item in export_countries]
if len(countries) > len(set(countries)):
    raise serializers.ValidationError(...)
```

reupen:
Some additional questions about the validation:

- if the feature flag is inactive, should we be making sure that the fields are blank to avoid letting unvalidated data through?
- do we need the validation that the export countries are unchanged during updates?

datahub/interaction/serializers.py
sekharpanja: This would mean, we don't care if `export_countries` is blank when `were_countries_discussed` is either blank/null or False.

reupen: Yes – but isn't that covered by the next rule?

datahub/core/validators/rules_based.py
```diff
@@ -36,6 +37,35 @@ def field(self):
         return self._field
 
 
+class IsFeatureFlagActive():
```
reupen: I think this should still inherit from `BaseRule` since `ValidationRule` etc. expect rules to be an instance of `AbstractRule`.

`field` is optional in `BaseRule.__init__()`, so I think it should be straightforward (i.e. we should be able to just add a call to `super().__init__()` if `BaseRule` is added to the list of base classes).

reupen:
@sekharpanja, did you see these questions above?

> Some additional questions about the validation:
> 
>   * if the feature flag is inactive, should we be making sure that the fields are blank to avoid letting unvalidated data through?
> 
>   * do we need the validation that the export countries are unchanged during updates?

sekharpanja:
> @sekharpanja, did you see these questions above?
> 
> > Some additional questions about the validation:
> > 
> > * if the feature flag is inactive, should we be making sure that the fields are blank to avoid letting unvalidated data through?
> > * do we need the validation that the export countries are unchanged during updates?

datahub/interaction/test/views/test_interaction.py
```diff
@@ -9,10 +9,17 @@
 from rest_framework.reverse import reverse
 from rest_framework.settings import api_settings
 
-from datahub.company.test.factories import AdviserFactory, CompanyFactory, ContactFactory
-from datahub.core.constants import Service
+from datahub.company.models import Company, CompanyExportCountry
+from datahub.company.test.factories import (
```
elcct: Is this change intentional? The imports seem to fit in one line.

datahub/interaction/serializers.py
```diff
@@ -100,6 +106,18 @@ class Meta:
         validators = []
 
 
+class InteractionExportCountrySerializer(serializers.ModelSerializer):
+    """
+    InteractionExportCountry serializer.
```
elcct: It looks like the doc string is short enough to be in one line.

datahub/interaction/serializers.py
```diff
@@ -289,6 +319,32 @@ def _save_dit_participants(self, interaction, validated_dit_participants):
         for adviser in old_advisers - new_advisers:
             old_adviser_mapping[adviser].delete()
 
+    def _save_export_countries(self, interaction, validated_export_countries):
+        """
+        Adds export countries related to an interaction
+        Update is not allowed yet
```
elcct: Maybe it will be better to say that update is not implemented and an attempt to update will raise `NotImplementedError` exception. Also it would be great to add full stops, so we know the comment is whole.

datahub/interaction/test/views/test_interaction.py
```diff
@@ -167,6 +211,118 @@ def test_add(self, extra_data, permissions):
...
+            'archived_by': None,
+            'archived_on': None,
+            'archived_reason': None,
+        }
+
+    @freeze_time('2017-04-18 13:25:30.986208')
+    @pytest.mark.parametrize('permissions', NON_RESTRICTED_ADD_PERMISSIONS)
+    def test_add_feature_flag_off(self, permissions):
+        """
+        Check export country fields are not allowed to be updated
```
elcct: Does this test do what comment says?

datahub/core/validators/rules_based.py
```diff
@@ -266,6 +266,24 @@ def __call__(self, combiner) -> bool:
         return all(rule(combiner) for rule in self._rules)
 
 
+class NotRule(BaseRule):
+    """
+    Field-less NOT rule that can be test on a rule in the `when` argument in
+    `ValidationRule.__init__`.
+    """
```
reupen: Optional: I think it might be nice to make this pass through the field from `self._rule` rather than making this rule this field-less (so i.e. inherit from `AbstractRule` instead and define a property like:

```python
    @property
    def field(self):
        return self._rule.field
```
)

datahub/interaction/serializers.py
```diff
@@ -500,6 +519,28 @@ class Meta:
...
+                        InRule(
+                            'theme',
+                            [Interaction.THEMES.export, Interaction.THEMES.other],
+                        ),
+                    ),
+                ),
+                ValidationRule(
+                    'invalid_for_update',
+                    OperatorRule('were_countries_discussed', is_blank),
+                    OperatorRule('export_countries', is_blank),
```
reupen: Don't we want to say here that the values shouldn't have changed, rather than the values are blank? (An added difficulty being that the order of the countries doesn't matter...)

datahub/interaction/serializers.py
```diff
@@ -413,6 +482,65 @@ class Meta:
...
+                        ),
+                    ),
+                ),
+                ValidationRule(
+                    'invalid_when_no_countries_discussed',
+                    OperatorRule('export_countries', is_blank),
+                    when=AndRule(
+                        IsObjectBeingCreated(),
+                        IsFeatureFlagActive(INTERACTION_ADD_COUNTRIES),
+                        EqualsRule('were_countries_discussed', False),
```
reupen: Should we make this:

```suggestion
                        OperatorRule('were_countries_discussed', not_),
```
?

datahub/interaction/serializers.py
```diff
@@ -413,6 +482,65 @@ class Meta:
...
+                    OperatorRule('were_countries_discussed', is_blank),
+                    OperatorRule('export_countries', is_blank),
+                    when=AndRule(
+                        IsObjectBeingCreated(),
+                        NotRule(IsFeatureFlagActive(INTERACTION_ADD_COUNTRIES)),
+                    ),
+                ),
+                ValidationRule(
+                    'invalid_for_investment',
+                    OperatorRule('were_countries_discussed', not_),
```
reupen: Do we need to add `OperatorRule('export_countries', not_)` here too?

datahub/interaction/serializers.py
sekharpanja: Wouldn't it be better not to allow those fields during update?

reupen: It would probably be easier to return an error if they are specified in an update (PATCH request) at all, yep, but the rule will look at existing data too. So this is saying interactions with those fields set can only be edited if those fields are made blank.

It would probably be easier to define a `validate_were_countries_discussed` and `validate_export_countries` for this (the rules were only meant for cross-field validation).

datahub/interaction/test/views/test_interaction.py
```diff
@@ -1137,38 +1737,197 @@ def test_draft_update_enforces_required_fields(self, permissions, data, error_re
...
+            (
+                {
+                    'were_countries_discussed': True,
                     'export_countries': [
                         {
                             'country': {
-                                'id': export_country.country.id,
-                                'name': export_country.country.name,
+                                'id': Country.greece.value.id,
+                                'name': Country.greece.value.name,
```
reupen: The name in `NestRelatedField`s isn't normally used in requests so would suggest leaving it out (to avoid confusion).

datahub/interaction/validators.py
```diff
@@ -190,3 +190,32 @@ def __call__(self, data):
...
+    within list of export_countries.
+    """
+
+    def __init__(self):
+        """Initialises the validator."""
+        self.instance = None
+
+    def set_context(self, serializer):
+        """Saves a reference to the model object."""
+        self.instance = serializer.instance
```
reupen: Was going to mention using DRF 3.11's `requires_context` when rebasing this, but actually it doesn't look like `self.instance` is used – can we remove these two methods?

PR #2392, opened 2019-12-17
Add utility function to rollback D&B updates

alixedi:
### Description of change

Add a utility function for rolling back DNB company updates for a given company.

I would recommend reviewing each commit separately.

### Checklist

### DIFF
--- a/datahub/dnb_api/test/test_utils.py
+++ b/datahub/dnb_api/test/test_utils.py
@@ -599,7 +599,7 @@ class TestRollbackDNBCompanyUpdate:
             assert getattr(company, field) == getattr(original_company, field)
 
         latest_version = Version.objects.get_for_object(company)[0]
-        latest_version.revision.comment = 'Reverted D&B update from: foo'
+        assert latest_version.revision.comment == 'Reverted D&B update from: foo'
 
     @pytest.mark.parametrize(
         'update_comment, error_message',
@@ -626,6 +626,6 @@ class TestRollbackDNBCompanyUpdate:
             update_descriptor='foo',
         )
 
-        with pytest.raises(RevisionNotFoundError) as e:
+        with pytest.raises(RevisionNotFoundError) as excinfo:
             rollback_dnb_company_update(company, update_comment)
-            assert str(e) == error_message
+            assert str(excinfo.value) == error_message
--- a/datahub/dnb_api/test/test_utils.py
+++ b/datahub/dnb_api/test/test_utils.py
@@ -2,6 +2,7 @@ from urllib.parse import urljoin
 from uuid import UUID
 
 import pytest
+import reversion
 from django.conf import settings
 from django.forms.models import model_to_dict
 from django.utils.timezone import now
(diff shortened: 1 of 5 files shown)


datahub/dnb_api/test/test_utils.py
```diff
@@ -554,3 +558,74 @@ def test_get_company_dnb_service_request_error(
...
+        """
+        company = CompanyFactory(duns_number=formatted_dnb_company['duns_number'])
+
+        update_company_from_dnb(
+            company,
+            formatted_dnb_company,
+            update_descriptor='foo',
+        )
+
+        with pytest.raises(RevisionNotFoundError) as e:
```
reupen: [Would suggest `excinfo` instead of `e` (it's an instance of ExceptionInfo)}(https://docs.pytest.org/en/latest/assert.html?highlight=raises#assertions-about-expected-exceptions):

```suggestion
        with pytest.raises(RevisionNotFoundError) as excinfo:
```

datahub/dnb_api/utils.py
```diff
@@ -319,3 +327,42 @@ def get_company_update_page(last_updated_after, next_page=None):
...
+):
+    """
+    Given a company, an update descriptor that identifies a particular update
+    patch and fields that default to ALL_DNB_UPDATE_FIELDS, rollback the record
+    to the state before the update was applied.
+    """
+    fields_to_update = fields_to_update or ALL_DNB_UPDATED_MODEL_FIELDS
+    update_comment = f'Updated from D&B [{update_descriptor}]'
+    rollback_version = _get_rollback_version(company, update_comment)
+    fields = rollback_version.field_dict if fields_to_update is None else {
```
reupen: Can `fields_to_update` be `None` here?

alixedi: Relic.

datahub/dnb_api/test/test_utils.py
```diff
@@ -554,3 +558,74 @@ def test_get_company_dnb_service_request_error(
...
+        """
+        Test that rollback_dnb_company_update will fail with the given error
+        message when there is an issue in finding the version to revert to.
+        """
+        company = CompanyFactory(duns_number=formatted_dnb_company['duns_number'])
+
+        update_company_from_dnb(
+            company,
+            formatted_dnb_company,
+            update_descriptor='foo',
```
mcbhenwood: Is the success of this test predicated upon the idea that (unlike in `test_rollback`) you don't initiate Reversion management on the original Company record? So then 'foo' is the base version because the `update_company_from_dnb` just acts on a versionless object?

If so then it feels a bit like this test is based too heavily on knowing how the implmentation works. Unless I am misunderstanding which is highly likely. Maybe this test is for a particular edge case I don't know about.

alixedi: This test is meant to tease out a couple of errors. One of those errors happens when there is no version before the update which is why I chose to not have a base version for `company`.

This is an unlikely corner case which is not easy to recover from so this test is essentially checking whether we are logging this properly.

PR #2397, opened 2019-12-18
interaction journey sync to company

sekharpanja:
### Description of change
This is third of three PR set, adding export countries to interaction journey.
First one is: https://github.com/uktrade/data-hub-api/pull/2358
Second is: https://github.com/uktrade/data-hub-api/pull/2370

This PR handles consolidation of `InteractionExportCountry` records into `CompanyExportCountry` model, allowing a `Company` to maintain countries with export interest independently.

If a country added to an interaction doesn't already exist at company level, it will be added. If in case that country already exists, following business rules apply:

* `Status` of `InteractionExportCountry` added to an interaction with current date overrides `CompanyExportCountry` with older status.
* Whereas `Status` of `InteractionExportCountry` added to an interaction with past date can't override `CompanyExportCountry` with newer status.
* An interaction added with future dated, will be treated as current date and existing rules apply.

### Checklist

### DIFF
--- a/datahub/company/models/company.py
+++ b/datahub/company/models/company.py
@@ -419,6 +419,28 @@ class Company(ArchivableModel, BaseModel):
         self.one_list_tier = None
         self.save()
 
+    def add_export_country(self, country, status, record_date, adviser):
+        """
+        Add a company export_country, if it doesn't exist.
+        If the company already exists and incoming status is different
+        check if incoming record is newer and update.
+        """
+        export_country, created = CompanyExportCountry.objects.get_or_create(
+            country=country,
+            company=self,
+            defaults={
+                'status': status,
+                'created_by': adviser,
+                'modified_by': adviser,
+            },
+        )
+
+        if not created:
+            if export_country.status != status and export_country.modified_on < record_date:
+                export_country.status = status
+                export_country.modified_by = adviser
+                export_country.save()
+
 
 class OneListCoreTeamMember(models.Model):
     """
--- a/datahub/interaction/serializers.py
+++ b/datahub/interaction/serializers.py
@@ -2,6 +2,7 @@ from collections import Counter
 from operator import not_
 
 from django.db.transaction import atomic
+from django.utils.timezone import now
 from django.utils.translation import gettext_lazy
 from rest_framework import serializers
 
@@ -164,7 +165,7 @@ class InteractionSerializer(serializers.ModelSerializer):
             "A theme can't be removed once set.",
         ),
         'invalid_when_feature_flag_off': gettext_lazy(
-            'export_countries fields are not valid when feature flag is off.',
+            'export countries related fields are not valid when feature flag is off.',
         ),
         'invalid_when_no_countries_discussed': gettext_lazy(
             'This field is only valid when countries were discussed.',
@@ -229,8 +230,8 @@ class InteractionSerializer(serializers.ModelSerializer):
 
     def validate_were_countries_discussed(self, were_countries_discussed):
         """
-        Make sure `were_countries_discussed` field is not being updated
-        when flag is inactive.
+        Make sure `were_countries_discussed` field is not being updated.
+        Updates are not allowed on this field.
         """
         if self.instance is None:
             return were_countries_discussed
@@ -239,7 +240,8 @@ class InteractionSerializer(serializers.ModelSerializer):
 
     def validate_export_countries(self, export_countries):
         """
-        Make sure `export_countries` field is not being updated
+        Make sure `export_countries` field is not being updated.
+        updates are not allowed on this field.
         """
         if self.instance is None:
             return export_countries
@@ -356,6 +358,8 @@ class InteractionSerializer(serializers.ModelSerializer):
         Adds export countries related to an interaction.
         Update is not allowed yet.
         An attempt to update will result in `NotImplementedError` exception.
+
+        Syncs interaction export countries into company export countries.
         """
         existing_country_mapping = {
             export_country.country: export_country
@@ -376,6 +380,16 @@ class InteractionSerializer(serializers.ModelSerializer):
                 interaction=interaction,
                 status=status,
                 created_by=interaction.created_by,
+            )
+            # Sync company_CompanyExportCountry model
+            # NOTE: current date is preferred over future interaction date
+            current_date = now()
+            record_date = current_date if interaction.date > current_date else interaction.date
+            interaction.company.add_export_country(
+                new_country,
+                status,
+                record_date,
+                interaction.created_by,
             )
 
     class Meta:
--- a/datahub/interaction/test/views/test_interaction.py
+++ b/datahub/interaction/test/views/test_interaction.py
@@ -10,7 +10,12 @@ from rest_framework.reverse import reverse
 from rest_framework.settings import api_settings
 
 from datahub.company.models import Company, CompanyExportCountry
-from datahub.company.test.factories import AdviserFactory, CompanyFactory, ContactFactory
+from datahub.company.test.factories import (
+    AdviserFactory,
+    CompanyExportCountryFactory,
+    CompanyFactory,
+    ContactFactory,
+)
 from datahub.core.constants import Country, Service
 from datahub.core.test_utils import APITestMixin, create_test_user, random_obj_for_model
 from datahub.event.test.factories import EventFactory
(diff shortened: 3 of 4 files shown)


datahub/interaction/serializers.py
```diff
@@ -185,13 +216,37 @@ class InteractionSerializer(serializers.ModelSerializer):
...
         if value and value.children.count() > 0:
             raise serializers.ValidationError(SERVICE_LEAF_NODE_NOT_SELECTED_MESSAGE)
         return value
 
+    def validate_were_countries_discussed(self, were_countries_discussed):
+        """
+        Make sure `were_countries_discussed` field is not being updated
+        when flag is inactive.
+        """
+        if self.instance is None:
```
mcbhenwood: I can't see how this is ensuring the feature flag is active (as suggested in the docstring), maybe I'm missing something? Looking at the test structure I'd say that actually you are simply validating against `were_countries_discussed` being in the payload for all updates, regardless of the feature flag setting. i.e. it's my guess that the docstring is wrong here, not the code.

datahub/interaction/serializers.py
```diff
@@ -413,6 +517,57 @@ class Meta:
                     OperatorRule('contacts', lambda value: len(value) <= 1),
                     when=OperatorRule('is_event', bool),
                 ),
+                ValidationRule(
+                    'invalid_when_feature_flag_off',
+                    OperatorRule('were_countries_discussed', is_blank),
+                    OperatorRule('export_countries', is_blank),
```
mcbhenwood: Wouldn't we be testing for non-blanks here rather than `is_blank`?

datahub/interaction/test/views/test_interaction.py
```diff
@@ -9,10 +9,18 @@
 from rest_framework.reverse import reverse
 from rest_framework.settings import api_settings
 
-from datahub.company.test.factories import AdviserFactory, CompanyFactory, ContactFactory
-from datahub.core.constants import Service
+from datahub.company.models import Company, CompanyExportCountry
```
mcbhenwood: I hate to be the one to say this but it really does look like this test module needs to be transformed into a package and split into smaller modules, e.g:

- test_interaction/test_add_interaction.py
- test_interaction/test_get_interaction.py
- test_interaction/test_update_interaction.py
- test_interaction/test_list_interaction.py

datahub/interaction/test/views/test_interaction.py
```diff
@@ -491,10 +1220,82 @@ def test_add(self, extra_data, permissions):
...
+                    'status':
+                        CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting,
+                },
+            ],
+        }
+        errors = {
+            'export_countries':
+                ['export_countries fields are not valid when feature flag is off.'],
+            'were_countries_discussed':
+                ['export_countries fields are not valid when feature flag is off.'],
```
mcbhenwood: I can see why you are using the same error message for both fields being present, that's fine. But I think you should change the text of the error message to give the group of fields a new name (i.e. not "export_countries fields") because at first glance it looked to me here like you'd accidentally pasted the error line from previous error. So maybe something like "export country related fields are not...."

mcbhenwood:
P.S. In the PR description I think you mean "This is *third* of three PR set"

datahub/interaction/serializers.py
```diff
@@ -289,6 +352,44 @@ def _save_dit_participants(self, interaction, validated_dit_participants):
...
+            if new_country in existing_country_mapping:
+                # TODO: updates are not supported yet
+                raise NotImplementedError()
+            InteractionExportCountry.objects.create(
+                country=new_country,
+                interaction=interaction,
+                status=status,
+                created_by=interaction.created_by,
+            )
+            # Sync comapny_CompanyExportCountry model
```
elcct: There is a typo.

datahub/interaction/serializers.py
sekharpanja: It is making sure both the fields are not blank in those conditions.

datahub/interaction/test/views/test_interaction.py
sekharpanja: Agreed. But can I please leave this for a refactoring PR.

changelog/interaction/interaction-add-export-countries.api.md
```diff
@@ -5,4 +5,10 @@ Interactions API `/v3/interaction` now allows to specify if there was a discussi
 `export_countries` field is of type `InteractionExportCountry` and takes a list of `country` and `status` combinations where `country` is of type `Country` and `status` is a choice of `Not interested`, `Currently exporting to` or `Future country of interest`.
 
 * Above details are only valid for non-investment themed interactions. Hence `were_countries_discussed` is mandatory field for both `export` and `other` themed interactions.
-* At least one country/status combination is mandatory if `were_countries_discussed` is set to True
\ No newline at end of file
+* At least one country/status combination is mandatory if `were_countries_discussed` is set to True
+
+`export_countries` added to an interaction are consolidated into `CompanyExportCountry` model, in order to maintain company level countries with various types export interests. If a country added to an interaction doesn't already exist at company level, it will be added. If in case that country already exists, following business rules apply:
```
elcct: Would it be possible to reword this? It is not quite clear what "level" means here.

datahub/company/models/company.py
```diff
@@ -419,6 +419,30 @@ def remove_from_one_list(self, modified_by):
...
+            defaults={
+                'status': status,
+                'created_by': adviser,
+                'modified_by': adviser,
+                'modified_on': record_date,
+            },
+        )
+
+        if not created:
+            if export_country.status is not status and export_country.modified_on < record_date:
```
reupen: `is not` is relying on the statuses being the same objects here – I would think `==` is more appropriate:
```suggestion
            if export_country.status == status and export_country.modified_on < record_date:
```

datahub/interaction/serializers.py
```diff
@@ -377,6 +381,15 @@ def _save_export_countries(self, interaction, validated_export_countries):
                 status=status,
                 created_by=interaction.created_by,
             )
+            # Sync company_CompanyExportCountry model
+            # NOTE: current date is preffered over future interaction date
```
reupen: Typo:
```suggestion
            # NOTE: current date is preferred over future interaction date
```

datahub/interaction/serializers.py
```diff
@@ -377,6 +381,15 @@ def _save_export_countries(self, interaction, validated_export_countries):
...
+            record_date = now() if interaction.date > now() else interaction.date
```
reupen: Would suggest evaluating `now()` only once here.

datahub/interaction/test/views/test_interaction.py
```diff
@@ -345,6 +351,166 @@ def test_add_with_export_countries_feature_flag_active(self, extra_data, permiss
...
+        response.json()['export_countries'] == [
+            {
+                'country': {
+                    'id': Country.canada.value.id,
+                    'name': Country.canada.value.name,
+                },
+                'status':
+                    CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting,
+            },
+        ]
```
reupen: I think it would be better to check `company.export_countries` directly (or the `CompanyExportCountry` objects) rather than make a second API request here (where the GET company endpoint works should be a separate test).

datahub/interaction/test/views/test_interaction.py
```diff
@@ -345,6 +351,166 @@ def test_add_with_export_countries_feature_flag_active(self, extra_data, permiss
...
+        assert company_response.status_code == status.HTTP_200_OK
+        response.json()['export_countries'] == [
+            {
+                'country': {
+                    'id': Country.canada.value.id,
+                    'name': Country.canada.value.name,
+                },
+                'status': expected_status,
+            },
+        ]
```
reupen: I think it would be better to check `company.export_countries` directly (or the `CompanyExportCountry` objects) rather than make a second API request here (where the GET company endpoint works should be a separate test).

datahub/company/models/company.py
```diff
@@ -419,6 +419,30 @@ def remove_from_one_list(self, modified_by):
...
+        check if incoming record is newer and update.
+        """
+        export_country, created = CompanyExportCountry.objects.get_or_create(
+            country=country,
+            company=self,
+            defaults={
+                'status': status,
+                'created_by': adviser,
+                'modified_by': adviser,
+                'modified_on': record_date,
```
reupen: Is setting `modified_on` here having an effect? (As `modified_on` uses `auto_now=True`.)

datahub/company/models/company.py
```diff
@@ -419,6 +419,30 @@ def remove_from_one_list(self, modified_by):
...
+            },
+        )
+
+        if not created:
+            if export_country.status is not status and export_country.modified_on < record_date:
+                export_country.status = status
+                export_country.modified_by = adviser
+                export_country.modified_on = record_date
```
reupen: Is setting `modified_on` here having an effect? (As `modified_on` uses `auto_now=True`.)

datahub/interaction/test/views/test_interaction.py
```diff
@@ -345,6 +351,162 @@ def test_add_with_export_countries_feature_flag_active(self, extra_data, permiss
...
+            ],
+        }
+
+        api_client = self.create_api_client(user=adviser)
+        response = api_client.post(url, request_data)
+
+        assert response.status_code == status.HTTP_201_CREATED
+
+        company_after_update = Company.objects.get(id=company.pk)
+        company_after_update.export_countries == [
```
reupen: This statement doesn't do anything – did you mean to put an assertion in?

datahub/interaction/test/views/test_interaction.py
```diff
@@ -345,6 +351,162 @@ def test_add_with_export_countries_feature_flag_active(self, extra_data, permiss
...
+                },
+            ],
+        }
+
+        api_client = self.create_api_client(user=adviser)
+        response = api_client.post(url, request_data)
+
+        assert response.status_code == status.HTTP_201_CREATED
+
+        company_after_update = Company.objects.get(id=company.pk)
```
reupen: Small thing, but it shouldn't be necessary to re-fetch the company since `company.export_countries.all()` will query the database directly anyway.

datahub/interaction/test/views/test_interaction.py
```diff
@@ -345,6 +351,162 @@ def test_add_with_export_countries_feature_flag_active(self, extra_data, permiss
...
+        company_after_update.export_countries == [
```
reupen: Same comments as for `test_add_interaction_add_company_export_country`.

datahub/interaction/test/views/test_interaction.py
```diff
@@ -395,17 +395,11 @@ def test_add_interaction_add_company_export_country(self, permissions):
...
-                },
-                'status':
-                    CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting,
-            },
-        ]
+        export_countries = company.export_countries.all()
+        assert export_countries.count() == 1
+        assert str(export_countries[0].country.id) == Country.canada.value.id
+        assert export_countries[0].status == \
+            CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting
```
reupen: Small point: wrapping expressions in `(` and `)` is preferred over `\` according to PEP 8 (could also use variables to shorten the statement though).

PR #2441, opened 2020-01-07
Add companies postcode filter

elcct:
### Description of change

`POST /v4/search/company`: A `uk_postcode` filter was added for the `address_postcode` and `registered_address_postcode` 
fields for UK based companies. The filter accepts a single or a partial postcode as well as an array of postcodes. Multiple postcodes are 
matched with `or` query.

Because we only want to analyse UK postcodes I created two properties - `uk_address_postcode` and `uk_registered_address_postcode` that only return value of corresponding postcode fields if given company is UK based. These properties are indexed and analysed with the analysers described below.

`Elasticsearch`: The search of UK postcodes has been improved so that any entity having a field storing UK postcode 
should use new `postcode_analyzer` and `postcode_search_analyzer` analyzers. It is now possible to perform partial 
search for area, area and district, area and sub-district and also area, district and sector and whole postcode. 
Postcodes are case insensitive and any spaces are filtered out before analysis.

### Checklist

### DIFF
--- a/changelog/elasticsearch-postcode-improvement.internal.md
+++ b/changelog/elasticsearch-postcode-improvement.internal.md
@@ -1,0 +1,4 @@
+`Elasticsearch`: The new `postcode_analyzer` and `postcode_search_analyzer` analyzers were added. Analyzers can be used 
+to enable partial search for area, district, sub-district, sector and a whole postcode. 
+
+The search is case insensitive and any spaces are filtered out before analysis.
--- a/datahub/search/company/models.py
+++ b/datahub/search/company/models.py
@@ -109,6 +109,8 @@ class Company(BaseESModel):
     turnover_range = fields.id_name_field()
     uk_region = fields.id_name_field()
     uk_based = Boolean()
+    uk_address_postcode = fields.PostcodeKeyword()
+    uk_registered_address_postcode = fields.PostcodeKeyword()
     vat_number = Keyword(index=False)
     duns_number = Keyword()
     website = Text()
@@ -131,6 +133,9 @@ class Company(BaseESModel):
             dict_utils.contact_or_adviser_dict,
         ),
         'latest_interaction_date': lambda obj: obj.latest_interaction_date,
+        'uk_address_postcode': lambda obj: obj.address_postcode if obj.uk_based else '',
+        'uk_registered_address_postcode':
+            lambda obj: obj.registered_address_postcode if obj.uk_based else '',
     }
 
     MAPPINGS = {
--- a/datahub/search/company/test/test_entity_search_views.py
+++ b/datahub/search/company/test/test_entity_search_views.py
@@ -91,6 +91,42 @@ def setup_data(es_with_collector):
 
 
 @pytest.fixture
+def company_names_and_postcodes(es_with_collector):
+    """Get companies with postcodes."""
+    (names, postcodes) = zip(*(
+        ('company_w1', 'w1 2AB'),  # AB in suffix to ensure not matched in AB tests
+        ('company_w1a', 'W1A2AB'),  # AB in suffix to ensure not matched in AB tests
+        ('company_w11', 'W112AB'),  # AB in suffix to ensure not matched in AB tests
+        ('company_ab1', 'AB11WC'),  # WC in suffix to ensure not matched in WC tests
+        ('company_ab10', 'ab10 1WC'),  # WC in suffix to ensure not matched in WC tests
+        ('company_wc2b', 'WC2B4AB'),  # AB in suffix to ensure not matched in AB tests
+        ('company_wc2n', 'WC2N9ZZ'),
+        ('company_wc1x', 'w  C   1 x0aA'),
+        ('company_wc1a', 'W C 1 A 1 G A'),
+        ('company_se1', 'SE13A J'),
+        ('company_se1_2', 'SE13AJ'),
+        ('company_se2', 'SE23AJ'),
+        ('company_se3', 'SE33AJ'),
+    ))
+
+    CompanyFactory.create_batch(
+        len(names),
+        name=factory.Iterator(names),
+        address_country_id=constants.Country.united_kingdom.value.id,
+        address_postcode=factory.Iterator(postcodes),
+        registered_address_country_id=constants.Country.united_kingdom.value.id,
+        registered_address_postcode=factory.Iterator(postcodes),
+    )
+
+    CompanyFactory(
+        name='non_uk_company_se1',
+        address_country_id=constants.Country.united_states.value.id,
+        address_postcode='SE13AJ',
+    )
+    es_with_collector.flush_and_refresh()
+
+
+@pytest.fixture
 def setup_headquarters_data(es_with_collector):
     """Sets up data for headquarter type tests."""
     CompanyFactory(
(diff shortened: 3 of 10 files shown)


reupen:
@elcct One of the requirements was that the filter only matches UK companies (and not foreign ones). Is that ensured here?

elcct:
@reupen I missed that. I have modified it so it now uses separate fields for UK company postcodes. Thank you!

datahub/search/elasticsearch.py
```diff
@@ -34,6 +34,59 @@
...
+        # Area and district
+        # Area + one or two numbers or a number and letter
+        '^([a-z]{1,2}(?:[0-9][a-z]|[0-9]{1,2}))[0-9][a-z]{2}',
+
+        # Area, district and sector
+        # Area + district + digit
+        '^([a-z]{1,2}(?:[0-9][a-z]|[0-9]{1,2})[0-9])[a-z]{2}',
+
+        # Whole postcode
+        '^([a-z]{1,2}(?:[0-9][a-z]|[0-9]{1,2})[0-9][a-z]{2})',
```
reupen: What are the advantages of trying to parse the entire postcode vs leaving `preserve_original` as `True` (the default)? Just thinking it might be better to be forgiving for the entire postcode match.

elcct: There are no advantages. I have removed the last pattern now and replaced it with `preserve_original`. Thank you!

datahub/company/models/company.py
```diff
@@ -302,6 +302,20 @@ def uk_based(self):
...
+        if self.uk_based:
+            return self.address_postcode
+        return ''
+
+    @property
+    def uk_registered_address_postcode(self):
+        """Return `registered_address_postcode` if company is uk based."""
+        if self.uk_based:
+            return self.registered_address_postcode
+        return ''
```
reupen: Are we able to use `COMPUTED_MAPPINGS` on the search model instead of adding properties here as these seem very specific to the search fields?

reupen:
Noticed that the new fields are appearing in responses – perhaps we could add them to `fields_to_exclude` on the view to prevent that.

datahub/search/fields.py
```diff
@@ -17,6 +17,12 @@
         'trigram': TrigramText(),
     },
 )
+# Keyword with normalisation that recognises UK postcodes to improve searching
+PostcodeKeyword = partial(
+    Text,
+    analyzer='postcode_analyzer',
+    search_analyzer='postcode_search_analyzer',
+)
```
reupen: We should be able to import the analysers and pass them in directly i.e.:

```python
PostcodeKeyword = partial(
    Text,
    analyzer=postcode_analyzer,
    search_analyzer=postcode_search_analyzer,
)
```

(That way, we don't need to add them to `ANALYZERS` in `datahub.search.elasticsearch`.)

datahub/search/elasticsearch.py
```diff
@@ -34,6 +34,57 @@
...
+
+postcode_filter = analysis.token_filter(
+    'postcode_filter',
+    type='pattern_capture',
+    # Match whole postcode
+    preserve_original=True,
+    patterns=[
+        # Area
+        # One or two letters
+        '^([a-z]{1,2})(?:[0-9][a-z]|[0-9]{1,2})[0-9][a-z]{2}',
```
reupen: Should we make these raw strings since they are regexes?

Also, I think it would be good to define them in named constants to rely on comments a bit less e.g.

```python
# First one or two letters
POSTCODE_AREA_REGEX = r'...'
```

datahub/search/elasticsearch.py
```diff
@@ -34,6 +34,57 @@
...
+
+        # Area and district (with sub-district code ignored)
```
reupen: Very minor thing but I think saying 'area and district' is a bit redundant (from [the PAF documentation](https://www.poweredbypaf.com/wp-content/uploads/2017/07/Latest-Programmers_guide_Edition-7-Version-6.pdf)):

image

(Would be good to add a comment saying something like 'See the Royal Mail programmer's guide for the exact definitions too.)

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -355,6 +399,30 @@ def test_response_body(self, es_with_collector):
...
+                [],
+            ),
+
+            # uk_postcode
+            (
+                {
+                    'uk_postcode': ['SW1', 'wc'],
+                },
+                ['abc defg ltd'],
+            ),
```
reupen: Is there a reason for testing the filter here and in `test_search_postcodes` below? I think it would be good to do the testing in one place. 

(Also would be good to add a case with a space in the search term if we don't have one already.)

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -372,12 +440,81 @@ def test_filters(self, setup_data, filters, expected_companies):
...
+            ('W1', ['company_w1', 'company_w1a']),
+            ('W11', ['company_w11']),
+            ('W1A', ['company_w1a']),
+            ('W', ['company_w1', 'company_w1a', 'company_w11']),
+            ('WC2', ['company_wc2b', 'company_wc2n']),
+            ('WC', ['company_wc2b', 'company_wc2n', 'company_wc1x', 'company_wc1a']),
+            ('AB1', ['company_ab1']),
+            ('AB10', ['company_ab10']),
+            ('AB', ['company_ab1', 'company_ab10']),
+            ('SE1', ['company_se1', 'company_se1_2']),
```
reupen: Perhaps these could be organised by whether an area, district, sub-district or sector is being searched for? Would make it a bit easier to see what's covered.

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -372,12 +440,81 @@ def test_filters(self, setup_data, filters, expected_companies):
...
+                    'company_w1',
+                    'company_w1a',
+                    'company_w11',
+                    'company_ab1',
+                    'company_ab10',
+                ],
+            ),
+
+            # Valid and invalid
+            (['SE1', 'Invalid'], ['company_se1', 'company_se1_2']),
```
reupen: I'm not sure there is much benefit to having so many array cases since the OR logic should be being tested elsewhere already. I think one or two arrays cases should be sufficient.

datahub/search/elasticsearch.py
elcct: I see your point, but I was commenting about individual parts that whole postcode is composed of  that is area code, district, sector and unit and which ones will go into group to make a token.
If think the table above doesn't explain well what makes e.g. Postcode sector (that is area + district + sector). I am happy to modify the comments though and link to Royal Mail documentation.

elcct: I have updated the comments now, I think together with the constants it looks a little better. Thank you!

datahub/search/company/test/test_entity_search_views.py
elcct: I added it because there are also other filters being tested, but since it also has a separate test, it will make sense to remove it from that test. Thank you!

datahub/search/company/test/test_entity_search_views.py
elcct: I agree. I removed some tests cases. Thank you!

datahub/search/company/test/test_entity_search_views.py
elcct: This is a good idea. I have organised the tests now. Thank you!

datahub/search/elasticsearch.py
elcct: I started doing this and realised maybe it is not as readable as I thought it will be, however, I see constants could help understand the patterns quicker, so I added it now. Thank you!

datahub/search/fields.py
elcct: I fixed that now. Thank you!

datahub/company/models/company.py
elcct: That makes sense. There is no point to have these properties on the model. I have updated the search model now. Thank you!

datahub/search/elasticsearch.py
```diff
@@ -41,30 +41,36 @@
...
     # Match whole postcode
     preserve_original=True,
     patterns=[
-        # Area
-        # One or two letters
-        '^([a-z]{1,2})(?:[0-9][a-z]|[0-9]{1,2})[0-9][a-z]{2}',
+        # Match postcode area
+        # See the Royal Mail programmer's guide for the exact definitions
+        f'^({POSTCODE_AREA_REGEX})(?:{POSTCODE_DISTRICT_REGEX})'
+        f'{POSTCODE_SECTOR_REGEX}{POSTCODE_UNIT_REGEX}',
```
reupen: Should be able to make this raw f-strings as well (by using an `rf` prefix).

datahub/search/company/models.py
```diff
@@ -131,6 +133,14 @@ class Company(BaseESModel):
...
         ),
         'latest_interaction_date': lambda obj: obj.latest_interaction_date,
+        'uk_address_postcode': dict_utils.conditional_attrgetter_with_default(
+            'uk_based',
+            'address_postcode',
+        ),
+        'uk_registered_address_postcode': dict_utils.conditional_attrgetter_with_default(
+            'uk_based',
+            'registered_address_postcode',
+        ),
```
reupen: I think lambdas would be sufficient and easier to follow than the generic function `conditional_attrgetter_with_default()` (e.g. `lambda obj: obj.address_postcode if obj.uk_based else None`).

datahub/search/company/test/test_entity_search_views.py
```diff
@@ -372,12 +414,65 @@ def test_filters(self, setup_data, filters, expected_companies):
...
         response_data = response.json()
         assert response_data['count'] == len(expected_companies)
-
         assert [
             result['name']
             for result in response_data['results']
         ] == expected_companies
 
+    @pytest.mark.parametrize(
+        'search_term,expected_companies',
```
reupen: Do we have a case for a lowercase `search_term`? Also, one with a full postcode?

datahub/search/elasticsearch.py
```diff
@@ -34,6 +34,63 @@
...
+
+        # Match postcode district (with sub-district code ignored)
+        # This is so `wc1` query would match `wc1ab` and `wc1a1ab`, but not `wc111ab`
+        # Area + one or two digits
+        f'^(({POSTCODE_AREA_REGEX}[0-9]){POSTCODE_SECTOR_REGEX}{POSTCODE_UNIT_REGEX}|'
+        f'({POSTCODE_AREA_REGEX}[0-9]{{2}}){POSTCODE_SECTOR_REGEX}{POSTCODE_UNIT_REGEX}|'
+        f'({POSTCODE_AREA_REGEX}[0-9])[a-z]?{POSTCODE_SECTOR_REGEX}{POSTCODE_UNIT_REGEX})',
```
reupen: Ah – actually I meant to put the entire regexes in named constants, but I think this is also a good idea. Maybe we could drop either `_REGEX` or `POSTCODE_` from these ones to make the names less repetitive?

changelog/elasticsearch-postcode-improvement.internal.md
```diff
@@ -0,0 +1,4 @@
+`Elasticsearch`: The search of UK postcodes has been improved so that any entity having a field storing UK postcode 
+should use new `postcode_analyzer` and `postcode_search_analyzer` analyzers. It is now possible to perform partial 
+search for area, area and district, area and sub-district and also area, district and sector and whole postcode.
```
reupen: I think we could make this more concise like elsewhere (or, alternatively, use bullet points) as it's a bit hard to parse at the moment.

datahub/search/elasticsearch.py
elcct: I have removed the `POSTCODE_` prefix. It is now more readable. Thank you!

datahub/search/company/models.py
elcct: Great idea. That function didn't really belong in dict_utils anyway, so using lambda makes perfect sense for this one-off use case. Thank you!

datahub/search/elasticsearch.py
elcct: That's true! Thank you!

changelog/elasticsearch-postcode-improvement.internal.md
elcct: I reworded it a bit, I hope it makes more sense now.

datahub/search/company/test/test_entity_search_views.py
elcct: I added few more test cases. Thank you!

PR #2461, opened 2020-01-13
Remove metadata for dnb_match models

alixedi:
### Description of change

I forgot to use `datahub.core.migration_utils.DeleteModelWithMetadata` in #2439.

### Checklist


datahub/dnb_match/migrations/0004_delete_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dnb_match', '0003_delete'),
+    ]
+
+    operations = [
+        migrations.RunPython(forward_func, reverse_func)
```
reupen: Could use `migrations.RunPython.noop` instead of `reverse_func`:

```suggestion
        migrations.RunPython(forward_func, migrations.RunPython.noop)
```

(Also I'd say this should technically have `elidable=True`, since we wouldn't want this if the migrations were squashed, but it doesn't matter if we're going to delete the app anyway.)

PR #2462, opened 2020-01-13
Adding new history log for CompanyExportCountry model

sekharpanja:
### Description of change
A new model `company_companyexportcountryhistory` was created to log all changes made to `company_companyexportcountry` model. This will maintain all inserts, updates and deletions to that model with appropriate `history_type` with values `insert`, `update` and `delete`.

### Checklist


datahub/company/signals.py
```diff
@@ -49,3 +53,41 @@ def notify_dnb_investigation_post_save(sender, instance, created, raw, **kwargs)
 def _notify_dnb_investigation_post_save(instance):
     logger.info(f'Company with ID {instance.id} is pending DNB investigation.')
     notify_new_dnb_investigation(instance)
+
+
+@receiver(post_save, sender=CompanyExportCountry)
```
reupen: Would be good to set `dispatch_uid` for this and the next receiver (as per https://docs.djangoproject.com/en/3.0/topics/signals/#preventing-duplicate-signals).

datahub/company/test/test_signals.py
```diff
@@ -85,3 +86,79 @@ def test_notify_signal_company_updated_no_investigation(self):
...
+        Test that creating new CompanyExportCountry record
+        sets up a corresponding history record.
+        """
+        export_country = CompanyExportCountryFactory()
+        history = CompanyExportCountryHistory.objects.filter(id=export_country.id)
+        assert len(history) == 1
+        assert export_country.id == history[0].id
+        assert export_country.company == history[0].company
+        assert export_country.country == history[0].country
+        assert export_country.status == history[0].status
```
reupen: It would be good to swap the sides of the comparisons here, so the actual value is on the left-hand side and the expected value is on the right-hand side, to be consistent with other tests.

datahub/company/signals.py
```diff
@@ -49,3 +53,41 @@ def notify_dnb_investigation_post_save(sender, instance, created, raw, **kwargs)
...
+def record_export_country_history_update(sender, instance, created, raw, **kwargs):
+    """
+    Record export country changes to history.
+    """
+    action = CompanyExportCountryHistory.HISTORY_TYPES.update
```
reupen: I think we probably don't want to do anything if `raw=True` (which happens during fixture loading), as the history objects should be included in the fixtures temselves.

datahub/company/models/company.py
```diff
@@ -521,3 +521,60 @@ def __str__(self):
...
+        related_name='+',
+    )
+    company = models.ForeignKey(
+        Company,
+        on_delete=models.CASCADE,
+        related_name='export_countries_history',
+    )
+    status = models.CharField(
+        null=True,
+        blank=True,
```
reupen: Are `null=True` and `blank=True` needed here?

datahub/company/models/company.py
```diff
@@ -521,3 +521,60 @@ def __str__(self):
...
+    )
+
+    history_id = models.UUIDField(
+        primary_key=True,
+        default=uuid.uuid4,
+    )
+    history_date = models.DateTimeField(db_index=True, null=True, blank=True, auto_now_add=True)
+    history_user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        null=True, blank=True,
```
reupen: Would be good to add a newline between the two arguments here.

PR #2467, opened 2020-01-14
Enable querying the search endpoints in API browser

elcct:
### Description of change

In our API browser it was not possible to make queries to the search endpoints mostly due to "non-standard" nature of our search app.
This PR makes API browser aware of the query serialisers, so that we can edit the request body and perform a request. 
In case of the basic search, the description of query parameters has been added.
The response schema is hidden, because we would have to create it manually and I think it is not worth the effort. 
Currently the AutoSchema is still not flexible to extend as vital methods are private.

Basic search before:
basic-search-before

Basic search after:
basic-search-after

Entity search before:
entity-search-before

Entity search after:
entity-search-after

### Checklist

### DIFF
--- a/datahub/search/views.py
+++ b/datahub/search/views.py
@@ -9,6 +9,7 @@ from django.utils.timezone import now
 from oauth2_provider.contrib.rest_framework.permissions import IsAuthenticatedOrTokenHasScope
 from rest_framework.generics import ListAPIView
 from rest_framework.response import Response
+from rest_framework.schemas.openapi import AutoSchema
 from rest_framework.views import APIView
 
 from datahub.core.csv import create_csv_response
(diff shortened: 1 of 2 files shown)


datahub/search/views.py
```diff
@@ -74,6 +137,10 @@ def get(self, request, format=None):
 
         return Response(data=response)
 
+    def get_serializer(self):
+        """Return query serializer for use with OpenAPI documentation."""
+        return BasicSearchQuerySerializer()
```
reupen: Is this being used?

changelog/enable-search-queries-for-api-browser.internal.md
```diff
@@ -0,0 +1 @@
+It is now possible to perform queries to search endpoints in the API browser.
```
reupen: I think this should be a `feature` fragment, since it's an externally visible change.

elcct: I have changed it now. Thank you!

datahub/search/views.py
elcct: Actually no! I removed it now. Thank you!

PR #2473, opened 2020-01-16
Improve delete_orphaned_versions command

elcct:
### Description of change

The `get_deleted` method wouldn't work in our environment so it has been replaced with a more efficient one that gives the same results.

### Checklist

### DIFF
--- a/datahub/cleanup/management/commands/delete_orphaned_versions.py
+++ b/datahub/cleanup/management/commands/delete_orphaned_versions.py
@@ -5,6 +5,8 @@ import reversion
 from django.apps import apps
 from django.core.management import BaseCommand
 from django.db import transaction
+from django.db.models import CharField
+from django.db.models.functions import Cast
 from reversion.models import Revision, Version
 
 logger = getLogger(__name__)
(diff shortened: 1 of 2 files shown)


datahub/cleanup/management/commands/delete_orphaned_versions.py
```diff
@@ -41,7 +43,28 @@ def handle(self, *args, **options):
...
+                'object_id',
+            ).difference(
+                all_model_ids,
+            )
+
+            """
+            The Version's `get_deleted` method was originally used here, but it was not working
+            quickly enough for our use case. The records to be deleted are now being determined
+            in a way that is easier on the database being used.
+            """
```
reupen: Shouldn't this be a normal comment?

changelog/delete-orphaned-versions.internal.md
```diff
@@ -0,0 +1,2 @@
+The `delete_orphaned_versions` command was modified to select records to be deleted in a fashion that is more
```
reupen: Small thing but I would see this as either a `feature` or `bugfix` news fragment (`internal` I would take to mean that there is no change to how the app behaves).

datahub/cleanup/management/commands/delete_orphaned_versions.py
elcct: Yes! I updated it now. Thank you.

changelog/delete-orphaned-versions.internal.md
elcct: I changed it to `bugfix`. Thank you!

elcct:
Thank you! I moved the query creation into a separate function now.

PR #2487, opened 2020-01-21
Patch MAX_DATABAG_BREADTH for sentry

alixedi:
### Description of change

This patches a constant in Sentry to raise the upper limit of e.g. elements in an array that we plonk in the `extra` kwarg [here](https://github.com/uktrade/data-hub-api/blob/develop/datahub/dnb_api/tasks.py#L89).

The original value for this constant was 10 which meant that we were getting:

Screenshot 2020-01-21 at 15 14 59

i.e. 10 IDs instead of 41.

I found the solution [here](https://forum.sentry.io/t/python-sdk-extra-data-capped-at-400-characters/6909).

Sentry also does some trimming server side and we would know when we run into it but for now, this should help a lot with monitoring automatic-updates.

### Checklist

### DIFF
--- a/config/settings/common_sentry.py
+++ b/config/settings/common_sentry.py
@@ -1,8 +1,12 @@
 import sentry_sdk
 from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
+from sentry_sdk import serializer
 
 from config.settings.common import *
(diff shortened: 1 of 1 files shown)


alixedi:
Had a chat with @reupen, I will put an admin filter for `dnb_modified_on` instead 👍

PR #2492, opened 2020-01-23
Adding readonly export_countries field to company API

sekharpanja:
### Description of change
This is first part of two part feature to allow GET/PATCH of CompanyExportCountry data within company API.

This adds readonly export_countries field to expose new CompanyExportCountry model data along with old export country fields.

### Checklist

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -195,6 +195,18 @@ class ContactSerializer(PermittedFieldsModelSerializer):
         }
 
 
+class CompanyExportCountrySerializer(serializers.ModelSerializer):
+    """
+    CompanyExportCountry serializer.
+    """
+
+    country = NestedRelatedField(meta_models.Country)
+
+    class Meta:
+        model = CompanyExportCountry
+        fields = ('country', 'status')
+
+
 class CompanySerializer(PermittedFieldsModelSerializer):
     """
     Base Company read/write serializer
@@ -269,6 +281,7 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         allow_null=True,
     )
     address = AddressSerializer(source_model=Company, address_source_prefix='address')
+    export_countries = CompanyExportCountrySerializer(many=True, read_only=True)
 
     # Use our RelaxedURLField instead to automatically fix URLs without a scheme
     serializer_field_mapping = {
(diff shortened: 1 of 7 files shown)


datahub/company/serializers.py
```diff
@@ -493,6 +507,7 @@ class Meta:
             'is_global_ultimate',
             'global_ultimate_duns_number',
             'dnb_modified_on',
+            'export_countries',
```
reupen: This doesn't appear to be having any effect (probably because the field is explicitly specified) from testing locally. So we will probably to swap out `required=False` for `read_only=True` on the field.

datahub/company/test/test_company_views_export_countries.py
```diff
@@ -0,0 +1,445 @@
...
+)
+from datahub.core.constants import Country
+from datahub.core.test_utils import APITestMixin, create_test_user, format_date_or_datetime
+from datahub.metadata.models import Country as CountryModel
+
+
+class TestCompaniesToCompanyExportCountryModel(APITestMixin):
+    """Tests for copying export countries from company model to CompanyExportCountry model"""
+
+    def test_get_company_with_export_countries(self):
```
reupen: I think, instead of adding a new test with the entire response body, it would be good to either parametrise the existing `TestGetCompany.test_get` test or possibly add a new test that only looks at the `export_countries` field in the response.

Also, the docstring for this class is `Tests for copying export countries from company model to CompanyExportCountry model`. This test seems out of place with that in mind; we should be able to delete the rest of the tests in the class once the old fields are removed.

reupen: The first part is resolved, but the test still doesn't fit in with the class:

```python
class TestCompaniesToCompanyExportCountryModel(APITestMixin):
    """Tests for copying export countries from company model to CompanyExportCountry model"""
```

So I think it should go in a separate class...

datahub/company/test/factories.py
```diff
@@ -93,6 +93,13 @@ def future_interest_countries(self):  # noqa: D102
         """
         return []
 
+    @to_many_field
+    def export_countries(self):  # noqa: D102
+        """
+        Add support for setting `export_countries`.
```
elcct: Is it a TODO?

datahub/company/test/test_company_views_export_countries.py
sekharpanja: Moved the test back into TestGetCompany, which is where it belongs.

datahub/company/test/factories.py
sekharpanja: This is to set `allow export_countries` during `Company` setup in the factory.

PR #2496, opened 2020-01-23
Add investment projects activity / SPI report dataset endpoint

elcct:
### Description of change

The `GET /v4/dataset/investment-projects-activity-dataset` has been added. The endpoint returns SPI report records for corresponding investment projects.

It updates the SPI report so that formatting of propositions can be changed, as the response of dataset is in JSON, so there is no need for `;` separated objects.

There will be a second PR that removes SPI report from Django admin. That should resolve duplicated `ist_adviser` and `propositions` fixtures in dataset app.

For a time being we will have to run both versions of the report (the endpoint and celery task). Once we remove the task and no longer need to support both looks, the implementation should become much cleaner.

### Checklist

### DIFF
--- a/datahub/core/query_utils.py
+++ b/datahub/core/query_utils.py
@@ -85,6 +85,7 @@ def get_array_agg_subquery(
     model,
     join_field_name,
     expression_to_aggregate,
+    filter=None,
     distinct=False,
     ordering=(),
 ):
@@ -112,6 +113,15 @@ def get_array_agg_subquery(
             ),
         )
 
+        Interaction.objects.annotate(
+            adviser_first_names=get_array_agg_subquery(
+                InteractionDITParticipant,
+                'interaction',
+                'first_name,
+                filter=Q('first_name__startswith='A'),
+            ),
+        )
+
     Note: The usage of this function differs from `get_string_agg_subquery()`, as this function
     omits the outer model from the subquery to avoid unwanted NULL values appearing in the
     returned arrays when rows don't have any values in the intermediate model being queried.
@@ -124,7 +134,12 @@ def get_array_agg_subquery(
     """
     return get_aggregate_subquery(
         model,
-        ArrayAgg(expression_to_aggregate, distinct=distinct, ordering=ordering),
+        ArrayAgg(
+            expression_to_aggregate,
+            distinct=distinct,
+            ordering=ordering,
+            filter=filter,
+        ),
         join_field_name=join_field_name,
     )
 
--- a/datahub/dataset/investment_project/pagination.py
+++ b/datahub/dataset/investment_project/pagination.py
@@ -1,0 +1,27 @@
+"""
+Pagination serializers determine the structure of the output that should
+be used for paginated responses.
+"""
+
+from rest_framework.response import Response
+
+from datahub.dataset.core.pagination import DatasetCursorPagination
+from datahub.dataset.investment_project.spi import SPIReportFormatter
+
+
+class InvestmentProjectActivityDatasetViewCursorPagination(DatasetCursorPagination):
+    """Cursor Pagination for SPI Report."""
+
+    ordering = ('created_on', 'pk')
+
+    def get_paginated_response(self, data):
+        """Get paginated response."""
+        spi_report = SPIReportFormatter()
+        results = spi_report.format(data)
+        return Response(
+            {
+                'next': self.get_next_link(),
+                'previous': self.get_previous_link(),
+                'results': results,
+            },
+        )
--- a/datahub/dataset/investment_project/spi.py
+++ b/datahub/dataset/investment_project/spi.py
@@ -1,0 +1,67 @@
+from dateutil.parser import parse as dateutil_parse
+
+from datahub.investment.project.proposition.models import PropositionStatus
+from datahub.investment.project.report.spi import SPIReport
+
+
+class SPIReportFormatter:
+    """
+    SPI Report Formatter changes the format of the SPI Report.
+
+    TODO: Once original SPI report is removed from Django admin, this format should be
+    implemented in the report itself.
+    """
+
+    required_fields_label_mapping = {
+        'Data Hub ID': 'investment_project_id',
+        'Enquiry processed': 'enquiry_processed',
+        'Enquiry type': 'enquiry_type',
+        'Enquiry processed by': 'enquiry_processed_by_id',
+        'Assigned to IST': 'assigned_to_ist',
+        'Project manager assigned': 'project_manager_assigned',
+        'Project manager assigned by': 'project_manager_assigned_by_id',
+        'Project moved to won': 'project_moved_to_won',
+        'Aftercare offered on': 'aftercare_offered_on',
+        'Propositions': 'propositions',
+    }
+
+    required_fields_value_mapping = {
+        'Enquiry processed by': lambda adviser: str(adviser),
+        'Project manager assigned by': lambda adviser: str(adviser.id),
+    }
+
+    def __init__(self):
+        """Initialise SPI Report Formatter."""
+        self._SPIReport = SPIReport(proposition_formatter=proposition_formatter)
+
+    def filter_fields(self, result):
+        """Filter results fields."""
+        return {
+            self.required_fields_label_mapping[key]: value
+            for key, value in result.items()
+            if key in self.required_fields_label_mapping
+        }
+
+    def format(self, investment_projects):
+        """
+        Enrich Investment Project record with SPI report data and only include required fields.
+        """
+        for investment_project in investment_projects:
+            spi_report_row = self._SPIReport.get_row(investment_project)
+            result = self.filter_fields(spi_report_row)
+            yield result
+
+
+def proposition_formatter(propositions):
+    """Returns a list of propositions with selected fields."""
+    return [
+        {
+            'deadline': dateutil_parse(proposition['deadline']).strftime('%Y-%m-%d'),
+            'status': proposition['status'],
+            'modified_on':
+                dateutil_parse(proposition['modified_on']).isoformat()
+                if proposition['status'] != PropositionStatus.ongoing else '',
+            'adviser_id': proposition['adviser_id'],
+        }
+        for proposition in propositions
+    ]
--- a/datahub/investment/project/report/spi.py
+++ b/datahub/investment/project/report/spi.py
@@ -12,18 +12,51 @@ SPI5_START  - when project has been moved to won
 SPI5_END    - earliest interaction when aftercare was offered, only for new investor,
               only for IST managed projects
 """
+from dateutil.parser import parse as dateutil_parse
+from django.db.models import Q
 
 from datahub.core.constants import InvestmentProjectStage as Stage, Service
 from datahub.core.csv import csv_iterator
+from datahub.core.query_utils import (
+    get_array_agg_subquery,
+    get_full_name_expression,
+    JSONBBuildObject,
+)
 from datahub.interaction.models import Interaction
 from datahub.investment.project.constants import InvestorType
 from datahub.investment.project.models import InvestmentProject
 from datahub.investment.project.proposition.constants import PropositionStatus
+from datahub.investment.project.proposition.models import Proposition
 from datahub.metadata.models import Team
+from datahub.metadata.query_utils import get_service_name_subquery
+
+SPI1_END_SERVICE_IDS = {
+    Service.investment_enquiry_requested_more_information.value.id,
+    Service.investment_enquiry_confirmed_prospect.value.id,
+    Service.investment_enquiry_assigned_to_ist_cmc.value.id,
+    Service.investment_enquiry_assigned_to_ist_sas.value.id,
+    Service.investment_enquiry_assigned_to_hq.value.id,
+    Service.investment_enquiry_transferred_to_lep.value.id,
+    Service.investment_enquiry_transferred_to_da.value.id,
+    Service.investment_enquiry_transferred_to_lp.value.id,
+}
+
+SPI2_START_SERVICE_IDS = {
+    Service.investment_enquiry_assigned_to_ist_cmc.value.id,
+    Service.investment_enquiry_assigned_to_ist_sas.value.id,
+}
+
+SPI5_END_SERVICE_IDS = {
+    Service.investment_ist_aftercare_offered.value.id,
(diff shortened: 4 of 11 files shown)


reupen:
@elcct I gave this a try locally.

First thing is that, for a page of 100 projects, it is executing over 300 queries:

image

My database is very bare so it is hard to know how that will behave in production. Can it be optimised?

Second, the response looks like this:

```json
{
  "next": "http://localhost:8000/v4/dataset/investment-projects-activity-dataset?cursor=cD1mN2FiZjA3NS01MDcxLTQ2YzYtYTMyZS03ZTM1MzhmZWNiY2I%3D&format=json",
  "previous": null,
  "results": [
    {
      "investment_project_id": "00195fc6-d1c1-4478-87f0-b57e67e65ec1"
    },
    {
      "investment_project_id": "0283955a-c9bc-45e5-b6c6-ab4a42122fbc"
    },
    {
      "investment_project_id": "07f41b5f-3adc-4f9f-9ccc-3c1c30596bc4"
    },
    {
      "investment_project_id": "0c4c31dd-ebec-42ff-889a-c80963ee6168"
    }
  ]
}
```

Are the result properties dynamic?

elcct:
@reupen It does make at least one query per investment project, that's why the number is quite high. This, however, works the same as the original report. This is also not meant to be called frequently - ideally once a day. I am looking into how I could improve the performance. 

The fields are created if corresponding value exists.

reupen:
> This, however, works the same as the original report.

However, the calling context is different as this is running in a web process and not Celery. I'm also not sure if there would be even more queries with more realistic data.

> The fields are created if corresponding value exists.

This doesn't seem like standard behaviour – could you elaborate on the reason for it? Does it fit in with how Data Workspace works?

elcct:
Fair point re calling context. I am going to try to move it into a single query. 

Re dynamic fields, I see that this is wrong - in case of CSV output, that would mean just an empty cell, but it doesn't make sense for JSON object. I am going to change that. Thank you!

elcct:
I have improved the report performance. For my 100 investment projects, the number of queries went down from 317 to 17. I am looking at ways to improve it further.
The endpoint is now returning all fields as well.

elcct:
I have updated the query and now it does just one.

datahub/dataset/investment_project/pagination.py
```diff
@@ -0,0 +1,77 @@
...
+from datahub.investment.project.proposition.models import PropositionStatus
+from datahub.investment.project.report.spi import SPIReport
+
+
+class InvestmentProjectActivityDatasetViewCursorPagination(DatasetCursorPagination):
+    """Cursor Pagination for SPI Report."""
+
+    ordering = ('id', )
+
+    required_fields_label_mapping = {
```
currycoder: Could there perhaps be some comments explaining the purpose of these two class attributes?

datahub/dataset/investment_project/pagination.py
```diff
@@ -0,0 +1,77 @@
...
+
+from dateutil.parser import parse as dateutil_parse
+from rest_framework.response import Response
+
+from datahub.dataset.core.pagination import DatasetCursorPagination
+from datahub.investment.project.proposition.models import PropositionStatus
+from datahub.investment.project.report.spi import SPIReport
+
+
+class InvestmentProjectActivityDatasetViewCursorPagination(DatasetCursorPagination):
```
currycoder: It's a little confusing to me that this pagination class has so much formatting logic within it.  I think it could be cleaner with neater separation of concerns if there was an additional `SPIReportFormatter` class of some kind which `get_paginated_response()` uses to build formatted results from data.

That might also allow for some extended comments/class documentation which explain how the translation between the SPIReport and the chosen formatting is achieved.  The formatter logic could also be tested in isolation.

datahub/investment/project/report/spi.py
```diff
@@ -171,16 +166,15 @@ def _format_propositions(self, propositions):
         deadline;ongoing;;adviser.name;deadline;completed;modified_on;adviser.name...
         """
         formatted = []
-
         for proposition in propositions:
-            formatted.append(proposition.deadline.isoformat())
-            formatted.append(proposition.status)
-            if proposition.status == PropositionStatus.ongoing:
+            formatted.append(dateutil_parse(proposition['deadline']).strftime('%Y-%m-%d'))
```
currycoder: Could you help me understand when this default proposition formatter might be used as opposed to `self.proposition_formatter`?  e.g. what business cases are each used for..?

elcct: The default formatter is being used in the CSV report that is generated by the Celery task. I understand it is weird, but that was the requirement for the propositions to look in the spreadsheet cell. Once the new endpoint is in use, we will be able to remove that.

datahub/dataset/investment_project/pagination.py
elcct: I agree that this is not ideal and thank you for suggestions. I'll have a look how to improve it.

datahub/investment/project/report/spi.py
currycoder: Ok I understand - thanks.

I think this leads me to my next thought. It feels a little bit like some of the complexity here is because of this situation: We have this existing unit of code whose contract is to generate CSV formatted SPI reports by interrogating the DB and formatting results.  Some of the complexity is from transforming this CSV output in to something that can work on an API endpoint.  Could we instead leave the legacy CSV implementation as it is and develop some new units which 1) grab the data 2) format it, and 3) expose it on an endpoint for DW

I realise that what I'm saying represents a big refactor and the current implementation works etc.  I'm just wondering whether we will end up in a situation where we are stuck with this CSV formatting code that won't ultimately make sense when we are completely using an API?

elcct: The end goal is to remove the CSV generating code and associated formatting, so it could greatly reduce the complexity. The reason I re-used the existing code is because I wanted to reduce possibility of having its output different from the (to be) legacy report, as it is important that the output is exactly the same (minus a couple of fields that are no longer needed and the formatting of propositions). So I'd rather not create a second implementation. I'd prefer to follow this PR with another that will remove any (to be) legacy code instead.

elcct:
Currently this PR is blocked until https://github.com/uktrade/data-hub-api/pull/2486 is merged, because with the current version of Django that is being used, the `ArrayAgg(filter=...)` doesn't work.

datahub/investment/project/report/spi.py
currycoder: Ok awesome - I understand your rationale here and glad to hear that the CSV generating code will eventually be ripped out.

datahub/core/query_utils.py
```diff
@@ -124,7 +125,12 @@ def get_array_agg_subquery(
     """
     return get_aggregate_subquery(
         model,
-        ArrayAgg(expression_to_aggregate, distinct=distinct, ordering=ordering),
+        ArrayAgg(
+            expression_to_aggregate,
+            distinct=distinct,
+            ordering=ordering,
+            filter=Q(**filter) if filter else None,
```
reupen: I think it would be better to just pass `filter` through (so the caller can pass a Q object directly).

datahub/dataset/investment_project/pagination.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _proposition_formatter(self, propositions):
+        """Returns a list of propositions with selected fields."""
+        return [{
+            'deadline': dateutil_parse(proposition['deadline']).strftime('%Y-%m-%d'),
+            'status': proposition['status'],
+            'modified_on': dateutil_parse(proposition['modified_on']).isoformat()
+            if proposition['status'] != PropositionStatus.ongoing else '',
+            'adviser_id': proposition['adviser_id'],
+        } for proposition in propositions]
```
reupen: Would be good to try and improve the formatting here e.g.:

```python
        return [
            {
                'deadline': dateutil_parse(proposition['deadline']).strftime('%Y-%m-%d'),
                'status': proposition['status'],
                'modified_on':
                    dateutil_parse(proposition['modified_on']).isoformat()
                    if proposition['status'] != PropositionStatus.ongoing else '',
                'adviser_id': proposition['adviser_id'],
            }
            for proposition in propositions
        ]
```

datahub/dataset/investment_project/pagination.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def get_paginated_response(self, data):
+        """Get paginated response."""
+        data = self._get_data(data)
+        results = self._map_results(data)
+        return Response(OrderedDict([
+            ('next', self.get_next_link()),
+            ('previous', self.get_previous_link()),
+            ('results', results),
+        ]))
```
reupen: Should be able to use a dict literal here (instead of `OrderedDict`) since we don't support old versions of Python.

datahub/core/query_utils.py
elcct: Good point, that should be responsibility of the caller. Updated. Thank you!

datahub/dataset/investment_project/pagination.py
elcct: Good point! Updated. Thank you!

datahub/dataset/investment_project/pagination.py
elcct: Yes, that will be an improvement! Updated, thank you!

datahub/dataset/investment_project/pagination.py
elcct: I moved the formatting code into a separate `SPIReportFormatter` class. It looks much cleaner now. I'll have a look into adding more comments now.

elcct:
Since Django 3.0.3 update has now been merged, I have rebased this PR so that aggregate queries can work correctly.

datahub/core/query_utils.py
```diff
@@ -158,7 +173,6 @@ def get_aggregate_subquery(model, expression, join_field_name='pk'):
     ).values(
         '_annotated_value',
     )
-
```
reupen: Is this change intended?

datahub/dataset/investment_project/pagination.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def get_paginated_response(self, data):
+        """Get paginated response."""
+        spi_report = SPIReportFormatter()
+        results = spi_report.format(data)
+        return Response({
+            'next': self.get_next_link(),
+            'previous': self.get_previous_link(),
+            'results': results,
+        })
```
reupen: Minor point, but we should break before and after `{` and `}` if we're following the code style.

datahub/dataset/investment_project/spi.py
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    required_fields_value_mapping = {
+        'Enquiry processed by': lambda adviser: str(adviser),
+        'Project manager assigned by': lambda adviser: str(adviser.id),
+    }
+
+    def __init__(self):
+        """Initialise SPI Report Formatter."""
+        self._SPIReport = SPIReport(proposition_formatter=proposition_formatter)
```
reupen: Maybe it would be simpler to use `proposition_formatter` in this class instead of passing it through to `SPIReport`? (It seems to be adding extra complexity that we might not need.)

datahub/investment/project/report/spi.py
```diff
@@ -259,7 +261,6 @@ def get_spi_data_for_investment_project(self, investment_project):
         data.update(self.get_spi2(investment_project, spi_interaction_dates))
         data.update(self.get_spi3(investment_project))
         data.update(self.get_spi5(investment_project, spi_interaction_dates))
-
```
reupen: Was this change intended?

datahub/investment/project/report/spi.py
```diff
@@ -226,10 +225,13 @@ def get_spi3(self, investment_project):
...
         data = {}
 
-        propositions = self._get_propositions(investment_project, ist_only=True)
-        if propositions:
-            data[self.SPI3] = self._format_propositions(propositions)
+        if not investment_project.spi_propositions:
+            return data
+
+        formatter = self.proposition_formatter \
+            if self.proposition_formatter else self._format_propositions
```
reupen: Minor point – if this remains, would be good to wrap this in `(` and `)` instead of using `\` (to be in line with PEP 8).

datahub/dataset/investment_project/pagination.py
```diff
@@ -0,0 +1,25 @@
...
+from rest_framework.response import Response
+
+from datahub.dataset.core.pagination import DatasetCursorPagination
+from datahub.dataset.investment_project.spi import SPIReportFormatter
+
+
+class InvestmentProjectActivityDatasetViewCursorPagination(DatasetCursorPagination):
+    """Cursor Pagination for SPI Report."""
+
+    ordering = ('id', )
```
reupen: The cursor pagination is really meant to be ordered by e.g. `('created_on', 'pk')`. Is there a reason we didn't stick to that here?

datahub/dataset/investment_project/spi.py
elcct: I think there is no access to raw propositions in the data returned by `SPIReport` , so I would have to parse `;` separated ones back into JSON if I was going to do it in this class. The goal is to remove the old formatting once people have switched to data returned by this endpoint, so that complexity will be removed then.

datahub/dataset/investment_project/pagination.py
elcct: Good point! There is no good reason - I have updated it to order by `created_on` and `pk`.

datahub/investment/project/report/spi.py
elcct: Ah! Updated it now. Thank you!

datahub/investment/project/report/spi.py
elcct: No. I have reversed it now. Thank you!

datahub/dataset/investment_project/pagination.py
elcct: I am too used to this formatting :( Updated it now. Thank you!

datahub/core/query_utils.py
elcct: No! I was too happy with deleting :-) Updated. Thank you.

datahub/dataset/investment_project/spi.py
reupen: :+1:

PR #2506, opened 2020-01-27
Allow editing export countries in Company api

sekharpanja:
### Description of change
This is the second part of the two part PR that allows exposing/editing export countries from the new model in Company API.

For now this is behind feature flag. So if feature flag is OFF, API will work as is updating old fields. And if the feature flag is ON, API will start updating new model instead. In both scenarios, data will be synced across to allow feature flag to be switched ON and OFF when required.

### Checklist


changelog/company/edit_company_export_countries.api.md
```diff
@@ -0,0 +1,3 @@
+`PATCH /v4/company/<pk>`: This will allow export countries to be added/updated/removed into the new `CompanyExportCountry` model, replacing old company export country fields: `export_to_countries` and `future_interest_countries`.
```
alixedi: I don't like this URL @sekharpanja 

the URL suggests that this endpoint allows users to patch the company entity. In reality, it (should) only allow them to patch a very specific subset of properties of the company entity i.e. `export_to_countries` and `future_interest_countries`.

Can we perhaps go PATCH `with /v4/company/<pk>/export-countries`? 

TY 🙏

datahub/company/serializers.py
```diff
@@ -300,47 +307,72 @@ def __init__(self, *args, **kwargs):
...
+            *(export_to or []),
+            *(future_interest or []),
+        }
+
+        if future_interest is not None:
+            self._sync_to_company_export_country_model(
                 company=instance,
                 adviser=adviser,
-                export_countries=future_interest_countries,
+                export_countries=future_interest,
```
reupen: I think `future_interest_countries` was a better name than `future_interest` (same for `export_to_countries` and `export_to`).

datahub/company/serializers.py
```diff
@@ -357,6 +389,61 @@ def _save_to_company_export_country_model(*, company, adviser, export_countries,
                 export_country.modified_by = adviser
                 export_country.save()
 
+    def _update_export_countries_model(self, company, validated_export_countries, adviser):
+        """
+        Adds/updates export countries related to a company within validated_export_countries.
+        And removes existing ones that are not in the list.
+        """
+        if not validated_export_countries:
+            return
```
reupen: If `validated_export_countries` equals `[]` shouldn't we be removing all countries?

datahub/company/serializers.py
```diff
@@ -300,47 +307,72 @@ def __init__(self, *args, **kwargs):
...
+        if export_to is not None:
+            self._sync_to_company_export_country_model(
                 company=instance,
                 adviser=adviser,
-                export_countries=export_to_countries,
+                export_countries=export_to,
                 status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.currently_exporting,
             )
 
+        not_interested_status = CompanyExportCountry.EXPORT_INTEREST_STATUSES.not_interested
```
reupen: I've seen similar things in other PRs – it would be good to (in a separate PR) rename or move `CompanyExportCountry.EXPORT_INTEREST_STATUSES` to make it more concise (e.g. rename to `CompanyExportCountry.STATUSES` or move it off the class).

datahub/company/serializers.py
```diff
@@ -300,47 +307,72 @@ def __init__(self, *args, **kwargs):
...
                 self.fields[field].read_only = True
 
+    @atomic
     def update(self, instance, validated_data):
         """
-        Using writable nested representations to copy export country elements
-        from the `Company` model into the `CompanyExportCountry`.
+        If the feature flag is ON, write data to export country model
+        and sync data back to compay export country fields.
+        When the feature flag is OFF, write data to company export coutry fields
```
currycoder: coutry -> country :)

datahub/company/serializers.py
```diff
@@ -300,47 +307,72 @@ def __init__(self, *args, **kwargs):
             for field in self.Meta.dnb_read_only_fields:
                 self.fields[field].read_only = True
 
+    @atomic
     def update(self, instance, validated_data):
         """
-        Using writable nested representations to copy export country elements
-        from the `Company` model into the `CompanyExportCountry`.
+        If the feature flag is ON, write data to export country model
+        and sync data back to compay export country fields.
```
currycoder: compay -> company :)

datahub/company/serializers.py
```diff
@@ -300,47 +307,72 @@ def __init__(self, *args, **kwargs):
             for field in self.Meta.dnb_read_only_fields:
                 self.fields[field].read_only = True
 
+    @atomic
     def update(self, instance, validated_data):
         """
-        Using writable nested representations to copy export country elements
-        from the `Company` model into the `CompanyExportCountry`.
+        If the feature flag is ON, write data to export country model
```
currycoder: My read of this is that regardless of whether the feature flag is on or off, the data gets written to the new model AND the old model.  But it seems like this is not the case when looking at the code below, which appears to only update the new fields when the feature flag is active.  Can you help clarify?

datahub/company/serializers.py
```diff
@@ -357,6 +389,61 @@ def _save_to_company_export_country_model(*, company, adviser, export_countries,
...
+            country = meta_models.Country.objects.get(id=item['country'].id)
+            status = item['status']
+            company.add_export_country(
+                country=country,
+                status=status,
+                record_date=company.modified_on,
+                adviser=adviser,
+            )
+
+        existing_countries = [item.country for item in CompanyExportCountry.objects.all()]
```
currycoder: Shouldn't the query here be bounded to the company in question, rather than **all** CompanyExportCountry records?

datahub/company/serializers.py
```diff
@@ -357,6 +389,61 @@ def _save_to_company_export_country_model(*, company, adviser, export_countries,
...
+        new_countries = [
+            meta_models.Country.objects.get(id=item['country'].id)
```
currycoder: Is there a way to do this as a single query rather than n-queries where n is the length of validated_export_countries?

changelog/company/edit_company_export_countries.api.md
sekharpanja: Added a new endpoint as suggested albeit with a slightly different name.

datahub/company/serializers.py
sekharpanja: This works already, but now made pop's default to be [] instead of None. Which should clear the confusion.

datahub/company/serializers.py
sekharpanja: If the flag is OFF, data gets written to old fields and synced to new model. When flag is ON, data gets written to the new model and syncs back to old fields. That is how it should work. We can go through it together if it is not clearer. :+1:

datahub/company/serializers.py
sekharpanja: Absolutely. I've fixed it.

datahub/company/serializers.py
sekharpanja: Thanks for pointing it out, I've updated it to use only country ids instead. Database queries is not required at all now.

datahub/company/constants.py
```diff
@@ -6,6 +6,7 @@
 
 NOTIFY_DNB_INVESTIGATION_FEATURE_FLAG = 'notify-dnb-investigations'
 AUTOMATIC_COMPANY_ARCHIVE_FEATURE_FLAG = 'automatic-company-archive'
+EXPORT_COUNTRIES_SWITCH_FEATURE_FLAG = 'export-countries-switch'
```
alixedi: The general scheme we follow for naming feature flags is: `<shiny_new_thing>_FEATURE_FLAG`.

I might be wrong but it seems that over here: `<shiny_new_thing> = EXPORT_COUNTRIES` and not `EXPORT_COUNTRIES_SWITCH`.

datahub/company/serializers.py
```diff
@@ -239,6 +245,9 @@ class CompanySerializer(PermittedFieldsModelSerializer):
         'uk_establishment_not_in_uk': gettext_lazy(
             'A UK establishment (branch of non-UK company) must be in the UK.',
         ),
+        'invalid_field': gettext_lazy(
```
alixedi: This error message, if not used elsewhere, is too generic. All of the error messages in `default_error_message` could be `invalid_field`. Can we please be more specific?

datahub/company/serializers.py
```diff
@@ -420,6 +431,24 @@ def validate_global_headquarters(self, global_headquarters):
 
         return global_headquarters
 
+    def validate_export_to_countries(self, export_to_countries):
+        """This field is invalid when feature flag is active"""
+        if is_feature_flag_active(EXPORT_COUNTRIES_SWITCH_FEATURE_FLAG):
```
alixedi: I must say, It is a bit unusual to check for a feature flag inside the validation method of a Serializer - which, among other things, is too deep in the call stack. Usually, we check for the feature flag in the view i.e. the first method in the call stack and select an appropriate code path. In this case, it would mean subclassing `CompanySerializer`, putting the modifications required for the feature flag in the subclass and choosing it for when the feature flag is enabled in the view.

This, in turn means that the code path for when a feature flag is enabled/disabled are appropriately isolated and therefore when you turn the feature flag off, there is a guarantee that the default behaviour would revert to standard.

Having said this, regrettably, it seems I am asking for a change that may result in more work at this stage so I will leave this with you.

datahub/company/serializers.py
currycoder: Cool - I've traced the logic through on the helpers now and this makes sense thanks.  It was initially unclear to me as either side of the if statement for the feature flag appeared to be mutually exclusive.

datahub/company/test/test_company_views.py
```diff
@@ -1641,3 +1677,518 @@ def test_adding_overlapping_countries_in_company_export_to_country_model(self):
...
             list(actual_future_interest_countries)[0].country,
         ] == list(set(new_future_interest_countries) - set(new_export_to_countries))
+
+    def test_edit_company_fields_check_not_interested_is_intact(self):
+        """
+        Check when in case feature flag is switched OFF
+        and updating old export country fields will not wipe off
+        not_interested countries in `CompanyExportCountry` model.
+        """
+        initial_countries = list(CountryModel.objects.order_by('?')[:5])
```
currycoder: It might be better to order this consistently rather than randomly, so that if there are test failures they can be reproduced deterministically.

datahub/company/test/test_company_views.py
```diff
@@ -1641,3 +1677,518 @@ def test_adding_overlapping_countries_in_company_export_to_country_model(self):
...
+            company=company,
+            status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.future_interest,
+        ).save()
+        CompanyExportCountry(
+            country=initial_countries[2],
+            company=company,
+            status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.not_interested,
+        ).save()
+
+        new_countries = list(CountryModel.objects.order_by('?')[:5])
```
currycoder: Same as above!

datahub/company/test/test_company_views.py
```diff
@@ -1641,3 +1677,518 @@ def test_adding_overlapping_countries_in_company_export_to_country_model(self):
...
+        CompanyExportCountry(
+            country=initial_countries[1],
+            company=company,
+            status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.future_interest,
+        ).save()
+        CompanyExportCountry(
+            country=initial_countries[2],
+            company=company,
+            status=CompanyExportCountry.EXPORT_INTEREST_STATUSES.not_interested,
+        ).save()
```
alixedi: It seems like we should consider parametrising this test.

datahub/company/test/test_company_views.py
```diff
@@ -1641,3 +1677,518 @@ def test_adding_overlapping_countries_in_company_export_to_country_model(self):
...
+                            ],
+                        },
+                    ],
+                },
+            ),
+        ),
+    )
+    def test_validation_error(self, flag, data, expected_error):
+        """Test validation scenarios."""
+        FeatureFlagFactory(code=EXPORT_COUNTRIES_SWITCH_FEATURE_FLAG, is_active=flag)
```
alixedi: Elsewhere in the codebase, we like to create a fixture for feature flags.

datahub/company/test/test_company_views.py
```diff
@@ -1641,3 +1677,518 @@ def test_adding_overlapping_countries_in_company_export_to_country_model(self):
...
+
+        assert current_countries_request == current_countries_response
+        assert future_countries_request == future_countries_response
+
+    def test_update_company_export_countries_with_pre_existing_company_fields_sync(self):
+        """
+        Test sync when company export_countries update, of a company with
+        currently_exporting_to and future_interest_countries preset.
+        """
+        initial_countries = list(CountryModel.objects.order_by('?')[:5])
```
currycoder: Same as above

datahub/company/test/test_company_views.py
```diff
@@ -1641,3 +1677,518 @@ def test_adding_overlapping_countries_in_company_export_to_country_model(self):
...
+
+        url = reverse('api-v4:company:update-export-details', kwargs={'pk': company.pk})
+        response = self.api_client.put(url, data=input_export_countries)
+        assert response.status_code == status.HTTP_200_OK
+
+        company.refresh_from_db()
+        company_export_countries = {
+            'export_countries': [],
+        }
+        assert company_export_countries == input_export_countries
```
currycoder: I don't think this tests the value in the DB - just two dicts created in memory!

datahub/company/constants.py
sekharpanja: We are switching from old fields that are within Company model to the new CompanyExportCountry model. Hence then name.

datahub/company/serializers.py
sekharpanja: I agree. I would have made an effort if this feature is long lasting. But we will be deprecating old export country fields within Company as a follow up task in the next sprint or so. This feature flag will disappear along with it. So hope its okay to keep it as it is until then.

datahub/company/test/test_company_views.py
sekharpanja: I am not quite sure if I understood it, please can you elaborate. :pray:

datahub/company/test/test_company_views.py
sekharpanja: I've simplified this test for clarity. It tests a specific use case of feature flag switch back.

datahub/company/constants.py
alixedi: @sekharpanja most feature flags are a `switch` from old fields (or business logic) to new ones. 

I think there is value in sticking to convention in this case 🙏

datahub/company/views.py
```diff
@@ -146,6 +147,17 @@ def remove_account_manager(self, request, *args, **kwargs):
...
+    @action(methods=['put'], detail=True)
+    def update_export_details(self, request, *args, **kwargs):
+        """
+        Update export related information for the company.
+        """
+        instance = self.get_object()
+        serializer = UpdateExportDetailsSerializer(instance=instance, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save(request.user)
+        return Response(None, status=status.HTTP_200_OK)
```
reupen: I was under the impression that we were switching to PUT + a list of countries, rather than PUT (or PATCH) and an object with a single property.

I think we should do the former (i.e. `PUT /v4/company/<uuid:pk>/export-country`) as it is self-describing and it would reduce complexity.

datahub/company/constants.py
elcct: Feature flag is boolean so it is a switch.

datahub/company/test/test_company_views.py
elcct: For example:

```
@pytest.fixture()
def export_countries_feature_flag():
    """Creates the export countries feature flag."""
    yield FeatureFlagFactory(code=EXPORT_COUNTRIES_FEATURE_FLAG)
```

PR #2511, opened 2020-01-27
Feature/check dnb currency

alixedi:
### Description of change

The `format_dnb_company` now includes a check for `annual_sales_currency`.

  If `annual_sales_currency` is not US Dollars, we do not propagate `annual_sales` or `is_annual_sales_estimated` fields downstream.

  All D&B records that we have encountered until now have `annual_sales` in US dollars but we would like to monitor this behavior and not ingest bad data in case there is an exception.

### Checklist

### DIFF
--- a/datahub/dnb_api/utils.py
+++ b/datahub/dnb_api/utils.py
@@ -180,10 +180,18 @@ def format_dnb_company(dnb_company):
     domain = dnb_company.get('domain')
     company_website = f'http://{domain}' if domain else ''
 
+    duns_number = dnb_company.get('duns_number')
+    is_turnover_in_usd = dnb_company.get('annual_sales_currency') == 'USD'
+
+    if not is_turnover_in_usd:
+        logger.error(f'D&B did not have USD turnover for: {duns_number}')
+        dnb_company.pop('annual_sales', None)
+        dnb_company.pop('is_annual_sales_estimated', None)
+
     return {
         'name': dnb_company.get('primary_name'),
         'trading_names': dnb_company.get('trading_names'),
-        'duns_number': dnb_company.get('duns_number'),
+        'duns_number': duns_number,
         'address': extract_address_from_dnb_company(dnb_company, 'address'),
         'registered_address': extract_address_from_dnb_company(
             dnb_company,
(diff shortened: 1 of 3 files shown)


datahub/dnb_api/utils.py
```diff
@@ -180,10 +180,18 @@ def format_dnb_company(dnb_company):
     domain = dnb_company.get('domain')
     company_website = f'http://{domain}' if domain else ''
 
+    duns_number = dnb_company.get('duns_number')
+    is_turnover_in_usd = dnb_company.get('annual_sales_currency') == 'USD'
+
+    if not is_turnover_in_usd:
+        logger.error(f'D&B did not have USD turnover for: {duns_number}')
+        dnb_company.pop('annual_sales')
+        dnb_company.pop('is_annual_sales_estimated')
```
reupen: Not too clear why `dict.pop()` is being used here – did you perhaps mean to [provide a default to avoid an exception when the key is missing](https://docs.python.org/3/library/stdtypes.html#dict.pop)? (e.g. `dnb_company.pop('annual_sales', None)`.)

datahub/dnb_api/utils.py
```diff
@@ -180,10 +180,18 @@ def format_dnb_company(dnb_company):
     domain = dnb_company.get('domain')
     company_website = f'http://{domain}' if domain else ''
 
+    duns_number = dnb_company.get('duns_number')
+    is_turnover_in_usd = dnb_company.get('annual_sales_currency') == 'USD'
+
+    if not is_turnover_in_usd:
+        logger.error(f'D&B did not have USD turnover for: {duns_number}')
```
currycoder: It might be good to provide some additional context for this error in `extra`.  I'm thinking it would be good to see the raw result for this company that triggered the error.  Though you have more knowledge about whether that fits within sentry's truncation logic :)

alixedi: I do not quite see how the raw JSON would benefit us here. Help me out 🙏

currycoder: Well maybe not the whole lot, but my thought is that this might expose what currency the company does have (if any).  We can of course look this up after the fact with a call to dnb-service, so perhaps this is redundant.

PR #2512, opened 2020-01-27
Add dnb company match utility function

currycoder:
### Description of change

A utility function `datahub.dnb_api.match.match_company_with_dnb` was added which
provides a mechanism for matching a Data Hub company record with a D&B company
record. The utility function saves the pertinent D&B data to the Data Hub record.

This will allow us to firstly provide an admin mechanism for quickly matching
companies, followed by an API endpoint.

### Checklist

### DIFF
--- a/changelog/company/dnb-match-logic.feature.md
+++ b/changelog/company/dnb-match-logic.feature.md
@@ -1,4 +1,4 @@
-A utility function `datahub.dnb_api.match.link_company_with_dnb` was added which
+A utility function `datahub.dnb_api.link_company.link_company_with_dnb` was added which
 provides a mechanism for linking a Data Hub company record with a D&B company
 record. The utility function saves the pertinent D&B data to the Data Hub record.
 
--- a/changelog/company/dnb-match-logic.feature.md
+++ b/changelog/company/dnb-match-logic.feature.md
@@ -1,0 +1,6 @@
+A utility function `datahub.dnb_api.match.link_company_with_dnb` was added which
+provides a mechanism for linking a Data Hub company record with a D&B company
+record. The utility function saves the pertinent D&B data to the Data Hub record.
+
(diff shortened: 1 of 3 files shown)


datahub/dnb_api/match.py
```diff
@@ -0,0 +1,37 @@
...
+    Given a Data Hub company ID and a duns number, save the company with this
+    duns number and update it's record from D&B.
+    """
+    company = Company.objects.get(id=dh_company_id)
+    if company.duns_number:
+        raise CompanyAlreadyDNBMatchedException(
+            f'Company {company.id} is already matched with duns number {company.duns_number}',
+        )
+    company.duns_number = duns_number
+    company.save()
```
reupen: Small observation – if we're saving the entire company object here, we should probably be setting `modified_by` as well.

datahub/dnb_api/match.py
```diff
@@ -0,0 +1,37 @@
...
+
+class CompanyAlreadyDNBMatchedException(Exception):
+    """
+    An exception to raise when a Data Hub company has already been matched with
+    a DNB company record.
+    """
+
+
+@transaction.atomic
+def match_company_with_dnb(
```
reupen: I think 'match' might imply that it tries to find a corresponding D&B record... Maybe something like `link_company_with_dnb_record` would be clearer?

currycoder: I think this is a good observation - thanks.  Especially since we have already used "match" to mean a certain thing in the backend when it comes to D&B integration.  @alixedi Are you happy to switch our verb for this to "link"?

elcct: I prefer `link` as well.

alixedi: Lets switch 👍

PR #2525, opened 2020-01-31
Refactor dnb update admin tool to break out reusable functionality

currycoder:
### Description of change

The `update_from_dnb` admin tool was refactored to break out some useful utilities
for use by other modules.  This will aid in the new "Link Company with D&B" tool
which will follow shortly.

This is the first PR to lay the foundations for the "Link Company with D&B" tool.  More will follow that build on this.

### Checklist

### DIFF
--- a/datahub/company/admin/update_from_dnb.py
+++ b/datahub/company/admin/update_from_dnb.py
@@ -1,0 +1,95 @@
+import logging
+
+from django.contrib.admin.templatetags.admin_urls import admin_urlname
+from django.core.exceptions import PermissionDenied, SuspiciousOperation
+from django.http import HttpResponseRedirect
+from django.template.response import TemplateResponse
+from django.urls import reverse
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext_lazy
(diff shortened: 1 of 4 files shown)


datahub/company/admin/update_from_dnb.py
```diff
@@ -0,0 +1,95 @@
...
+    except (
+        DNBServiceError,
+        DNBServiceConnectionError,
+        DNBServiceTimeoutError,
+        DNBServiceInvalidResponse,
+    ):
+        message = 'Something went wrong in an upstream service.'
+        raise AdminException(message, company_change_page)
+
+    except DNBServiceInvalidRequest:
```
elcct: It seems a bit strange that invalid request means the match was not found. Could other thing happen that is not "match not found" under this exception?

currycoder: The API call that we are making is essentially to just GET a company that has a particular duns number.  Duns number is the only input that we have for this API, so we are safe to respond with this in the case of a 400 response.  It just infers that either the duns number was invalid, or the company with that duns number could not be found.

PR #2535, opened 2020-02-03
Link company with dnb admin step 1

currycoder:
### Description of change

A view was added to allow admin users to select company ID/duns number for D&B company linking. This forms the first step for the admin D&B link tool. As the tool is not yet complete, it is not linked to from elsewhere in the admin yet.

### Checklist

### DIFF
--- a/datahub/company/admin/dnb_link/review_changes.py
+++ b/datahub/company/admin/dnb_link/review_changes.py
@@ -1,0 +1,15 @@
+from django.contrib.admin.templatetags.admin_urls import admin_urlname
+from django.http import HttpResponseRedirect
+from django.urls import reverse
+
+
+def dnb_link_review_changes(model_admin, request):
+    """
+    View to allow users to review changes that would be applied to a record before linking it.
+    POSTs make the link and redirect the user to view the updated record.
+    """
+    # TODO: Remove this temporary redirect
+    company_change_url = reverse(
+        admin_urlname(model_admin.model._meta, 'changelist'),
+    )
+    return HttpResponseRedirect(company_change_url)
--- a/datahub/company/admin/dnb_link/select_ids.py
+++ b/datahub/company/admin/dnb_link/select_ids.py
@@ -1,5 +1,91 @@
+from django import forms
+from django.contrib.admin.templatetags.admin_urls import admin_urlname
+from django.core.exceptions import PermissionDenied, ValidationError
+from django.http import HttpResponseRedirect
+from django.template.response import TemplateResponse
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext_lazy
+from django.views.decorators.csrf import csrf_protect
+
+from datahub.company.models import Company
+from datahub.core.admin import RawIdWidget
+from datahub.core.utils import reverse_with_query_string
+
+
+class SelectIdsToLinkForm(forms.Form):
+    """
+    Form used for selecting Data Hub company id and D&B duns number for linking.
+    """
+
+    COMPANY_ALREADY_DNB_LINKED = gettext_lazy(
+        'This company has already been linked with a D&B company.',
+    )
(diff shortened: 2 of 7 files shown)


datahub/company/test/admin/dnb_link/test_select_ids.py
```diff
@@ -0,0 +1,125 @@
...
+            (
+                {
+                    'company': '731bdcc1-f685-4c8e-bd66-b356b2c16995',
+                    'duns_number': '123456789',  # duns_number already linked
+                },
+                'This duns number has already been linked with a Data Hub company.',
+            ),
+        ),
+    )
+    def test_invalid_input_responds_error(self, company_link_data, data, expected_error):
```
reupen: I think the tests could be simplified and made more explicit: it looks like the main thing to parametrise is the existing DUNS number (rather than the company ID). So we could e.g. make the parameter 'existing_duns_number' and instantiate a `CompanyFactory` in the test body.

datahub/company/admin/dnb_link/select_ids.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@method_decorator(csrf_protect)
+def dnb_link_select_ids(model_admin, request):
+    """
+    View to select Data Hub company id and D&B duns number for linking records.
+    Valid POSTs redirect to the change review page.
+    """
+    if not model_admin.has_change_permission(request):
+        raise PermissionDenied
```
reupen: Minor suggestion:
```suggestion
        raise PermissionDenied()
```

datahub/company/test/admin/dnb_link/test_select_ids.py
currycoder: That's a fair criticism, thanks.  I've gotten rid of the data fixture and broken down this test case a little to make things easier to follow.

datahub/company/admin/dnb_link/review_changes.py
```diff
@@ -0,0 +1,5 @@
+def dnb_link_review_changes(model_admin, request):
+    """
+    View to allow users to review changes that would be applied to a record before linking it.
+    POSTs make the link and redirect the user to view the updated record.
+    """
```
reupen: I think even as a stub this should still return an empty response (it 500s at the moment...)

currycoder: Good idea.

datahub/company/admin/dnb_link/select_ids.py
```diff
@@ -0,0 +1,91 @@
...
+    DUNS_NUMBER_ALREADY_LINKED = gettext_lazy(
+        'This duns number has already been linked with a Data Hub company.',
+    )
+
+    company = forms.ModelChoiceField(
+        Company.objects.all(),
+        widget=RawIdWidget(Company),
+        label='Data Hub Company',
+    )
+    duns_number = forms.CharField(min_length=9, max_length=9)
```
alixedi: IDK if we are doing this consistently across the codebase but would be nice if we can make the validation consistent with: 

https://github.com/uktrade/data-hub-api/blob/616ac79f89613e3036cfba2db20bb4c7f841f79c/datahub/dnb_api/serializers.py#L71-L75

currycoder: Agreed.  I'll do this as a follow-up PR once the bulk of the feature is in, as discussed IRL.

currycoder:
> Is there a reason for not going for a link when viewing a company (similar to 'Update from D&B', with the company ID in the path) rather than making the company a field in the form? Will leave it with you to consider...

This is a good question.  Some context: The frontend linking mechanism generates a zendesk ticket which essentially has a data hub company ID and a duns number within it's body.  This tool aims to solve the problem where an admin user receiving this ticket needs to link the company - the expectation is that the values will be copy/pasted in to the form.

PR #2537, opened 2020-02-03
Add list company referrals endpoint

elcct:
### Description of change

A new endpoint, `GET /v4/company-referral` has been added. It lists referral sent or received by the adviser. 
Refer to the API documentation for the schema.

### Checklist

* [x] Do any added or updated endpoints appear in the API documentation? See [docs/Maintaining the API documentation.md](https://github.com/uktrade/data-hub-api/blob/develop/docs/Maintaining&#32;the&#32;API&#32;documentation.md) for more details

### DIFF
--- a/datahub/company_referral/test/test_views.py
+++ b/datahub/company_referral/test/test_views.py
@@ -23,6 +23,107 @@ def _item_url(pk):
     return reverse('api-v4:company-referral:item', kwargs={'pk': pk})
 
 
+class TestListCompanyListsView(APITestMixin):
+    """Tests for listing user's referrals."""
+
+    def test_returns_401_if_unauthenticated(self, api_client):
+        """Test that a 401 is returned if the user is unauthenticated."""
+        response = api_client.get(collection_url)
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    @pytest.mark.parametrize(
+        'permission_codenames,expected_status',
+        (
+            ([], status.HTTP_403_FORBIDDEN),
+            (['view_companyreferral'], status.HTTP_200_OK),
+        ),
+    )
+    def test_permission_checking(self, permission_codenames, expected_status, api_client):
+        """Test that the expected status is returned for various user permissions."""
+        user = create_test_user(permission_codenames=permission_codenames, dit_team=None)
+        api_client = self.create_api_client(user=user)
+        response = api_client.get(collection_url)
+        assert response.status_code == expected_status
+
+    def test_returns_empty_list_when_no_lists(self):
+        """Test that no results are returned when the user has no referrals."""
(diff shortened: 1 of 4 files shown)


datahub/company_referral/views.py
```diff
@@ -12,6 +14,12 @@ class CompanyReferralViewSet(CoreViewSet):
...
-        'recipient__team',
+        'created_by__dit_team',
+        'recipient__dit_team',
     )
+
+    def get_queryset(self):
+        """Get a queryset filtered to the authenticated user's sent and received referrals."""
+        return super().get_queryset().filter(
+            Q(created_by=self.request.user) | Q(recipient=self.request.user),
+        )
```
reupen: I think we need to be careful with doing the filtering in this was, since it will affect getting a single object as well (when it shouldn't).

Since the filtering is not permissions-related, maybe we could use a query parameter like the list interactions endpoint (would also mean we could build different views in future if needed)? Or if we stick with this approach, then we should restrict the filtering to listing only (e.g. check `self.action`).

datahub/company_referral/test/test_views.py
```diff
@@ -11,13 +11,115 @@
...
+        response = self.api_client.get(collection_url)
+
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['results'] == []
+
+    def test_doesnt_return_other_users_referrals(self):
+        """Test that other users' referrals are not returned."""
+        # Create some referrals sent or received by other users
+        CompanyReferralFactory.create_batch(5)
```
reupen: I think 1 or 2 objects should be enough (as 2 should not be any different from 5).
```suggestion
        CompanyReferralFactory.create_batch(2)
```

datahub/company_referral/test/test_views.py
```diff
@@ -11,13 +11,115 @@
...
+        recipient2 = CompanyReferralFactory(recipient=self.user)
+
+        response = self.api_client.get(collection_url)
+        assert response.status_code == status.HTTP_200_OK
+
+        results = response.json()['results']
+        assert len(results) == 3
+
+        expected_referrals = {str(sender.pk), str(recipient1.pk), str(recipient2.pk)}
+        assert {result['id'] for result in results} == expected_referrals
```
currycoder: This is neat

datahub/company_referral/views.py
elcct: This is a very good point! I have restricted the filtering to the `list` action now. I think we could change it later on if we are going to need to filter by other fields.

datahub/company_referral/test/test_views.py
elcct: Good point! Yes, having just two objects will be sufficient. Thank you!

PR #2541, opened 2020-02-03
Link company with dnb admin step 2

currycoder:
### Description of change
A view was added to allow admin users to review and confirm changes that will be made for D&B company linking.

I'd recommend reviewing the commits in order.

### Checklist

### DIFF
--- a/datahub/company/admin/dnb_link/review_changes.py
+++ b/datahub/company/admin/dnb_link/review_changes.py
@@ -1,15 +1,122 @@
+from django.contrib import messages
 from django.contrib.admin.templatetags.admin_urls import admin_urlname
+from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseRedirect
+from django.template.response import TemplateResponse
 from django.urls import reverse
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext_lazy
+from django.views.decorators.csrf import csrf_protect
+from rest_framework import serializers
+
+from datahub.company.admin.dnb_link.forms import SelectIdsToLinkForm
+from datahub.company.admin.utils import (
+    AdminException,
+    format_company_diff,
+    redirect_with_message,
+)
+from datahub.dnb_api.link_company import link_company_with_dnb
+from datahub.dnb_api.utils import (
+    DNBServiceConnectionError,
+    DNBServiceError,
+    DNBServiceInvalidRequest,
+    DNBServiceInvalidResponse,
+    DNBServiceTimeoutError,
+    get_company,
+)
 
 
+def _build_error_message(all_errors):
+    error_message = ''
+    for field, errors in all_errors.items():
+        concatenated_errors = ', '.join(errors)
+        error_message += f'{field}: {concatenated_errors} '
+    return error_message
+
+
+def _link_company_with_dnb(dh_company_id, duns_number, user, error_url):
+    # We don't need to catch CompanyAlreadyDNBLinkedException as our form will
+    # do this validation for us
+    try:
+        link_company_with_dnb(dh_company_id, duns_number, user)
+    except serializers.ValidationError:
+        message = 'Data from D&B did not pass the Data Hub validation checks.'
+        raise AdminException(message, error_url)
+    except (
+        DNBServiceError,
+        DNBServiceConnectionError,
+        DNBServiceTimeoutError,
+        DNBServiceInvalidResponse,
+    ):
+        message = 'Something went wrong in an upstream service.'
+        raise AdminException(message, error_url)
+    except DNBServiceInvalidRequest:
+        message = 'No matching company found in D&B database.'
+        raise AdminException(message, error_url)
+
+
+def _get_company(duns_number, error_url):
+    try:
+        return get_company(duns_number)
+
+    except (
+        DNBServiceError,
+        DNBServiceConnectionError,
+        DNBServiceTimeoutError,
+        DNBServiceInvalidResponse,
+    ):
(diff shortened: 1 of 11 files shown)


datahub/dnb_api/tasks.py
```diff
@@ -25,17 +25,26 @@
...
     except DNBServiceError as exc:
-        if is_server_error(exc.status_code):
+        if is_server_error(exc.status_code) and retry_failures:
             raise task.retry(exc=exc, countdown=60)
         raise
     except (DNBServiceConnectionError, DNBServiceTimeoutError) as exc:
-        raise task.retry(exc=exc, countdown=60)
+        if retry_failures:
+            raise task.retry(exc=exc, countdown=60)
+        else:
```
reupen: Should be able to remove this `else:`.

datahub/company/admin/dnb_link/review_changes.py
```diff
@@ -1,15 +1,122 @@
...
+        link_company_with_dnb(dh_company_id, duns_number, user)
+    except serializers.ValidationError:
+        message = 'Data from D&B did not pass the Data Hub validation checks.'
+        raise AdminException(message, error_url)
+    except (
+        DNBServiceError,
+        DNBServiceConnectionError,
+        DNBServiceTimeoutError,
+        DNBServiceInvalidResponse,
+    ):
```
reupen: I think we could use an exception hierarchy to avoid having to handle four exceptions here (but might be something for another PR).

datahub/company/admin/dnb_link/review_changes.py
```diff
@@ -1,15 +1,122 @@
...
     )
-    return HttpResponseRedirect(company_change_url)
+
+    form = SelectIdsToLinkForm(data=request.GET)
+    if not form.is_valid():
+        message = _build_error_message(form.errors)
+        raise AdminException(message, company_list_page)
+
+    dh_company = form.cleaned_data['company']
+    duns_number = request.GET.get('duns_number', form.cleaned_data['duns_number'])
```
alixedi: I presume we are doing this as opposed to:

```
duns_number = form.cleaned_data['duns_number']
```

to enable a direct link to the "review changes" page? If so, it would be nice to include a comment here.

datahub/company/admin/dnb_link/review_changes.py
```diff
@@ -1,15 +1,122 @@
...
+    get_company,
+)
 
+
+def _build_error_message(all_errors):
+    error_message = ''
+    for field, errors in all_errors.items():
+        concatenated_errors = ', '.join(errors)
+        error_message += f'{field}: {concatenated_errors} '
+    return error_message
```
reupen: It might be simpler to flatten the list of errors completely e.g.:

```python

messages = [
    f'{field}: {error}'
    for field, errors in all_errors.items()
    for error in errors
]

return ', '.join(messages)  # or maybe newline delimited
```

But, I also think there might be too much display formatting here. I think it would be even better to be able to attach a list of error messages to the exception and have the exception handler deal with them (for example, it might call `messages.add_message()` multiple times).

datahub/company/admin/dnb_link/review_changes.py
reupen: Yes, it would be good to understand this, as `request.GET` is fed into the form.

datahub/company/admin/dnb_link/review_changes.py
```diff
@@ -1,15 +1,122 @@
...
+@method_decorator(csrf_protect)
 def dnb_link_review_changes(model_admin, request):
     """
     View to allow users to review changes that would be applied to a record before linking it.
     POSTs make the link and redirect the user to view the updated record.
     """
-    # TODO: Remove this temporary redirect
-    company_change_url = reverse(
+    if not model_admin.has_change_permission(request):
+        raise PermissionDenied
```
reupen:
```suggestion
        raise PermissionDenied()
```

datahub/company/admin/dnb_link/review_changes.py
currycoder: My bad!  This is a leftover after refactoring from a point where this view didn't use a form - Ali is right, it should be:
`duns_number = form.cleaned_data['duns_number']`

datahub/company/admin/dnb_link/review_changes.py
currycoder: By exception hierarchy, do you mean ensuring that these all inherit from a common base class?

datahub/company/admin/dnb_link/review_changes.py
currycoder: I'll have a think about this.  Could this perhaps wait for a future cleanup PR?

currycoder: I do like the idea of calling `messages.add_message()` multiple times.

datahub/company/admin/dnb_link/review_changes.py
reupen: Yep, that's what I meant.

datahub/company/admin/dnb_link/review_changes.py
reupen: > I'll have a think about this. Could this perhaps wait for a future cleanup PR?
Yep :+1:

datahub/company/admin/dnb_link/review_changes.py
currycoder: Ok neat - I've added this to a cleanup ticket which I'll get going with shortly after the feature gets released.

PR #2575, opened 2020-02-10
Add company referral activity stream endpoint.

elcct:
### Description of change

This adds `/v3/activity-stream/company-referral` endpoint that lists company referrals. This endpoint will be consumed by the activity stream.

Company referral sender is an `actor`, recipient is a `target` and if referral is closed or completed, the relevant person will be appended to `attributedTo` list.

### Checklist

### DIFF
--- a/datahub/activity_stream/company_referral/serializers.py
+++ b/datahub/activity_stream/company_referral/serializers.py
@@ -1,0 +1,57 @@
+from datahub.activity_stream.serializers import ActivitySerializer
+from datahub.company_referral.models import CompanyReferral
+
+
+class CompanyReferralActivitySerializer(ActivitySerializer):
+    """Company Referral serialiser for activity stream."""
+
+    class Meta:
+        model = CompanyReferral
+
+    def _get_adviser_with_team_and_role(self, adviser, role):
+        adviser = self._get_adviser_with_team(adviser, adviser.dit_team)
+        adviser['dit:DataHubCompanyReferral:role'] = role
+        return adviser
+
+    def to_representation(self, instance):
+        """
+        Serialize the interaction as per Activity Stream spec:
+        https://www.w3.org/TR/activitystreams-core/
+        """
+        company_referral_id = f'dit:DataHubCompanyReferral:{instance.pk}'
+        company_referral = {
+            'id': f'{company_referral_id}:Announce',
+            'type': 'Announce',
+            'published': instance.modified_on,
+            'generator': self._get_generator(),
+            'object': {
+                'id': company_referral_id,
+                'startTime': instance.created_on,
+                'dit:subject': instance.subject,
+                'dit:status': str(instance.status),
+                'attributedTo': [
+                    self._get_company(instance.company),
+                    self._get_adviser_with_team_and_role(instance.created_by, 'sender'),
+                    self._get_adviser_with_team_and_role(instance.recipient, 'recipient'),
+                ],
+                'url': instance.get_absolute_url(),
+            },
+        }
+
+        if instance.completed_by:
+            company_referral['object']['dit:completedOn'] = instance.completed_on
+            company_referral['object']['attributedTo'].append(
+                self._get_adviser_with_team_and_role(instance.completed_by, 'completer'),
+            )
+        if instance.closed_by:
+            company_referral['object']['dit:closedOn'] = instance.closed_on
+            company_referral['object']['attributedTo'].append(
+                self._get_adviser_with_team_and_role(instance.closed_by, 'closer'),
+            )
+
+        if instance.contact:
+            company_referral['object']['attributedTo'].append(
+                self._get_contact(instance.contact),
+            )
+
+        return company_referral
--- a/datahub/activity_stream/company_referral/views.py
+++ b/datahub/activity_stream/company_referral/views.py
@@ -1,0 +1,33 @@
+from datahub.activity_stream.company_referral.serializers import CompanyReferralActivitySerializer
+from datahub.activity_stream.pagination import ActivityCursorPagination
+from datahub.activity_stream.views import ActivityViewSet
+from datahub.company_referral.models import CompanyReferral
+
+
+class CompanyReferralCursorPagination(ActivityCursorPagination):
+    """
+    Cursor pagination for Company Referral.
+
+    `modified_on` is mutable. Most recently updated company referrals would be consumed first,
+    so they get a chance to appear in the Activity Feed quicker.
+    """
+
+    ordering = ('modified_on', 'pk')
+    summary = 'Company Referral Activities'
+
+
+class CompanyReferralActivityViewSet(ActivityViewSet):
+    """
+    Interaction ViewSet for the activity stream
+    """
+
+    pagination_class = CompanyReferralCursorPagination
+    serializer_class = CompanyReferralActivitySerializer
+    queryset = CompanyReferral.objects.select_related(
+        'company',
+        'contact',
+        'created_by__dit_team',
+        'closed_by__dit_team',
+        'completed_by__dit_team',
+        'recipient__dit_team',
+    )
--- a/changelog/company/company-referral-activity-stream.api.md
+++ b/changelog/company/company-referral-activity-stream.api.md
@@ -1,0 +1,2 @@
+It is now possible to get a list of company referrals created in activity-stream using 
+following URL `/v3/activity-stream/company-referral`.
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -460,6 +460,7 @@ DATAHUB_FRONTEND_BASE_URL = env('DATAHUB_FRONTEND_BASE_URL', default='http://loc
 #  support urls that have object pks within the url.
 DATAHUB_FRONTEND_URL_PREFIXES = {
     'company': f'{DATAHUB_FRONTEND_BASE_URL}/companies',
+    'companyreferral': f'{DATAHUB_FRONTEND_BASE_URL}/company-referrals',
     'contact': f'{DATAHUB_FRONTEND_BASE_URL}/contacts',
     'event': f'{DATAHUB_FRONTEND_BASE_URL}/events',
     'interaction': f'{DATAHUB_FRONTEND_BASE_URL}/interactions',
--- a/datahub/activity_feed/views.py
+++ b/datahub/activity_feed/views.py
@@ -25,6 +25,7 @@ class ActivityFeedView(APIView):
     permission_classes = (IsAuthenticatedOrTokenHasScope,)
 
     ACTIVITY_MODELS_PERMISSIONS_REQUIRED = (
+        'company_referral.view_companyreferral',
         'interaction.view_all_interaction',
         'investment.view_all_investmentproject',
         'order.view_order',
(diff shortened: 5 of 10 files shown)


reupen:
@elcct Some of the tests failed on CircleCI.

datahub/activity_stream/company_referral/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+            'type': 'Announce',
+            'published': instance.created_on,
+            'generator': self._get_generator(),
+            'actor': self._get_adviser(instance.created_by),
+            'target': self._get_adviser(instance.recipient),
+            'object': {
+                'id': company_referral_id,
+                'dit:subject': instance.subject,
+                'dit:status': str(instance.status),
+                'dit:notes': instance.notes,
```
reupen: Do we need `notes` in here? They might be quite large.

datahub/activity_stream/company_referral/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+            'id': f'{company_referral_id}:Announce',
+            'type': 'Announce',
+            'published': instance.created_on,
+            'generator': self._get_generator(),
+            'actor': self._get_adviser(instance.created_by),
+            'target': self._get_adviser(instance.recipient),
+            'object': {
+                'id': company_referral_id,
+                'dit:subject': instance.subject,
+                'dit:status': str(instance.status),
```
reupen: [We previously mapped interaction kinds](https://github.com/uktrade/data-hub-api/blob/c48ac34f3acf1555fbcac5d652498db5748acf20/datahub/activity_stream/interaction/serializers.py#L8-L11), I believe to be compliant with https://www.w3.org/TR/activitystreams-core/. Do we need to do that here? I think @alixedi or @michalc might be able to confirm if we do or if it actually matters. (I think if we can avoid it, it would probably keep things simpler.)

datahub/activity_stream/company_referral/views.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class CompanyReferralActivityViewSet(ActivityViewSet):
+    """
+    Interaction ViewSet for the activity stream
+    """
+
+    pagination_class = CompanyReferralCursorPagination
+    serializer_class = CompanyReferralActivitySerializer
+    queryset = CompanyReferral.objects
```
reupen: Would be good to use `select_related` to reduce the number of queries.

elcct:
@reupen yes, I had a typo in one of the apps, but it is now fixed. Thanks!

datahub/activity_stream/company_referral/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+        if instance.completed_by:
+            company_referral['object']['dit:completed_on'] = instance.completed_on
+            company_referral['object']['attributedTo'].append(
+                self._get_adviser(instance.completed_by),
+            )
+        if instance.closed_by:
+            company_referral['object']['dit:closed_on'] = instance.closed_on
+            company_referral['object']['attributedTo'].append(
+                self._get_adviser(instance.closed_by),
+            )
```
reupen: Is putting `completed_by` and `closed_by` in `attributedTo` clear enough for the front end?

Also, it seems this will affect how the 'My activity' filter works: https://github.com/uktrade/data-hub-frontend/blob/79a80fd4a5a3737367c74e7778d5456d1a521e96/src/apps/companies/apps/activity-feed/builders.js#L35-L39

Do we know which referrals should show when the 'My activity' filter is used?

datahub/activity_stream/company_referral/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        company_referral_id = f'dit:DataHubCompanyReferral:{instance.pk}'
+        company_referral = {
+            'id': f'{company_referral_id}:Announce',
+            'type': 'Announce',
+            'published': instance.created_on,
+            'generator': self._get_generator(),
+            'actor': self._get_adviser(instance.created_by),
+            'target': self._get_adviser(instance.recipient),
+            'object': {
```
reupen: I think we need `startTime` in `object` as well as it seems to be what is used for sorting: https://github.com/uktrade/data-hub-frontend/blob/79a80fd4a5a3737367c74e7778d5456d1a521e96/src/apps/companies/apps/activity-feed/builders.js#L28

datahub/activity_stream/company_referral/serializers.py
elcct: You're right. I was having doubts about it too. I have removed it now. Thank you!

datahub/activity_stream/company_referral/views.py
elcct: Yes! I forgot about this. It is now added. Thank you!

datahub/activity_stream/company_referral/serializers.py
elcct: Good catch! I have added it now. Thank you!

datahub/activity_stream/company_referral/serializers.py
elcct: I was thinking about giving it under `dit:completed_by` etc. or adding another object to the collection with e.g.:

```
{
      "type": "Complete",
      "actor": { <adviser data> },
      "object": <company referral ID>
}
```

but I am not sure if that would be too complex.
Having the adviser who either completed or closed in `attributedTo` was fine to FE dev. 
What do you think?

reupen: I think we need to find out the answer to:

> Do we know which referrals should show when the 'My activity' filter is used?

first.

elcct: It looks like we should show all items that the adviser is involved in.

reupen: So, if I'm not mistaken, that currently means all of `created_by`, `recipient`, `completed_by` and `closed_by` should be in `attributedTo`...

datahub/activity_stream/company_referral/serializers.py
reupen: Actually, those were in `type`, so it doesn't look applicable here.

datahub/activity_stream/company_referral/serializers.py
elcct: I have updated the format so all advisers are in `attributedTo` and the information about performed action is stored in the `context`.

datahub/activity_stream/company_referral/serializers.py
```diff
@@ -0,0 +1,59 @@
...
+            'published': instance.created_on,
+            'generator': self._get_generator(),
+            'object': {
+                'id': company_referral_id,
+                'startTime': instance.created_on,
+                'dit:subject': instance.subject,
+                'dit:status': str(instance.status),
+                'attributedTo': [
+                    self._get_company(instance.company),
+                    self._get_contact(instance.contact),
```
reupen: `contact` is optional, it missing doesn't seem to be handled at the moment.

datahub/activity_stream/company_referral/views.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    pagination_class = CompanyReferralCursorPagination
+    serializer_class = CompanyReferralActivitySerializer
+    queryset = CompanyReferral.objects.select_related(
+        'company',
+        'contact',
+        'created_by',
+        'recipient',
+    )
```
reupen: Would be good to add `completed_by` and `closed_by` too.

reupen:
@elcct Could you explain why we decided to make `context` an object with a timestamp instead of something simpler?

Also, don't we need team info with each the adviser as well?

datahub/activity_stream/company_referral/serializers.py
elcct: That's true. It is now checked I also added a test. Thank you!

elcct:
@reupen I thought that this information gives a context into why particular adviser is included in the list and I thought this will be easier to process when such object attached to adviser object. I initially wanted to include that information at the referral object level, but I thought this will complicate parsing. If you have better way in mind, I am happy to change it.

Regarding teams, I was too focused on general shape of data I missed this detail. We indeed need this information, so I have added it now. Thank you!

datahub/activity_stream/company_referral/views.py
elcct: I added it now and also for `dit_team`. Thank you!

datahub/activity_stream/company_referral/views.py
```diff
@@ -0,0 +1,38 @@
...
+    Interaction ViewSet for the activity stream
+    """
+
+    pagination_class = CompanyReferralCursorPagination
+    serializer_class = CompanyReferralActivitySerializer
+    queryset = CompanyReferral.objects.select_related(
+        'company',
+        'contact',
+        'created_by',
+        'created_by__dit_team',
```
reupen: `created_by__dit_team` implies `created_by`, so we could get rid of `created_by` etc.

datahub/activity_stream/company_referral/serializers.py
```diff
@@ -0,0 +1,74 @@
...
+    def to_representation(self, instance):
+        """
+        Serialize the interaction as per Activity Stream spec:
+        https://www.w3.org/TR/activitystreams-core/
+        """
+        activity_type = self._get_activity_type(instance)
+
+        company_referral_id = f'dit:DataHubCompanyReferral:{instance.pk}'
+        company_referral = {
+            'id': f'{company_referral_id}:{activity_type}',
```
reupen: I'm not sure about varying the ID like this depending on status. Do we have a risk of duplicates until the activity stream service completes a full re-index?

elcct: As per our conversation I understand this will cause unintended behaviour, so I removed this change. Thank you!

datahub/activity_stream/company_referral/views.py
elcct: That is true! Updated. Thank you.

elcct:
Your proposed way will make it even simpler to parse, so I replaced the context with a role and moved the date of either completion or closure to the referral object. I am not sure whether the role would be generic enough, so I namespaced it. Thank you!

datahub/activity_stream/company_referral/views.py
```diff
@@ -0,0 +1,34 @@
...
+from datahub.activity_stream.pagination import ActivityCursorPagination
+from datahub.activity_stream.views import ActivityViewSet
+from datahub.company_referral.models import CompanyReferral
+
+
+class CompanyReferralCursorPagination(ActivityCursorPagination):
+    """
+    Cursor pagination for Company Referral.
+
+    `modified_on` is no unchanging but we have decided to use it because the benefits of being
```
currycoder: Could this read better?

"`modified_on` is mutable"

currycoder: What would the alternative have been which would have meant 10s queries?  Might be good to fill in the blanks for readers of this comment who don't have all the context.

elcct: This is actually a copy & paste comment from `InteractionCursorPagination` and I missed that. I have changed it so it describes why `modified_on` is being used, I hope it makes more sense now.

PR #2589, opened 2020-02-14
update elastic search query builder to support multiple entities

sekharpanja:
### Description of change
This is an enabler PR for to show export country history, where we need to show company export country edits on export tab as well as export countries tagged with interaction.

By enabling elastic search query builder to search across entities, we can then obtain results from company export country history and interaction at the same time.

Exact change is to convert `get_search_by_entity_query` to `get_search_by_entities_query` and instead of taking one `entity` into consideration, take multiple `entities`.

### Checklist

### DIFF
--- a/datahub/search/query_builder.py
+++ b/datahub/search/query_builder.py
@@ -70,8 +70,8 @@ def get_basic_search_query(
     return search[offset:offset + limit]
 
 
-def get_search_by_entity_query(
-        entity,
+def get_search_by_entities_query(
+        entities,
         term=None,
         filter_data=None,
         composite_field_mapping=None,
@@ -81,12 +81,13 @@ def get_search_by_entity_query(
         fields_to_exclude=None,
 ):
     """
-    Performs filtered search for the given term in the given entity.
+    Performs filtered search for the given term across given entities.
     """
     filter_data = filter_data or {}
-    query = [Term(_type=entity._doc_type.name)]
+    query = []
     if term != '':
-        query.append(_build_term_query(term, fields=entity.SEARCH_FIELDS))
+        for entity in entities:
+            query.append(_build_term_query(term, fields=entity.SEARCH_FIELDS))
 
     filters, ranges = _split_range_fields(filter_data)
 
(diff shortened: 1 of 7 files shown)


datahub/search/query_builder.py
```diff
@@ -81,20 +81,27 @@ def get_search_by_entity_query(
...
     """
-    Performs filtered search for the given term in the given entity.
+    Performs filtered search for the given term across given entities.
     """
     filter_data = filter_data or {}
-    query = [Term(_type=entity._doc_type.name)]
+    query = [
+        Term(_type=entity._doc_type.name)
+        for entity in entities
+    ]
```
reupen: Does this mean that each document has to have multiple types when multiple models (entities) are passed in? (Which wouldn't be possible, of course...)

I can't see why we're querying type at all, actually, as each index only has one type. It may be a remnant of when we had multiple types per index...

datahub/search/test/test_query_builder.py
```diff
@@ -559,8 +559,8 @@ def test_get_search_by_entity_query(
     expected_query,
 ):
     """Tests for the get_search_by_entity_query function."""
-    query = get_search_by_entity_query(
-        SimpleModelSearchApp.es_model,
+    query = get_search_by_entities_query(
+        [SimpleModelSearchApp.es_model],
```
reupen: Could we add a test with multiple models passed in? (Just one will probably do.)

datahub/search/test/test_query_builder.py
```diff
@@ -559,8 +559,8 @@ def test_get_search_by_entity_query(
     expected_query,
 ):
     """Tests for the get_search_by_entity_query function."""
```
reupen: The text `get_search_by_entity_query` needs updating here and in the test function name... (There are a few other mentions in the code base to update too.)

datahub/search/query_builder.py
sekharpanja: removed querying the type.

datahub/search/test/test_query_builder.py
```diff
@@ -558,9 +557,9 @@ def test_get_search_by_entity_query(
     fields_to_exclude,
     expected_query,
 ):
-    """Tests for the get_search_by_entity_query function."""
-    query = get_search_by_entity_query(
-        SimpleModelSearchApp.es_model,
+    """Tests for the get_search_by_entities_query function."""
```
currycoder: This method should be called `test_get_search_by_entities_query` now

PR #2591, opened 2020-02-14
Add complete a company referral endpoint

reupen:
### Description of change

This adds a new view at `POST /v4/company-referral/<ID>/complete` for completing a company referral.

Completing a company referral involves creating an interaction (as well as updating the status of the referral).

(The view itself is relatively simple – most of the additions are tests.)

### Checklist

* [x] Do any added or updated endpoints appear in the API documentation? See [docs/Maintaining the API documentation.md](https://github.com/uktrade/data-hub-api/blob/develop/docs/Maintaining&#32;the&#32;API&#32;documentation.md) for more details

### DIFF
--- a/datahub/company_referral/models.py
+++ b/datahub/company_referral/models.py
@@ -2,6 +2,7 @@ from uuid import uuid4
 
 from django.conf import settings
 from django.db import models
+from django.utils.timezone import now
 
 from datahub.core.models import BaseModel
 from datahub.core.utils import get_front_end_url
@@ -76,3 +77,11 @@ class CompanyReferral(BaseModel):
     def get_absolute_url(self):
         """URL to the object in the Data Hub internal front end."""
         return get_front_end_url(self)
+
+    def mark_as_complete(self, interaction, user):
+        """Mark this referral as complete."""
+        self.status = CompanyReferral.Status.COMPLETE
+        self.interaction = interaction
+        self.modified_by = user
+        self.completed_by = user
+        self.completed_on = now()
--- a/datahub/company_referral/test/test_complete_view.py
+++ b/datahub/company_referral/test/test_complete_view.py
@@ -1,0 +1,377 @@
+from datetime import datetime
+from uuid import uuid4
+
+import pytest
+from django.utils.timezone import utc
+from freezegun import freeze_time
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from datahub.company.test.factories import AdviserFactory, CompanyFactory, ContactFactory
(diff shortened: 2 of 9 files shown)


datahub/company_referral/models.py
```diff
@@ -71,3 +72,11 @@ class Status(models.TextChoices):
     def __str__(self):
         """Human-friendly representation (for admin etc.)."""
         return f'{self.company} – {self.subject}'
+
+    def mark_as_complete(self, interaction, user):
+        """Mark this referral as complete."""
+        self.status = CompanyReferral.Status.COMPLETE
+        self.interaction = interaction
```
elcct: In other models calling `mark_*` method would save the object. Why we don't do it here?
I think it is also possible to pass `None` for interaction, so we could end up with invalid object. I think this shouldn't be possible with the way it is being used now, but would be nice to check?

reupen: > In other models calling mark_* method would save the object. Why we don't do it here?

I don't mind calling `save()` here, but I didn't go with that because:

- you can't control the arguments to `save()`
- `save()` ends up being called multiple times when you want to do multiple things
- the calls to `save()` end up being split between the serialiser and the model

> I think it is also possible to pass None for interaction, so we could end up with invalid object. I think this shouldn't be possible with the way it is being used now, but would be nice to check?

Is there any particular scenario you are envisaging? That's already possible via the admin site and I think we we want to be strict about that, we should probably use a `CheckConstraint` (but I'm not sure we should without a good reason, maybe we will find cases where we need to break the link between them?)

elcct: I see. Let's leave it then, as arguments against it are good.

Regarding the scenario, I was just thinking that a user (developer) of the model could call it in a way that doesn't not necessarily ensures that interaction exists, but that seems rather unlikely and probably would have been pointed out during a review.

datahub/company_referral/test/test_complete_view.py
```diff
@@ -0,0 +1,368 @@
...
+                {
+                    'kind': Interaction.Kind.SERVICE_DELIVERY,
+                    'is_event': True,
+                    'event': EventFactory,
+                },
+                id='service-delivery',
+            ),
+        ),
+    )
+    def test_creates_an_interaction(self, extra_data):
```
currycoder: Could it be worth adding a param here which sets a different company ID and asserting that this is ignored and the referral's company id is used?

reupen: I've added one :+1:

PR #2596, opened 2020-02-17
Set PostgreSQL password in CircleCI build

reupen:
### Description of change

The `postgres` Docker image now requires a password to be set. This change sets a password in the CircleCI build to meet that requirement.

See https://github.com/docker-library/postgres/issues/681 for more information.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,8 +29,8 @@ job_defaults: &job_defaults
     ADMIN_OAUTH2_LOGOUT_PATH: '/o/logout'
     ACTIVITY_STREAM_ACCESS_KEY_ID: some-id
     ACTIVITY_STREAM_SECRET_ACCESS_KEY: some-secret
-    DATABASE_URL: postgresql://postgres@localhost/datahub
-    MI_DATABASE_URL: postgresql://postgres@mi-postgres/mi
+    DATABASE_URL: postgresql://postgres:datahub@postgres/datahub
+    MI_DATABASE_URL: postgresql://postgres:datahub@mi-postgres/mi
     DEBUG: 'True'
     DJANGO_SECRET_KEY: changeme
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,8 +29,8 @@ job_defaults: &job_defaults
     ADMIN_OAUTH2_LOGOUT_PATH: '/o/logout'
     ACTIVITY_STREAM_ACCESS_KEY_ID: some-id
     ACTIVITY_STREAM_SECRET_ACCESS_KEY: some-secret
-    DATABASE_URL: postgresql://postgres@localhost/datahub
-    MI_DATABASE_URL: postgresql://postgres@mi-postgres/mi
+    DATABASE_URL: postgresql://postgres:datahub@localhost/datahub
+    MI_DATABASE_URL: postgresql://postgres:datahub@mi-postgres/mi
```
currycoder: My suspicion is that you don't need this...  My understanding is that `POSTGRES_PASSWORD` is just the superuser password and not the password for a DB instance.

elcct: I think this is a password for the DB superuser and would be required if you connect from a different host.

reupen: Yes – without it you get `psycopg2.OperationalError: fe_sendauth: no password supplied`: https://circleci.com/gh/uktrade/data-hub-api/15619

currycoder: Awesome, thank you for validating.  I didn't read the bit where we were authenticating as the `postgres` user :man_facepalming: Please carry on :+1:

reupen: It seems like it might be only required for `mi-postgres` as that one isn't connecting over `localhost`. Still, I think it's easier to leave it in.

reupen: I'll name the main `postgres` container as well – I think it'll be less confusing and will keep things consistent.

PR #2602, opened 2020-02-18
Mechanism for syncing fields on outdated Data Hub company records with D&B

currycoder:
### Description of change

A celery task `datahub.dnb_api.tasks.sync_outdated_companies_with_dnb` was added which provides a mechanism for syncing fields on outdated Data Hub company records with D&B.

This can be used to "backfill" older Company records with newly recorded D&B data; e.g. for rolling out company hierarchies to all D&B linked companies by syncing the `global_ultimate_duns_number` field.

The idea is that we will run this task nightly to roll out company hierarchies, with progressively increasing limits until we reach an agreed target (e.g. 250 calls per night).  This will slowly fill in the data that we need while keeping us well below our D&B API limits.

### Possible points of contention

- The rate limiting enforced by celery is not explicitly tested.  My understanding is that this would be hard for us to do with our current test suite set up due to the dependency on `CELERY_ALWAYS_EAGER`.  To mitigate this risk, we will firstly run the task in simulation mode and interrogate the log timestamps to ensure that they have 1 second gaps.
- `datahub/dnb_api/tasks.py` (and it's test file) is getting long.  I'll follow up this PR with a separate PR to break those files down.

### Checklist

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -4,6 +4,7 @@ from celery import shared_task
 from celery.result import ResultSet
 from celery.utils.log import get_task_logger
 from django.conf import settings
+from django.db.models import F, Max, Q
 from django.utils.timezone import now
 from django_pglocks import advisory_lock
 from rest_framework.status import is_server_error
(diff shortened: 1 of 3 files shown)


datahub/dnb_api/tasks.py
```diff
@@ -80,6 +81,82 @@ def sync_company_with_dnb(
...
+    """
+    dnb_modified_on_before = datetime.strptime(dnb_modified_on_before, '%Y-%m-%dT%H:%M:%S')
+
+    company_ids = Company.objects.filter(
+        Q(dnb_modified_on__lte=dnb_modified_on_before) | Q(dnb_modified_on__isnull=True),
+        duns_number__isnull=False,
+    ).annotate(
+        most_recent_interaction_date=Max('interactions__date'),
+    ).order_by(
+        '-most_recent_interaction_date',
```
reupen: I think this will have nulls first – should we put them last? (I.e. `F('most_recent_interaction_date').desc(nulls_last=True)`.)

datahub/dnb_api/tasks.py
```diff
@@ -80,6 +81,82 @@ def sync_company_with_dnb(
...
+    """
+    A rate limited wrapper around the sync_company_with_dnb task. This task
+    can be used for bulk tasks to ensure that we do not exceed our agreed
+    rate limit with D&B.
+    """
+    message = f'Syncing dnb-linked company: {company_id}'
+    if simulate:
+        logger.info(f'[SIMULATION] {message}')
+        return
+    sync_company_with_dnb.apply(kwargs={
```
reupen: Small point, but would be good to [add a line break before kwargs](https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md#indentation).

datahub/dnb_api/tasks.py
```diff
@@ -80,6 +81,82 @@ def sync_company_with_dnb(
...
+    limit=100,
+    simulate=True,
+):
+    """
+    Sync company records with data sourced from DNB which are determined as outdated.
+    This task will filter dnb-matched companies which have a `dnb_modified_on` date which is before
+    `dnb_modified_on_before` and will then interact with dnb-service to get the latest data to sync
+    these companies.
+    """
+    dnb_modified_on_before = datetime.strptime(dnb_modified_on_before, '%Y-%m-%dT%H:%M:%S')
```
reupen: Could we use `isoparse` from `dateutil`? https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse

reupen: Also note this warning in the build log: `RuntimeWarning: DateTimeField Company.dnb_modified_on received a naive datetime (2019-01-02 11:12:13) while time zone support is active.`

datahub/dnb_api/test/test_tasks.py
```diff
@@ -821,3 +824,340 @@ def test_update_company_from_dnb_data_fails_validation(dnb_response_uk, caplog):
...
+    company = CompanyFactory(
+        duns_number='123456789',
+        dnb_modified_on=now() - timedelta(days=5),
+    )
+    original_company = Company.objects.get(id=company.id)
+
+    task_result = sync_outdated_companies_with_dnb.apply_async(
+        kwargs={
+            'fields_to_update': ['global_ultimate_duns_number'],
+            'dnb_modified_on_before': (now() - timedelta(days=1)).strftime('%Y-%m-%dT%H:%M:%S'),
```
reupen: Are we able to use ISO format instead of using a custom format string?
```suggestion
            'dnb_modified_on_before': (now() - timedelta(days=1)).isoformat(),
```

elcct: I also think it would be great if we could use `isoformat`.

reupen: Another option would be a relative value e.g. number of days ago.

currycoder: I just thought about it and realised that I was hung up with making the `dnb_modified_on_before` argument a string for no reason.  I've changed it to expect a datetime instead which allows us to do away with this whole conversion business - thanks for making me think about this properly!

reupen: Looks like Celery serialises it in ISO format automatically – works out fine though, so seems good to me.

PR #2610, opened 2020-02-20
Switch on global ultimate number ingestion in simulation

currycoder:
### Description of change

A task was added to celery beat to run daily syncs of global ultimate duns numbers. Initially, this task is run with the simulation flag so that we can observe that it is behaving correctly.

### Checklist

* [X] If this is a releasable change, has a news fragment been added?

  <details>
  
  </details>
  
* [X] Has this branch been rebased on top of the current `develop` branch?

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -11,6 +11,8 @@ import base64
 import os
 import stat
 from datetime import timedelta
+from datetime import datetime
+from pytz import utc  # Note: importing django.utils.timezone.utc would cause a circular import
 from urllib.parse import urlencode
 
 import environ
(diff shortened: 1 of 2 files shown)


config/settings/common.py
```diff
@@ -400,6 +402,18 @@ def _build_redis_url(base_url, db_number, **query_args):
                 'simulate': False,
             }
         },
+        'global_ultimate_duns_number_rollout': {
```
alixedi: I think `dnb_heirarchies_backfill` might be a better name because at the job level, we should describe the outcome and not the implementation.

config/settings/common.py
```diff
@@ -400,6 +402,18 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+            'task': 'datahub.dnb_api.tasks.sync_outdated_companies_with_dnb',
+            'schedule': crontab(minute=0, hour=1,),
```
alixedi: Have you considered running this weekly instead of daily?

config/settings/common.py
```diff
@@ -400,6 +402,18 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+            'kwargs': {
+                'dnb_modified_on_before': datetime(
+                    year=2019, month=10, day=24, hour=23, minute=59, second=59, tzinfo=utc,
```
alixedi: Would be useful to put in some comments about why this specific date?

config/settings/common.py
currycoder: I'd prefer to get quick feedback so that we can iterate removing the simulate flag and upping the numbers quickly - e.g. daily feedback.
Perhaps when we are feeling confident enough to switch this on fully, we could switch it to be a weekly task.

config/settings/common.py
currycoder: Good point!

PR #2619, opened 2020-02-24
Add company referral by interaction ID endpoint

elcct:
### Description of change

This adds a new endpoint that helps the front-end display company referral detail when the viewed interaction has been created as a result of referral completion.
There is 1:1 relation between completed referral and interaction.
The endpoint is under the following URL:

`GET /v4/interaction/<uuid:interaction_id>/company-referral`

The response is the same as with the company referral item endpoint.

Only a completed referral can have assigned interaction, so the endpoint is only able to return complete one, otherwise the result will be 404.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company_referral/test/test_views.py
+++ b/datahub/company_referral/test/test_views.py
@@ -461,14 +461,18 @@ class TestGetCompanyReferralByInteractionID(APITestMixin):
         response = api_client.get(url)
         assert response.status_code == expected_status
 
-    def test_returns_404_for_non_existent_referral(self):
-        """Test that a 404 is returned for a non-existent referral ID."""
+    def test_returns_404_for_non_existent_interaction(self):
+        """Test that a 404 is returned for a non-existent interaction."""
         url = _item_by_interaction_id_url(uuid4())
         response = self.api_client.get(url)
         assert response.status_code == status.HTTP_404_NOT_FOUND
 
     def test_retrieve_a_complete_referral(self):
-        """Test that a complete referral can be retrieved by interaction ID."""
+        """
+        Test that a complete referral can be retrieved by interaction ID.
+
+        Only complete referral has interaction assigned.
+        """
         referral = CompleteCompanyReferralFactory()
         url = _item_by_interaction_id_url(referral.interaction.pk)
 
--- a/datahub/company_referral/test/test_views.py
+++ b/datahub/company_referral/test/test_views.py
@@ -31,6 +31,13 @@ def _item_url(pk):
     return reverse('api-v4:company-referral:item', kwargs={'pk': pk})
 
 
+def _item_by_interaction_id_url(interaction_id):
+    return reverse(
+        'api-v4:company-referral:item-by-interaction-id',
+        kwargs={'interaction_id': interaction_id},
+    )
+
+
 class TestListCompanyListsView(APITestMixin):
     """Tests for listing user's referrals."""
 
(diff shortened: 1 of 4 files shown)


datahub/company_referral/test/test_views.py
```diff
@@ -424,6 +431,78 @@ def test_retrieve_a_referral(self, factory):
...
+        referral = CompleteCompanyReferralFactory()
+        user = create_test_user(permission_codenames=permission_codenames)
+
+        url = _item_by_interaction_id_url(referral.interaction.pk)
+        api_client = self.create_api_client(user=user)
+
+        response = api_client.get(url)
+        assert response.status_code == expected_status
+
+    def test_returns_404_for_non_existent_referral(self):
```
currycoder: Could it be valuable to add a similar test to this one which tests that nothing is returned when the referral is not completed?

elcct: Incomplete referral will no have attached interaction, so this test in a way covers this as well. I could perhaps add a comment about it?

elcct: I renamed this test and updated other test with a comment that only complete referral has interaction assigned.

elcct:
My reason for using `/v4/interaction/<interaction_id>/company_referral` is because in this context company referral is an additional interaction detail and given that I'd have only an interaction id as a starting point, I'd check if any of interaction endpoints would give me this information first. We already use this pattern when there is 1:1 relationship to get related object, for example in OMIS `/v3/omis/order/{order_pk}/invoice`. I considered using the filter `/v4/company-referral?interaction_id=<id>` but we would always have maximum one record returned, so I wasn't sure if a filtered list would be good application for this. The second option I considered is as you suggest to use read-only field on the interaction, but I thought that would make already complex area even more complex.

datahub/company_referral/test/test_views.py
```diff
@@ -424,6 +431,82 @@ def test_retrieve_a_referral(self, factory):
...
+        api_client = self.create_api_client(user=user)
+
+        response = api_client.get(url)
+        assert response.status_code == expected_status
+
+    def test_returns_404_for_non_existent_interaction(self):
+        """Test that a 404 is returned for a non-existent interaction."""
+        url = _item_by_interaction_id_url(uuid4())
+        response = self.api_client.get(url)
+        assert response.status_code == status.HTTP_404_NOT_FOUND
```
mcbhenwood: In the PR comments it says "Only a completed referral can have assigned interaction, so the endpoint is only able to return complete one, otherwise the result will be 404."

I can see that this test block here logically tests that situation but inherent in the test structure is an assumption that the rule "only a completed referral can have an interaction". Is there any way we could harden the tests to verify that actual statement? Because as it stands it feels like it's based on an assumption about the implementation and that may give rise to problems later.

I may be asking to test the impossible here, in which case fair enough, but let's give it a few minutes' thought at least.

mcbhenwood: Oh I just realised I am echoing @currycoder 's question above. My questions still stands though.

mcbhenwood:
re the discussion about the path etc. I too feel that `/v4/interaction/<interaction_id>/company_referral` is a bit kludgy. I think I favour the 'property on the interaction' approach. I'm with you @elcct in not liking `/v4/company-referral?interaction_id=<id>` bc it will only ever provide zero or one referral.

While I favour the 'property on interaction' approach I'm not sure I favour it enough to get you to rip up this whole implementation at PR time. API / ROA design is _hard_ and sometimes we just have kludgy paths.

What does everybody else think?

reupen:
I'm not against the path per se, but more the awkward source code structure that results. I think OMIS is different as everything is under OMIS there, and those objects are also never accessed outside the context of an order and don't already have separate listing endpoints.

I think my first preference would be an extra read-only field on the interaction serialiser, though. 

Regarding the query parameter, I think handling zero results might well be be less awkward in the FE than handling a 404, and query parameters tend to be more flexible (since they can be combined). Nested resources can easily get awkward when you get several levels deep as you have to check that each object in the path exists (in order to return a 404) and we've had bugs relating to that before.

elcct:
My second favourite option (after current implementation) is to add read-only field on interaction. This adds a benefit for the front-end to not have to make a call to see if the interaction has been created as a result of company referral at a cost of increased complexity of the interaction endpoint (that I wanted to avoid). I am happy to make this change, as I see the current implementation was not met with enthusiasm.

currycoder:
I must admit that I missed that this was adding a `/interaction` endpoint under the company referral app in my first pass.  I can't add much to the good discussion that has already been had, other than my opinion that adding a read-only referral ID to the existing interaction endpoint seems like a good compromise :+1:

elcct:
Implementation will change, so I am closing this one.

PR #2623, opened 2020-02-25
Configure company hierarchy rollout task to run without simulate flag

currycoder:
### Description of change
The daily company hierarchy rollout task was adjusted so that it is no longer in simulation mode. It will ingest a number of companies per night up to a limit configured by the `DAILY_HIERARCHY_ROLLOUT_LIMIT` environment variable.

A `ENABLE_DAILY_HIERARCHY_ROLLOUT` environment variable has been added so that we can limit this task to run in production only.  We observed that the simulated run ran on staging and production around the same time which might lead to us exceeding our rate limit.

### Checklist

* [X] If this is a releasable change, has a news fragment been added?

  <details>
  
  </details>
  
* [X] Has this branch been rebased on top of the current `develop` branch?

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/dnb_api/tasks.py
+++ b/datahub/dnb_api/tasks.py
@@ -136,6 +136,7 @@ def sync_outdated_companies_with_dnb(
     `dnb_modified_on_before` and will then interact with dnb-service to get the latest data to sync
     these companies.
     """
+    logger.info('sync_outdated_companies_with_dnb task started.')
     company_ids = Company.objects.filter(
         Q(dnb_modified_on__lte=dnb_modified_on_before) | Q(dnb_modified_on__isnull=True),
         duns_number__isnull=False,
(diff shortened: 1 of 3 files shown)


datahub/dnb_api/tasks.py
```diff
@@ -155,6 +160,7 @@ def sync_outdated_companies_with_dnb(
                 'simulate': simulate,
             },
         )
+    logger.info('sync_outdated_companies_with_dnb task finished.')
```
alixedi: Correct me if I am wrong but the calls to `sync_company_with_dnb_rate_limited` are not blocking and so this line will not add anything to our understanding.

currycoder: I agree in theory, but I would personally like to validate that understanding.  I believe that we will rip out this logging later anyway.

currycoder: Additionally, it will help give an indication about how long the query above it takes.

PR #2624, opened 2020-02-25
Celery task to communicate with DIT Consent Service

r4vi:
### Description of change

We are creating a central service for all DIT systems to store marketing consent for DIT contacts. This will be the central source of truth for great.gov.uk, data-hub, email marketing. 

In the interest of keeping PRs small and manageable this is just one PR of many.  All of which will be labeled with `consent-service`. The [entire plan](https://docs.google.com/document/d/1XrDAtweJQQdRh6-tdhF8ujVUIyAbSiwfMUz-R-npGlo/edit) is available to view for people with @digital.trade.gov.uk accounts.

This PR creates a new task, which is behind a feature flag to update email
marketing consent in the DIT consent service. The flag should not be flipped
until the consent service in production (with SRIO approval for handling production data).

There will be a follow up PR make the `company.Contact` serializer or manager call this task on update/create.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

* Make sure the README is updated e.g. when adding new environment variables :+1: 

</details>


r4vi:
I'm not entirely convinced by splitting up the `tasks.py` into a module and if others think it should just stay as a single `.py` I'm happy to change it back.

datahub/company/tasks/__init__.py
```diff
@@ -0,0 +1,4 @@
+from datahub.company.tasks.company import automatic_company_archive
+from datahub.company.tasks.contact import update_contact_consent
+
+__all__ = [automatic_company_archive, update_contact_consent]
```
r4vi: to allow celery to autodiscover these tasks

datahub/company/tasks/contact.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def _update_contact_consent(email_address, accepts_dit_email_marketing, modified_at=None):
+    if not all([settings.CONSENT_SERVICE_HAWK_ID,
+                settings.CONSENT_SERVICE_HAWK_KEY,
+                settings.CONSENT_SERVICE_BASE_URL]):
+        raise ImproperlyConfigured(
+            'The all CONSENT_SERVICE_* settings must be set to use this task.')
+
+    auth = HawkAuth(id=settings.CONSENT_SERVICE_HAWK_ID, key=settings.CONSENT_SERVICE_HAWK_KEY)
```
alixedi: Please consider promoting this to a module-level constant. 

I might be wrong but we shouldn't have to do this every time this function is called.

alixedi:
> I'm not entirely convinced by splitting up the `tasks.py` into a module and if others think it should just stay as a single `.py` I'm happy to change it back.

I like the looks of it.

That said, we should also refactor the tests to reflect this. This can happen in a separate PR though.

changelog/contact/new-celery-task-to-post-email-consent.md
```diff
@@ -0,0 +1,3 @@
+A new celery task to update the consent service has been added but it's
```
currycoder: I believe the file should be called `new-celery-task-to-post-email-consent.feature.md`

datahub/company/test/test_tasks.py
```diff
@@ -322,3 +336,69 @@ def test_investment_projects(
...
+    @pytest.mark.parametrize(
+        'email_address, accepts_dit_email_marketing, modified_at',
+        (
+            ('example@example.com', True, None),
+            ('example@example.com', False, None),
+            ('example@example.com', True, now()),
+            ('example@example.com', False, now()),
+        ),
+    )
+    def test_task_makes_http_request(
```
currycoder: Could it also be worth adding a test case which codifies how the task should behave when the consent service API call returns various error statuses?

datahub/company/tasks/contact.py
r4vi: yep you're right, since it doesn't change unless env vars change, I'll make it module level

r4vi:
> > I'm not entirely convinced by splitting up the `tasks.py` into a module and if others think it should just stay as a single `.py` I'm happy to change it back.
> 
> I like the looks of it.
> 
> That said, we should also refactor the tests to reflect this. This can happen in a separate PR though.

I've taken the opportunity to split the tests up here too so it doesn't get missed later

datahub/company/test/test_tasks.py
r4vi: Yes, I'll add some for the actual post / function but testing the celery auto-retry mechanism might be tricky

datahub/company/tasks/__init__.py
reupen: AFAIK `__all__`  usually contains strings: https://docs.python.org/3/reference/simple_stmts.html?highlight=__all__#the-import-statement

r4vi: :face_with_head_bandage: good to know, weird how it continued to work anyway. changing

reupen:
It would be good to use `datahub.core.api_client.APIClient` as it does some useful stuff like logging etc. consistently.

Also have a look at https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md#indentation – some of the indentation/line wrapping isn't in line with that at the moment.

datahub/company/tasks/contact.py
r4vi: @alixedi  I've moved this out, but it would require the ENV vars to be set even if the feature flag is off. Do you think I should make the env vars required now (with dummy values), and leave it at module level or move it back in. 

There is a plan to remove the feature flag once it stops being optional so at that point it can be made a module level constant?

datahub/company/tasks/company.py
```diff
@@ -67,6 +66,8 @@ def _automatic_company_archive(limit, simulate):
     priority=9,
     max_retries=3,
     queue='long-running',
+    name='datahub.company.tasks.automatic_company_archive',
+    # name set explicitly to maintain backwards compatibility
```
reupen: Would suggest putting the comment before the line it relates to (or get rid of this, the task runs once a week on a Saturday so don't think we should run into any problems).

datahub/company/tasks/contact.py
```diff
@@ -0,0 +1,61 @@
...
+        timeout=(connect_timeout, read_timeout),
+        json=body,
+    )
+    response.raise_for_status()
+
+    logger.debug(response.json())
+
+
+@shared_task(
+    retry_kwargs={'max_retries': 5},
```
reupen: Any reason for not passing `max_retries` directly?

r4vi: no reason, I'll move to the simpler syntax

datahub/company/tasks/contact.py
```diff
@@ -0,0 +1,61 @@
...
+    autoretry_for=(requests.exceptions.RequestException,),
+    retry_backoff=True,
```
reupen: I think we would be better off setting this to an integer e.g. 30, [as `True` means `1` here](https://docs.celeryproject.org/en/stable/userguide/tasks.html#Task.retry_backoff) (so delays of roughly 1, 2, 4, 8, 16 seconds...).

r4vi: nice, I think 30,60,120 etc makes more sense than 1,2,4. Will change

r4vi:
> It would be good to use `datahub.core.api_client.APIClient` as it does some useful stuff like logging etc. consistently.
> 
> Also have a look at https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md#indentation – some of the indentation/line wrapping isn't in line with that at the moment.

Handy, I'll switch to using that

PR #2632, opened 2020-02-27
Include interactions to export country history search API

sekharpanja:
### Description of change
Expose Interactions with export countries within export country history endpoint, along with existing export country history that is generated as a result of edits to export countries within export tab of a company.

This would mean enabling `export_country_history` search app to expose multiple entities: `export_country_history` and `interaction`.

Also adding couple of specific filters to satisfy business needs: 
- include interactions with export countries: `kind: interaction and were_countries_discussed: True`
- include only inserts and deletes within export country history items (exclude UPDATEs): `history_type: [INSERT, DELETE]`

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/search/export_country_history/test/test_views.py
+++ b/datahub/search/export_country_history/test/test_views.py
@@ -210,11 +210,8 @@ class TestSearchExportCountryHistory(APITestMixin):
             url,
             data={
                 'company': company_id,
-                # 'were_countries_discussed': True,
-                # 'history_type': ['insert', 'delete'],
             },
         )
-        # print(response.json())
         assert response.status_code == status.HTTP_200_OK
         assert response.json()['count'] == count
         assert all(
@@ -263,8 +260,6 @@ class TestSearchExportCountryHistory(APITestMixin):
             url,
             data={
                 'company': company_id,
-                # 'kind': 'interaction',
-                # 'were_countries_discussed': True,
             },
         )
 
--- a/datahub/search/export_country_history/test/test_views.py
+++ b/datahub/search/export_country_history/test/test_views.py
@@ -137,7 +137,7 @@ def setup_data():
             history_type=CompanyExportCountryHistory.HistoryType.DELETE,
         )
 
-    yield str(benchmark_company.id)
+    yield benchmark_company
 
 
 class TestSearchExportCountryHistory(APITestMixin):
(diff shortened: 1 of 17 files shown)


reupen:
@sekharpanja The change of the type of `history_type` to `Keyword(index=True)` will need to be done and released before `history_type` can be used in a query as the data will need to be reindexed (otherwise the query will fail as the field wasn't indexed previously).

So we'll probably want to do that and add the `date` field in a separate PR first so the fields are in place before they start to be queried.

datahub/search/export_country_history/views.py
```diff
@@ -22,7 +24,49 @@ class ExportCountryHistoryView(SearchAPIView):
...
+                    'match', **{'history_type': 'insert'},
+                ) | Q(
+                    'match', **{'history_type': 'delete'},
+                ),
+            ]),
+        )
+        base_query.query.filter.append(
+            Bool(should=should_filters, minimum_should_match=1),
+        )
+        return base_query
```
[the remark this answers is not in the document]
reupen: Also, is it correct to exclude service deliveries (by checking that `kind == DBInteraction.Kind.INTERACTION`)? We collect have export countries for them as well, right?

datahub/search/export_country_history/test/test_views.py
```diff
@@ -44,24 +49,100 @@ def setup_data():
         CompanyExportCountryHistoryFactory(
```
reupen: It seems like this fixture creates quite a few objects – do all of the tests that use the fixture need all of these?

datahub/search/export_country_history/test/test_views.py
```diff
@@ -188,34 +334,112 @@ def test_filtering_by_company_and_country_on_export_country_history_search(
...
         )
+        for result in response.json()['results']:
+            if result.get('kind', '') == 'interaction':
+                result_countries = [item['country']['id'] for item in result['export_countries']]
+                assert expected_data['country']['id'] in result_countries
         assert all(
             result['country']['id'] == expected_data['country']['id']
             for result in response.json()['results']
+            if result.get('kind', '') != 'interaction'
         )
```
reupen: I think we should be checking that the correct objects are returned (i.e. the expected interaction and export country history IDs) rather than the country IDs (which could relate to any object).

datahub/search/export_country_history/test/test_views.py
```diff
@@ -188,34 +334,112 @@ def test_filtering_by_company_and_country_on_export_country_history_search(
...
+        assert response.json()['count'] == 6
+        for result in response.json()['results']:
+            if result.get('kind', '') == 'interaction':
+                result_countries = [item['country']['id'] for item in result['export_countries']]
+                assert expected_data['country']['id'] in result_countries
+        assert all(
+            result['country']['id'] == expected_data['country']['id']
+            for result in response.json()['results']
+            if result.get('kind', '') != 'interaction'
+        )
```
reupen: I think we should be checking that the correct objects are returned (i.e. the expected interaction and export country history IDs) rather than the country IDs (which could relate to any object).

datahub/search/export_country_history/test/test_views.py
```diff
@@ -188,34 +334,112 @@ def test_filtering_by_company_and_country_on_export_country_history_search(
...
+
+    def test_country_filter_sorting(self, es_with_collector, setup_data):
+        """Tests the sorting of country history search response."""
```
reupen: I don't think this test is useful as it is – sorting and filtering are discrete operations, and testing every combination of them doesn't seem particularly sustainable.

But, the assertion in `test_filtering_by_country` could easily expect a specific order of results also.

datahub/search/export_country_history/test/test_views.py
```diff
@@ -93,72 +166,170 @@ def test_export_country_history_search_with_empty_request(self, es_with_collecto
...
+        Check search works when there are only interactions, no export country history.
+        """
+        benchmark_country_canada = Country.objects.get(
+            pk=CountryConstant.canada.value.id,
+        )
+        benchmark_company = CompanyFactory()
+
+        interaction = ExportCountriesInteractionFactory(
+            company=benchmark_company,
+            were_countries_discussed=countries_discussed,
```
reupen: Looks like this will be creating an export country even when `were_countries_discussed=False`. Also seems odd to pass `were_countries_discussed=False` to `ExportCountriesInteractionFactory`. Maybe it would be better to use a different factory depending on `were_countries_discussed`?

datahub/search/export_country_history/test/test_views.py
```diff
@@ -93,72 +166,170 @@ def test_export_country_history_search_with_empty_request(self, es_with_collecto
...
+        ]
+        assert date_times == [
+            FROZEN_DATETIME_3,
+            FROZEN_DATETIME_2,
+            FROZEN_DATETIME_2,
+            FROZEN_DATETIME_2,
+            FROZEN_DATETIME_1,
+            FROZEN_DATETIME_1,
+            FROZEN_DATETIME_1,
+        ]
```
reupen: I really think these assertions are too complicated – IMO all that is needed for checking that the right items are returned is:

```python
results = response.json()['results']
expected_result_ids = [...]
assert [result['id'] for result in results] == expected_result_ids
```

reupen:
I'm working on updates to this.

reupen:
I've pushed some updates. They:

- stop excluding service deliveries
- rejiig the tests to make them simpler
- add tests for the response bodies
- update the permissions required
- restrict the interaction fields returned to only the ones required

reupen:
I'm going to split this to make it easier to review. Closing for now.

PR #2648, opened 2020-03-02
Trigger consent celery task on create/update of `company.Contact` when using ContactSerializer [1 of 3]

r4vi:
#### Goal
When a Contact is updated using the Contact Serializer, we want to trigger the celery task. This approach is quite readable and easy to reason about but probably not that elegant.  We also want to keep this behind a feature flag until we have signoff to put production data into the consent service.

#### Request for feedback on approach
This is the next increment of the consent service integration with data-hub. We previously created a celery task to allow setting consent, the function to do this was alongside the task in the same python file.

* A consent service module, a utility of module level functions to call the consent service api
* Updates ContactSerializer to call the celery task on `.save` and `.update`

NOTE: 
The feature flag should not be turned on until we have SIRO sign off

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/company/consent.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        Takes configuration from django settings by default but they can be
+        optionally passed in via kwargs
+        """
+        if not base_url:
+            base_url = settings.CONSENT_SERVICE_BASE_URL
+        if not hawk_id:
+            hawk_id = settings.CONSENT_SERVICE_HAWK_ID
+        if not hawk_key:
+            hawk_key = settings.CONSENT_SERVICE_HAWK_KEY
```
reupen: Do we have a use case for setting these directly instead of via settings?

datahub/company/consent.py
```diff
@@ -0,0 +1,111 @@
...
+        logger.debug(response.json())
+
+    def lookup(self, emails):
+        """
+        Bulk lookup consent for a list of emails
+
+        :param emails: List of email addresses
+        :return: dict of email address to consent status
+        """
+        logger.info(f'get: {emails}')
```
reupen: AFAIK we avoid putting email addresses in logs – so if we need this would give it the debug log level.

datahub/company/consent.py
```diff
@@ -0,0 +1,111 @@
...
+        response = self.api_client.request(
+            'POST',
+            CONSENT_SERVICE_PERSON_PATH_LOOKUP,
+            json=body,
+            params={'limit': len(emails)},
+        )
+        data = {}
+        for result in response.json().get('results'):
+            data[result['email']] = CONSENT_SERVICE_EMAIL_CONSENT_TYPE in result['consents']
+        return data
```
reupen: Could we use a dict comprehension here? E.g.:

```python
return {
    result['email']: CONSENT_SERVICE_EMAIL_CONSENT_TYPE in result['consents']
    for result in response.json().get('results')
}
```

datahub/company/consent.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        response = self.api_client.request(
+            'POST',
+            CONSENT_SERVICE_PERSON_PATH_LOOKUP,
+            json=body,
+            params={'limit': len(emails)},
+        )
+        data = {}
+        for result in response.json().get('results'):
```
reupen: Won't `.get()` default to `None` here (which would fail)? If the `results` key should always be there I think `response.json()['results']` would be better.

datahub/company/serializers.py
```diff
@@ -211,6 +213,40 @@ class Meta:
...
 
+    def notify_consent_service(self, validated_data):
+        """
+        Trigger the update_contact_consent task with the current version
+        of `validated_data`. The actual enqueuing of the task happens in the
+        on_commit hook so it won't actually notify the consent service unless
+        the database transaction was successful.
+        """
+        on_commit(
+            lambda: update_contact_consent.delay(
```
reupen: The rest of the code base [uses `. apply_async()`](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async) (since it allows configuration options to be passed) – would be good to keep that consistent.

datahub/company/serializers.py
```diff
@@ -211,6 +213,40 @@ class Meta:
...
+                validated_data['email'],
+                validated_data['accepts_dit_email_marketing'],
```
reupen: If it's a `PATCH` request (partial update) these might not both be here. You could use [`DataCombiner`](https://github.com/uktrade/data-hub-api/blob/fd246654c548ab27c5c6b12eaf66e0baa3c66244/datahub/core/validate_utils.py#L7) to fall back to the model instance if they aren't in the PATCH request body.

datahub/company/consent.py
r4vi: not really apart from maybe making testing easier but there are different ways to achieve that

datahub/company/consent.py
r4vi: my mistake, that's leftover from my debugging it - will remove it

datahub/company/consent.py
r4vi: my thinking was if there's no `results` it will return a None, the loop won't run and an empty dict will be returned, but in theory there should always be a `results` in a request to that endpoint

reupen: If `results` is absent I expect `for result in response.json().get('results'):` will give us `TypeError: 'NoneType' object is not iterable`.

I think if it's meant to be there, it's good to fail since otherwise we might return the wrong result.

datahub/company/consent.py
reupen: Yep, I think if we have this then the override logic will need testing too. Could add values to  `config.settings.test` to make sure the tests are using a consistent value.

r4vi:
>  Have left some comments.
> 
> > * ConsentService class, a utility to call the consent service api, is it required or overkill.
> 
> I think it's fine (but a set of model-level functions would probably be fine as well).
> 
> Some additional comments on the class though:
> 
> * would be good to think about more descriptive names for the `lookup` and `get` methods to make it clear that one is for many email addresses and the other is for a single one – e.g. `get_many` and `get_one`
> * we could have a single instance of the class so that we don't have to instantiate it everywhere
> 
> > * Using `on_commit` signal, this ensures the task isn't triggered unless the database save worked but it makes testing a bit harder
> 
> It is correct as the transaction could be rolled back as you say, so would keep it – you can use `@pytest.mark.usefixtures('synchronous_on_commit')` for tests (`on_commit()` is used elsewhere in the code base already).
> 
> > * Is there a more elegant way to hook into the serializer than explicitly overriding `.create`, and `.update`? Could be a mixin but it wouldn't be very generic? `TriggerCeleryTaskOnUpdateSerializerMixin` seems like something that'd only ever be used here.
> 
> While I agree overriding `update()` and `create()` is a bit messy, I think it's fine (and what DRF expects you to do AFAIK).

Thanks for the comments @reupen , I'll wait for other people to chime in for a bit before fixing the things you've brought up. In the meantime I'll probably create a couple of new draft PRs to this PR that show ways to:

1. Get consent info via API for Contact detail view
2. Get consent info in bulk for ContactsDatasetView
3. Get consent info in bulk for SearchContactExportAPIView
4. Update consent when a contact's primary email address is changed in admin form

I want to get the high level approaches to these things agreed early, before locking down the exact implementations.

datahub/company/consent.py
currycoder: I agree that a dependency injection approach is noble, but I'd err on the side of just using `override_settings` in tests: https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.override_settings

datahub/company/consent.py
```diff
@@ -0,0 +1,111 @@
...
+from datahub.company.constants import CONSENT_SERVICE_EMAIL_CONSENT_TYPE
+from datahub.core.api_client import APIClient, HawkAuth
+
+logger = logging.getLogger(__name__)
+
+CONSENT_SERVICE_PERSON_PATH = 'api/v1/person/'
+CONSENT_SERVICE_PERSON_PATH_LOOKUP = f'{CONSENT_SERVICE_PERSON_PATH}bulk_lookup/'
+
+
+class ConsentService:
```
currycoder: I'd be tempted to just make this a singleton - e.g. instantiate with `consent_service = ConsentService()` below this class.

datahub/company/serializers.py
```diff
@@ -211,6 +213,40 @@ class Meta:
...
+            ),
+        )
+
+    def create(self, validated_data):
+        """
+        Create a new instance using this serializer
+
+        :return: created instance
+        """
+        self.notify_consent_service(validated_data)
```
currycoder: Forgive me if this has already been discussed before this PR..  An alternative here would be to instead add a post save signal here which would fire the celery task: https://github.com/uktrade/data-hub-api/blob/develop/datahub/company/signals.py

This would have the added benefit that if the contact was saved through the admin (or through some other mechanism that is not directly using the API/serializer in the future), the celery task would still be fired.

r4vi: I think we might have discussed it one of the many google docs. 

The field will eventually be removed the the datahub model entirely so if we put in a post save handler, at the point we remove the field, we'll have to figure out what to do again. 

I do find it annoying that we'll end up with 2/3 places that need to know about this - maybe there is a way to centralise this in the model post_save in a way it'll continue to work after the field is removed.

r4vi: actually I've thought about it more, and centralising to the model is breaking abstraction levels. This field is something that it composed from an external service by the serializer and shouldn't really leak any deeper into datahub than the serializer layer

datahub/company/consent.py
```diff
@@ -0,0 +1,111 @@
...
+                'The all CONSENT_SERVICE_* settings must be set to use this task.')
+
+        # once feature flag is remove this can be moved to be a module level constant.
+        auth = HawkAuth(
+            api_id=settings.CONSENT_SERVICE_HAWK_ID,
+            api_key=settings.CONSENT_SERVICE_HAWK_KEY,
+            verify_response=settings.CONSENT_SERVICE_HAWK_VERIFY_RESPONSE,
+        )
+
+        connect_timeout, read_timeout = 5.0, 30.0
```
elcct: Maybe these could be defined at the top e.g. `CONSENT_SERVICE_CONNECT_TIMEOUT` so it is easier to find.

datahub/company/serializers.py
r4vi: good point but I don't think `DataCombiner` is the best idea here, once the accepts_dit_email_marketing field is removed from datahub's database then `DataCombiner` won't be able to reconstruct the required info anyway. 

This is going to be tricky, we might need to be able to look up the old value from the API when it's a patch with just an email address or if it's a change of email address then assume False (as we've not had explicit opt-in)

So if (given a patch):
| email | accepts_dit_email_marketing | instance_email | *Action* |
|-------------|-----------------------------|----------------|-------------------------------------------------------------------------------|
| foo@bar.com | True | foo@bar.com | Notify consent service |
|  | True | foo@bar.com | Notify consent service (use `email` from instance) |
| foo@bar.com |  | foo@bar.com | Do nothing? |
| foo@bar.com |  | old@old.com | Notify consent service (use `email` from PATCH, set consent to `False`) |
|  |  | foo@bar.com | Do nothing (PATCH didn't have either `email` or `accepts_dit_email_marketing` |

@reupen  This is turning into a bit of a mess, can you think of a way to avoid this? Maybe disable PATCH if it isn't used by thing?

datahub/company/consent.py
r4vi: thanks, agreed - I've turned it into a bunch of module level functions so nothing needs to be instantiated

datahub/company/serializers.py
r4vi: actually `DataCombiner` is still useful to reconstruct the email address if it wasn't in the patch body... just need to figure out if the email address changed, if it did then assume False or whatever was in the PATCH body and notify consent service

reupen: `PATCH` is used exclusively for updates in Data Hub.

At the moment the front end probably sends all fields in the HTML form, but I think the direction of travel (with React) is to only send modified form fields. So, I'm not sure about inferring that '`email` changed but `accepts_dit_email_marketing` omitted' means an opt-out.

One option would be to require `accepts_dit_email_marketing` in the body if `email` is present. I think it's a bit risky though as it's a bit of a hidden requirement and could easily be missed by testing.

r4vi: Having considered this again, I think we can side step the complexity entirely. If there is no `accepts_dit_email_marketing` in the patch, then don't send anything. This will work if we assume all 404s from the consent service imply false.

Therefore if we change an email address but don't provide consent status then it is, not in the consent service - when the new email address is queried it will return false.

If the new email already exists in the consent service it will return the last known consent.

r4vi:
> It is correct as the transaction could be rolled back as you say, so would keep it – you can use `@pytest.mark.usefixtures('synchronous_on_commit')` for tests (`on_commit()` is used elsewhere in the code base already).

Weirdly this didn't work for me and I spent almost a day figuring out why, eventually I have up and just patched `on_commit` where I was using it

reupen:
> Weirdly this didn't work for me and I spent almost a day figuring out why, eventually I have up and just patched on_commit where I was using it

It's because `on_commit` is being imported directly rather than the wntire `transaction` module (i.e. `from django.db.transaction import on_commit` rather than `from django.db import transaction`).

I'd probably do the latter as it seems to be the convention in this case: https://docs.djangoproject.com/en/3.0/topics/db/transactions/

r4vi:
> > Weirdly this didn't work for me and I spent almost a day figuring out why, eventually I have up and just patched on_commit where I was using it
> 
> It's because `on_commit` is being imported directly rather than the entire `transaction` module (i.e. `from django.db.transaction import on_commit` rather than `from django.db import transaction`).
> 
> I'd probably do the latter as it seems to be the convention in this case: https://docs.djangoproject.com/en/3.0/topics/db/transactions/

Thanks! Will change over to that way and try it :pray:

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -0,0 +1,165 @@
...
+            address_county='London',
+            address_country_id=constants.Country.united_kingdom.value.id,
+            address_postcode='SW1A1AA',
+            notes='lorem ipsum',
+            accepts_dit_email_marketing=False,
+        )
+        return contact
+
+    @patch('datahub.company.serializers.on_commit', side_effect=immediate)
+    @patch('datahub.company.serializers.update_contact_consent.apply_async')
```
reupen: Most of the tests use pytest's `monkeypatch` rather than `unittest.mock.patch` – it's not critical but it would be good to do the same if possible. (There are examples in the code base – it's often still used with `Mock` and you can do it as a fixture if it'll be used multiple times.)

datahub/company/consent.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if modified_at:
+        body['modified_at'] = modified_at
+
+    response = _get_client().request(
+        'post',
+        CONSENT_SERVICE_PERSON_PATH,
+        json=body,
+    )
+    response.raise_for_status()
```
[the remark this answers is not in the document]
r4vi: :+1:

datahub/company/consent.py
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+
+def get_one(email):
+    """
+    Get consent for single email address
+    :return: bool indicating if we have consent to send email marketing
+    to an address
+    """
+    return get_many([email]).get(email, False)
```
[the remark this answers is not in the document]
r4vi: We decided to never return None after a chat with Billy, unless we have recorded explicit consent we're going to return False

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -0,0 +1,165 @@
...
+    Tests for the Contact Serializer. Checking that update / create notify the
+    consent service correctly.
+    """
+
+    @property
+    def user(self):
+        """Return the user."""
+        if not hasattr(self, '_user'):
+            self._user = get_default_test_user()
+        return self._user
```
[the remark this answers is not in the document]
r4vi: I thought I'd use it a few times but yeah can just call it once

datahub/company/test/test_consent_client.py
```diff
@@ -0,0 +1,101 @@
...
+@pytest.fixture
+def consent_client():
+    """
+    Fixture to return the consent module
+
+    Using a fixture instead of importing it globally
+    to insure that it doesn't cache any config between test runs
+    """
+    from datahub.company import consent
+    yield consent
```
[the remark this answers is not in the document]
r4vi: the tests are using `consent_client` but I'm overriding the settings in `settings.test` so it's probably not required anymore

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -0,0 +1,165 @@
...
+            telephone_alternative='987654321',
+            address_same_as_company=False,
+            address_1='Foo st.',
+            address_2='adr 2',
+            address_town='London',
+            address_county='London',
+            address_country_id=constants.Country.united_kingdom.value.id,
+            address_postcode='SW1A1AA',
+            notes='lorem ipsum',
+            accepts_dit_email_marketing=False,
```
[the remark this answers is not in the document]
r4vi: I guess they don't need to be, just copied the data from another test :+1:

datahub/company/serializers.py
```diff
@@ -207,6 +210,47 @@ class Meta:
...
+    def _notify_consent_service(self, validated_data):
+        """
+        Trigger the update_contact_consent task with the current version
+        of `validated_data`. The actual enqueuing of the task happens in the
+        on_commit hook so it won't actually notify the consent service unless
+        the database transaction was successful.
+        """
+        combiner = DataCombiner(self.instance, validated_data)
+        # If consent value in POST, notify
+        if 'accepts_dit_email_marketing' in validated_data:
```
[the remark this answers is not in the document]
r4vi: saves running the datacombiner too, I like it

datahub/company/test/serializers/test_contact_serializer.py
r4vi: I'll update for consistency

r4vi:
@reupen I think I've addressed all of your comments, can I get a :+1: ?

datahub/company/serializers.py
```diff
@@ -207,6 +209,51 @@ class Meta:
...
 
+    def _notify_consent_service(self, validated_data):
+        """
+        Trigger the update_contact_consent task with the current version
+        of `validated_data`. The actual enqueuing of the task happens in the
+        on_commit hook so it won't actually notify the consent service unless
+        the database transaction was successful.
+        """
+        if 'accepts_dit_email_marketing' not in validated_data:
+            # If no consent value in POST body
```
reupen: I think this comment could be clearer as it could be a POST or PATCH. Maybe 'request body' would be more generic.
```suggestion
            # If no consent value in request body
```

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -0,0 +1,132 @@
...
+@freeze_time('2020-03-13T14:21:24.367265+00:00')
+class TestContactSerializer:
+    """
+    Tests for the Contact Serializer. Checking that update / create notify the
+    consent service correctly.
+    """
+
+    def _make_contact(self):
+        contact = ContactFactory(
+            adviser=get_default_test_user(),
```
reupen: I don' think setting `adviser` is necessary (I don't believe this field is even used now...)

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture
+def update_contact_task_mock(monkeypatch):
+    """Mocks the apply_async method of the update_contact_consent celery task"""
+    m = Mock()
+    monkeypatch.setattr('datahub.company.serializers.update_contact_consent.apply_async', m)
+    yield m
+
+
+@freeze_time('2020-03-13T14:21:24.367265+00:00')
```
reupen: Might be nice to do `FROZEN_TIME = ...` above and then reference that here and in the tests.

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -0,0 +1,132 @@
...
+            'address_county': 'London',
+            'address_country': {
+                'id': constants.Country.united_kingdom.value.id,
+            },
+            'address_postcode': 'SW1A1AA',
+            'notes': 'lorem ipsum',
+            'accepts_dit_email_marketing': True,
+        }
+        c = ContactSerializer(data=data)
+        c.is_valid()
```
reupen: Since the return value is not checked, maybe:
```suggestion
        c.is_valid(raise_exception=True)
```

datahub/company/consent.py
```diff
@@ -0,0 +1,108 @@
...
+CONSENT_SERVICE_READ_TIMEOUT = 30.0
+
+
+def _get_client():
+    """
+    Get configured API client for the consent service,
+
+    not set module level as it makes using test utilities like
+    @override_settings impossible because the settings will be loaded
+    at import time if this was top-level.
```
elcct: I think would be great to clarify this as without context this comment sounds confusing.

r4vi: hopefully it's clearer now

adamledwards: I'm also confused about the context of this comment

r4vi: have updated it, let me know if you're still confused

adamledwards: Ahh yes I get it now.  So in the dnb service and company matching service the configuration is checked right before the request is made. So what you could do is.
This is just information purposes
```python
api_client = APIClient(....args)

def make_request(url, extras):
    if not all([
        settings.CONSENT_SERVICE_HAWK_ID,
        settings.CONSENT_SERVICE_HAWK_KEY,
        settings.CONSENT_SERVICE_BASE_URL,
    ]):
        raise ImproperlyConfigured(
            'CONSENT_SERVICE_* environment variables must be set, see README',
        )
    api_client.request(
        'post',
        CONSENT_SERVICE_PERSON_PATH,
       **extras
    )
```

adamledwards:
Also the links went out of sync due to a force push. I would say try avoid until its approved

r4vi:
> Also the links went out of sync due to a force push. I would say try avoid until its approved

yeah sorry, been trying to keep it rebased on top of develop hence the force pushes because it's been open so long

PR #2676, opened 2020-03-10
Company Matching API

adamledwards:
#### Context
This PR is part of a bigger piece of work around export wins. We plan to expose wins from the export-wins-data project and display them in DataHub.
Eventually DataHub will call export-wins and expose wins related to a company.
#### Changes in this PR
PR adds `company_matching_service_api` to a match_id from the [matching algorithm service](https://github.com/uktrade/company-matching-service/blob/develop/api.md).
I've taken inspiration from the dnb_api on structuring the api, testing and handling errors.

In the test you may notice here is a monkeypatch for the Hawk authentication to prevent Hawk authentication attempting to validate the response as this results in an error.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/company/company_matching_api.py
```diff
@@ -0,0 +1,125 @@
...
+        raise ImproperlyConfigured('The all COMPANY_MATCHING_SERVICE_* setting must be set')
+
+    response = api_client.request(
+        'POST',
+        'api/v1/company/match/',
+        json=query_param,
+        timeout=3.0,
+    )
+    """Question to reviewers do I need the below?"""
+    # statsd.incr(f'company_matching.match.{response.status_code}')
```
adamledwards: Should I be tracking the response code??

datahub/company/company_matching_api.py
```diff
@@ -0,0 +1,125 @@
...
+def model_to_match_payload(company):
+    """Format the Company model to json for the POST body."""
+    description = {
+        'id': str(company.id),
+        'company_name': company.name,
+        'companies_house_id': company.company_number,
+        'duns_number': company.duns_number,
+        'postcode': company.address_postcode,
+        # Question to reviewer is the cdms_ref and reference_code the same?
+        'cdms_ref': company.reference_code,
```
adamledwards: Is the cdms_ref and reference_code the same have I used the correct mapping?

datahub/company/serializers.py
```diff
@@ -743,3 +743,30 @@ class OneListCoreTeamMemberSerializer(serializers.Serializer):
...
+    """
+    Serializer for formating Company object for POST data to the company matching service
+    """
+
+    id = serializers.UUIDField()
+    company_name = serializers.CharField(source='name', required=False)
+    companies_house_id = serializers.CharField(source='company_number', required=False)
+    duns_number = serializers.CharField(required=False)
+    postcode = serializers.CharField(source='address_postcode', required=False)
+    cdms_ref = serializers.CharField(source='reference_code', required=False)
```
adamledwards: Is this correct is mapping for cdms_ref and reference_code

datahub/company/serializers.py
```diff
@@ -743,3 +743,30 @@ class OneListCoreTeamMemberSerializer(serializers.Serializer):
 
     adviser = NestedAdviserWithEmailAndTeamGeographyField()
     is_global_account_manager = serializers.BooleanField()
+
+
+class CompanyMatchingRequestSerializer(serializers.Serializer):
```
reupen: Are we planning to use this serialiser for anything else? I'm not sure it's giving us much benefit here – we don't seem to be using much of its features (type conversion, validation, persistence etc.) so a plain function might be simpler.

datahub/company/test/test_company_matching_api.py
```diff
@@ -0,0 +1,163 @@
...
+        requests_mock,
+        build_company,
+        expected_result,
+    ):
+        """
+        Test that the function maps the Company object to JSON correctly
+        also stripping out falsy values.
+        """
+        company = build_company()
+        monkeypatch.setattr(HawkAuth, '__call__', lambda self, r: r)
```
reupen: Rather than disabling HawkAuth, I think the correct thing to do is add the `Server-Authorization` header to the requests_mock response headers (e.g. using [a dynamic response](https://requests-mock.readthedocs.io/en/latest/response.html#dynamic-response) and [`mohawk`](https://mohawk.readthedocs.io/en/latest/usage.html#responding-to-a-request)).

But, if that's too difficult, I think it would be better to make the patch more specific e.g. patch `_verify_responses` on the auth instance on the APIClient instance.

datahub/company/serializers.py
reupen: `reference_code` is the CDMS reference, yes.

datahub/company/company_matching_api.py
```diff
@@ -0,0 +1,112 @@
...
+    ]):
+        raise ImproperlyConfigured('The all COMPANY_MATCHING_SERVICE_* setting must be set')
+
+    response = api_client.request(
+        'POST',
+        'api/v1/company/match/',
+        json=query_param,
+        timeout=3.0,
+    )
+    return response
```
adamledwards: shall I add `statsd.incr(f'company_matching.match.{response.status_code}')` ?

datahub/company/serializers.py
adamledwards: No I'm not planning to use this anywhere and see your point I notice that @alixedi had done a similar implementation on his PR . 
https://github.com/uktrade/data-hub-api/pull/2685/commits/03c6946aa1eefe5cc41bb512870477351d838567#diff-fe79eed29fbe7689da8347e725629b5bR262

reupen: Regarding `ChangeRequestSerializer`, I suggested that there as that one's taking external input from the view so it is important to validate types etc. before digging into the data structure.

datahub/company/test/test_company_matching_api.py
adamledwards: So I've made a test utility to help with this that validates the responses

datahub/company/company_matching_api.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+
+class CompanyMatchingServiceConnectionError(CompanyMatchingServiceException):
+    """
+    Exception for when an error was encountered when connecting to Company matching service.
+    """
+
+
+def request_match_companies(query_param):
```
currycoder: `query_param` made me thought this was going to be a URL query parameter, could it be renamed to something a bit less ambiguous?

datahub/company/company_matching_api.py
```diff
@@ -0,0 +1,105 @@
...
+    description = {
+        'id': str(company.id),
+        'company_name': company.name,
+        'companies_house_id': company.company_number,
+        'duns_number': company.duns_number,
+        'postcode': company.address_postcode,
+        'cdms_ref': company.reference_code,
+    }
+
+    """Function to remove all falsy values from the dict."""
```
currycoder: I think these can just be standard inline comments e.g. `# Function to remove all falsy values from the dict`

We generally use the multiline string for docstrings on functions

datahub/company/test/test_company_matching_api.py
```diff
@@ -0,0 +1,141 @@
...
+                    'company_name': 'Name 1',
+                    'companies_house_id': '00000000',
+                    'postcode': 'W1',
+                },
+            ),
+        ),
+    )
+    def test_model_to_match_payload(
+        self,
+        monkeypatch,
```
currycoder: Looks like this isn't used?

adamledwards: ah yes left over from a refactor

datahub/company/company_matching_api.py
elcct: I agree this could be renamed as it may be confusing what it is expected to be received.

adamledwards: let's go with json_body

datahub/core/test_utils.py
```diff
@@ -487,3 +487,62 @@ def resolve_value(value):
         return value
 
     return resolve_data(data, value_resolver=resolve_value)
+
+
+class HawkMockJSONResponse:
```
reupen: Nice one!

datahub/core/test_utils.py
```diff
@@ -487,3 +487,62 @@ def resolve_value(value):
...
+    dynamic_reponse.set_reponse({'content': 'Hello'})
+
+    requests_mock.post(
+        'some/api/',
+        status_code=status.HTTP_200_OK,
+        json=dynamic_reponse,
+    )
+    """
+
+    def __init__(self, id, key, algorithm='sha256'):
```
reupen: `id` here shadows the built-in `id()` function – would suggest `id_` or a prefix.

datahub/core/test_utils.py
```diff
@@ -487,3 +487,62 @@ def resolve_value(value):
...
+            'algorithm': algorithm,
+        }
+        self.response = {}
+
+    def set_reponse(self, response):
+        """
+        Set a dict that can be serialized to json
+        this reponse body will be validated and returned in the mock reponse.
+        """
+        self.response = {}
```
reupen: Maybe it would be simpler to just let `response` be passed to `__init__()`?

datahub/core/test_utils.py
```diff
@@ -487,3 +487,62 @@ def resolve_value(value):
...
+            request.method,
+            content=request.text,
+            content_type=request.headers['Content-Type'],
+        )
+        receiver.respond(
+            content=response,
+            content_type=request.headers['Content-Type'],
+        )
+        context.headers['Server-Authorization'] = receiver.response_header
+        context.headers['Content-Type'] = request.headers['Content-Type']
```
reupen: I think we shouldn't copy the content type from the request – would suggest adding `content_type='application/json'` to `__init__()` (so there is a sensible default but it's still overridable).

PR #2680, opened 2020-03-11
Add sso email user id to Adviser model

elcct:
### Description of change

This adds `sso_email_user_id` field to Adviser model. This is an unique user id field in email format that is being used by Staff SSO and will be used to match Data Hub users. 
The field has 254 character length as this is the length of the field in Staff SSO.

This is a first step towards improving integration with Staff SSO and making it the source of truth.  More PRs will follow.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


mcbhenwood:
Would this benefit from just one test which verifies that `sso_email_user_id` is now persisted on the Adviser?

datahub/company/models/adviser.py
```diff
@@ -87,6 +87,12 @@ class Advisor(AbstractBaseUser, PermissionsMixin):
     )
     date_joined = models.DateTimeField('date joined', default=now)
 
+    sso_email_user_id = models.EmailField(
+        default='',
+        max_length=254,  # The length of the field in Staff SSO is 254 characters.
+        help_text='A user ID in email compatible format used for Staff SSO authentication',
+    )
```
reupen: I think we should have `unique=True` (which means `null=True` too unfortunately). Should be able to remove the default if we add `blank=True`.

datahub/company/models/adviser.py
```diff
@@ -87,6 +87,12 @@ class Advisor(AbstractBaseUser, PermissionsMixin):
     )
     date_joined = models.DateTimeField('date joined', default=now)
 
+    sso_email_user_id = models.EmailField(
+        default='',
+        max_length=254,  # The length of the field in Staff SSO is 254 characters.
+        help_text='A user ID in email compatible format used for Staff SSO authentication',
```
reupen: Could we be more explicit about what this is since we'll be exposing it in the admin site later one? E.g. `This is the 'Email user ID' that is shown for this user in Staff SSO.`

datahub/company/models/adviser.py
elcct: I wanted to avoid it, but on a second thought you're right. I have updated it.

datahub/company/models/adviser.py
elcct: I have updated the help text, thank you!

datahub/company/models/adviser.py
```diff
@@ -87,6 +87,14 @@ class Advisor(AbstractBaseUser, PermissionsMixin):
     )
     date_joined = models.DateTimeField('date joined', default=now)
 
+    sso_email_user_id = models.EmailField(
+        blank=True,
```
reupen: Would be good to set `verbose_name` to `'SSO email user ID'` actually (as otherwise we'll probably get `Sso email user id` in the admin site).

elcct: Nice! I have updated the model now. Thank you.

elcct:
@mcbhenwood good point! I added tests for the new field now.

PR #2682, opened 2020-03-11
Add workaround for Django 3 memory leak under gevent

reupen:
### Description of change

We've been suffering from a problematic memory leak since upgrading to Django 3 which is believed to be caused by https://github.com/django/asgiref/issues/144:

image

The memory leak is causing Gunicorn worker processes to exhaust available memory and crash and restart.

This adds a workaround which is inactive by default. This is so it can be we can verify that it fixes the problem in e.g. the dev environment.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/config/gunicorn.py
+++ b/config/gunicorn.py
@@ -41,6 +41,9 @@ worker_connections = os.environ.get('GUNICORN_WORKER_CONNECTIONS', '10')
 _enable_async_psycopg2 = (
     os.environ.get('GUNICORN_ENABLE_ASYNC_PSYCOPG2', 'true').lower() in ('true', '1')
 )
+_patch_asgiref = (
+    os.environ.get('GUNICORN_PATCH_ASGIREF', 'false').lower() in ('true', '1')
+)
 
 
 def post_fork(server, worker):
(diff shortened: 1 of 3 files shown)


config/gunicorn.py
```diff
@@ -49,6 +52,17 @@ def post_fork(server, worker):
...
     """
-    if worker_class == 'gevent' and _enable_async_psycopg2:
-        patch_psycopg()
-        worker.log.info('Enabled async Psycopg2')
+    if worker_class == 'gevent':
+        if _enable_async_psycopg2:
+            patch_psycopg()
+            worker.log.info('Enabled async Psycopg2')
+
+        # Temporary workaround for https://github.com/django/asgiref/issues/144.
```
elcct: Maybe would be good to add `TODO: ` and reword slightly so it is easy to find, when looking for temporary things across the codebase.

reupen: :+1:

reupen:
The fix appears to have been successful:

image

PR #2685, opened 2020-03-12
Add serializer for change-requests

alixedi:
### Description of change

This enriches the stub `DNBCompanyChangeRequestView` introduced in #2678 with a Serializer for validating the change-requests translating them to the `dnb-service` format.

At the moment, the view returns validated and translated change-request payload. This is temporary & deliberate.

In the following PR, I will introduce utility functions that will send the change-requests over to `dnb-service` and return the response of the `dnb-service`.

### Checklist

* [x] If this is a releasable change, has a news fragment been added?

  <details>
  <summary>Explanation</summary>
  
  A news fragment is required for any releasable change (i.e. code that runs in or affects production) so that a corresponding changelog entry is added when releasing.
  
  Check [changelog/README.md](https://github.com/uktrade/data-hub-api/blob/master/changelog/README.md) for instructions.
  
  </details>
  
* [ ] Has this branch been rebased on top of the current `develop` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [ ] Is the CircleCI build passing?

### General points

<details>
<summary><strong>Other things to check</strong></summary><p></p>

* Make sure `fixtures/test_data.yaml` is maintained when updating models
* Consider the admin site when making changes to models
* Use select-/prefetch-related field lists in views and search apps, and update them when fields are added
* Make sure the README is updated e.g. when adding new environment variables

</details>

See [docs/CONTRIBUTING.md](https://github.com/uktrade/data-hub-api/blob/develop/docs/CONTRIBUTING.md) for more guidelines.

PR #2688, opened 2020-03-12
Send change-requests to dnb-service

alixedi:
### Description of change

This PR follows from #2685 which set up the Serializer for validating change-requests.

This PR adds utility functions for sending these change-requests to `dnb-service` and returning the response back to the client.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/dnb_api/test/test_views.py
+++ b/datahub/dnb_api/test/test_views.py
@@ -8,7 +8,10 @@ from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.test.utils import override_settings
 from freezegun import freeze_time
-from requests.exceptions import ConnectionError
+from requests.exceptions import (
+    ConnectionError,
+    Timeout,
+)
 from rest_framework import status
 from rest_framework.reverse import reverse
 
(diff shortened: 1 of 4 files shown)


datahub/dnb_api/test/test_views.py
```diff
@@ -1508,20 +1579,150 @@ def test_valid(
...
+            status_code=status.HTTP_201_CREATED,
+            json=dnb_response,
+        )
+
         response = self.api_client.post(
             reverse('api-v4:dnb-api:company-change-request'),
             data=change_request,
         )
 
+        assert json.loads(requests_mock.last_request.body) == dnb_request
```
adamledwards: You could use `matcher.last_request.json()`

datahub/dnb_api/utils.py
```diff
@@ -389,3 +389,49 @@ def rollback_dnb_company_update(
...
+        error_message = 'Encountered an error connecting to DNB service'
+        logger.error(error_message)
+        raise DNBServiceConnectionError(error_message) from exc
+
+    except Timeout as exc:
+        error_message = 'Encountered a timeout interacting with DNB service'
+        logger.error(error_message)
+        raise DNBServiceTimeoutError(error_message) from exc
+
+    if dnb_response.status_code != status.HTTP_201_CREATED:
```
adamledwards: Just a question does this endpoint ever return any other 20* status.
just checking its not a possibility to get this message
`DNB service returned an error status: 200`

alixedi: The upstream endpoint returns `201` on success.

currycoder: I think it would be good to allow 200's as well.  It might be that some subtle change happens on the service later which means a 200 is returned and it would be good if this didn't cause a 500 error.

datahub/dnb_api/utils.py
```diff
@@ -389,3 +389,49 @@ def rollback_dnb_company_update(
...
+        error_message = f'DNB service returned an error status: {dnb_response.status_code}'
+        logger.error(error_message)
```
currycoder: Does this error need logging in addition to the raised error in the view class which calls this util?  Just thinking we might be doubling up on error logs here...

datahub/dnb_api/utils.py
reupen: Yes I agree. Error codes are only 4xx and 5xx. `raise_for_status()` handles this well, responses have an `ok` attribute too.

datahub/dnb_api/utils.py
```diff
@@ -389,3 +389,49 @@ def rollback_dnb_company_update(
...
+        dnb_response = _request_changes(
+            {
+                'duns_number': duns_number,
+                'changes': changes,
+            },
+        )
+
+    except ConnectionError as exc:
+        error_message = 'Encountered an error connecting to DNB service'
+        logger.error(error_message)
```
reupen: Would consider `logger.exception()` for these.

PR #2695, opened 2020-03-16
Add Staff SSO introspection API request utility function

reupen:
### Description of change

This adds a utility function for making introspection API requests to Staff SSO.

This will be used as part of the new introspection logic that will replace what we're currently using via django-oauth-toolkit.

Some new environment variables were added as part of this (to eventually replace `RESOURCE_SERVER_INTROSPECTION_URL` and `RESOURCE_SERVER_AUTH_TOKEN`). They are only required when `SSO_ENABLED=True`. (They'll need to be added to the CircleCI configuration in the front end too – I'll open a PR there for that.)

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/oauth/sso_api_client.py
```diff
@@ -0,0 +1,63 @@
...
+
+    As this deals with authentication, this aims to be more robust than might be the case
+    for other API requests.
+    """
+    api_client = _get_api_client()
+
+    try:
+        response = api_client.request(method, path, **kwargs)
+    except RequestException as exc:
+        raise SSORequestError('SSO request failed') from exc
```
elcct: Should we end the messages with a full stop?

reupen: Good question. Seems like we aren't consistent about that – all the ones I can make Python itself and the standard library emit seem to not end in full stops. Any other reference saying that they should?

elcct: That is true. I would think that punctuation mark would tell me that I received complete message about the error. In the past, I would encounter a log message written to disk incomplete when the app completely crashes, but I understand this could unlikely happen. I was trying to find any reference, but surprisingly I couldn't apart from debates like this: https://stackoverflow.com/questions/1136829/do-you-end-your-exception-messages-with-a-period
Would be great if we could become consistent, so if standard library doesn't use stops, let's not use them. I'll try to remember that :)

PR #2697, opened 2020-03-16
Add OMIS invoice detail Hawk authenticated endpoint

elcct:
### Description of change

As a part of deeper Staff SSO integration work we need to change public OMIS endpoints that are currently being authenticated with OAuth, to use Hawk authentication instead.
This is the first of a number of PRs that will follow. This moves existing OMIS public invoice detail view to corresponding legacy files and creates new Hawk authenticated view with exactly the same functionality.
Once all views are re-created with Hawk authentication, the existing legacy views will be deprecated and public facing OMIS app will be updated to use new endpoints. The last step will be to remove all legacy OMIS endpoints.

New URLs for Hawk authenticated OMIS endpoints will have a following format:

`/v3/public/omis/...`

We have a pattern for "public" search endpoints that start with `/v4/public` prefix.

The endpoint has not been upgraded to `v4` because that may require additional changes in the public facing app to support `v4`'s address format. This work could be done once migration is completed.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/config/settings/types.py
+++ b/config/settings/types.py
@@ -14,3 +14,4 @@ class HawkScope(Enum):
     public_company = auto()
     data_flow_api = auto()
     metadata = auto()
+    public_omis = auto()
--- a/README.md
+++ b/README.md
@@ -321,10 +321,12 @@ Data Hub API can run on any Heroku-style platform. Configuration is performed vi
 | `MI_DATABASE_SSLCERT` | No | base64 encoded client certificate for MI database connection. |
 | `MI_DATABASE_SSLKEY` | No | base64 encoded client private key for MI database connection. |
 | `MI_FDI_DASHBOARD_TASK_DURATION_WARNING_THRESHOLD` | No | Threshold (in seconds) for emitting warnings about long transfer duration (default=600). |
+| `OMIS_PUBLIC_ACCESS_KEY_ID` | No | A non-secret access key ID, corresponding to `OMIS_PUBLIC_SECRET_ACCESS_KEY`. The holder of the secret key can access the OMIS public endpoints by Hawk authentication. |
 | `OMIS_NOTIFICATION_ADMIN_EMAIL`  | Yes | |
 | `OMIS_NOTIFICATION_API_KEY`  | Yes | |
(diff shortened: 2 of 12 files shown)


config/settings/types.py
```diff
@@ -14,3 +14,4 @@ class HawkScope(Enum):
     public_company = auto()
     data_flow_api = auto()
     metadata = auto()
+    omis = auto()
```
reupen: Maybe `public_omis` would make it clearer that's it's not all of OMIS?
```suggestion
    public_omis = auto()
```

datahub/omis/invoice/test/views/test_public_invoice_details.py
```diff
@@ -14,9 +13,76 @@
...
+        url = reverse(
+            'api-v3:public-omis:invoice:detail',
+            kwargs={'public_token': order.public_token},
+        )
+        public_omis_api_client.set_http_x_forwarded_for('1.1.1.1')
+        response = public_omis_api_client.get(url)
+
+        assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+    @pytest.mark.parametrize('verb', ('post', 'patch'))
```
reupen: Did we mean to lose `'delete'` from this?

elcct: I was thinking to add `delete` as a separate PR, as this could be added to another test that is not related to OMIS.

config/settings/types.py
elcct: Good idea! Thank you

datahub/omis/invoice/test/views/test_public_invoice_details.py
reupen: You mean as `HawkAPITestClient` doesn't have a delete method? I think since the old test was testing delete, we should sort that if it's easy.

elcct: Okay I added that case, it was actually small change. Thank you!

PR #2705, opened 2020-03-19
Add new authentication class with SSO email user ID support

reupen:
### Description of change

This adds a new Django Rest Framework authentication class (that uses OAuth token introspection with Staff SSO). This will eventually replace the currently used authentication class provided by Django OAuth Toolkit.

For further background [have a look at this document](https://docs.google.com/document/d/1n6aQXXyfX8auecUnF4dqQaX8-QJUgAhPwsbnkq0NobM).

The main reason for the new class to help us migrate to using the SSO email user ID to link advisers in Data Hub with users in Staff SSO.

Some behaviour of the new class:

- The new class has support for looking up advisers using the unique SSO email user ID rather than the user's primary application email address. It also contains some migration logic for users that do not have an SSO email user ID set.

- The new class also uses Redis to cache introspected access tokens rather than a Django model.

- The new class does not automatically create stub users at present. This will hopefully not be needed if a better add adviser flow is added to the admin site.

- The new class doesn't bother with OAuth scopes, as we no longer need them with machine-to-machine endpoints (that don't involve a user) now using Hawk.

The new class is disabled by default until further related functionality is also added. However, it can be enabled using the `STAFF_SSO_USE_NEW_INTROSPECTION_LOGIC` setting.

This is deployed to the UAT environment (with `STAFF_SSO_USE_NEW_INTROSPECTION_LOGIC=True`) so it can be tested there. (Testing locally with a real Staff SSO instance is more difficult, but if you'd like to, let me know and I'll try and help.)

[This is the main part of the equivalent code in Django OAuth Toolkit](https://github.com/jazzband/django-oauth-toolkit/blob/fd0a0d8a260f9ca41785d26a03925811c9462ad2/oauth2_provider/oauth2_validators.py#L271-L382) if you'd like to compare.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/oauth/auth.py
+++ b/datahub/oauth/auth.py
@@ -1,0 +1,134 @@
+import time
+from logging import getLogger
+
+from django.conf import settings
+from django.core.cache import cache
+from rest_framework.authentication import BaseAuthentication
+from rest_framework.exceptions import AuthenticationFailed
+
+from datahub.company.models import Advisor
+from datahub.oauth.sso_api_client import introspect_token, SSORequestError, SSOTokenDoesNotExist
+
+NO_CREDENTIALS_MESSAGE = 'Authentication credentials were not provided.'
+INCORRECT_CREDENTIALS_MESSAGE = 'Incorrect authentication credentials.'
+INCORRECT_SCHEME = 'Incorrect authentication scheme.'
+INVALID_CREDENTIALS_MESSAGE = 'Invalid authentication credentials.'
+
+
+logger = getLogger(__name__)
+
+
+class SSOIntrospectionAuthentication(BaseAuthentication):
+    """OAuth token introspection (RFC 7662) authentication class, with Staff SSO extensions."""
+
+    def authenticate_header(self, request):
+        """
+        The value for the WWW-Authenticate for when credentials aren't provided.
+
+        This returns the same value as django-oauth-toolkit.
+        """
+        return 'Bearer realm="api"'
+
+    def authenticate(self, request):
+        """
+        Authenticate the user using token introspection.
+
+        This first checks if the token is cached. If it's not cached, the token is looked
+        up in Staff SSO. An adviser is then looked up using the retrieved token data.
+        """
+        try:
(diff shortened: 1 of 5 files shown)


datahub/oauth/auth.py
```diff
@@ -0,0 +1,134 @@
...
+    This first checks the cache, and falls back to querying Staff SSO if the token isn't cached.
+    """
+    cache_key = f'access_token:{token}'
+    cached_token_data = cache.get(cache_key)
+
+    if cached_token_data:
+        return cached_token_data
+
+    try:
+        token_data = introspect_token(token)
```
elcct: I wonder if we could cache invalid token somehow, so that a repeated request, that cannot be authenticated, wouldn't cause service congestion by triggering subsequent introspection calls.

reupen: Good thinking. But, I think:

- the previous implementation wasn't doing that
- it might open up the possibility of someone requesting lots of different tokens and filling up Redis
- if there's a temporary problem with Staff SSO, we might end up caching that

datahub/oauth/auth.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # No match on sso_email_user_id – fall back to looking up using the email field
+    # TODO: Remove the below logic once active users have sso_email_user_id set
+    try:
+        # Note: The email field uses CICharField so this lookup is case-insensitive
+        adviser = _get_adviser(email=username, sso_email_user_id__isnull=True)
+    except Advisor.DoesNotExist:
+        return None
+
+    adviser.sso_email_user_id = sso_email_user_id
```
currycoder: Could it be better to not call `save()` here when the adviser was retrieved with their `sso_email_user_id`?  My understanding is that there's no value to update the field in this case.

reupen: There's a return on line 111, so (if I've understood the point correctly) that shouldn't happen. Open to any suggestions to make that clearer...

PR #2722, opened 2020-03-25
ContactSerializer: lookup consent from consent service [4 of 4]

r4vi:
### Description of change

Update ContactSerializer so that `ContactViewSet.detail` calls the consent service to determine if a contact `accepts_dit_email_marketing`. This lookup will only happen if the feature flag `GET_CONSENT_FROM_CONSENT_SERVICE` is active, otherwise this will not change the existing behaviour

depends on: https://github.com/uktrade/data-hub-api/pull/2657

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -8,8 +8,10 @@ from django.utils.timezone import now
 from django.utils.translation import gettext_lazy
 from rest_framework import serializers
 
+from datahub.company import consent
 from datahub.company.constants import (
     BusinessTypeConstant,
+    GET_CONSENT_FROM_CONSENT_SERVICE,
     OneListTierID,
 )
 from datahub.company.models import (
@@ -47,6 +49,7 @@ from datahub.core.validators import (
     RulesBasedValidator,
     ValidationRule,
 )
+from datahub.feature_flag.utils import is_feature_flag_active
 from datahub.metadata import models as meta_models
 from datahub.metadata.serializers import TeamWithGeographyField
 
@@ -149,6 +152,19 @@ class ContactSerializer(PermittedFieldsModelSerializer):
     archived_reason = serializers.CharField(read_only=True)
     archived_by = NestedAdviserField(read_only=True)
     primary = serializers.BooleanField()
+
+    def to_representation(self, instance):
+        """
+        Convert instance to dict.
+
+        Optionally lookup from consent service if feature flag
+        is enabled.
+        """
+        representation = super().to_representation(instance)
+        if is_feature_flag_active(GET_CONSENT_FROM_CONSENT_SERVICE):
+            representation['accepts_dit_email_marketing'] = \
+                consent.get_one(representation['email'])
+        return representation
 
     class Meta:
         model = Contact
--- a/datahub/company/test/serializers/test_contact_serializer.py
+++ b/datahub/company/test/serializers/test_contact_serializer.py
@@ -18,9 +18,9 @@ FROZEN_TIME = '2020-03-13T14:21:24.367265+00:00'
 @pytest.fixture
 def update_contact_task_mock(monkeypatch):
     """Mocks the apply_async method of the update_contact_consent celery task"""
-    m = Mock()
-    monkeypatch.setattr('datahub.company.serializers.update_contact_consent.apply_async', m)
-    yield m
+    mock = Mock()
+    monkeypatch.setattr('datahub.company.serializers.update_contact_consent.apply_async', mock)
+    yield mock
 
 
 @pytest.fixture
--- a/datahub/company/test/serializers/test_contact_serializer.py
+++ b/datahub/company/test/serializers/test_contact_serializer.py
@@ -3,9 +3,11 @@ from unittest.mock import Mock
 import pytest
 from freezegun import freeze_time
 
+from datahub.company.constants import GET_CONSENT_FROM_CONSENT_SERVICE
 from datahub.company.serializers import ContactSerializer
 from datahub.company.test.factories import CompanyFactory, ContactFactory
 from datahub.core import constants
+from datahub.feature_flag.test.factories import FeatureFlagFactory
 
 # mark the whole module for db use
 pytestmark = pytest.mark.django_db
@@ -18,6 +20,14 @@ def update_contact_task_mock(monkeypatch):
     """Mocks the apply_async method of the update_contact_consent celery task"""
     m = Mock()
     monkeypatch.setattr('datahub.company.serializers.update_contact_consent.apply_async', m)
+    yield m
+
+
+@pytest.fixture
+def consent_get_one_mock(monkeypatch):
+    """Mocks the .get_one function of the `datahub.company.consent` module"""
+    m = Mock()
+    monkeypatch.setattr('datahub.company.consent.get_one', m)
     yield m
 
 
@@ -129,4 +139,27 @@ class TestContactSerializer:
         update_contact_task_mock.assert_called_once_with(
             args=(data['email'], data['accepts_dit_email_marketing']),
             kwargs={'modified_at': FROZEN_TIME},
+        )
+
+    def test_to_representation_feature_flag_off(self, consent_get_one_mock):
+        """Ensure that when feature flag is off, the consent service is not called"""
+        FeatureFlagFactory(code=GET_CONSENT_FROM_CONSENT_SERVICE, is_active=False)
+        contact = self._make_contact()
+        c = ContactSerializer(instance=contact, partial=True)
+        _ = c.data
+        assert not consent_get_one_mock.called
+
+    @pytest.mark.parametrize('val', (True, False))
+    def test_to_representation_feature_flag_on(self, consent_get_one_mock, val):
+        """
+        When feature flag enabled ensure that value returned
+        from consent service is reflected in `ContactSerializer.data`
+        """
+        consent_get_one_mock.return_value = val
+        FeatureFlagFactory(code=GET_CONSENT_FROM_CONSENT_SERVICE, is_active=True)
+        contact = self._make_contact()
+        c = ContactSerializer(instance=contact)
+        assert c.data['accepts_dit_email_marketing'] == val
+        consent_get_one_mock.assert_called_once_with(
+            c.instance.email,
         )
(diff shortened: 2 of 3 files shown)


r4vi:
:warning: Merge https://github.com/uktrade/data-hub-api/pull/2657/files first, before merging this PR

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -130,3 +140,26 @@ def test_serializer_create_calls_task(self, update_contact_task_mock, synchronou
...
+    def test_to_representation_feature_flag_off(self, consent_get_one_mock):
+        """Ensure that when feature flag is off, the consent service is not called"""
+        FeatureFlagFactory(code=GET_CONSENT_FROM_CONSENT_SERVICE, is_active=False)
+        contact = self._make_contact()
+        c = ContactSerializer(instance=contact, partial=True)
+        _ = c.data
+        assert not consent_get_one_mock.called
+
+    @pytest.mark.parametrize('val', (True, False))
+    def test_to_representation_feature_flag_on(self, consent_get_one_mock, val):
```
elcct: Would be good to rename `val` to represent what actually it is e.g. `has_consented` or something similar.

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -130,3 +140,26 @@ def test_serializer_create_calls_task(self, update_contact_task_mock, synchronou
...
+        """
+        When feature flag enabled ensure that value returned
+        from consent service is reflected in `ContactSerializer.data`
+        """
+        consent_get_one_mock.return_value = val
+        FeatureFlagFactory(code=GET_CONSENT_FROM_CONSENT_SERVICE, is_active=True)
+        contact = self._make_contact()
+        c = ContactSerializer(instance=contact)
```
elcct: Would be good to rename `c` to something more descriptive e.g. `contact_serializer`.

datahub/company/serializers.py
```diff
@@ -150,6 +153,19 @@ class ContactSerializer(PermittedFieldsModelSerializer):
...
+        """
+        Convert instance to dict.
+
+        Optionally lookup from consent service if feature flag
+        is enabled.
+        """
+        representation = super().to_representation(instance)
+        if is_feature_flag_active(GET_CONSENT_FROM_CONSENT_SERVICE):
+            representation['accepts_dit_email_marketing'] = \
+                consent.get_one(representation['email'])
```
elcct: I understand this will make a request to the consent service every time contact is being serialised?

datahub/company/test/serializers/test_contact_serializer.py
```diff
@@ -21,6 +23,14 @@ def update_contact_task_mock(monkeypatch):
     yield m
 
 
+@pytest.fixture
+def consent_get_one_mock(monkeypatch):
+    """Mocks the .get_one function of the `datahub.company.consent` module"""
+    m = Mock()
```
elcct: Would be good to rename `m` to e.g. `mock` so it is easier to read. Looks like it would be great to change single letter variables to something more descriptive in all places.

datahub/company/serializers.py
r4vi: yes, that will be the source of truth and will get unsubscribes from the email marketing system, so everytime we're sending a representation of this object, that value should come from the consent service

adamledwards: I have reservation for this approach I'm afraid. The ContactSerializer is used by CompanySerializer that means it will be called often. I'm concerned about a performance bottleneck when calling `v4/company` e.g 10 companies with 3 contacts = 30 API calls in one request. 

One approach you can do is a custom list_seializer_class that gets all the emails related to one company. This will reduce the calls to one api call per company. Using the previous example this will bring the api calls down to 10.

However I think maybe we should look where accepts_dit_email_marketing is used (Think only the detail view in contacts)
and  find a solution where the api calls is only triggered in a detail view.

adamledwards: I'm also concerned about the lack of error handling in consent. `get_many` will be used in a api view there should be sufficient error handling at the moment if the consent service blows up it will take `v4/company` and the contact detail page down with it.  
https://github.com/uktrade/data-hub-api/blob/8ef5e62f7a96e868a36f9c08556022a3d0556a09/datahub/company/consent.py#L81-L101

adamledwards: Also I would like to add my apologies for not looking at this sooner.

reupen: > The ContactSerializer is used by CompanySerializer that means it will be called often.

That is a good point. We had an item on the tech board to remove the list of contacts from the company view – but it needs front end work as last time I looked the front end was using that field in some forms (for the contact drop-down)...

I was also under the impression that we agreed not to inflate contacts with the email marketing status in `GET /v4/contact` (i.e. when listing contacts) – but I can't see how that is avoided here at the moment...

mcbhenwood: Yes @reupen we did agree originally that the Contact list would omit that flag BUT the decision got reversed when (I believe) Anthony pointed out that we must surface that flag in the Contacts download.

mcbhenwood: So as is often the way in a project where people have light, infrequent touches, the decisions diverged. It's a sad and common antipattern in fragmented workloads :(

r4vi: The contact list download uses a different serializer and does bulk lookups per page (see https://github.com/uktrade/data-hub-api/pull/2719 ) so it would be quite reasonable to remove the `accepts_dit_email_marketing` from here. 

You might be able to split `ContactSerializer` used by `CompanySerializer` which wouldn't return that field and `CompanySerializer` used to provide the Company detail endpoint which would retain this field.

reupen: > Yes @reupen we did agree originally that the Contact list would omit that flag BUT the decision got reversed when (I believe) Anthony pointed out that we must surface that flag in the Contacts download.

I think we're talking about two different things – the search CSV download (which doesn't use `ContactSerializer`) and `GET /v3/contact` (which uses `ContactSerializer`).

We should still have the doc to refer to if in doubt :slightly_smiling_face:

adamledwards: IMO I think going forward we can do the following.

1. Remove `accepts_dit_email_marketing` from the `ContactSerializer `
2. Create 
```python
class ContactSerializerDetail(ContactSerializer):
    accepts_dit_email_marketing = SerializerMethodField()
   
   def get_accepts_dit_email_marketing(self):
        # api call logic

```` 
this is where the api call could be made.
3. Introduce more error handling to `get_many` so the page can still render if the consent service returns an error.
4. Inform the FE that `accepts_dit_email_marketing ` will now be optional with the values being `Boolean|Null`  (null being an error scenario)

My question is would we need to put in feature flags and omit the fields on `def to_representation`
Suggestions welcome?

@r4vi @markhigham @reupen what do you think?

Comment (r4vi) [reply datahub/company/serializers.py:166]: sounds good to me, 

we went around in circles a bit with what the front-end does in a failure scenario, I think we agreed that if we didn't get a response from consent it is best to default to `False`. 

If telling the user something failed is useful then obviously the FE needs to do something to handle the `null`, if we don't feel that is particularly useful the API could just send `False` and the FE doesn't have to handle the optional `null` and the contract stays the same.

Comment (adamledwards) [reply datahub/company/serializers.py:166]: > If telling the user something failed is useful then obviously the FE needs to do something to handle the null, if we don't feel that is particularly useful the API could just send False and the FE doesn't have to handle the optional null and the contract stays the same.

Ok that sounds ok, False it is.

Comment (reupen) [reply datahub/company/serializers.py:166]: >  3. Introduce more error handling to get_many so the page can still render if the consent service returns an error.

I think we need to define what should happen in that case, as if we introduce soft-failures then the front end needs to understand that and also do something to the form (disable the field?)

> My question is would we need to put in feature flags and omit the fields on def to_representation

It might be possible to do it wherever the serialiser class is being swapped out (in `get_serializer_class` on the view set)?

But I think we also need to consider that the update logic is currently in `ContactSerlializer`, and if that would mean splitting the update and read logic for `accepts_dit_email_marketing` over two serialiser classes, that might not be ideal. 

So we might want all of:

- POST /v3/contact
- GET /v3/contact/<id>
- PATCH /v3/contact/<id>

to use the same serialiser with the `accepts_dit_email_marketing` logic (also considering that they all return a single contact record).

Comment (mcbhenwood) [reply datahub/company/serializers.py:166]: I feel that before deciding finally, we should discuss this on a call. @reupen, @adamledwards and myself. There's a good chance that @adamledwards 's time will be pulled from this task so we need to make sure all the thinking is captured. It's hard to get involved in decisions like this when they're made in various threads and / or IM conversations unless you're one of the key contributors to the conversation. Will contact you separately to set a time.

**Comment by adamledwards** [comment]
It looks like we will be closing this and rasing two tickets. 
Ticket 1.
Add error handling around the API calls to the consent service. I believe there is a conversation around how API status and connection errors are handled in the frontend.

Ticket 2.
Remove accepts_dit_email_marketing field in the listing view only.

```python
class ContactSerializer(PermittedFieldsModelSerializer):
   """
   Used by
   POST /v3/contact
    GET /v3/contact
    PATCH /v3/contact
    and CompanySerializer
    
   remove accepts_dit_email_marketing from the listing serializer 
   """
    def to_representation(self, instance):
        """
        Optionally lookup from consent service if feature flag
        is enabled.
        """
        representation = super().to_representation(instance)
        if is_feature_flag_active(GET_CONSENT_FROM_CONSENT_SERVICE) and 'accepts_dit_email_marketing' in representation:
            representation['accepts_dit_email_marketing'] = consent.get_one(representation['email'])
        return representation


class ContactListSerializer(ContactSerializer):
   """
   Used by
    GET /v3/contact/
   """
   class Meta(ContactSerializer.Meta)
      exclude = ['accepts_dit_email_marketing']

                   
class ContactViewSet:
   #....
   serializer_class = ContactSerializer
   def get_serializer_class(self):
       if self.action == 'retrieve' and is_feature_flag_active(GET_CONSENT_FROM_CONSENT_SERVICE):
           return ContactListSerializer

        return super(ContactViewSet, self).get_serializer_class(
```

PR #2737, opened 2020-03-30
Record user event on introspection in SSOIntrospectionAuthentication

reupen:
### Description of change

This updates `SSOIntrospectionAuthentication` to create a `UserEvent` whenever an SSO access token introspection occurs.

This is for reporting needs (where reports were previously inferring this information from `AccessToken` objects).

It also makes some minor changes to the `ModelAdmin` configuration for `UserEvent`.

I'd recommend reviewing the commits individually.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/oauth/auth.py
+++ b/datahub/oauth/auth.py
@@ -1,5 +1,6 @@
 import time
 from logging import getLogger
+from typing import Optional, Tuple
 
 from django.conf import settings
 from django.core.cache import cache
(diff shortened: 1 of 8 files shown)


datahub/oauth/auth.py
```diff
@@ -49,52 +52,59 @@ def authenticate(self, request):
...
     This first checks the cache, and falls back to querying Staff SSO if the token isn't cached.
+
+    :returns: a 2-tuple of: (token data, was the token cached)
     """
     cache_key = f'access_token:{token}'
     cached_token_data = cache.get(cache_key)
 
     if cached_token_data:
-        return cached_token_data
+        return cached_token_data, True
```
alixedi: Given that the boolean is True in one case only and False otherwise, can we not call `record_user_event ` here and perhaps return the token only?

reupen: `record_user_event()` requires an adviser object which hasn't been looked up at this point. So the only way to do that would be to do the adviser look-up here and is-active check at both return points which would create a more confusing flow IMO (currently it is pretty much sequential).

PR #2749, opened 2020-04-02
Allow clients of interactions to set page size

samuelhwilliams:
### Description of change
In data-flow, we scrape interactions from this API every morning -
approx 3 mil records as it stands currently. At the moment, this is
paginated in groups of 100 records at a time, which obviously results in
a significant amount of requests. By increasing page size we hope to
make the task generally run faster. At the moment each request seems to
take around 0.1s, so increasing page size by a factor of 10 should be
fine. A factor of 100 should be manageable, though we don't intend to
use that as a client yet.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/changelog/interactions-endpoint-page-size.api.md
+++ b/changelog/interactions-endpoint-page-size.api.md
@@ -1,0 +1,1 @@
+`GET /v4/dataset/*`: enabled a `page_size` parameter that allows clients to change the default page size from 100, up to a maximum of 10,000.
--- a/datahub/dataset/core/pagination.py
+++ b/datahub/dataset/core/pagination.py
@@ -14,3 +14,10 @@ class DatasetCursorPagination(CursorPagination):
     # is greater than the `offset_cutoff` the `next_page` in the response would
     # point to the current page creating an infinite loop.
     offset_cutoff = None
+
+    # Allow clients to override the default page size (100) for paginated responses
+    page_size_query_param = 'page_size'
+
+    # Set a maximum page size that we'll allow, so clients can't do something
+    # completely outrageous.
+    max_page_size = 10_000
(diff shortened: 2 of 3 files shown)


datahub/dataset/interaction/pagination.py
```diff
@@ -0,0 +1,15 @@
...
+Pagination serializers determine the structure of the output that should
+be used for paginated responses.
+"""
+
+from datahub.dataset.core.pagination import DatasetCursorPagination
+
+
+class InteractionsDatasetViewCursorPagination(DatasetCursorPagination):
+    # Allow clients to override the default page size (100) for paginated responses
+    page_size_query_param = 'page-size'
```
reupen, line 11: I believe all the current query parameters use an `_` between words, so would suggest `page_size` here.

samuelhwilliams:
@reupen I'm happy to do the latter - and would think it's generally useful. But wanted to propose the smallest possible change in case there were other endpoints where 1k/10k page size would be less reasonable.

I'll update.

datahub/dataset/interaction/pagination.py
samuelhwilliams: Ah - ok. I'll update.

changelog/interactions-endpoint-page-size.api.md
```diff
@@ -0,0 +1 @@
+GET /v4/dataset/*: enabled a `page_size` parameter that allows clients to change the default page size from 100, up to a maximum of 10,000.
```
reupen, line 1: Would be careful with the `*` since it's for italics – might need to use backticks or escape it with `\`:

```suggestion
`GET /v4/dataset/*`: enabled a `page_size` parameter that allows clients to change the default page size from 100, up to a maximum of 10,000.
```

samuelhwilliams: Good spot.

PR #2772, opened 2020-04-15
Fix export country history feed

PippoRaimondi:
### Description of change

When in Exports tab and editing export countries information, if we had a couple of countries present, and when either removing, moving or adding a new country, instead of creating a single export country history record (which you can view in the "View full export country history" link), we were creating new history records for all existing countries present, creating a lot of noise and duplication. This fix addressed the above.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/company/test/test_signals.py
```diff
@@ -148,26 +151,64 @@ def test_company_export_country_history_update(self):
...
+            CompanyExportCountry.Status.FUTURE_INTEREST,
+            company.created_on,
+            adviser,
+            True,
+        )
+
+        history = CompanyExportCountryHistory.objects.filter(
+            id=export_country.id,
+        )
+        assert history.count() == 1
```
sekharpanja, line 209: It is worth checking history count is same before and after add_export_country call.

PippoRaimondi: makes sense! will update along with the suggestions above!

reupen:
@PippoRaimondi Could you expand on the description on the PR to explain the problem being fixed in more detail?

datahub/company/test/test_signals.py
```diff
@@ -148,26 +151,67 @@ def test_company_export_country_history_update(self):
         company.add_export_country(
             country,
             CompanyExportCountry.Status.CURRENTLY_EXPORTING,
-            company.created_on,
+            utc.localize(datetime.now()),
```
reupen, line 154: We should generally use the `now()` from `django.utils.timezone` rather than other options (as Django has its own time zone handling).

Also, `pytz` is not a direct dependency so we shouldn't be importing it (unless we make it a direct dependency).

datahub/company/test/test_signals.py
```diff
@@ -212,3 +256,8 @@ def test_delete_company_export_country_no_signal(self):
             history_type=CompanyExportCountryHistory.HistoryType.DELETE,
         )
         assert history.count() == 0
+
+    def _assert_history_records(self, country_id, history_type, history_count):
+        history = CompanyExportCountryHistory.objects.filter(id=country_id)
+        assert history.count() == history_count
+        assert history[0].history_type == history_type
```
reupen, line 263: I don't think this method adds much and it obfuscates what the tests are doing. I would suggest removing it.

datahub/company/test/test_signals.py
```diff
@@ -148,26 +151,67 @@ def test_company_export_country_history_update(self):
...
         assert history[1].history_type == CompanyExportCountryHistory.HistoryType.UPDATE
 
+    def test_company_export_country_history_update_with_no_change(self):
+        """
+        Test that submitting an update for a CompanyExportCountry record
+        that doesn't change any field (i.e status) does not create
+        a history record.
+        """
+        company = CompanyFactory()
+        country = CountryModel.objects.order_by('name')[0]
```
reupen, line 181: Generally it's better to use e.g. `.first()` rather than `[0]` on querysets as the former will only retrieve a single object from the database.

However, we added `random_obj_for_model()` for this purpose so I would suggest using that (which country it is should make no difference).

PippoRaimondi:
> @PippoRaimondi Could you expand on the description on the PR to explain the problem being fixed in more detail?

Done!

reupen:
@PippoRaimondi This will need (presumably) a `bugfix` news fragment – have a look at https://github.com/uktrade/data-hub-api/blob/master/changelog/README.md for instructions.

datahub/company/serializers.py
```diff
@@ -674,6 +674,7 @@ def validate(self, data):
         # check for duplicate countries
         export_countries = data.get('export_countries', [])
         countries = [item['country'] for item in export_countries]
+
```
reupen, line 677: This change does not look relevant so would be good to remove it.

changelog/company/fix-export-country-history.bugfix.md
```diff
@@ -0,0 +1 @@
+Ensure export country history doesn't create duplicate records when removing, moving or adding a new country in company Exports tab.
```
sekharpanja, line 1: I recommend something like this: Fix to ensure `CompanyExportCountryHistory` records are tracked only when `CompanyExportCountry` records are being created or updated.

changelog/company/fix-export-country-history.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix to ensure `CompanyExportCountryHistory` records are tracked only when `CompanyExportCountry` records are being created or updated.
```
reupen, line 1: FYI this should be reworded to follow https://github.com/uktrade/data-hub-api/blob/develop/changelog/README.md. So, something like '`CompanyExportCountryHistory` records are now only created when...'

(Also, it's redundant to say it's a bug fix as it will be under that heading.)

PR #2804, opened 2020-04-23
Allow authorised users to change an investment project to any stage 

mforner13:
### Description of change

This PR adds a new investment project permission `can_change_to_any_stage_investmentproject` and a new endpoint `v3/investment/<uuid:pk>/update-stage`. Users with this permission will be able to update the stage of the investment project freely, so can move from any one stage to any other stage, through a patch request to this endpoint. This request will bypass any field validation checks. 

The status of the project is also be updated if the current or new stage is 'won'.

Currently the investment verification team can change a project's stage in Django admin, whereas this change will allow them to make these changes in the front end and reduce the number of people who need access to Django admin.

There is currently a permission `can_change_stage_to_won` which allows only the verification team to move an investment project from verify win -> won. Once this change is released, and the relevant users have been assigned the new permission, `can_change_stage_to_won` will no longer be needed and will be removed.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


changelog/investment/allow_authorised_users_to_change_project_to_any_stage.md
```diff
@@ -0,0 +1,3 @@
+To allow the investment verification team to move projects between stages in the front end, rather than in Dajngo Admin, a new permission `change_to_any_stage_investmentproject` is now available.
```
reupen, line 1: FYI this news fragment needs a type in the file name (e.g. `.api.md` or `.feature.md`).

Also, small typo :slightly_smiling_face: 

```suggestion
To allow the investment verification team to move projects between stages in the front end, rather than in Django Admin, a new permission `change_to_any_stage_investmentproject` is now available. 
```

datahub/investment/project/serializers.py
```diff
@@ -604,3 +603,29 @@ def many_init(cls, *args, **kwargs):
...
+
+    def change_stage(self):
+        """
+        Change the stage of a project and update the status if moving from
+        or to 'won'.
+        """
+        data = self.validated_data
+        current_stage = self.instance.stage_id
+        new_stage = self.validated_data['stage'].id
+        if UUID(InvestmentProjectStage.won.value.id) in {new_stage, current_stage}:
```
reupen, line 625: `current_stage` could technically be a string as Django allows assigning strings to UUID fields – I would suggest using `force_uuid()` from `datahub.core.utils` on it (and in place of `UUID` below).

datahub/investment/project/serializers.py
```diff
@@ -153,6 +154,23 @@
 ALL_FIELDS = CORE_FIELDS + VALUE_FIELDS + REQUIREMENTS_FIELDS + TEAM_FIELDS + SPI_FIELDS
 
 
+def update_status(instance, data):
+    """Updates the project status when the stage changes to or from Won."""
+    old_stage = instance.stage if instance else None
```
reupen, line 159: IMO the blank lines that were in the this code when it was on the serialiser helped readability as it grouped the lines into logical chunks and highlighted the early return. So would suggest keeping them.

datahub/investment/project/urls.py
```diff
@@ -77,6 +79,11 @@
         audit_item,
         name='audit-item',
     ),
+    path(
+        'investment/<uuid:pk>/update-stage',
+        update_stage_of_item,
+        name='update_stage_of_item',
```
reupen, line 85: Would be good to use hyphens in the name to keep this consistent with other routes.
```suggestion
        name='update-stage-of-item',
```

datahub/investment/project/test/test_views.py
```diff
@@ -1578,6 +1578,95 @@ def test_revert_stage_to_verify_win(self):
...
+        )
+
+        response = api_client.patch(url, data=request_data)
+        assert response.status_code == status.HTTP_200_OK
+        response_data = response.json()
+        assert response_data['stage'] == {
+            'id': constants.InvestmentProjectStage.prospect.value.id,
+            'name': constants.InvestmentProjectStage.prospect.value.name,
+        }
+        assert response_data['status'] == 'ongoing'
```
reupen, line 1624: Would be good to use the status constants here instead of `'ongoing'` i.e. `InvestmentProject.Status.ONGOING`.

(I realise this came from another test, but looks like that one was wrong...)

datahub/investment/project/serializers.py
```diff
@@ -604,3 +603,29 @@ def many_init(cls, *args, **kwargs):
     class Meta:
         model = InvestmentProjectTeamMember
         fields = ('investment_project', 'adviser', 'role')
+
+
+class IProjectChangeStageSerializer(serializers.ModelSerializer):
```
reupen, line 608: Is anything from `serializers.ModelSerializer` actually being used or can this inherit from `serializers.Serializer` instead?

datahub/investment/project/test/test_views.py
```diff
@@ -1578,6 +1578,95 @@ def test_revert_stage_to_verify_win(self):
         }
         assert response_data['status'] == 'ongoing'
 
+    def test_move_project_back_a_stage_by_unauthorised_user(self):
```
reupen, line 1581: Would consider combining this test with `test_move_project_forward_multiple_stages_by_unauthorised_user` by parametrising `factory` and `new_stage_id` (using `pytest.mark.parametrize`).

Having said that, permission behaviour varying based on the request body is in general unusual, so IMO just one of `test_move_project_back_a_stage_by_unauthorised_user` and `test_move_project_back_a_stage_by_unauthorised_user` is sufficient.

datahub/investment/project/test/test_views.py
```diff
@@ -1578,6 +1578,95 @@ def test_revert_stage_to_verify_win(self):
...
+        url = reverse('api-v3:investment:update_stage_of_item', kwargs={'pk': project.pk})
+        request_data = {
+            'stage': {
+                'id': constants.InvestmentProjectStage.prospect.value.id,
+            },
+        }
+        response = self.api_client.patch(url, data=request_data)
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+
+    def test_move_project_back_a_stage_success(self):
```
reupen, line 1596: Would also consider parametrising this to combine it with `test_move_project_forward_multiple_stages_success`. (So would parametrise `factory`, `new_stage_id` and `expected_status`.)

datahub/investment/project/views.py
```diff
@@ -131,6 +137,28 @@ def get_serializer_context(self):
...
 
+    @action(methods=['patch'], detail=True)
+    def change_stage(self, request, *args, **kwargs):
+        """Change the stage of an investment project"""
+        current_user = self.request.user
+        change_to_any_stage_permission = (
+            f'investment.{InvestmentProjectPermission.change_to_any_stage}'
+        )
+        if not current_user.has_perm(change_to_any_stage_permission):
+            return Response(status=status.HTTP_403_FORBIDDEN)
```
reupen, line 148: I would suggest using permission classes for this – e.g.:

https://github.com/reupen/data-hub-api/blob/04bded4a45efd2eb8dca561f2bea236ad3dd883e/datahub/company_referral/views.py#L43-L50

datahub/investment/project/views.py
```diff
@@ -131,6 +137,28 @@ def get_serializer_context(self):
             'current_user': self.request.user if self.request else None,
         }
 
+    @action(methods=['patch'], detail=True)
```
reupen, line 140: Technically, if this is a PATCH, it means that not all request body fields are required, while in this case they are. So I think this should be a POST.

datahub/investment/project/test/test_views.py
mforner13: That's a very good point, testing an unauthorised user sending a request to this endpoint once seems sufficient 👍

mforner13:
@reupen I've made the changes requested, let me know what you think.

datahub/investment/project/serializers.py
```diff
@@ -604,3 +610,24 @@ def many_init(cls, *args, **kwargs):
...
+    def change_stage(self, user):
+        """
+        Change the stage of a project and update the status if moving from
+        or to 'won'.
+        """
+        data = self.validated_data
+        current_stage = force_uuid(self.instance.stage_id)
+        new_stage = self.validated_data['stage'].id
+        if force_uuid(InvestmentProjectStage.won.value.id) in {new_stage, current_stage}:
+            self.instance.status = _get_updated_status(instance=self.instance, data=data)
```
reupen, line 629: If both the old and new stages are Won, will this try to set the status to None?

datahub/investment/project/serializers.py
```diff
@@ -153,6 +154,26 @@
 ALL_FIELDS = CORE_FIELDS + VALUE_FIELDS + REQUIREMENTS_FIELDS + TEAM_FIELDS + SPI_FIELDS
 
 
+def _get_updated_status(instance, data):
+    """Updates the project status when the stage changes to or from Won."""
+    old_stage = instance.stage if instance else None
+    new_stage = data.get('stage')
+
+    if not new_stage or new_stage == old_stage:
+        return
```
reupen, line 163: For info, if one `return` returns a value, PEP 8 says they all should so this should be `return None` (search for 'Be consistent in return statements' in PEP 8).

datahub/investment/project/permissions.py
```diff
@@ -23,7 +23,6 @@ class _PermissionTemplate(StrEnum):
 class InvestmentProjectModelPermissions(ViewBasedModelPermissions):
     """
     Custom permissions class for investment views.
-
```
reupen, line 26: Would be good to remove these blank line removals :wink:

datahub/investment/project/serializers.py
mforner13: @reupen Ahh yes you're right! Although this method would only be used when a user is changing stage, so I don't think that situation would ever occur. Would it be good to guard against it anyway?

reupen: Yes, it's possible to call the API in that manner, so it shouldn't fail. It might happen if the same project is open in multiple tabs, for example.

datahub/investment/project/serializers.py
```diff
@@ -153,6 +153,26 @@
...
+    new_status = combiner.get_value('status')
+
+    if str(new_stage.id) == InvestmentProjectStage.won.value.id:
+        return InvestmentProject.Status.WON
+    elif (
+        old_stage and str(old_stage.id) == InvestmentProjectStage.won.value.id
+        and new_status == InvestmentProject.Status.WON
+    ):
+        return InvestmentProject.Status.ONGOING
+
```
reupen, line 174: Would be good to add a `return None` at the end as well (for the same reason as above).

mforner13: Cheers Reupen - all done 👍

PR #2814, opened 2020-04-28
Create a new Pipeline Item

sekharpanja:
#### Background

As part of improving information and process within Data Hub, yellow team is implementing a Pipeline for users to add companies they are dealing with and their current status within the pipeline (Lead, In Progress or Win). They can then maintain its status as they progress.

Idea of MVP is to show a separate Pipeline tab next to My company lists and My referrals, allowing the users maintain a dashboard of companies separated by predefined statuses. Users can add companies and also be able to move items between statuses.

#### This PR
API endpoint is `/v4/pipeline-item`. Expects a POST with JSON:

```
{
     'company': <company.pk>,
     'status': <pipeline_status>,
}
```

and returns 201 with following response upon success:

```
{
        'id': <company.pk>,
        'name': <company.name>,
        'export_potential': <company.export_potential>,
        'turnover': <company.turnover>,
    },
    'status': <pipeline_status>,
    'created_on': <created datetime>,
}
```

It can raise 
- 401 for unauthenticated request, 
- 403 if the user doesn't have enough permissions. 
- 400 if the company is archived, company already exists for the user, company doesn't exist or status is not one of the predefined.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/changelog/adviser/pipelineitem.api.md
+++ b/changelog/adviser/pipelineitem.api.md
@@ -1,0 +1,31 @@
+`POST /v4/pipeline-item`
+
+Expects a POST with JSON:
+
+```
+{
+     'company': '<company.pk>',
+     'status': '<pipeline_status>'
+}
+```
+
+and returns 201 with following response upon success:
+
+```
+{
+    {
+        'id': '<company.pk>',
+        'name': '<company.name>',
+        'export_potential': '<company.export_potential>',
+        'turnover': '<company.turnover>'
+    },
+    'status': '<pipeline_status>',
+    'created_on': '<created datetime>'
+}
+```
+
+It can raise:
+
+- 401 for unauthenticated request
+- 403 if the user doesn't have enough permissions
+- 400 if the company is archived, company already exists for the user, company doesn't exist or status is not one of the predefined.
--- a/datahub/user/company_list/serializers.py
+++ b/datahub/user/company_list/serializers.py
@@ -1,8 +1,17 @@
+from django.utils.translation import gettext_lazy
 from rest_framework import serializers
+from rest_framework.validators import UniqueTogetherValidator
 
 from datahub.company.models import Company
 from datahub.core.serializers import NestedRelatedField
 from datahub.user.company_list.models import CompanyList, CompanyListItem, PipelineItem
+
+CANT_ADD_ARCHIVED_COMPANY_TO_PIPELINE_MESSAGE = gettext_lazy(
+    "An archived company can't be added to the pipeline.",
+)
+COMPANY_ALREADY_EXISTS_IN_PIPELINE_MESSAGE = gettext_lazy(
+    'This company already exists in the pipeline for this user.',
+)
 
 
 class CompanyListSerializer(serializers.ModelSerializer):
@@ -60,7 +69,7 @@ class CompanyListItemSerializer(serializers.ModelSerializer):
 
 
 class PipelineItemSerializer(serializers.ModelSerializer):
-    """Serialiser for pipeline list items."""
+    """Serialiser for pipeline item."""
 
     company = NestedRelatedField(
         Company,
@@ -68,11 +77,34 @@ class PipelineItemSerializer(serializers.ModelSerializer):
         # call in the queryset module
         extra_fields=('name', 'turnover', 'export_potential'),
     )
+    adviser = serializers.HiddenField(default=serializers.CurrentUserDefault())
+
+    def validate_company(self, company):
+        """Make sure company is not archived"""
+        if company.archived:
+            raise serializers.ValidationError(CANT_ADD_ARCHIVED_COMPANY_TO_PIPELINE_MESSAGE)
+
+        return company
 
     class Meta:
         model = PipelineItem
         fields = (
+            'id',
             'company',
-            'created_on',
             'status',
+            'adviser',
+            'created_on',
         )
+        read_only_fields = (
+            'id',
+            'adviser',
+            'created_on',
+        )
+
+        validators = [
+            UniqueTogetherValidator(
+                queryset=PipelineItem.objects.all(),
+                fields=['company', 'adviser'],
+                message=COMPANY_ALREADY_EXISTS_IN_PIPELINE_MESSAGE,
+            ),
+        ]
--- a/datahub/user/company_list/test/test_pipeline_views.py
+++ b/datahub/user/company_list/test/test_pipeline_views.py
@@ -1,12 +1,14 @@
+from uuid import uuid4
+
 import pytest
+from freezegun import freeze_time
 from rest_framework import status
 from rest_framework.reverse import reverse
 
-from datahub.company.test.factories import CompanyFactory
+from datahub.company.test.factories import ArchivedCompanyFactory, CompanyFactory
 from datahub.core.test_utils import APITestMixin, create_test_user, format_date_or_datetime
 from datahub.user.company_list.models import PipelineItem
 from datahub.user.company_list.test.factories import PipelineItemFactory
-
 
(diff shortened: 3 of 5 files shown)


datahub/user/company_list/serializers.py
```diff
@@ -57,3 +67,57 @@ class Meta:
...
+            raise serializers.ValidationError(errors)
+
+        if PipelineItem.objects.filter(
+            company=company,
+            adviser=adviser,
+        ).exists():
+            errors = {
+                api_settings.NON_FIELD_ERRORS_KEY: COMPANY_ALREADY_EXISTS_MESSAGE,
+            }
+            raise serializers.ValidationError(errors)
```
adamledwards, line 107: Should we be using the UniqueTogetherValidator?
https://www.django-rest-framework.org/api-guide/validators/#uniquetogethervalidator

datahub/user/company_list/serializers.py
```diff
@@ -57,3 +67,57 @@ class Meta:
...
+        Create a new instance using this serializer
+        :return: created instance
+        validates to make sure:
+        - an archived company will not be added
+        - existing company is not added to the same adviser
+        """
+        company = validated_data.get('company')
+        adviser = validated_data.get('adviser')
+
+        if company.archived:
```
adamledwards, line 94: Could we reuse this in validators? 
from datahub.core.validators.not_archived import NotArchivedValidator

datahub/user/company_list/test/test_pipeline_views.py
```diff
@@ -0,0 +1,221 @@
...
+        response = api_client.post(
+            pipeline_collection_url,
+            data={
+                'company': str(company.pk),
+                'status': PipelineItem.Status.IN_PROGRESS,
+            },
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+    @freeze_time('2017-04-19 15:25:30.986208')
```
adamledwards, line 171: Do we need freeze_time here? Should we only use it when testing code that rely on dates?

datahub/user/company_list/serializers.py
sekharpanja: I think `NotArchivedValidator` works only the model that the serializer applies to. In this instance `PipelienItem`. But we want to check `Company` instead. Correct me if I am wrong.

adamledwards: Yes you are right.

datahub/user/company_list/views.py
```diff
@@ -198,3 +203,47 @@ def _get_company_list_or_404(self, request, company_list_pk):
...
+
+
+class PipelineItemViewSet(CoreViewSet):
+    """A view set for returning the contents of pipeline item."""
+
+    required_scopes = (Scope.internal_front_end,)
+    permission_classes = (
+        IsAuthenticatedOrTokenHasScope,
+        PipelineItemAPIPermissions,
+    )
```
reupen, line 230: It doesn't look like a custom permissions class is needed here – you could use `DjangoCrudPermission` and `HasPermissions` if you want to require `view_company` as well.

datahub/user/company_list/serializers.py
```diff
@@ -57,3 +67,57 @@ class Meta:
...
+        model = PipelineItem
+        fields = (
+            'id',
+            'company',
+            'status',
+            'created_on',
+        )
+        read_only_fields = (
+            'id',
+            'company',
```
reupen, line 121: I don't think this applies to explicitly defined fields. But, also, `company` doesn't seem to be read-only.

datahub/user/company_list/serializers.py
reupen: Yes, I think so. (The company list stuff didn't because `company` isn't a field on the serialiser there.)

datahub/user/company_list/serializers.py
```diff
@@ -57,3 +67,57 @@ class Meta:
...
+        - existing company is not added to the same adviser
+        """
+        company = validated_data.get('company')
+        adviser = validated_data.get('adviser')
+
+        if company.archived:
+            errors = {
+                api_settings.NON_FIELD_ERRORS_KEY: CANT_ADD_ARCHIVED_COMPANY_MESSAGE,
+            }
+            raise serializers.ValidationError(errors)
```
reupen, line 98: Not sure about putting validation logic in `create()` – this one looks like a good candidate for `validate_company`.

datahub/user/company_list/test/test_pipeline_views.py
```diff
@@ -0,0 +1,217 @@
...
+        company = ArchivedCompanyFactory()
+
+        response = api_client.post(
+            pipeline_collection_url,
+            data={
+                'company': str(company.pk),
+                'status': PipelineItem.Status.IN_PROGRESS,
+            },
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
```
reupen, line 184: It would be good to check the contents of `response.json()` in addition to the status.

datahub/user/company_list/test/test_pipeline_views.py
```diff
@@ -0,0 +1,217 @@
...
+                    'company': ['This field may not be null.'],
+                },
+                id='company is empty string',
+            ),
+        ),
+    )
+    def test_validation(self, request_data, expected_errors):
+        """Test validation."""
+        permission_codenames = ['view_company', 'add_pipelineitem']
+        user = create_test_user(permission_codenames=permission_codenames, dit_team=None)
```
reupen, line 105: The permissions are tested above so I don't think we need to explicitly set them in the other tests. Does just using `self.api_client` work?

datahub/user/company_list/test/test_pipeline_views.py
```diff
@@ -0,0 +1,217 @@
...
+        PipelineItemFactory(company=company, adviser=user)
+
+        response = api_client.post(
+            pipeline_collection_url,
+            data={
+                'company': str(company.pk),
+                'status': PipelineItem.Status.IN_PROGRESS,
+            },
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
```
reupen, line 168: It would be good to check the contents of `response.json()` in addition to the status.

datahub/user/company_list/views.py
sekharpanja: Would it be the same case if we want to implement similar permission structure as company_lists, where LEPs and DAs don't have access to company lists?

reupen: Yes, the LEP and DA permissions are determined through groups.

datahub/user/company_list/views.py
```diff
@@ -209,13 +209,13 @@ class PipelineItemViewSet(CoreViewSet):
...
     def initial(self, request, *args, **kwargs):
         """
-        Raise an Http404 if user has no pipeline items.
+        Raise an Http404 if user has no pipeline items during list action.
         """
         super().initial(request, *args, **kwargs)
 
-        if not self.get_queryset().exists():
+        if self.action == 'list' and not self.get_queryset().exists():
             raise Http404()
```
reupen, line 217: It's not strictly related to this PR, but I don't think the 404 is typical behaviour for listing either – normally listing would return no results rather than a 404...

changelog/adviser/pipelineitem.api.md
```diff
@@ -0,0 +1,30 @@
...
+```
+{
+        'id': '<company.pk>',
+        'name': '<company.name>',
+        'export_potential': '<company.export_potential>',
+        'turnover': '<company.turnover>'
+    },
+    'status': '<pipeline_status>',
+    'created_on': '<created datetime>'
+}
```
reupen, line 23: Looks like the braces are mismatched here.

PR #2875, opened 2020-05-18
Add daily job to track auto-archive backlog

alixedi:
### Description of change

We have been running the `automatic_company_archive` job for a while. Lately, we have noticed that the number of companies archived automatically have exceeded our original estimates.

The estimates in this case were only valid for a point-in-time because the criteria includes clauses that mean the set of companies eligible for automatic archiving is a rolling one.

We would like to track the number of companies that are eligible for archiving. It would be interesting to take note of how this set grows etc.

This introduces a cron entry that runs the `automatic_company_archive` job in simulation mode every night. This should provide us with the visibility we need.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -441,6 +441,16 @@ if REDIS_BASE_URL:
                 'simulate': False,
             }
         },
+        # The purpose of the following is to keep us posted on the backlog of
+        # companies that are eligible for automatic archiving
+        'simulate_automatic_company_archive': {
+            'task': 'datahub.company.tasks.automatic_company_archive',
+            'schedule': crontab(minute=0, hour=19),
+            'kwargs': {
+                'limit': 20000,
+                'simulate': True,
+            }
+        },
     }
 
     if env.bool('ENABLE_DAILY_HIERARCHY_ROLLOUT', False):
--- a/datahub/company/tasks/company.py
+++ b/datahub/company/tasks/company.py
@@ -89,4 +89,7 @@ def automatic_company_archive(self, limit=1000, simulate=True):
             return
 
         archive_count = _automatic_company_archive(limit, simulate)
-        send_realtime_message(f'{self.name} archived: {archive_count}')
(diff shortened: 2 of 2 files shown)


reupen:
Perhaps the message emitted by `send_realtime_message()` in the task could be adjusted when simuate=True to say it's a simulation?

config/settings/common.py
```diff
@@ -441,6 +441,16 @@ def _build_redis_url(base_url, db_number, **query_args):
                 'simulate': False,
             }
         },
+        # The purpose of the following is to keep us posted on the backlog of
+        # companies that are eligible for automatic archiving
+        'simulate_automatic_company_archive': {
+            'task': 'datahub.company.tasks.automatic_company_archive',
+            'schedule': crontab(minute=0, hour=19),
+            'kwargs': {
+                'limit': 100000,
```
currycoder, line 450: Given this new ceiling, could it be worth using: https://docs.djangoproject.com/en/3.0/ref/models/querysets/#iterator

e.g. in https://github.com/uktrade/data-hub-api/blob/develop/datahub/company/tasks/company.py#L46

Make it: `for company in companies_to_be_archived.iterator():`

My worry is that if - for whatever reason - this thing grabs 100K companies, it will load all 100K companies in to memory at once.

alixedi: It seems, the downside to this would be issuing 100k queries. 

Can we perhaps keep the ceiling to 20k? We would lose some benefits but we will keep the number of queries down.

reupen: It will use a server-side cursor so it won't issue 100k queries as such – but it will use more resources on the database server as the results end up being cached there instead of in the process.

So you could add e.g. a `.only('pk')` to the queryset to keep the load on the database down as well (doesn't seem like any fields are actually read).

([There are also some pytest-django fixtures](django_assert_num_queries) that can be used to add assertions about the number of queries if we want to add some of those.)

PR #2915, opened 2020-06-02
Create new interaction feature usage command

PippoRaimondi:
### Description of change

The intent of this PR is automate reports that are generated every other week.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/interaction/management/commands/get_interaction_usage.py
+++ b/datahub/interaction/management/commands/get_interaction_usage.py
@@ -1,0 +1,107 @@
+from logging import getLogger
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+from datahub.interaction.models import Interaction, InteractionExportCountry
+
+logger = getLogger(__name__)
+cursor = connection.cursor()
+
+
+def _get_export_country_interaction_with_discussed_countries():
+    """
+    Number of export country interactions with countries tagged
+    where countries were discussed.
+    """
+    interactions = Interaction.objects.filter(
+        were_countries_discussed=True,
+    ).count()
+
+    return interactions
+
+
+def _get_all_export_interaction():
+    """
+    Number of interactions with countries tagged.
+    """
+    interactions = InteractionExportCountry.objects.count()
+
+    return interactions
+
+
+def _get_all_countries_with_a_company():
+    """
+    Number of all countries that had a company added.
+    """
+    countries = InteractionExportCountry.objects.all(
+    ).select_related(
+        'interaction',
+    ).distinct('interaction__company_id').count()
+
+    return countries
+
+
+def _get_team_usage():
+    """
+    Count of all teams that used the new feature.
+    """
+    cursor.execute(
+        """
+        select count(1)
+        from interaction_interaction ii
+        inner join company_advisor ca on ii.created_by_id = ca.id
+        inner join metadata_team mt on ca.dit_team_id = mt.id
+        left join metadata_teamrole m on mt.role_id = m.id
+        where ii.were_countries_discussed=true;
+        """,
+    )
+
+    return cursor.fetchone()
+
+
+def _get_top_5_team_usage():
+    """
+    Top 5 teams that used the new feature.
+    """
+    cursor.execute(
+        """
+        select mt.name "Team", m.name "Team type",
+        count(*) "Number of interactions with export countries"
+        from interaction_interaction ii
+        inner join company_advisor ca on ii.created_by_id = ca.id
+        inner join metadata_team mt on ca.dit_team_id = mt.id
+        left join metadata_teamrole m on mt.role_id = m.id
+        where ii.were_countries_discussed=true
+        group by 1, 2
+        order by "Number of interactions with export countries" desc
+        limit 5;
+        """,
+    )
+
+    return cursor.fetchone()
+
+
+class Command(BaseCommand):
+    """
+    Command to query usage stats from the new interaction journey.
+    """
+
+    def handle(self, *args, **options):
+        """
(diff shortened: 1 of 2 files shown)


elcct:
I have few concerns with this:
1) I am not sure if `dbmaintenance` app is a best place for such command, as this is intended for data maintenance and not reporting. I would consider whether this needs its own `report` app either on its own or within `interaction` app.
2) Why this needs to be a command and not e.g. an export in `dataset` app to be consumed by Data workspace?
3) As a minimum, it would be good to see some tests to ensure this produces expected results and doesn't have negative impact on the database.

PippoRaimondi:
> I have few concerns with this:
> 
> 1. I am not sure if `dbmaintenance` app is a best place for such command, as this is intended for data maintenance and not reporting. I would consider whether this needs its own `report` app either on its own or within `interaction` app.
> 2. Why this needs to be a command and not e.g. an export in `dataset` app to be consumed by Data workspace?
> 3. As a minimum, it would be good to see some tests to ensure this produces expected results and doesn't have negative impact on the database.

I will address the concerns above, I wasn't sure where to have this in first place either... thanks for the feedback!

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
...
+    ).distinct('interaction__company_id').count()
+
+    return countries
+
+
+def _get_team_usage():
+    """
+    Count of all teams that used the new feature.
+    """
+    cursor.execute(
```
mcbhenwood, line 49: Not sure we should be writing raw SQL anywhere in Data Hub. We should be leveraging the Django ORM for lots of reasons: maintability, change impact analysis, consistency...

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
+from logging import getLogger
+
+from django.core.management.base import BaseCommand
```
mcbhenwood, line 3: Thos ought probably instead to be a Celery Beat task. If a thing is going to be used regularly rather than occasionally it oughtn't be a management command really.

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
...
+    ).select_related(
+        'interaction',
+    ).distinct('interaction__company_id').count()
+
+    return countries
+
+
+def _get_team_usage():
+    """
+    Count of all teams that used the new feature.
```
mcbhenwood, line 47: It's only "new" now; we should write this code on the basis that it will live forever (even if we're _convinced_ that's not the case). We should remove all references to "new" in docstrings, code and output.

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
...
+
+    return cursor.fetchone()
+
+
+def _get_top_5_team_usage():
+    """
+    Top 5 teams that used the new feature.
+    """
+    cursor.execute(
+        """
```
mcbhenwood, line 68: see earlier comment about ORM > SQL

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
...
+        where ii.were_countries_discussed=true;
+        """,
+    )
+
+    return cursor.fetchone()
+
+
+def _get_top_5_team_usage():
+    """
+    Top 5 teams that used the new feature.
```
mcbhenwood, line 65: see earlier comment about removing "new" from narrative

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
...
+        limit 5;
+        """,
+    )
+
+    return cursor.fetchone()
+
+
+class Command(BaseCommand):
+    """
+    Command to query usage stats from the new interaction journey.
```
mcbhenwood, line 87: see earlier comment about removing "new" from narrative

datahub/interaction/management/commands/get_interaction_usage.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    def handle(self, *args, **options):
+        """
+        Run the query and output the results as an info message to the log file.
+        """
+        interaction_usage_stats = """
+        {result_1} uses of new interaction journey
```
mcbhenwood, line 95: see earlier comment about removing "new" from narrative

PippoRaimondi:
> Sorry to say but I think we need some changes here. As noted inline:
> 
> 1. This ought to be a scheduled Celery task, if it's any more than a one-off (which it is).
> 2. We shouldn't write raw SQL in Django
> 3. References to the thing under measurement being "new" should be removed because after a while it's not "new" (code has a habit of outliving its expected lifetime)
> 4. (optional) This might be a nice use for the newly-implemented "send a message to Slack" feature which is already used by the Company Archiving and DnB Update scheduled jobs. See here: https://github.com/uktrade/data-hub-api/blob/develop/datahub/core/realtime_messaging.py

Thanks for the feedback, I agree with all points especially with the raw sql (despite it being awkward to use the ORM when there is no foreign key relationship), it felt wrong and I was looking at alternatives...

mcbhenwood:
Thanks @PippoRaimondi If you need any ORM help then please do ask in the [BE Slack channel](https://ditdigitalteam.slack.com/archives/CAXBZGS95)

PR #2920, opened 2020-06-02
Add update_sector_parent management command

abbas123456:
### Description of change

This adds an update_sector_parent management command to allow Sectors to be moved to new parents.

The current Data Hub sector list was inherited from previous systems and has never been reviewed. It includes a number of duplicates (e.g. ‘Environment and Water’ / ‘Environment’ / ‘Water’) as well as out-of-date options (e.g. ‘3G services’).
Updating the list will allow us to use a simpler, more up-to-date set of options and to better link data across systems.

Our team have worked with representatives from all sector clusters and the analyst team to define an updated sector list. This has been approved by the DIT Reference Data Board.

As part of updating sectors in Data Hub, we need to migrate historical records once the update has taken place. In order to do this three management commands will be added. The `update_sector_segment`(#2899), `create_sector`(#2907), and `update_sector_parent` (this PR) management commands will make it possible to rename, split, and move sector segments respectively.

### Checklist

* [x] If this is a releasable change, has a news fragment been added?

  <details>
  <summary>Explanation</summary>
  
  A news fragment is required for any releasable change (i.e. code that runs in or affects production) so that a corresponding changelog entry is added when releasing.
  
  Check [changelog/README.md](https://github.com/uktrade/data-hub-api/blob/master/changelog/README.md) for instructions.
  
  </details>
  
* [x] Has this branch been rebased on top of the current `develop` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [x] Is the CircleCI build passing?

### General points

<details>
<summary><strong>Other things to check</strong></summary><p></p>

* Make sure `fixtures/test_data.yaml` is maintained when updating models
* Consider the admin site when making changes to models
* Use select-/prefetch-related field lists in views and search apps, and update them when fields are added
* Make sure the README is updated e.g. when adding new environment variables

</details>

See [docs/CONTRIBUTING.md](https://github.com/uktrade/data-hub-api/blob/develop/docs/CONTRIBUTING.md) for more guidelines.

PR #2907, opened 2020-05-29
Add create_sector management command

abbas123456:
### Description of change

This adds a create_sector management command to allow Sectors to be created and attached to a parent.

The current Data Hub sector list was inherited from previous systems and has never been reviewed. It includes a number of duplicates (e.g. ‘Environment and Water’ / ‘Environment’ / ‘Water’) as well as out-of-date options (e.g. ‘3G services’).
Updating the list will allow us to use a simpler, more up-to-date set of options and to better link data across systems.

Our team have worked with representatives from all sector clusters and the analyst team to define an updated sector list. This has been approved by the DIT Reference Data Board.

As part of updating sectors in Data Hub, we need to migrate historical records once the update has taken place. In order to do this three management commands will be added. The `update_sector_segment`(#2899), `create_sector` (this PR), and `update_sector_parent` (#2920) management commands will make it possible to rename, split, and move sector segments respectively.

### Checklist

* [x] If this is a releasable change, has a news fragment been added?

  <details>
  <summary>Explanation</summary>
  
  A news fragment is required for any releasable change (i.e. code that runs in or affects production) so that a corresponding changelog entry is added when releasing.
  
  Check [changelog/README.md](https://github.com/uktrade/data-hub-api/blob/master/changelog/README.md) for instructions.
  
  </details>
  
* [x] Has this branch been rebased on top of the current `develop` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [x] Is the CircleCI build passing?

### General points

<details>
<summary><strong>Other things to check</strong></summary><p></p>

* Make sure `fixtures/test_data.yaml` is maintained when updating models
* Consider the admin site when making changes to models
* Use select-/prefetch-related field lists in views and search apps, and update them when fields are added
* Make sure the README is updated e.g. when adding new environment variables

</details>

See [docs/CONTRIBUTING.md](https://github.com/uktrade/data-hub-api/blob/develop/docs/CONTRIBUTING.md) for more guidelines.

PR #2899, opened 2020-05-27
Add update_sector_segment management command

abbas123456:
### Description of change

This adds an `update_sector_segment` management command to allow `Sector` segments to be renamed.

### Checklist

* [x] If this is a releasable change, has a news fragment been added?

  <details>
  <summary>Explanation</summary>
  
  A news fragment is required for any releasable change (i.e. code that runs in or affects production) so that a corresponding changelog entry is added when releasing.
  
  Check [changelog/README.md](https://github.com/uktrade/data-hub-api/blob/master/changelog/README.md) for instructions.
  
  </details>
  
* [x] Has this branch been rebased on top of the current `develop` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [x] Is the CircleCI build passing?

### General points

<details>
<summary><strong>Other things to check</strong></summary><p></p>

* Make sure `fixtures/test_data.yaml` is maintained when updating models
* Consider the admin site when making changes to models
* Use select-/prefetch-related field lists in views and search apps, and update them when fields are added
* Make sure the README is updated e.g. when adding new environment variables

</details>

See [docs/CONTRIBUTING.md](https://github.com/uktrade/data-hub-api/blob/develop/docs/CONTRIBUTING.md) for more guidelines.

PR #2960, opened 2020-06-16
Add delete_sector management command

abbas123456:
### Description of change

This adds a delete_sector management command to allow Sectors to be deleted if they are no longer required.

This will allow orphaned Sectors to be tidied up after the update_sector_segment(#2899), create_sector(#2907), and update_sector_parent (#2920) commands are run. This will also ensure that they do not appear on the front end.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/dbmaintenance/management/commands/delete_sector.py
```diff
@@ -0,0 +1,32 @@
...
+class Command(CSVBaseCommand):
+    """Command to delete Sector."""
+
+    def _process_row(self, row, simulate=False, **options):
+        """Process one single row."""
+        pk = parse_uuid(row['id'])
+        sector = Sector.objects.get(pk=pk)
+
+        # Check that the sector is not referenced by any objects:
+        if sector not in get_unreferenced_objects_query(Sector).all():
```
elcct, line 21: That seems to be quite inefficient, but I guess there is not going to be many to be deleted?

abbas123456: We are planning to delete 87 sectors. As the `CSVBaseCommand` is designed to process one row at a time there isn't the option to do a bulk check unless we subclass `BaseCommand` instead but this will result in code duplication.

abbas123456: As discussed in the meeting on 19th June, I will update the command to do the get_unreferenced_objects_query once by defining an __init__ command

elcct: Another way could be to add something like a call to `post_process` function to the `_handle` function in the `CSVBaseCommand` after the processing loop, and then in the subclass make `process_row` collect the sectors and do deletion in `post_process`. This will make it reusable if we need such pattern in the future.

datahub/dbmaintenance/management/commands/delete_sector.py
```diff
@@ -0,0 +1,40 @@
...
+
+        # Check that the sector is not referenced by any objects:
+        if sector not in self.unreferenced_sectors:
+            logger.warning(f'Not deleting sector {sector} as it is referenced by another object')
+            return
+
+        if simulate:
+            return
+
+        # This will attempt to delete all descendants as well, but as the
```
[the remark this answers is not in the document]
abbas123456: Orginally yes, the CSV was ordered by path descending meaning children would get deleted before their parents.

However, now that the command collects all unreferenced sectors before processing, this would still fail. Therefore we have had to split the delete_sector.csv into one file per level, deleting level 4's before level 3's etc.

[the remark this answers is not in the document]
currycoder: Awesome - thank you for explaining.

PR #2969, opened 2020-06-18
Forward Zipkin headers with APIClient

rafenden:
### Description of change

Zipkin headers coming from the API client will be now forwarded to any services called via APIClient.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/serializers.py
+++ b/datahub/company/serializers.py
@@ -30,6 +30,7 @@ from datahub.company.validators import (
     has_no_invalid_company_number_characters,
     has_uk_establishment_number_prefix,
 )
+from datahub.core.api_client import get_zipkin_headers
 from datahub.core.constants import Country
 from datahub.core.constants import HeadquarterType
 from datahub.core.serializers import (
@@ -258,6 +259,7 @@ class ContactDetailSerializer(ContactSerializer):
         accepts_dit_email_marketing = validated_data.pop('accepts_dit_email_marketing')
         # If consent value in POST, notify
         combiner = DataCombiner(self.instance, validated_data)
+        request = self.context.get('request', None)
         transaction.on_commit(
             lambda: update_contact_consent.apply_async(
                 args=(
@@ -266,6 +268,7 @@ class ContactDetailSerializer(ContactSerializer):
                 ),
                 kwargs={
                     'modified_at': now().isoformat(),
+                    'zipkin_headers': get_zipkin_headers(request),
                 },
             ),
         )
--- a/datahub/core/api_client.py
+++ b/datahub/core/api_client.py
@@ -89,6 +89,7 @@ class APIClient:
         accept=DEFAULT_ACCEPT,
         default_timeout=None,
         raise_for_status=True,
+        request=None,
     ):
         """Initialises the API client."""
         self._api_url = api_url
@@ -96,6 +97,7 @@ class APIClient:
         self._accept = accept
         self._default_timeout = default_timeout
         self._raise_for_status = raise_for_status
+        self._request = request
 
     def request(self, method, path, **kwargs):
         """Makes an HTTP request."""
@@ -107,6 +109,8 @@ class APIClient:
         headers = kwargs.pop('headers', {})
         if self._accept:
             headers['Accept'] = self._accept
+        if self._request:
+            headers.update(get_zipkin_headers(self._request))
 
         try:
             response = requests.request(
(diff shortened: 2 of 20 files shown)


datahub/core/api_client.py
```diff
@@ -8,6 +8,8 @@
 
 logger = getLogger(__name__)
 
+api_client_session = requests.session()
```
currycoder, line 11: So I'm a little worried about this being a way for state to bleed between sequential HTTP requests to data-hub-api.

The docs on it say "It also persists cookies across all requests made from the Session instance, and will use urllib3’s connection pooling." - so I guess e.g. if we call something with cookie based auth, there's some chance that subsequent calls could use the same cookie unexpectedly..?
https://requests.readthedocs.io/en/master/user/advanced/#session-objects

One potential way that I can think of avoiding this is to "reset" the session variable via the `ZipkinHeadersMiddleware` at the end of servicing each HTTP request?

currycoder: Regardless, this sort of pattern does mean having some "global state" for a request.  I don't fully disagree with that idea, but it does feel like it could be error prone so we need to be super careful here.  I fully agree that to pass around these headers throughout the codebase (i.e. do all this stuff through runtime parameters) would be tiresome and difficult to maintain.

A safer alternative could be to copy the way that django does translations/current language activation:
- A thread local is created https://github.com/django/django/blob/master/django/utils/translation/trans_real.py#L25
- On activation of a new language, it's value is changed: https://github.com/django/django/blob/master/django/utils/translation/trans_real.py#L279
- Middleware is used to activate the current language on request, and deactivate it on response: https://github.com/django/django/blob/stable/1.4.x/django/middleware/locale.py#L11

So applying that same pattern here, we could have a `ZipkinHeaders` singleton.  `ZipkinHeadersMiddleware` could set the values on request and unset them on response.  The `APIClient` could interrogate the `ZipkinHeaders` singleton to get the current header values when it needs to make a request.

Of course this couples the `APIClient` and `ZipkinHeaders` objects, but I feel like this at least vastly reduces the scope of global variables.  What do you think?

mcbhenwood: Yep I would instinctively say that instantiating this object at the module level will result in non-threadsafe activity. Also not sure if this kind of object can expire / go bad over time but in this implementation it is expected to last for the life of the Python process running this worker, living on between unrelated requests "forever".

I don't have any alternative solutions at this time, just been thinking about it.

My best guess would be to set some predictably-but-uniquely-named Redis object in middleware and retrieve that same object during `ApiClient.request()`.

mcbhenwood: I guess this is a good example of how it's better to discuss potentially contentious implementations with the team in advance.

currycoder: To be fair, Raf and I discussed this a bit before implementation and I didn't have many bright ideas then :)

mcbhenwood: I probably would have made similar noises to this if I'd been involved at the time.

datahub/company/serializers.py
```diff
@@ -240,6 +240,7 @@ def _notify_consent_service(self, validated_data):
                 ),
                 kwargs={
                     'modified_at': now().isoformat(),
+                    'request': self.context.get('request', None),
```
mcbhenwood, line 243: Everything looks good except I can't see how the `request` object gets into the `context` for the `ContactDetailSerializer`. I would expect something in the `ContactViewSet` maybe overriding `get_serializer()` but I couldn't see anything. Therefore I suspect this kwarg will always be `None`.

Plus both tests (`test_serializer_update_call_task` and `test_serializer_create_calls_task `) seem to expect a lack of `request` object also (they assert it was `None`), so it seems like the Zipkin headers will not make it into the calls made by the Celery task (which they need to do as per the ticket).

I could be wrong though.

mcbhenwood (commented):
In general looks great, but there's this one Q regarding having access to the `request` object during the Consent Service task

datahub/company/serializers.py
rafenden: The request is added to the context by DRF:
https://github.com/encode/django-rest-framework/blob/5828d8f7ca167b11296733a2b54f9d6fca29b7b0/rest_framework/generics.py#L135

mcbhenwood: Thanks for reminding me about the default DRF context enrichment; I had forgotten.

mcbhenwood (approved):

ambrozic (approved):

PR #2972, opened 2020-06-19
Add error handling to the functions that retrieve consent status from the Legal Basis API

LaurenQurashi:
### Description of change
 This PR is one of many involved with integrating the Consent Service. The overall aim of the project is to source the `accepts_dit_email_marketing` field from the Legal Basis API rather than the Datahub API. If you would like to read more about the roll out process, please see [here](https://docs.google.com/document/d/1csAPy3fPaDGSS0BManpozZmiZBXm6HN0U7a28Nryfrc/edit#). 

This PR adds error handling for the `get_many()` function. This function is involved with retrieving whether a contact accepts DIT email marketing. After some previous discussions [on this PR](https://github.com/uktrade/data-hub-api/pull/2722/files), It's been decided that if the LB API goes down these functions should hide the error and simply return that none of the contacts accept email marketing. 

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/consent.py
+++ b/datahub/company/consent.py
@@ -7,6 +7,7 @@ import logging
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from requests.exceptions import HTTPError
 
 from datahub.company.constants import CONSENT_SERVICE_EMAIL_CONSENT_TYPE
 from datahub.core.api_client import APIClient, HawkAuth
@@ -22,7 +23,6 @@ CONSENT_SERVICE_READ_TIMEOUT = 30.0
 def _get_client():
     """
     Get configured API client for the consent service,
-
     _api_client could've just been set as a top level attribute
     of this module but that makes testing harder as the settings are
     loaded at `import` time rather than at run time. Which breaks test
@@ -81,20 +81,31 @@ def update_consent(email_address, accepts_dit_email_marketing, modified_at=None)
 def get_many(emails):
     """
     Bulk lookup consent for a list of emails
-
     :param emails: List of email addresses
     :return: dict of email address to consent status
     """
     body = {
         'emails': emails,
     }
+    api_client = _get_client()
 
-    response = _get_client().request(
-        'POST',
-        CONSENT_SERVICE_PERSON_PATH_LOOKUP,
-        json=body,
-        params={'limit': len(emails)},
-    )
+    try:
+        response = api_client.request(
+            'POST',
+            CONSENT_SERVICE_PERSON_PATH_LOOKUP,
+            json=body,
+            params={'limit': len(emails)},
+        )
+    except HTTPError as exc:
+        status = exc.response.status_code
+        if status >= 500:
+            return {
+                email: False
+                for email in emails
+            }
+        else:
+            raise
+
     return {
         result['email']: CONSENT_SERVICE_EMAIL_CONSENT_TYPE in result['consents']
         for result in response.json()['results']
--- a/datahub/company/test/test_consent_client.py
+++ b/datahub/company/test/test_consent_client.py
@@ -1,5 +1,6 @@
 import pytest
 from django.conf import settings
+from requests.exceptions import ConnectionError, HTTPError, Timeout
 from rest_framework import status
 
 from datahub.company import consent as consent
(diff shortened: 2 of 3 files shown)


datahub/company/consent.py
```diff
@@ -81,24 +81,31 @@ def update_consent(email_address, accepts_dit_email_marketing, modified_at=None)
...
+    api_client = _get_client()
+
+    try:
+        response = api_client.request(
+            'POST',
+            CONSENT_SERVICE_PERSON_PATH_LOOKUP,
+            json=body,
+            params={'limit': len(emails)},
+        )
+        response.raise_for_status()
```
currycoder, line 99: I don't think you need this line.  Doesn't the `request()` method (called above) already trigger `raise_for_status()`?

datahub/company/consent.py
```diff
@@ -81,24 +81,31 @@ def update_consent(email_address, accepts_dit_email_marketing, modified_at=None)
...
-        params={'limit': len(emails)},
-    )
-    return {
-        result['email']: CONSENT_SERVICE_EMAIL_CONSENT_TYPE in result['consents']
-        for result in response.json()['results']
-    }
+    api_client = _get_client()
+
+    try:
+        response = api_client.request(
```
currycoder, line 93: Please try to reduce this try block down to only contain the code that we need to wrap with error handling.

This will help make things more readable for future maintainers and reduce the scope of the block so that we don't accidentally catch unexpected exceptions.

datahub/company/consent.py
```diff
@@ -81,24 +81,31 @@ def update_consent(email_address, accepts_dit_email_marketing, modified_at=None)
...
+            CONSENT_SERVICE_PERSON_PATH_LOOKUP,
+            json=body,
+            params={'limit': len(emails)},
+        )
+        response.raise_for_status()
+        return {
+            result['email']: CONSENT_SERVICE_EMAIL_CONSENT_TYPE in result['consents']
+            for result in response.json()['results']
+        }
+    except HTTPError:
```
currycoder, line 104: Does this also trigger for 400s?  A 4XX response generally means "you did something wrong when making the request" - I think that we want to know about this ASAP instead of quietly handling things and saying that consent is `False`.

Would be good in the first instance to add a test case which triggers a [400, 401, etc] response and see whether it gets raised or is captured here.

datahub/company/test/test_consent_client.py
```diff
@@ -88,3 +88,63 @@ def test_update(self, requests_mock, accepts_marketing):
         result = consent.update_consent('foo@bar.com', accepts_marketing)
         assert result is None
         assert matcher.called_once
+
+    def test_consent_flag_false_when_service_fails_get_one(self, requests_mock):
+        """
+        When the feature flag is enabled, but the Consent Service is down,
```
currycoder, line 94: I don't think this test does anything with feature flags, does it..?

datahub/company/test/test_consent_client.py
```diff
@@ -88,3 +88,63 @@ def test_update(self, requests_mock, accepts_marketing):
...
+            },
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        )
+
+        response = consent.get_one('foo@bar.com')
+        assert response is False
+
+    def test_consent_flag_false_when_service_fails_get_many(self, requests_mock):
+        """
+        When the feature flag is enabled, but the Consent Service is down,
```
currycoder, line 118: Same comment as above

datahub/company/test/test_consent_client.py
```diff
@@ -88,3 +88,63 @@ def test_update(self, requests_mock, accepts_marketing):
...
+                'results': [
+                    {
+                        'email': 'foo@bar.com',
+                        'consents': [
+                            CONSENT_SERVICE_EMAIL_CONSENT_TYPE,
+                        ],
+                    },
+                ],
+            },
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
```
currycoder, line 110: It would be good to test some other statuses by using `pytest.mark.parametrize` on this test case.https://docs.pytest.org/en/stable/parametrize.html

currycoder: I'm thinking 502, 503, etc

datahub/company/consent.py
currycoder: It's also possible for the underlying `requests.request()` call that the api client makes to raise more exceptions here.

I'm thinking about connection errors and timeouts.  I believe that if we want to make this bullet proof, we need to handle those too.

An example: https://github.com/uktrade/data-hub-api/blob/develop/datahub/dnb_api/utils.py#L110

datahub/company/test/test_consent_client.py
```diff
@@ -88,3 +88,63 @@ def test_update(self, requests_mock, accepts_marketing):
...
+
+    def test_consent_flag_false_when_service_fails_get_one(self, requests_mock):
+        """
+        When the feature flag is enabled, but the Consent Service is down,
+        It returns an email with false as the value.
+        """
+        requests_mock.post(
+            f'{settings.CONSENT_SERVICE_BASE_URL}'
+            f'{consent.CONSENT_SERVICE_PERSON_PATH_LOOKUP}',
+            json={
```
currycoder, line 100: I don't think that the response will return JSON that looks like this on error - it's a little confusing to have this in right now and would be better to have no body at all.

datahub/company/test/test_consent_client.py
```diff
@@ -88,3 +88,63 @@ def test_update(self, requests_mock, accepts_marketing):
...
+
+    def test_consent_flag_false_when_service_fails_get_many(self, requests_mock):
+        """
+        When the feature flag is enabled, but the Consent Service is down,
+        It returns a dictionary of emails with False as the value.
+        """
+        requests_mock.post(
+            f'{settings.CONSENT_SERVICE_BASE_URL}'
+            f'{consent.CONSENT_SERVICE_PERSON_PATH_LOOKUP}',
+            json={
```
currycoder, line 124: Same comment as above.

datahub/company/test/test_consent_client.py
```diff
@@ -88,3 +89,135 @@ def test_update(self, requests_mock, accepts_marketing):
...
+            f'{settings.CONSENT_SERVICE_BASE_URL}'
+            f'{consent.CONSENT_SERVICE_PERSON_PATH_LOOKUP}',
+            status_code=res_status,
+        )
+        emails = ['foo1@bar.com', 'foo2@bar.com', 'foo3@bar.com']
+        with pytest.raises(HTTPError):
+            consent.get_many(emails)
+
+    @pytest.mark.parametrize(
+        'req_exception',
```
elcct, line 185: I think it will be better to not use abbreviations as it makes code less readable, so I'd change `req_exception` to `request_exception`. Same with others, e.g. `res_status`.

PR #3037, opened 2020-07-22
Strip invalid emails from consent payload

adamledwards:
### Description of change
We have been experiencing error 400 BAD REQUEST responses from the legal-basis service since we turn the feature on this is due to invalid data that the consent service rejects. 
In datahub there are empty emails and also invalid the latest example was an email in the following format`fake.name.@example-com` (real name and domain striped) it had a hyphen between the domain name and the top level domain instead of a dot.

In the PR has introduced a `is_valid_email` method that uses Django internal email validators to strip out invalid emails.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/dataset/contact/views.py
+++ b/datahub/dataset/contact/views.py
@@ -1,3 +1,6 @@
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_email
+
 from datahub.company import consent
 from datahub.company.constants import GET_CONSENT_FROM_CONSENT_SERVICE
 from datahub.company.models.contact import Contact
@@ -14,6 +17,14 @@ class ContactsDatasetView(BaseDatasetView):
     various reports to the users out of flattened table and let analyst to work on denormalized
     table to get more meaningful insight.
     """
+
+    def _is_valid_email(self, value):
+        """Validate if emails are valid and return a boolean flag."""
+        try:
+            validate_email(value)
+            return True
+        except ValidationError:
+            return False
 
     def get_dataset(self):
         """Returns list of Contacts Dataset records"""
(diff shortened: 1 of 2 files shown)


datahub/dataset/contact/views.py
```diff
@@ -47,10 +58,12 @@ def get_dataset(self):
 
     def _enrich_data(self, data):
         """
-        Get the marketing consent from the consent service
+        Get the marketing consent from the consent service.
+        Strip invalid emails from the data, old data may be
```
mcbhenwood, line 62: This is very picky but there ought to be a blank line between the one-line title of the docstring and its following body paragrapah(s), as per https://www.python.org/dev/peps/pep-0257/

datahub/dataset/contact/views.py
```diff
@@ -47,10 +58,13 @@ def get_dataset(self):
...
         """
-        Get the marketing consent from the consent service
+        Get the marketing consent from the consent service.
+
+        Strip invalid emails from the data, old data may be
+        empty or invalid due to validation implemented at a later date.
         """
         if is_feature_flag_active(GET_CONSENT_FROM_CONSENT_SERVICE):
-            emails = [item['email'] for item in data if item['email']]
+            emails = [item['email'] for item in data if self._is_valid_email(item['email'])]
```
sekharpanja, line 67: What will be consent value of the email that is not valid? Would it be False by default?

adamledwards: Yes false by default. Let me assert this in the test

sekharpanja: perfect.

PR #3048, opened 2020-07-28
Consent service: Transform emails to lowercase

adamledwards:
### Description of change

A bug was resolved that resulted in emails stored in mixed casing was not found by the consent service. This is because the consent service will transform saved data into lowercase. With this fix datahub will transform the emails to lowercase when performing a lookup.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/consent.py
+++ b/datahub/company/consent.py
@@ -44,6 +44,22 @@ class ConsentAPIConnectionError(ConsentAPIException):
     """
 
 
+class CaseInsensitiveDict(dict):
+    """Inherit dict class to make keys case insitive."""
+
+    def __setitem__(self, key, value):
+        """Tranform key to lower case when setting an item."""
+        super().__setitem__(key.lower(), value)
+
+    def __getitem__(self, key):
+        """Tranform key to lower case when getting an item."""
+        return super().__getitem__(key.lower())
+
+    def get(self, key, default=False):
+        """Tranform key to lower case when getting an item using a dict get method."""
+        return super().get(key.lower(), default)
+
+
 def _get_client():
     """
     Get configured API client for the consent service,
@@ -88,7 +104,7 @@ def update_consent(email_address, accepts_dit_email_marketing, modified_at=None)
         'consents': [
             CONSENT_SERVICE_EMAIL_CONSENT_TYPE,
         ] if accepts_dit_email_marketing else [],
-        'email': email_address,
+        'email': email_address.lower(),
     }
 
     if modified_at:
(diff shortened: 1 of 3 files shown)


ambrozic:
Could email values ever be an invalid string or of a non-string type? Should we check for validity or are we fine with errors in this cases?

adamledwards:
@ambrozic there is frontend and DRF validation and we are using the EmailField Django model. There are some records that was imported and did bypass the validation. But we catch 400 error from the consent service

datahub/company/consent.py
```diff
@@ -44,6 +44,18 @@ class ConsentAPIConnectionError(ConsentAPIException):
...
 
 
+class ConsentMapping:
+    """Class around dict to transform keys to lowercase for the consent service."""
+
+    def __init__(self, results):
+        """Initialise the readonly storage class with consent data."""
+        self._storage = results
+
+    def get(self, key, default=False):
```
mcbhenwood, line 54: If you extended the `dict` class and overrode special methods `__setitem__` and `__getitem__` you could do away with this `get()` method and the need for the `self._storage` too:
```
class ConsentMapping(dict):
    def __setitem__(self, key, value):
        super().__setitem__(key.lower(), value)
    def __getitem__(self, key):
        return super().__getitem__(key.lower())
```
Then
```
d = ConsentMapping()
d['FOO@fOo.cOm'[ = 'hiya'
assert d['foo@foo.com'] = 'hiya'
```

datahub/company/consent.py
```diff
@@ -88,7 +100,7 @@ def update_consent(email_address, accepts_dit_email_marketing, modified_at=None)
         'consents': [
             CONSENT_SERVICE_EMAIL_CONSENT_TYPE,
         ] if accepts_dit_email_marketing else [],
-        'email': email_address,
+        'email': email_address.lower(),
```
mcbhenwood, line 103: In my override `dict` suggestion, you wouldn't need this

datahub/company/consent.py
```diff
@@ -106,9 +118,32 @@ def get_many(emails):
...
+                "created_at": "2020-07-27T11:54:51.874350Z",
+                "current": true
+                }
+            ]
+        }
+    }
     """
     body = {
-        'emails': emails,
+        'emails': [email.lower() for email in emails],
```
mcbhenwood, line 146: see comment about overriding `dict`, this would go

datahub/company/consent.py
adamledwards: I was doing some reading on this before I came to my solution. 
In the codebase the consent service is using python Dictionary get() method

```python
d = ConsentMapping()
d['FOO@fOo.cOm']= 'hiya'
assert d['foo@foo.com'] == 'hiya'

assert d.get('foo@foo.com') == 'hiya' #AssertionError

```
This is my thoughts behind the wrapper I implemented.

adamledwards: I will also implement the get method too actually

datahub/company/consent.py
adamledwards: same with this 
The lower is still needed for outbound request. The dict also transforms the keys not the value and because this is a list I want to avoid type checking in the custom dict class.

PR #3062, opened 2020-08-07
Removal/accepts dit email marketing

adamledwards:
### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/search/contact/views.py
```diff
@@ -169,5 +169,15 @@ def _add_consent_response(self, rows):
 
     def _get_rows(self, ids, search_ordering):
         """Get row queryset for constent service if the feature flag is enabled."""
-        rows = super()._get_rows(ids, search_ordering)
+        db_ordering = self._translate_search_ordering_to_django_ordering(search_ordering)
+        field_titles = self.field_titles.copy()
+        del field_titles['accepts_dit_email_marketing']
```
elcct, line 174: The comment in this method seems to be outdated - it talks about feature flag, but one doesn't seem to be used. 

Should `accept_dit_email_marketing` field be removed from `field_titles` rather that here?

I understand this field will no longer exist in the contact search export?

adamledwards: so the accept_dit_email_marketing comes from the consent service. The reason why it is deleted is because we don't want to lookup from the db but we would like to keep the column in the csv. See line 180
https://github.com/uktrade/data-hub-api/blob/2267ca67b0cf8be6c8fd9a728c2877f308444b03/datahub/search/contact/views.py#L174-L182

elcct: That makes sense, thank you!

datahub/company/serializers.py
```diff
@@ -244,14 +254,16 @@ def _notify_consent_service(self, validated_data):
         if 'accepts_dit_email_marketing' not in validated_data:
             # If no consent value in request body
             return
-
+        accepts_dit_email_marketing = validated_data['accepts_dit_email_marketing']
+        # Remove the accepts_dit_email_marketing from validated_data
+        del validated_data['accepts_dit_email_marketing']
```
ambrozic, line 259: Should we rather do `pop` here to achieve the same in a single line?
```
accepts_dit_email_marketing = validated_data.pop('accepts_dit_email_marketing')
```

adamledwards: cheers

PR #3080, opened 2020-08-19
Minimal Standalone API

ambrozic:
### Description of change

This is draft pull request where we review changes and discuss their suitability or recommend new ones.

#### Goals
 - minimal set of required services for successful startup
   - postgres/datahub
   - postgres/mi

 - manage errors from services
   - search over Elasticsearch
   - company matching service (dnb)
   - external dnb api service
   - consent service
   - activity stream
   - sso service
   - export wins api

 - added timeout for all API clients communicating with external services
 - external services using datahub API will have to handle this errors

#### Questions
 - should we raise 502 or 503 error?
 - we could make it work with a single postgres instance and two databases (datahub and mi)


mcbhenwood:
My answers to the Qs:

# 502 vs 503....

Tricky. I still think 502 is the answer. Even though Data Hub isn't just acting as a dumb proxy, it is reporting the unavailability of a downstream service. 503 is more for "go away I'm too busy right now", i.e. it reflects a state in the immediate server, rather than its downstream. I guess we're just falling into the gaps left in an protocol which was defined during an era of quite monolithic architectures

# One Postgres instance with 2 dbs

Yes I think this is a great idea.

config/settings/common.py
```diff
@@ -340,6 +340,8 @@
 
 APPEND_SLASH = False
 
+DEFAULT_SERVICE_TIMEOUT = float(env('DEFAULT_SERVICE_TIMEOUT', default=5.0))
```
mcbhenwood, line 343: Perhaps append the time unit to the property name (which I assume to be `_SEC`) but not entirely sure)

datahub/company/consent.py
```diff
@@ -158,7 +159,7 @@ def get_many(emails):
             json=body,
             params={'limit': len(emails)},
         )
-    except ConnectionError as exc:
+    except APIServiceUnavailableException as exc:
         logger.error(exc)
         error_message = 'Encountered an error connecting to Consent API'
```
mcbhenwood, line 164: Can we reword this to say "Legal Basis API"? 

I appreciate that the whole Python module is incorrectly named and there probably ought to be a ticket to remedy that but hey

docker-compose-minimal.yml
```diff
@@ -0,0 +1,19 @@
...
+      - "5432:5432"
+
+  mi-postgres:
+    image: postgres:9.6
+    restart: always
+    environment:
+      - POSTGRES_DB=mi
+      - POSTGRES_PASSWORD=mi
+    ports:
+      - "54321:5432"
```
mcbhenwood, line 19: I can see you've had to side-step a port clash here. Do these two instances _need_ to be exposed on the Docker host? Or are you catering for people who bring up the API natively rather than inside Docker?

(Noted that your "one Postgres" solution would entirely negate the need for this port choice)

config/settings/common.py
ambrozic: i am fine with either. other timeout settings have no unit. another option would be to add a comment ie `#  in seconds` to all timeout settings (mind blowing, i know)

docker-compose-minimal.yml
ambrozic: this instances do not need to be exposed outside docker network. it is a matter of developer's workflow (native or not). personally i am not a fan of run-everything-in-docker and prefer a native approach. but it is a small change to make it work in native mode.

what i will do:
 - make it run with a single postgres instance with 2 pre-created databases
 - leave ports unexposed (native users can change this manually)
 - change `sample_env` settings to work on closed docker network (native users can change this manually)

config/settings/common.py
elcct: I think our other timeout environment variables use seconds as a unit across, but don't manifest that in their name. I think to be consistent just the DEFAULT_SERVICE_TIMEOUT should be okay.

docker-compose-minimal.yml
elcct: This assumes that developer will have the port 54321 free. Maybe we could use an environment variable for that, e.g. `MI_DATABASE_EXTERNAL_PORT` and then we could also provide a default value like:

```
    ports:
      - "${MI_DATABASE_EXTERNAL_PORT:-54321}:5432"
```
That I think needs bumping the version to `'3'`.

ambrozic: @elcct will try to have both databases in same, single postgres instance

regarding docker compose file versioning - why are we using `2` in most cases? any particular reason for not using `3.x`?

elcct: I can think that maybe because `3.x` was not available when the file has been created and then it was never upgraded.

config/settings/common.py
mcbhenwood: ok conceded

mcbhenwood (approved):

cgsunkel (approved):

PR #3340, opened 2021-03-02
Add US states to metadata API

faissalMT:
### Description of change
This adds the US State field for metadata along with the relevant seed data

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


.gitignore
```diff
@@ -79,3 +79,7 @@ docker-volumes/
 
 # celery
 celerybeat-schedule
+
+# pycharm
```
vfarah-if, line 83: Remove for the global file

datahub/metadata/metadata.py
```diff
@@ -7,7 +7,7 @@
     InvestmentProjectStageSerializer,
     SectorSerializer,
     ServiceSerializer,
-    TeamSerializer,
+    TeamSerializer, StateSerializer,
```
vfarah-if, line 10: Move this one line below

ccp92: Resolved as part of the fixes for flake8

datahub/omis/order/pricing.py
```diff
@@ -20,9 +20,11 @@ def get_pricing_from_order(order, in_pence=True):
     :returns: an instance of OrderPricing from the obj `order`
     """
     if in_pence:
-        transform = lambda x: x  # noqa: E731
+        def transform(x):
+            return x  # noqa: E731
     else:
-        transform = lambda x: x / 100  # noqa: E731
+        def transform(x):
+            return x / 100  # noqa: E731
```
tomkingston, line 27: Looks good - I think you can remove the noqa comment now that you are using funcs instead of lambdas

tomkingston (approved):
Looks good to me

.gitignore
```diff
@@ -79,3 +79,4 @@ docker-volumes/
 
 # celery
 celerybeat-schedule
+
```
ambrozic, line 82: this is unnecessary change

config/settings/local.py
```diff
@@ -1,10 +1,10 @@
+from config.settings.common import *
 import socket
 import environ
 
 environ.Env.read_env()  # reads the .env file
 env = environ.Env()
 
-from config.settings.common import *
```
ambrozic, line 7: this does not work and it breaks settings. this import order is actually important so that env variables are loaded before common settings module.

config/settings/test.py
```diff
@@ -1,9 +1,9 @@
+from config.settings.common import *
 import environ
 
 environ.Env.read_env()  # reads the .env file
 env = environ.Env()
 
-from config.settings.common import *
```
ambrozic, line 6: same as in `local.py`

datahub/metadata/migrations/0007_initial_us_states.yaml
```diff
@@ -0,0 +1,364 @@
...
+    name: Wyoming
+    country: 81756b9a-5d95-e211-a939-e4115bead28a  
+    state_code: WY
+- model: metadata.state
+  pk: 4ad22904-60b4-42af-a3b3-5db055d8f395
+  fields: 
+    disabled_on: null 
+    name: District of Columbia
+    country: 81756b9a-5d95-e211-a939-e4115bead28a  
+    state_code:
```
ambrozic, line 364: no newline at the end of file

tests.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash -xe
-pytest --cov -s $@
+pytest --cov -s $@
```
ambrozic, line 2: no newline at the end of file

datahub/metadata/models.py
```diff
@@ -111,6 +111,22 @@ class Meta(BaseConstantModel.Meta):
...
+
+    state_code = models.CharField(max_length=2, blank=True)
+    country = models.ForeignKey(
+        Country,
+        null=True,
+        blank=True,
+        related_name='states',
+        on_delete=models.PROTECT,
+    )
+
```
ambrozic, line 125: We need to discuss where and how is `State` model going to be used. Will it be used ie on `Company` model as a new field `address_state` including `country` but same `country` would be duplicated as `address_country` on model `Company`?

Also, `state_code` field in migrations is nullable but here it is not. Seems like `State` model was changes after  migration was created?

ambrozic (changes requested):

ambrozic:
Once all issues are addressed we should also squash as many commits as possible. This is small and focused amount of work and maybe even single commit is enough.

config/settings/common.py
```diff
@@ -704,7 +706,8 @@ def _add_hawk_credentials(id_env_name, key_env_name, scopes):
     },
 }
 
-DIT_EMAIL_INGEST_BLACKLIST = [email.lower() for email in env.list('DIT_EMAIL_INGEST_BLACKLIST', default=[])]
+DIT_EMAIL_INGEST_BLACKLIST = [email.lower()
+                              for email in env.list('DIT_EMAIL_INGEST_BLACKLIST', default=[])]
```
elcct, line 710: I would revert those code formatting changes as it makes code look out of place in comparison to the rest of the code base. There are some guidelines: https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md

datahub/metadata/models.py
elcct: I am not sure about the `State` name. It seems too generic and may be confusing without context. Maybe `CountryState` or similar would be more descriptive?

datahub/omis/order/pricing.py
elcct: I would not change the code that does not relate to the goal of the PR

datahub/metadata/models.py
elcct: @ambrozic has a good point regarding the relation to Country. Is there a reason why the relation is "state has a country" rather than "country has a state"?

elcct:
One more thing - we seem to have the US states already in the metadata - https://github.com/uktrade/data-hub-api/blob/9a176dfdab82077ba886ab181e8ff702abb99e5a/datahub/metadata/migrations/0001_administrative_areas.yaml
Is there a reason why this can't be reused?

config/settings/common.py
```diff
@@ -663,6 +663,7 @@ def _add_hawk_credentials(id_env_name, key_env_name, scopes):
 
 # To read data from Activity Stream
 ACTIVITY_STREAM_OUTGOING_URL = env('ACTIVITY_STREAM_OUTGOING_URL', default=None)
+
```
ambrozic, line 666: this change here is still unnecessary

datahub/metadata/admin.py
```diff
@@ -114,6 +114,17 @@ class CountryAdmin(ViewOnlyAdmin):
     list_filter = (DisabledOnFilter, 'overseas_region')
 
 
+@admin.register(models.CountryState)
+class StateAdmin(ViewOnlyAdmin):
```
ambrozic, line 118: this class should be named `CountryStateAdmin`

datahub/metadata/migrations/0007_initial_us_states.py
```diff
@@ -0,0 +1,41 @@
...
+        ('metadata', '0006_update_services'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='CountryState',
+            fields=[
+                ('disabled_on', models.DateTimeField(blank=True, null=True)),
+                ('id', models.UUIDField(default=uuid.uuid4, primary_key=True, serialize=False)),
+                ('name', models.TextField()),
```
ambrozic, line 30: `name` field on `CountryState` model inherited from `BaseConstantModel` should have `blank=True` but here in migration is going to have it set to `blank=False`. this migration should be recreated (or new one added that sets `blank=True` which is unnecessary)

ambrozic:
i hate to be that guy but we are also missing changelog news fragments file. because [rules](https://github.com/uktrade/data-hub-api/blob/9a176dfdab82077ba886ab181e8ff702abb99e5a/changelog/README.md).

ccp92:
Following discussions in this PR, work has been re-done in administrative areas in #3342

PR #3364, opened 2021-03-19
Feature/cds 94 segment

vfarah-if:
### Checklist

* [ x] If this is a releasable change, has a news fragment been added?

  <details>
  
  </details>
  
* [ x] Has this branch been rebased on top of the current `develop` branch?

  <details>
  
  </details>

* [x ] Is the CircleCI build passing?

### General points

<details>

</details>


datahub/company/test/admin/test_company.py
```diff
@@ -28,6 +28,8 @@ def get_required_company_form_data(company=None):
         company = CompanyFactory.build()
 
     data = {
+        'export_segment': '',
```
vfarah-if, line 31: @mforner13 this is odd behaviour as a result of the empty string default. Going forward this will need to be explicitly set, which I think is a downgrade from the default null values. So we are taking up a character of space and we need to fix tests that don't care about export_segment values, which will become hard to understand for developers as the error is subtle. @ambrozic will be able to elaborate more in this

ambrozic: this is what [django documentation](https://docs.djangoproject.com/en/3.1/ref/models/fields/#null) advises. this whole case is a weird combination of a string field that is a choice and usage of admin endpoint to update the model with blank empty strings for this fields. further more if we want to follow on this in whole repository it will involve a lot of migrations and change tests.

rachaelcodes (approved):

datahub/core/constants.py
```diff
@@ -362,3 +362,38 @@ class InvestmentStrategicDriver(Enum):
...
+    """ExportSegment type constants."""
+
+    hep = Constant('hep', 'High export potential')
+    non_fdi = Constant('non-hep', 'Not high export potential')
+
+
+class ExportSubSegment(Enum):
+    """ExportSubSegment type constants."""
+
+    sustain_nurture_and_grow = Constant('sustain_nurture_and_grow',
```
elcct, line 377: Code convention is to have arguments each in a separate line if it is too long, e.g.:

```
  sustain_nurture_and_grow = Constant(
      'sustain_nurture_and_grow',
      'Sustain: nurture & grow',
  )
```

elcct (approved):

PR #3378, opened 2021-03-23
Fix US postcodes and generate address area data

vfarah-if:
### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/company/test/factories.py
```diff
@@ -110,6 +110,16 @@ class Meta:
         model = 'company.Company'
 
 
+class USCompanyFactory(CompanyFactory):
```
elcct, line 113: This should rather be a pytest.fixture. We tend to use specific factories for different kind of objects, e.g. CompanyFactory or SubsidiaryFactory and fixtures for specific attributes of an object.

datahub/core/constants.py
```diff
@@ -21,7 +939,13 @@ class Country(Enum):
     japan = Constant('Japan', '85756b9a-5d95-e211-a939-e4115bead28a')
     montserrat = Constant('Montserrat', '1350bdb8-5d95-e211-a939-e4115bead28a')
     united_kingdom = Constant('United Kingdom', '80756b9a-5d95-e211-a939-e4115bead28a')
-    united_states = Constant('United States', '81756b9a-5d95-e211-a939-e4115bead28a')
+    united_states = CountryConstant(
```
elcct, line 942: Why united_states have a specific constant different from other countries? I am not sure if this is a good idea to involve constants into postcode renaming - shouldn't that be its own thing?

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
```diff
@@ -0,0 +1,149 @@
...
+        self.fix_postcodes_and_address_areas()
+
+    def fix_postcodes_and_address_areas(self):
+        """
+        Does update on the postcode address table
+        """
+        with reversion.create_revision():
+            self.update_address_postcode()
+            self.update_registered_address_postcode()
+            reversion.set_comment('Address Postcode Fix.')
```
elcct, line 41: It seems like tests that check whether a new revision is created are missing.

datahub/core/constants.py
vfarah-if: United states is the only country at the moment that utilises the zip code pattern so we will slowly bring the others inline one country at a time and in the end, they will all utilise the CountryConstant

elcct: If the number of countries is going to grow, eventually this will become a copy of Country metadata, but enhanced with postcode information. Why not adding it to metadata or making it its own thing?

datahub/core/constants.py
```diff
@@ -3,6 +3,924 @@
...
+    'CountryConstant',
+    (
+        'name',
+        'id',
+        'postcode_pattern',
+        'postcode_replacement',
+    ),
+)
+
+US_ZIP_STATES = (
```
elcct, line 16: How this is going to look like for other countries?
Can this be generalised and stored as new metadata objects?

datahub/core/constants.py
vfarah-if: We don't know how this will work in other countries or what priorities this will take, so for now we prefer to keep it the way it is or leave it inside the command file the way it originally was. The only reason this has been extended to constants was based on another review where  where you asked us to create constants in the constants file instand of the command class. I like what you are proposing, just think it can change a little further down the line.

elcct: You defined `UNITED_STATES_ID` in the command which was a duplicate of a constant, so I suggested that you use a constant that is already defined.
I don't think it is a good idea to have multiple schemas for the country constant, as this is confusing and breaks separation of concerns.

datahub/core/constants.py
vfarah-if: At the moment Canada would be the same as United States, but I have no idea if all the other countries would follow a different pattern which is why I can't answer your question. I think this is too early on for us to make a decision which is why the constants where originally inside the command itself.

datahub/core/constants.py
vfarah-if: I have no idea what UNITED_STATES_ID in the command, may have been historical or rectified now.

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
vfarah-if: Fixed but it seems the revision component does not work with batch updates. So I have changed the US postcode to do linear updates through Python which is very bad for database tasks but good for auditing.

docs/Database migration guidelines.md
```diff
@@ -0,0 +1,15 @@
...
+
+1. This utilises a well defined **Django command pattern** meaning you can execute what ever your file name is called as a command, externally and internally, utilising Django e.g. `python manage.py <your_command_name>` or `docker-compose run api python manage.py <your_command_name>`
+
+2. Make sure whatever is done can be rolled back or **audited** using the [Django reversion](https://django-reversion.readthedocs.io/en/stable/) library. This does however come with its own limitations and is probably the main reason this document exists. Django reversion **does not support batch updates**, meaning you always need to do all Postgres updates inefficiently through Python code sequentially one row at a time and cant support either a raw SQL or ORM multi update/create. As slow and inefficient as this seems, and it will be, the only thing that triggers changes on the reversion audit and middleware that listens for changes, is on single changes or commits.
```
elcct, line 7: I am not sure if this document or this paragraph is helpful. The fact that reversion doesn't support bulk updates is mentioned on the first page of reversion documentation. It is a limitation of Django ORM itself as it doesn't send signals that Reversion listens for during a bulk update. It is just a trade off. Doing a single update over entire table may lock it for other operations and you still will end up with corrupted audit log. An update per row does not take a lot of memory and is short enough so that it shouldn't interfere with other operations and as I understand this update doesn't not have to be done in near real-time anyway. So "inefficient" may mean different things depending on the context and here sounds misleading.

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
```diff
@@ -0,0 +1,148 @@
...
+            return True
+        return False
+
+    def us_administrative_area_by_code(self, area_code):
+        """
+        Gets United States Administrative Area by Area Code
+        @param area_code:
+        @return: First Administrative Area Found
+        """
+        return AdministrativeArea.objects.filter(
```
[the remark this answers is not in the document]
vfarah-if: We don't want to go through all the data once something has been updated - happy with this working as is for optimisation.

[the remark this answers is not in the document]
vfarah-if: Production will have 16000 records to go through - this will reduce it for next time

datahub/dbmaintenance/test/commands/test_fix_us_company_address_postcode_for_company_address_area.py
```diff
@@ -0,0 +1,288 @@
...
+    @param post_code: POSTCODE good
+    @param area_code: Area Code to be generated from Command
+    @param area_name: Area Name to describe area code
+    """
+    company = setup_us_company_with_all_addresses(post_code)
+    assert company.registered_address_area is None
+
+    call_command('fix_us_company_address_postcode_for_company_address_area')
+
+    current_company = Company.objects.first()
```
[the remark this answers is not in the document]
vfarah-if: That is much better thanks will fix all references

datahub/core/constants.py
```diff
@@ -3,6 +3,14 @@
 
 Constant = namedtuple('Constant', ('name', 'id'))
 OrderedConstant = namedtuple('OrderedConstant', ('name', 'id', 'order'))
+CountryPostcodeConstant = namedtuple(
```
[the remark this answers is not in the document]
vfarah-if: Fixed. Great suggestion, felt that most files seem to be too large but was thinking this is the pattern for most things here. Will make sure that happens again.

datahub/core/constants.py
```diff
@@ -33,6 +950,17 @@ class Country(Enum):
     united_states = Constant('United States', '81756b9a-5d95-e211-a939-e4115bead28a')
 
 
+class CountryPostcodeFormat(Enum):
+    """Country Postcode Replacement Format"""
+
+    united_states = CountryPostcodeConstant(
+        Country.united_states.value.id,
+        (r'^.*?(?:(\d{5}-\d{4})|(\d{5}\s-\s\d{4})'
```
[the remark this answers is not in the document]
vfarah-if: The expression had a comment with https://regex101.com/r/4XUuax/2 full explanation of the regex but shifting it about I misplaced this next to the file.

docs/Database migration guidelines.md
vfarah-if: In my case I missed this in the documentation because I was not looking for what works and doesn't, I was looking for specific things in the document I may need. I expected this to work for batching as iterating through everything using python would be a last resort, in my opinion, for dealing with database activities. My first solution was a SQL solution because of the dynamic nature of the data and annotations and so this seemed elegant and efficient. The backup mechanism could have been just as simple by selecting into a backup table and restoring from that if need be. The ORM version was a little less elegant but did not work with revision. The final solution is iteration through 16000 records 1 row at a time and seems a very inefficient way of doing what a database does better. Imagine this to be the full company, that would have been even worse, which is what I imply by inefficient sequential update.

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
```diff
@@ -0,0 +1,151 @@
...
+from datahub.core.validate_utils import is_not_blank
+from datahub.metadata.models import AdministrativeArea
+
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Command to query for the ids of duns-linked companies for a particular set of
+    one list tiers and (optionally) particular set of account managers.
```
mforner13, line 20: I think this docstring has accidentally been left in from copying another command :)

mforner13 (commented):
Looks good - comments are around updating/improving wording :)

mforner13:
More details in the PR description would be really useful for reviewers too :) E.g. Why the change is happening, the logic used and why particular choices were made over others.

docs/Database migration guidelines.md
elcct: Proposed other solutions would leave the audit log in corrupted state and in case of pure SQL there was also insufficient test coverage, so given that, you couldn't say they were more efficient, because they didn't fully work. 
The number of records that have to be processed given these constraints is small and should take reasonable time. 
If you could come up with a better way of updating this data while maintaining the integrity of audit log and interoperability that would be amazing and something that could be looked into.

ambrozic: Can we be honest with ourselves and remove this document from PR? It adds no value for an average datahub / django developer.

elcct: I also think it should be removed as it doesn't add value.
Another point is that as the bulk update does not trigger ORM signals, it wouldn't synchronise the records with Elasticsearch as a result and you would have to think how to synchronise the updated records afterwards.

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
```diff
@@ -0,0 +1,151 @@
...
+                        f'"{us_company.registered_address_postcode}"')
+            us_company.save(force_update=True)
+            return True
+        return False
+
+    def format_postcode(self, postcode):
+        """
+        Format postcode with postcode pattern for united states
+        @param postcode:
+        @return: Formatted us postcode value
```
tomkingston, line 132: Sorry - just adding a quick comment to the review, but this docstring format isn't used elsewhere in the repo - other places use the [sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html). For example:
```
:param postcode: the US postcode
:return: formatted US postcode
```

datahub/dbmaintenance/test/commands/test_fix_us_company_address_postcode_for_company_address_area.py
```diff
@@ -0,0 +1,289 @@
...
+    assert actual_result == expected_result
+
+
+@pytest.mark.parametrize('post_code, area_code, area_name',
+                         [('00589', 'NY', 'New York'),
+                          ('00612-1234', 'PR', 'Puerto Rico'),
+                          ('01012', 'MA', 'Massachusetts'),
+                          ('02823', 'RI', 'Rhode Island'),
+                          ])
+def test_us_company_with_unique_zips_generates_valid_address_area(
```
tomkingston, line 112: What you have written here is valid pep8, but the indentation style does have a few disadvantages - a few of which are described in this [SO answer](https://stackoverflow.com/a/39459117), this would be preferred:
```
@pytest.mark.parametrize(
    'post_code, area_code, area_name',
    [
        ('00589', 'NY', 'New York'),
        ('00612-1234', 'PR', 'Puerto Rico'),
        ('01012', 'MA', 'Massachusetts'),
        ('02823', 'RI', 'Rhode Island'),
    ]
)
```

tomkingston: There has been some suggestion of using [Black](https://github.com/psf/black) for strict formatting rules - just trying this out and it would enforce the style I have just suggested.

elcct: We have this doc https://github.com/uktrade/data-hub-api/blob/develop/docs/Coding%20style.md that addresses the style of formatting. 
Not sure if Black can be setup to support this flavour, otherwise we would rather have to reformat the whole repository to be consistent.

docs/Database migration guidelines.md
vfarah-if: Greg and Pawel, the team is now made up of developers who have strengths in the front-end and backend, so it makes sense to not make any assumptions and document a space I didn't find obvious. I agree that I am no Django expert, so I create this document for someone who has the same level as me  and may be hard to understand. The elastic data can be indexed with **python manage.py sync_es**. I will delete this and add it to our confluence.

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
vfarah-if: Thanks Tom will fix up all the references with this issue

datahub/dbmaintenance/management/commands/fix_us_company_address_postcode_for_company_address_area.py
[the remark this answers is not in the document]
vfarah-if: Yes there are duplicate areas when all international data is loaded, which was apparent and obvious when all the international data was in this branch. As this was moved out it seems strange because you don't see the other international data, but going forward, if this method is to be idempotent and consistent, we need to check the country too.

datahub/dbmaintenance/test/commands/test_fix_us_company_address_postcode_for_company_address_area.py
vfarah-if: Fixed

tomkingston (approved):
My comments have been addressed so marking as approved from me (subject to @elcct and @mforner13 approvals)

mforner13 (approved):
👍

PR #3385, opened 2021-03-24
Add v4 events endpoint with two new required fields

faissalMT:
### Description of change

This adds two new fields: `related_trade_agreements_exist` and `related_trade_agreements`.
These are not required fields on the existing v3 events endpoint but are marked as required on the new v4 events endpoint.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/event/test/test_views.py
```diff
@@ -438,6 +455,32 @@ def test_create_end_date_before_start_date(self):
             'end_date': ['End date cannot be before start date.'],
         }
 
+    def test_create_omitted_trade_agreement(self):
+        """Tests specifying an end date before the start date."""
```
tomkingston, line 459: This docstring doesn't relate to the test

datahub/event/serializers.py
```diff
@@ -27,6 +27,9 @@ class EventSerializer(serializers.ModelSerializer):
     teams = NestedRelatedField('metadata.Team', many=True, allow_empty=False)
     address_country = NestedRelatedField('metadata.Country')
     uk_region = NestedRelatedField('metadata.UKRegion', required=False, allow_null=True)
+    related_trade_agreements = NestedRelatedField(
+        'event.TradeAgreement', many=True, required=True, allow_empty=True,
```
tomkingston, line 31: By making this a required field, you are introducing a breaking change to the API - I'm not sure what other services depend on this endpoint, but they will all break if they do not provide 'related_trade_agreements'. It seems like it would work out best all round if this field were not required.

tomkingston (commented):
Looks good, but this introduces a breaking change by making `related_trade_agreements` a required field on an existing endpoint - it seems like it would be best to make this field not required (unless you introduce a new endpoint with a new serializer).

datahub/event/serializers.py
ccp92: I've pinged @ian-leggett a message about this.
The reason we marked the field as required is that the existing validation on the front-end form is driven by whether a field is marked as required on the API side.
Hopefully there will be a way to manage this solely on the front-end side as we don't want to introduce a breaking change if it can be avoided.

ccp92: @tomkingston after chatting to Ian and looking at the [GDS design docs](https://design-system.service.gov.uk/patterns/validation/), we're probably going to go down the route of creating a new endpoint and serializer.
Do you think that would be ok to add to this existing PR or should we split that off?

ccp92:
Please note when reviewing that a large chunk of this PR is the duplication of the V3 events tests to the V4 events tests.
The main difference between these two suites is that `related_trade_agreements_exist` and `related_trade_agreements` are not required on the V3 endpoint but are required on the V4 endpoint.

Similarly in the serializer file, the V4 class is the same as the V3 class but the fields are only required in the V4 class.

datahub/event/urlsv4.py
```diff
@@ -0,0 +1,18 @@
...
+})
+
+itemv4 = EventViewSetV4.as_view({
+    'get': 'retrieve',
+    'patch': 'partial_update',
+})
+
+urlpatterns = [
+    path('event', collectionv4, name='collectionv4'),
+    path('event/<uuid:pk>', itemv4, name='itemv4'),
```
tomkingston, line 17: There is no need to name these `itemv4` and `collectionv4` - they are already url namespaced at the root with `api-v4`.

For consistency, it would be best to define both url types (`urls_v3` and `urls_v4`) in the 'urls.py' file - see https://github.com/uktrade/data-hub-api/blob/develop/datahub/search/urls.py for an example.

datahub/event/test/test_views_v4.py
```diff
@@ -0,0 +1,787 @@
+from operator import itemgetter
```
tomkingston, line 1: It would be best to avoid having a completely new test file for v4 especially as it is so similar to v3 - you could just mark the majority of the v3 tests with [pytest.parrametrize](https://docs.pytest.org/en/stable/parametrize.html) - to test each of the two urls - you probably want something similar to what is here: https://github.com/uktrade/data-hub-api/blob/develop/datahub/admin_report/test/test_views.py#L16

datahub/event/test/test_views_v4.py
```diff
@@ -0,0 +1,787 @@
...
+
```
tomkingston, line 2: You should add a test case for when `related_trade_agreements_exist` is True, but there are no trade agreements specified - presumably it should give a 400 response to say that `related_trade_agreements_exist` but no `related_trade_agreements` were specified.

datahub/event/models.py
```diff
@@ -57,6 +57,8 @@ class Event(BaseModel, DisableableModel):
         related_name='+',
     )
     teams = models.ManyToManyField('metadata.Team', blank=True, related_name='+')
+    related_trade_agreements_exist = models.BooleanField(null=True, blank=True)
```
tomkingston, line 60: `has_related_trade_agreements` sounds a bit better to me - there doesn't seem to be a convention across the project though so no need to worry about this one.

tomkingston (changes requested):
Hi - looks good generally - I've left a few change requests, but hopefully they won't take too long.

datahub/event/serializers.py
```diff
@@ -27,6 +27,134 @@ class EventSerializer(serializers.ModelSerializer):
     teams = NestedRelatedField('metadata.Team', many=True, allow_empty=False)
     address_country = NestedRelatedField('metadata.Country')
     uk_region = NestedRelatedField('metadata.UKRegion', required=False, allow_null=True)
+    related_trade_agreements_exist = serializers.BooleanField(required=False)
+    related_trade_agreements = NestedRelatedField(
+        'event.TradeAgreement', many=True, required=False, allow_empty=True,
+    )
```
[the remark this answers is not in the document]
faissalMT: There are things that are using that data from the V3 endpoint, I think it'd be better if we kept it so things using V3 can still access that data

datahub/event/models.py
elcct: I agree with @tomkingston, I also prefer the `has_...`.

datahub/event/test/test_views_v4.py
faissalMT: I'd like to avoid this just because it might make the relevant tests harder to find as well as change as these two apis diverge going forward

datahub/event/test/test_views_v4.py
faissalMT: If the user doesn't select anything on the frontend despite clicking 'yes' a `null` is sent back, the only time no trade agreements are sent are if 'no' is selected

datahub/cleanup/test/commands/test_delete_orphans.py
```diff
@@ -286,7 +287,7 @@ def test_simulate(
     Test that if --simulate is passed in, the command only simulates the action
     without making any actual changes.
     """
-    # Set up the state before running the command
+    # Set up uthe state before running the command
```
tomkingston, line 290: Typo

datahub/event/serializers.py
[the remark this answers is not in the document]
tomkingston: I'm not quite sure what would be using these fields as they have only just been added? Since they are not required I don't mind them being left in though.

datahub/event/test/test_views_v4.py
tomkingston: Given that the v3 serializer, endpoint and tests will need to be deleted once this is implemented I think that should be fine. However, it does make it very difficult to see what new tests have been added in this PR and the functionality in most cases should be exactly the same, so this does temporarily add a huge amount of code duplication (ie double the maintenance).

datahub/event/test/test_views_v4.py
tomkingston: It sounds like that is dependent on the frontend implementation stopping you do bad things, but the backend should be agnostic of that. At the moment you could end up with bad data being stored in the database, the API could be called with this input...
```
{
    has_related_trade_agreements: True,
    related_trade_agreements: [],
    ...
}
```
Which would result in non-sensical / inconsistent data being stored in the database - it should raise a 400 error to specify that this is bad input. 
Ideally you would also have test cases for the `PATCH` method to make sure that the input does not conflict with what is already stored in the database - ie when an existing event has `has_related_trade_agreements` set to True and then you PATCH with `related_trade_agreements = [] ` (or vice-versa), the data will end up being inconsistent, so this should also raise a 400 error.

tomkingston: I think Chris mentioned just putting `has_related_trade_agreements` on the serializer but not the database model - that would get over the problem of inconsistent data so it would be less important to raise 400 errors when inconsistent data is received.

datahub/event/serializers.py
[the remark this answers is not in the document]
faissalMT: The frontend changes use it at the moment for displaying the data

datahub/event/constants.py
```diff
@@ -11,6 +11,13 @@ class Programme(Enum):
     grown_in_britain = Constant('Grown in Britain', 'd352a68f-aaf4-4c43-b39d-9bca67a8322e')
 
 
+class TradeAgreement(Enum):
+    """Trade agreement constants"""
+
+    uk_australia = Constant('Australia', '50370070-71f9-4ada-ae2c-cd0a737ba5e2')
+    uk_japan = Constant('Japan', '05587f64-b976-425e-8763-3557c7936632')
```
tomkingston, line 18: Should Canada, India, Mexico, New Zealand, USA also appear here?

faissalMT: We're not really using them for any tests, and as far as I know these enums aren't for anything other than tests

datahub/event/serializers.py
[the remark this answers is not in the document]
tomkingston: Oh, sorry, should have been a bit more specific, I just meant the two related_trade_agreements fields which have just been added, but it looks like you have now moved these to only the v4 serializer so all good :+1:

datahub/event/constants.py
tomkingston: Ah, I see, yeah, it looks like you only need these for tests etc so no need to add any more, unless anyone else thinks differently.

datahub/event/serializers.py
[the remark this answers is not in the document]
faissalMT: The v3 serializer still has `related_trade_agreements`, it just doesn't have `related_trade_agreements_exist` since the frontend doesn't really use it for anything aside from validation when submitting.

tomkingston (approved):

ambrozic (approved):

PR #3439, opened 2021-04-15
Add per user feature flag

tomkingston:
### Description of change

Adds a per user feature flag so that we can enable features for specific users

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/company/models/adviser.py
```diff
@@ -96,23 +96,37 @@ class Advisor(AbstractBaseUser, PermissionsMixin):
         help_text='This is the `Email user ID` that is shown for this user in Staff SSO.',
     )
 
+    features = models.ManyToManyField(
+        'feature_flag.UserFeatureFlag',
+        related_name='advisers',
+        blank=True,
+        help_text=(
+            "User's feature flags. Note that the feature flag also needs to be set to active."
```
ambrozic, line 104: double quotes could be replaced with single ones

datahub/company/models/adviser.py
```diff
@@ -96,23 +96,37 @@ class Advisor(AbstractBaseUser, PermissionsMixin):
...
 
-    @cached_property
-    def name(self):
-        """Full name shorthand."""
-        return join_truthy_strings(self.first_name, self.last_name)
+    class Meta:
+        indexes = [
+            models.Index(fields=['first_name', 'last_name']),
+            models.Index(fields=['date_joined', 'id']),
+        ]
```
ambrozic, line 119: was there a particular reason for this indexes? could we add few comments about why and where this indexes are useful?

ambrozic (approved):

datahub/company/models/adviser.py
tomkingston: Hi Gregor - I used the double quotes so that I could use a single quote for User's - would you suggest I just delimit that instead, so 'User\'s...'?

ambrozic: ah, i knew i missed something. it is better as it is.

datahub/company/models/adviser.py
tomkingston: I just moved up the existing Meta class to be above all of the model methods, so the indexes were there previously too. I'm not sure what the indexes are useful for - I can imagine that there might be some lookups by full name, so the first one seems like it could be useful, but I don't really understand the second index since id is a unique index on its own.

PippoRaimondi (approved):

PR #3609, opened 2021-07-06
Prevent blank investment projects attempting to sync to elastic search

mforner13:
### Description of change

This PR fixes a bug which was found in Sentry - [list of instances of this happening](https://sentry.ci.uktrade.digital/organizations/dit/issues/36982/events/c640b3e88f174b0896c817bd01456f2c/events/?environment=production&project=69&statsPeriod=90d). 

Currently, when an interaction is saved, the post-save signal `investment_project_sync_es_interaction_change` gets the primary key of the related investment project and the primary key of any of its previous versions, and then schedules the task `sync_object_async` for each of those primary keys.

However, the signal doesn't prevent this happening when there is no investment project linked to the interaction which has just been saved. This has lead to many tasks failing with the error `InvestmentProject.DoesNotExist: InvestmentProject matching query does not exist.`. 

This PR adds an if statement which first checks whether the interaction has an investment_project linked and returns if it doesn't, or syncs that investment project to elastic search if it does.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/search/investment/signals.py
+++ b/datahub/search/investment/signals.py
@@ -29,7 +29,7 @@ def investment_project_sync_es(instance):
 
 def investment_project_sync_es_interaction_change(instance):
     """
-    Sync investment projects in elastic search when related investments change.
+    Sync investment projects in elastic search when related interactions change.
 
     When an interaction changes, the elastic search index is also updated for
     the related investment project. The previous version also needs to be
@@ -51,7 +51,7 @@ def investment_project_sync_es_interaction_change(instance):
     ):
         pks.append(previous_version.field_dict['investment_project_id'])
 
-    for pk in set(pks):
+    for pk in set([pk for pk in pks if pk is not None]):
         sync_object_async(InvestmentSearchApp, pk)
 
 
--- a/changelog/investment/es-sync-signal-fix.bugfix.md
+++ b/changelog/investment/es-sync-signal-fix.bugfix.md
@@ -1,0 +1,1 @@
+Logic was added to make sure only interactions with an investment project related to them trigger the syncing of an investment project to elastic search.
--- a/datahub/search/investment/test/test_signals.py
+++ b/datahub/search/investment/test/test_signals.py
@@ -1,12 +1,16 @@
 from datetime import date, datetime
 from decimal import Decimal
+from unittest import mock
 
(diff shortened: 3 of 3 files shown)


ambrozic (approved):

cgsunkel (approved):

datahub/search/investment/signals.py
```diff
@@ -29,17 +29,19 @@ def sync_es_wrapper():
...
     """
-    Sync investment projects in elastic search when related investments change.
+    Sync investment projects in elastic search when related interactions change.
 
     When an interaction changes, the elastic search index is also updated for
     the related investment project. The previous version also needs to be
     checked to make sure that if the investment project changes, the old
     investment project is also updated in the index.
     """
+    if not instance.investment_project:
```
elcct, line 39: The function may still generate tasks with empty primary keys, because the `and 'investment_project_id' in previous_version.field_dict` only checks if the `investment_project_id` exists in the `field_dict`, but it can still be `None`. So if there is any previous version without the investment_project_id, the tasks still will be sent. Probably also when the investment project gets removed from the interaction, it will not get updated. 
I think I would leave the function as original, except change `for pk in set(pks):` to something like `for pk in set([pk for pk in pks if pk is not None]):`, so that it will only pass primary keys that exist.

mforner13: Ahh good spot thanks!!

To avoid processing lines unnecessarily, what if I changed the field_dicts section to: ?

```
if (
        previous_version is not None
        and previous_version.field_dict['investment_project_id'] is not None
    ):
```

elcct: That could work too, but I would change to `and previous_version.field_dict.get('investment_project_id') is not None` in case `investment_project_id` doesn't exist. 
I would also leave the original
```
 if instance.investment_project is not None:
        pks.append(instance.investment_project.pk)
```
so that it will still update an investment project if it is removed from interaction.

mforner13: I've gone for your suggestion of filtering the pks in the for loop + checking that the previous version's project id is not None :) Let me know what you think

elcct (approved):

PR #3655, opened 2021-08-02
CDS-248 Added new EU-UK trade agreement to metadata

chloehkwong1:
### Description of change
A new EU-UK trade agreement has been added to trade agreements metadata via a migration. 

We had an issue with getting the migration to work on local development environments and Circle CI due to the migrations running before the `loadinitialmetadata` command. This led to the following error: `failed on setup with "datahub.core.management.commands.loadinitialmetadata.ExistingDataFoundError: Cannot run loadinitialmetadata when metadata already exists. Existing data found for the metadata.tradeagreement model."`

To get round this issue, we have added a `--force` flag wherever `loadinitialmetadata` is called.

The places where `loadinitialmetadata` are called are only used in the development environment so there shouldn't be any implications on production. 

`start.sh` has also been renamed to `start-dev.sh` to ensure it will not be run on production by accident.

Any thoughts on whether we should just default `loadinitialmetadata` to force are welcome. 

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/fixtures/metadata/trade_agreements.yaml
+++ b/fixtures/metadata/trade_agreements.yaml
@@ -22,3 +22,6 @@
 - model: metadata.tradeagreement
   pk: 20e08a38-95a8-4250-bd5b-9d7f0dfc9387
   fields: {disabled_on: null, name: 'UK-United States Mutual Recognition Agreement'}
+- model: metadata.tradeagreement
+  pk: 72b0f84a-77cd-4216-841f-0a75b56f8174
+  fields: {disabled_on: null, name: 'EU-UK Trade Co-operation Agreement'}
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,4 +18,4 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 ADD . /app/
 
-CMD ./start.sh
+CMD ./start-dev.sh
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ shellplus:
 	docker-compose run api python manage.py shell_plus --ipython
 
 load-metadata:
-	docker-compose run api python manage.py loadinitialmetadata
+	docker-compose run api python manage.py loadinitialmetadata --force
 
 setup-flake8-hook:
 	python3 -m venv env
--- a/changelog/add-new-trade-agreement-to-metadata.feature.md
+++ b/changelog/add-new-trade-agreement-to-metadata.feature.md
@@ -1,0 +1,1 @@
+It is now possible to view a new trade agreement `EU-UK Trade Co-operation Agreement` in trade agreements metadata.
--- a/conftest.py
+++ b/conftest.py
@@ -45,9 +45,8 @@ def pytest_sessionstart(session):
 @pytest.fixture(scope='session')
 def django_db_setup(pytestconfig, django_db_setup, django_db_blocker):
     """Fixture for DB setup."""
-    reuse_db = pytestconfig.getoption('reuse_db')
     with django_db_blocker.unblock():
-        call_command('loadinitialmetadata', force=reuse_db)
+        call_command('loadinitialmetadata', force=True)
 
 
 @pytest.fixture(scope='session', autouse=True)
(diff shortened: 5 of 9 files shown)


start.sh
```diff
@@ -4,7 +4,7 @@ python /app/manage.py migrate --database mi --noinput
 python /app/manage.py migrate_es
 # Load initial metadata - ignore errors as we may have already loaded it in to
 # this DB
-python /app/manage.py loadinitialmetadata || true
+python /app/manage.py loadinitialmetadata --force || true
```
ambrozic, line 7: will this now overwrite existing data in database every time server starts?

tomkingston (commented):
I agree with Gregor's comment, but I'm also not sure whether this is the correct approach. 

As I understand it, we have done things like this before without resorting to the force flag. Perhaps @elcct has some insight on this?

start.sh
Staberinde: Annoyingly there's no explicit guidance in the documentation as to what the behavior is on running `loaddata` (which this command calls through to) when existing data is present. 

But, Django itself has test cases that test this behaviour:
See https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/tests/fixtures_model_package/tests.py#L37 and https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/tests/fixtures_model_package/fixtures/model_package_fixture1.json and https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/tests/fixtures_model_package/fixtures/model_package_fixture2.json for an example of the behavior in this situation; which is that the only case where an overwrite of existing data will happen is when there's a primary key collision.

Since this command would hopefully only be run in the local (and maybe dev/staging) environment(s) any data lost this way would be non-critical. It might be annoying.

I think the merit of this approach is that it utilises the existing django migration process for the data migration, and thus ensures that all environments are kept up-to-data with this additional constant value. It also means that no custom scripts which may touch existing data may have to be run on production. 

My inclination is that a data migration is the best approach to the problem of adding the new constant, but I'm happy to be challenged on that. If we do go down that path I can't see another nicer way of having the existing metadata populated in the context of the local (and maybe dev/staging) environments, but I'd welcome your thoughts on this :)

Staberinde: @ambrozic Just thinking about what we were talking about this morning, about the reason for not making the change to `start.sh` being a concern that it might be run on production. Looking in cloudfoundry, it looks like the entry-point for production is `web.sh` which is independent from `start.sh`, so I don't think the change here has any risk of being run on production.

ambrozic (approved):

fixtures/metadata/trade_agreements.yaml
```diff
@@ -1,3 +1,4 @@
+# This list has been updated to include 'EU-UK Trade Co-operation Agreement' via a migration located in `datahub/metadata/migrations/0016_update_trade_agreements.py`
```
mforner13, line 1: Sorry if this has already been discussed - but if this is a fixture file, why not add in the new trade agreement rather than leaving a comment? :)

chloehkwong1: @mforner13 If we were to put the new trade agreement in the fixture file, the `loaddata` command would need to be run on production for it to appear. However, last time we tried this Gregor ended up fixing stuff for 2 weeks 😅  Therefore we decided to add the new trade agreement via a migration and left a comment in the fixtures file so anyone looking would know there is an extra trade agreement :) Hope this explains it?

mforner13: Ahh okay, so if you had the new trade agreement in the migration and the fixture file, outside of prod it would try to add the same fixture twice?

chloehkwong1: That's a good point! I've added it to the fixtures file too now, thanks Miriam :)

mforner13 (approved):
👍

PR #3668, opened 2021-08-13
CDS-277 Add validation to make postcode mandatory for us/canadian companies

chloehkwong1:
### Description of change

This PR adds new validation on company address fields which make postcode mandatory for both US and Canadian addresses.

This is currently placed behind feature-flag: `address-postcode-company-required-field`

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/test/test_company_views.py
+++ b/datahub/company/test/test_company_views.py
@@ -1348,7 +1348,7 @@ class TestAddCompany(APITestMixin):
     """Tests for adding a company."""
 
     @pytest.mark.parametrize(
-        'data,expected_response',
+        'data, expected_response, feature_flags',
         (
             # uk company
             (
@@ -1395,13 +1395,15 @@ class TestAddCompany(APITestMixin):
                         'name': HeadquarterType.ghq.value.name,
                     },
                 },
+                [],
             ),
-            # non-UK
+            # US company
             (
                 {
                     'address': {
                         'line_1': '75 Stramford Road',
                         'town': 'Cordova',
+                        'postcode': '91210',
                         'area': {
                             'id': AdministrativeArea.texas.value.id,
                         },
@@ -1416,7 +1418,7 @@ class TestAddCompany(APITestMixin):
                         'line_2': '',
                         'town': 'Cordova',
                         'county': '',
-                        'postcode': '',
+                        'postcode': '91210',
                         'area': {
                             'id': AdministrativeArea.texas.value.id,
                             'name': AdministrativeArea.texas.value.name,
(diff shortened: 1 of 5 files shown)


datahub/company/test/test_company_views.py
```diff
@@ -1420,7 +1420,7 @@ class TestAddCompany(APITestMixin):
     """Tests for adding a company."""
 
     @pytest.mark.parametrize(
-        'data,expected_response',
+        'data, expected_response, feature_flags',
```
Staberinde, line 1423:
```suggestion
        'data,expected_response,feature_flags',
```
Pytest used to only allow for this first parameter to [`pytest.mark.parameterize`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.python.Metafunc.parametrize) to be separated by `,` and not `, `. If the function handles spaces in the string in the desired way right now then that's cool, but the fact it's not expressly included in the linked documentation makes me think that it's implicit functionality that might be removed in a future release. Plus the other usages of this function in this module don't have spaces in the equivalent argument, so it's good to keep the same conventions

datahub/company/test/test_company_views.py
```diff
@@ -1797,6 +1846,50 @@ def test_required_fields(self):
...
+            (
+                {
+                    'company_number': '123',
+                    'address': {
+                        'line_1': '1000 Mulholland Drive',
+                        'town': 'LA',
+                        'country': {
+                            'id': Country.united_states.value.id,
+                        },
+                        'address_area': 'Burbank',
```
Staberinde, line 1859: This field doesn't look right; for one the other keys in this dictionary don't have `address_` prepended, and shouldn't the value here be in a similar format to that of the `country` field? i.e. `{"id": AdministrativeArea.california.value.id}`?

datahub/core/serializers.py
```diff
@@ -377,36 +378,69 @@ class MultiAddressMeta(self.Meta):
...
-                            f'{self.address_source_prefix}_country',
-                            (
-                                CountryEnum.united_states.value.id,
-                                CountryEnum.canada.value.id,
-                            ),
-                        ),
-                    ),
-                ),
-            )
+            self.area_validator(validators)
```
Staberinde, line 437: I think it's probably worth making this function name more explicit by prepending an explicit action. e.g. `add_area_validator`

datahub/core/serializers.py
```diff
@@ -377,36 +378,69 @@ class MultiAddressMeta(self.Meta):
...
+
+        if (
+            self.postcode_can_be_required
+            and is_feature_flag_active('address-postcode-company-required-field')
+        ):
+            self.postcode_validator(validators)
```
Staberinde, line 443: Same as above

datahub/core/serializers.py
chloehkwong1: Good suggestion, thanks Toby

datahub/company/test/test_company_views.py
chloehkwong1: Hi @Staberinde, in regards to the `address_` prefix we just followed the existing pattern for area in the other tests. Also, `add_area_validator()` seems to suggest that it is looking for `address_area` rather than `area`. I agree it is strange but we'll go with this for now as based on the information we have it seems correct.

Definitely a valid statement about address_area being a constant, that has been changed now!

Thanks :)

LaurenQurashi (approved):

dredmonds (approved):

PR #3691, opened 2021-08-31
Upgrade Django to 3.2.

elcct:
### Description of change

This updates the Django to 3.2.7.

The upgrade required an update to app labels, which in turn would have changed the database table names. 
Renaming existing database tables could affect external apps, particularly `mi`, therefore the affected models are
now pointing at legacy names. 
This change requires faking those migrations for existing environment and update `django_migrations` and `django_content_type` tables.
The commands that perform the database update are in the `web.sh` script.

First it runs app label renaming, for the main database and then for mi and then it fakes migrations that would have changed the new table names to legacy names (which already exist in a live environment).

After the app is released, another PR should follow that reinstates the original `web.sh` file.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


conftest.py
```diff
@@ -127,19 +126,13 @@ def s3_stubber():
...
-    monkeypatch.setitem(
-        settings.CACHES,
-        'default',
-        {'BACKEND': 'django.core.cache.backends.locmem.LocMemCache'},
-    )
-    cache_handler = CacheHandler()
-    monkeypatch.setattr('django.core.cache.caches', cache_handler)
+def local_memory_cache():
+    """Get local memory cache."""
+    from django.core.cache import cache
```
faissalMT, line 131: Why import in the function?

datahub/omis/market/migrations/0002_alter_market_table.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('omis_market', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterModelTable(
+            name='market',
+            table='omis-market_market',
```
faissalMT, line 15: If I've understood these changes correctly, wouldn't this need to be `omis_market_market` instead?

web.sh
```diff
@@ -4,17 +4,33 @@
...
+./manage.py distributed_migrate omis_payment 0009_auto_20210816_1601 --fake --database mi
+./manage.py distributed_migrate omis_market 0002_alter_market_table --fake
+./manage.py distributed_migrate omis_market 0002_alter_market_table --fake --database mi
+./manage.py distributed_migrate omis_invoice 0013_alter_invoice_table --fake
+./manage.py distributed_migrate omis_invoice 0013_alter_invoice_table --fake --database mi
+./manage.py distributed_migrate omis_quote 0009_auto_20210816_1601 --fake
+./manage.py distributed_migrate omis_quote 0009_auto_20210816_1601 --fake --database mi
+
+# TODO: the migration commands below should be re-enabled after Django 3.2 upgrade
+# ./manage.py distributed_migrate --noinput
```
faissalMT, line 23: It might be a good idea to control all this with an environment variable instead of relying on a subsequent commit

datahub/omis/market/migrations/0002_alter_market_table.py
elcct: When running a fresh environment, Django "thinks" that the original table name is `omis_market_market` and it is going to create it like that, but in live we are still using `omis-market_market`, so this migration only takes local environment in line with live and that's why it has to be faked when running against production, as the table is already called `omis-market_market`.

tomkingston (approved):
Looks good - must have taken a while to fix all of it!

cgsunkel (approved):

datahub/core/management/commands/rename_apps.py
```diff
@@ -0,0 +1,87 @@
+from django.core.management import BaseCommand, CommandError
+from django.db import connections, DEFAULT_DB_ALIAS
+from django_pglocks import advisory_lock
+
+
+class AppLabelAlreadyRenamedError(CommandError):
+    """Error raised when new app_label or app is found when executing rename_apps command."""
+
+
+class Command(BaseCommand):
```
Staberinde, line 10: Out of interest, what was your reasoning for making this a `Command` rather than a django database migration?

elcct: This command is intended to update a live environment only and it will be removed once upgrade is completed. I think migration would make it overly complex, as there would be no need to apply it for any new environment.

web.sh
elcct: Actually the script would have the same effect if ran multiple times, so I uncommented the migration commands.

datahub/core/management/commands/rename_apps.py
Staberinde: Fair enough, I have no concerns about it being a command. I'm just always interested in peoples reasoning when the approaches differ from my own in case there's some hidden bit of complexity i'm not seeing :)

PR #3692, opened 2021-08-31
Add summaries to investment search endpoint

tomkingston:
### Description of change

Adds a summary to the investment search endpoint when the `show_summary` data param is set to true. This includes a count of projects in each stage.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/search/investment/views.py
+++ b/datahub/search/investment/views.py
@@ -124,6 +124,52 @@ class SearchInvestmentProjectAPIViewMixin:
 class SearchInvestmentProjectAPIView(SearchInvestmentProjectAPIViewMixin, SearchAPIView):
     """Filtered investment project search view."""
 
+    def get_base_query(self, request, validated_data):
+        """Add aggregations to show the number of projects at each stage."""
+        base_query = super().get_base_query(request, validated_data)
+        if validated_data.get('show_summary'):
+            base_query.aggs.bucket('stage', 'terms', field='stage.id')
+        return base_query
+
+    def enhance_response(self, results, response):
+        """Add summary data to the response."""
+        if 'stage' not in results.aggs:
+            return response
+
+        response['summary'] = {
+            InvestmentProjectStage.prospect.name: {
+                'label': 'Prospect',
+                'id': InvestmentProjectStage.prospect.value.id,
+                'value': 0,
+            },
+            InvestmentProjectStage.assign_pm.name: {
+                'label': 'Assign PM',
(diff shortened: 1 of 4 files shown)


datahub/search/investment/views.py
```diff
@@ -124,6 +124,52 @@ def get_extra_filters(self, validated_data):
...
+                'value': 0,
+            },
+            InvestmentProjectStage.won.name: {
+                'label': 'Won',
+                'id': InvestmentProjectStage.won.value.id,
+                'value': 0,
+            },
+        }
+        for stage_summary in results.aggs['stage'].buckets:
+            stage = InvestmentProjectStage.get_by_id(stage_summary['key'])
```
ambrozic, line 167: would it be possible to query all InvestmentProjectStages at once?

tomkingston: The `InvestmentProjectStage` here is just an enumerated type instead of a database model so there won't be any querying - it's possible there's a neater way of doing it though.

datahub/search/investment/views.py
```diff
@@ -124,6 +124,52 @@ def get_extra_filters(self, validated_data):
...
+        return base_query
+
+    def enhance_response(self, results, response):
+        """Add summary data to the response."""
+        if 'stage' not in results.aggs:
+            return response
+
+        response['summary'] = {
+            InvestmentProjectStage.prospect.name: {
+                'label': 'Prospect',
```
elcct, line 141: Why `label`? We tend to use `id` and `name` pairs, so `id` and `label` looks odd.

tomkingston: Hi Pawel - the label is there for the front end to render something human readable, so whereas the name might be `assign_pm`, the label is `Assign PM`. I would have left it up to the frontend to convert the `name` into title case and replace underscores for spaces, but that doesn't always work that well (in this case it would end up with `Assign Pm`). 

I could possibly render this as a list instead of a dict though, so something like:
```
[
  {
    'name': 'assign_pm',
    'label': 'Assign PM',
    'id': <uuid>,
    'value': 5,
  },
  ...
]
```
That might need some tweaks on the frontend though (we are already using the data format here for the Investment Summary component on the personalised dashboard).

elcct: Fair point!

elcct (approved):

ambrozic (approved):

PR #3797, opened 2021-11-05
CDS-335 Include turnover_gbp in company API responses

Staberinde:
### Description of change

* `turnover_gbp` has been added to response for company API GET requests (`/v4/company` and `/v4/company/{id}`
  * Current `turnover` field (in USD) remains unchanged

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/dnb_api/serializers.py
```diff
@@ -89,6 +89,58 @@ class DNBCompanySerializer(CompanySerializer):
     )
 
     class Meta(CompanySerializer.Meta):
+        # Overriding fields to remove turnover_gbp field not present in DnB API response
+        fields = (
```
mforner13, line 93: Would `exclude = ['turnover_gbp']` work in this case?

mforner13 (commented):
One small question :)

datahub/dnb_api/serializers.py
Staberinde: So I did have a look at going down that path, and it turns out that there's a lot of other fields that are excluded from the serializer response. Rather than having the parent serializer defined by an inclusive list (`fields`) and the child serializer defined by an exclusive list (`exclude`) it seemed less complex to make both inclusive lists. 

I know we wind up with a lot of duplication doing it this way, but there is the additional advantage that we can safely add fields to `CompanySerializer` without having those fields implicitly utilized in the `DNBCompanySerializer`.

To be honest I think there's probably some technical debt here around refactoring `datahub/company/serializers.py` and `datahub/dnb_api/serializers.py` to use some kind of Mixin pattern, to reduce the complexity of this whole hierarchy.

mforner13: Cool, seems like the sensible option :)

faissalMT (approved):

PR #3814, opened 2021-11-15
Accept turnover in GBP for DNB change requests

faissalMT:
### Description of change

In order to move currency conversion out of the frontend and to avoid having multiple sources of truth for conversion rates, we need to be able to convert from gbp to usd for D&B change requests before sending them to the D&B API. Sending currencies as usd via the `turnover` parameter is still possible, but will be deprecated in favor of providing `turnover_gbp` to the change request going forward.

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>


datahub/metadata/migrations/0019_add_gbp_exchange_rate.yaml
```diff
@@ -0,0 +1,6 @@
+- model: metadata.exchangerate
+  pk: e7927ac0-41c0-41b1-99b9-5de3497ac73b
+  fields:
+    from_currency_code: 'GBP'
```
Staberinde, line 4: You don't need to store both USD->GBP and GBP->USD exchange rate; the GBP->USD exchange rate is equal to 1 / USD->GBP exchange rate

craigjbass (approved):

datahub/metadata/utils.py
```diff
@@ -7,8 +7,26 @@ def convert_usd_to_gbp(usd):
...
+    """get_latest_exchange_rate.
+
+    :returns: A numeric value for latest USD to GBP exchange rate
+    """
+    return ExchangeRate.objects.filter(
         from_currency_code='USD',
         to_currency_code='GBP',
-    ).order_by('-created_on').first()
-    return usd * exchange_rate.exchange_rate
+    ).order_by('-created_on').first().exchange_rate
```
tomkingston, line 32: I suppose you would get a TypeError if no USD / GBP exchange rates exist (None doesn't have an exchange_rate property), but you probably don;t need to worry about that

datahub/dnb_api/serializers.py
```diff
@@ -296,7 +297,18 @@ class ChangeRequestSerializer(serializers.Serializer):
     name = serializers.CharField(source='primary_name', required=False)
     trading_names = serializers.ListField(required=False)
     number_of_employees = serializers.IntegerField(source='employee_number', required=False)
+
+    # the turnover field is deprecated, in favor of the turnover_gbp field which
```
tomkingston, line 301: British spelling is favour :)

tomkingston: Is turnover actually deprecated, or would that still be used if we want the turnover in USD?

tomkingston (approved):
Approved - a couple of queries but nothing that should stop it getting merged

datahub/dnb_api/serializers.py
ccp92: I think the message here could be clarified. I believe `turnover` is deprecated in the sense that it will no longer be used in the frontend. However, I believe the D&B data will still be passed into the original turnover field.
I'll look to clarify this and get the comment updated 👍

datahub/metadata/utils.py
ccp92: We're moving the hard-coded front-end exchange rate into the backend for now, so it should always be present. (There's a future ticket of work to get that rate update on a more regular basis, but as long as we're not erasing the existing value, there should always be an exchange rate)

PR #4004, opened 2022-03-03
WIP: Remove user feature flag from Fuzzy search

LaurenQurashi:
### Description of change

This PR removes the user feature flag that fuzzy search was hidden behind, in order to open it up for all users. 

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/search/test/test_views.py
+++ b/datahub/search/test/test_views.py
@@ -364,7 +364,7 @@ class TestBasicSearch(APITestMixin):
         """
         SimpleModel.objects.create(name='Smaxtec Limited', address='')
         SimpleModel.objects.create(name='Newsmax Media (HQ Florida)', address='')
-        SimpleModel.objects.create(name='Smart Notebooks', address='Maxet House')
+        SimpleModel.objects.create(name='Smooth Notebooks', address='Smaxet House')
         SimpleModel.objects.create(name='Other Notebooks', address='Maxet House')
 
         opensearch_with_collector.flush_and_refresh()
@@ -390,7 +390,7 @@ class TestBasicSearch(APITestMixin):
         assert [
             'Smaxtec Limited',
             'Newsmax Media (HQ Florida)',
-            'Smart Notebooks',
+            'Smooth Notebooks',
         ] == [result['name'] for result in response.data['results']]
 
     def test_partial_match(self, opensearch_with_collector, search_support_user):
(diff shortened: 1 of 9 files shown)


cgsunkel (approved):

datahub/search/test/test_views.py
```diff
@@ -357,22 +357,22 @@ def test_fuzzy_quality_cross_fields(
...
-        fuzzy_search_user,
+        search_support_user,
     ):
         """
         Tests that name is more important than other fields in cross field matches.
         """
         SimpleModel.objects.create(name='Smaxtec Limited', address='')
         SimpleModel.objects.create(name='Newsmax Media (HQ Florida)', address='')
-        SimpleModel.objects.create(name='Smart Notebooks', address='Maxet House')
+        SimpleModel.objects.create(name='Smooth Notebooks', address='Smaxet House')
```
LaurenQurashi, line 367: This test has been flaky for a while. When Open Search's algorithm is deciding what results are similar, it applies a score to the result to determine what order they should be listed in. We added weighting to the name field, hence why this test is checking for names with matches before addresses, but Smart Notebooks had partial matches in both addresses field and name fields, which would sometimes tip its score just slightly above Newsmax every now and then (scores aren't the same each time they're run weirdly).  To keep this test checking that names come before addresses, I've changed the address to be a direct match and reduced the amount of matching letters in the name field. Testing wise, this is important as we have to dictate where the results should be in a list, but the real-life implication of this is a small coin flip of whether it's the second item or the third.

datahub/search/views.py
```diff
@@ -127,10 +125,8 @@ def get(self, request, format=None):
             *(self.fields_to_exclude or ()),
         )
 
-        fuzzy_search_enabled = is_user_feature_flag_active(
-            user=request.user,
-            code=FUZZY_SEARCH_USER_FEATURE_FLAG,
-        )
+        fuzzy_search_enabled = True
```
LaurenQurashi, line 128: Should fuzzy search ever have any issues, we can turn this switch to False and the old stricter search will be used instead.

tomkingston: I think best just to pass this directly into the function on line 137 (`fuzzy=True`) instead of assigning it and then passing it through

tomkingston (approved):

PR #4165, opened 2022-06-06
Feature/det 176 add events to activity stream

richmadetech:
### Description of change

We want to be able to save Data Hub Events as objects into Activity Stream (interactions are currently all that are saved)

This work now adds the list of events to activity stream, following a similar pattern to interactions (including a serialiser, queryset, and associated test)

### Checklist

* [x] Added add-events-to-activity-stream.feature.md to the events file in the Changelog fodler 

    
* [x] Is the CircleCI build passing


datahub/activity_stream/event/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+    def to_representation(self, instance):
+        event_id = f'dit:DataHubEvent:{instance.pk}'
+        event = {
+            'id': f'{event_id}:Announce',
+            'type': 'Announce',
+            'published': instance.created_on,
+            'generator': self._get_generator(),
+            'object': {
+                'id': event_id,
+                'type': 'dit:Event',
```
Harriethw, line 20: question: I notice for the other activity stream serializers (e.g. company_referral) we're saving the `object.type` field as an array of strings, even if it's only one string. From the Activity Stream specs, it doesn't look like it needs to be an array, but maybe we should be consistent with the others?

Harriethw, line 20: question: I notice for the other activity stream serializers (e.g. company_referral) we're saving the `object.type` field as an array of strings, even if it's only one string. From the Activity Stream specs, it doesn't look like it needs to be an array, but maybe we should be consistent with the others?

datahub/event/queryset.py
```diff
@@ -0,0 +1,18 @@
+from datahub.event.models import Event
+
+
+def get_base_event_queryset():
+    """
+    Gets the base eveb queryset.
```
Harriethw, line 6: minor: typo on event here :)

Harriethw, line 6: minor: typo on event here :)

Harriethw:
LGTM! v cool to see it coming through on the `/v3/activity-stream/event` endpoint. let me know if you'd like help tidying up the commits.

Harriethw:
LGTM! v cool to see it coming through on the `/v3/activity-stream/event` endpoint. let me know if you'd like help tidying up the commits.

onyela-o (approved):

PR #4170, opened 2022-06-09
Added None checks to catch possible empty relational data for the eve…

richmadetech:
…nts information

### Previously merged wasn't working on Develop due to "None" type error

Have added additional checks to ensure that attributes that are None type are not attempted to be read.

### Checklist

### DIFF
--- a/datahub/activity_stream/event/serializers.py
+++ b/datahub/activity_stream/event/serializers.py
@@ -32,7 +32,6 @@ class EventActivitySerializer(ActivitySerializer):
                 'dit:address_postcode': instance.address_postcode,
                 'dit:address_country': {'name': instance.address_country.name},
                 'dit:disabledOn': instance.disabled_on,
-                'dit:service': {'name': instance.service.name},
                 'dit:archivedDocumentsUrlPath': instance.archived_documents_url_path,
                 'dit:eventType': {'name': instance.event_type.name},
             },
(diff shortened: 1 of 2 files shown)


datahub/activity_stream/event/serializers.py
```diff
@@ -21,20 +21,15 @@ def to_representation(self, instance):
...
                 'dit:address_1': instance.address_1,
                 'dit:address_2': instance.address_2,
                 'dit:address_town': instance.address_town,
                 'dit:address_county': instance.address_county,
                 'dit:address_postcode': instance.address_postcode,
-                'dit:address_country': {'name': instance.address_country.name},
-                'dit:leadTeam': {'name': instance.lead_team.name},
-                'dit:organiser': {'name': instance.organiser.name},
                 'dit:disabledOn': instance.disabled_on,
                 'dit:service': {'name': instance.service.name},
```
Harriethw, line 34: looking again at the Interaction serializer and model, `service` could also be null/blank so we probable want a check around it also

Harriethw (approved):

datahub/activity_stream/event/serializers.py
richmadetech: Yea good spot, I mainly missed it, I've also removed two others as they can't be set to Null.

cgsunkel (approved):

PR #4181, opened 2022-06-13
Refactor Celery tasks that manually do retries and apply synchronously

vfarah-if:
### Description of change

Remove Celery code that looks like it may potentially override internal Celery logic and that will make it easier to map on RQ

### Checklist

  <details>
  
  </details>
  
  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/dnb_api/link_company.py
+++ b/datahub/dnb_api/link_company.py
@@ -30,7 +30,7 @@ def link_company_with_dnb(
     company.duns_number = duns_number
     company.modified_by = modified_by
     company.save()
-    sync_company_with_dnb.apply(
+    sync_company_with_dnb.apply_async(
         args=(company.id,),
         kwargs={
             'update_descriptor': update_descriptor,
--- a/datahub/dnb_api/link_company.py
+++ b/datahub/dnb_api/link_company.py
@@ -30,13 +30,10 @@ def link_company_with_dnb(
     company.duns_number = duns_number
     company.modified_by = modified_by
     company.save()
-    sync_company_with_dnb.apply_async(
-        args=(company.id,),
-        kwargs={
-            'update_descriptor': update_descriptor,
-            'retry_failures': False,
-        },
-        throw=True,
+    sync_company_with_dnb(
+        company.id,
+        update_descriptor=update_descriptor,
+        retry_failures=False,
     )
     company.refresh_from_db()
     return company
--- a/datahub/dnb_api/management/commands/update_companies_from_dnb_service.py
+++ b/datahub/dnb_api/management/commands/update_companies_from_dnb_service.py
@@ -43,7 +43,7 @@ class Command(BaseCommand):
         """
         Run the celery task.
         """
-        get_company_updates.apply(
+        get_company_updates.apply_async(
             kwargs={
                 'last_updated_after': options['last_updated_after'],
                 'fields_to_update': options['fields'],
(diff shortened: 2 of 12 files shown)


datahub/dnb_api/test/test_views.py
```diff
@@ -1052,7 +1082,7 @@ def test_dnb_service_error(
             },
         )
 
-        assert response.status_code == status.HTTP_502_BAD_GATEWAY
+        assert response.status_code == status.HTTP_200_OK
```
tomkingston, line 1085: Looks like the behaviour has changed here - could this be a problem?

To avoid repetition you could parametrize so that you only need one function.

tomkingston (commented):
Looks like a good improvement - I think the final test needs checking to see that this is definitely the behaviour you want

docker-compose.yml
```diff
@@ -44,6 +44,8 @@ services:
     environment:
       - POSTGRES_DB=datahub
       - POSTGRES_PASSWORD=datahub
+    ports:
```
tomkingston, line 47: Is this needed? postgres should be available internally without exposing the ports here - with this line there will be a port conflict with locally running instances of postgres

tomkingston (approved):
Looks good - just the query on exposing the docker postgres port

dredmonds (approved):

datahub/dnb_api/link_company.py
```diff
@@ -30,7 +30,7 @@ def link_company_with_dnb(
     company.duns_number = duns_number
     company.modified_by = modified_by
     company.save()
-    sync_company_with_dnb.apply(
+    sync_company_with_dnb.apply_async(
```
elcct, line 33: The logic in this function relies on the synchronous execution of the task, as it returns the company record that should include the updates done by the task. If you use `apply_async` this way, the task will be sent off to the worker without waiting for its completion and so the returned company record most likely will not have any changes applied, depending on how busy is the worker.
Probably something like:
```
result = sync_company_with_dnb.apply_async(...
result.wait()
```
would make it work as expected, but I am not sure then if it makes sense to use a task at all here. Maybe this needs to be refactored.

datahub/dnb_api/management/commands/update_companies_from_dnb_service.py
```diff
@@ -43,7 +43,7 @@ def handle(self, *args, **options):
         """
         Run the celery task.
         """
-        get_company_updates.apply(
+        get_company_updates.apply_async(
```
elcct, line 46: Similar issue - this is not going to wait for the task to complete and so there will be no information whether there was any problem with the updates.

datahub/dnb_api/tasks/sync.py
```diff
@@ -97,7 +95,7 @@ def sync_company_with_dnb_rate_limited(
         return
 
     try:
-        sync_company_with_dnb.apply(
+        sync_company_with_dnb.apply_async(
```
elcct, line 98: That function probably shouldn't be running a task, but rather just calling the `_sync_company_with_dnb` function directly.

datahub/dnb_api/management/commands/update_companies_from_dnb_service.py
vfarah-if: On a get company updates I am also happy to simply execute the task here and remove celery logic

datahub/dnb_api/tasks/sync.py
vfarah-if: I thought the same but maybe the retry error handling is why it exists but I also feel the error handling is faulted, may be worth a chat initself

docker-compose.yml
vfarah-if: Or can I expose this as 543211?

PR #4296, opened 2022-08-23
Fix CloudFoundry issue and change master to main

marijnkampf:
### Description of change

Fix deployment error
Renamed master to main

### Notes
To fix the issue [fatal error: Python.h: No such file or directory](https://github.com/cloudfoundry/python-buildpack/issues/574) setuptools needed to be included in the requirements. This was previously specifically excluded. Once the underlying issue is resolved, the setuptools should be removed from the requirements file again.

### Checklist

* [X] Has this branch been rebased on top of the current `main` branch?
* [X] Is the CircleCI build passing?
* [ ] Rename master to main on [Jenkins default value](https://jenkins.ci.uktrade.digital/view/Datahub/job/datahub-api/configure)


requirements.txt
```diff
@@ -232,3 +232,4 @@ zope-interface==5.4.0
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
+setuptools==65.1.1
```
elcct, line 235: Is this change coming from `requirements.in` and it was not compiled before?

marijnkampf: Due to a change in CloudFoundry an error occurred whilst deploying. Setuptools was previously not necessary, but is now included as a temporary fix until the underlying issue is resolved. See https://github.com/cloudfoundry/python-buildpack/issues/574. I've added a comment in the code.

elcct: Hmm my point is that it should be added to `requirements.in` and then to `requirements.txt` through `pip-compile`. Otherwise if someone adds a new dependency and do `pip-compile` the `setuptools` entry may disappear.

https://github.com/uktrade/data-hub-api/blob/master/docs/Managing%20dependencies.md#how-to-add-a-new-dependency

marijnkampf: Aahh, I see. I've moved it to requirements.in, good catch.

elcct (approved):

paulgain (approved):

PR #4298, opened 2022-08-25
Store email notification id against a reminder.

elcct:
### Description of change

We need to track the delivery of reminder emails. Emails are being sent using GOV.UK Notify service. The service returns a `notification_id`, which then can be used to poll the Notify API for the email delivery status (using added `get_notification_by_id` function).
Given that an email corresponds to a reminder object, it seems natural to store the `notification_id` in the reminder.
Two issues - one is that the user can delete a reminder - so the user could also delete the `notification_id` with it and we wouldn't be able to track the delivery. This PR adds a soft delete for reminders. Instead of reminder being deleted, it gets a `Dismissed` status and will no longer be returned by the reminder API, however it will still keep the `notification_id` and we can still look for the delivery status update.
The second issue is that for estimated land date reminders, when a number of reminders exceeds the threshold (5), then we only send one summary email containing all the reminders. In this situation we are updating all reminders with the `notification_id`, so that we can find out which reminders were sent in a summary email by grouping by `notification_id`.

This PR will be followed by a second one that is polling the Notify API and updates the delivery status.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


marijnkampf (commented):
Would it be worth to add a tests for the code changes in core.py?

paulgain (approved):

elcct:
> Would it be worth to add a tests for the code changes in core.py?

@marijnkampf great point. I removed these changes as these are not actually used in this PR and will introduce it in the following one.

marijnkampf (approved):

PR #4468, opened 2023-02-03
Feature/rr 633 activate users to receive notifications

chopkinsmade:
### Description of change
- Added a task to auto migrate ITA user's meeting the criteria to receive notification reminders
- Added a task to auto migrate POST user's meeting the criteria to receive notification reminders
- Added a cron schedule for the tasks to run every day between midnight and 1 am. This is to allow new users to be migrated before the email sending schedule starts at 8am
- Updated the selection criteria for getting advisor companies to include the overseas post accounts
- Added a config setting to enable/disable the auto migration of users

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/reminder/tasks.py
```diff
@@ -397,7 +452,10 @@ def generate_no_recent_interaction_reminders_for_subscription(subscription, curr
                     continue
                 send_reminder = not qs.filter(created_on__date__gt=threshold).exists()
             else:
-                send_reminder = project.created_on.date() == threshold
+                if project.created_on is None:
+                    send_reminder = False
+                else:
+                    send_reminder = project.created_on.date() == threshold
```
elcct, line 458: Maybe it's better to have `send_reminder = project.created_on and project.created_on.date() == threshold` that would get read of if/else and should be more readable.

datahub/reminder/tasks.py
```diff
@@ -943,8 +1065,209 @@ def _get_active_projects(adviser):
...
+            )
+            & (
+                Q(one_list_tier_id=OneListTierID.tier_d_international_trade_advisers.value)
+                | Q(one_list_tier_id=OneListTierID.tier_d_overseas_post_accounts.value)
+            )
+        ),
     )
+
+
+class UserMigrationTasks:
```
elcct, line 1090: Maybe rather than creating a class, the methods could be moved to their own file like `migration_tasks.py`.
I see this object is just a collection of rather unrelated functions, so perhaps this could be changed to have functions on their own (dropping the class) - otherwise it may be confusing.

datahub/reminder/tasks.py
elcct: There seem to be a few more instances of this, so this could be applied throughout.

.gitignore
```diff
@@ -91,3 +91,4 @@ tags
 
 # IDE
 .vscode/
+.idea/
```
elcct, line 94: Not sure if it's a good .idea to add developer specific ignores there - but we already have .vscode in so, maybe that's fine. 
It's good to use global gitignore on one's machine, so that there is no need to change it for each individual project.

datahub/reminder/tasks.py
chopkinsmade: Yeh this looks neater

richmadetech (approved):

PaulWheatcroft (approved):

backslashbaker (approved):

PR #4469, opened 2023-02-03
fix: archive companies with duns number as long as they have no active subsidiaries

JosefSmith:
### Description of change

As per https://data-services-help.trade.gov.uk/data-hub/updates/announcements/data-hub-data-retention-approach/, all companies (so including those that have a duns number and so part of a "tree" of companies) should be archived as long as they don't have any active subsidiaries or subsidiaries that match the other conditions for archiving.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/tasks/company.py
+++ b/datahub/company/tasks/company.py
@@ -27,7 +27,7 @@ def _automatic_company_archive(limit, simulate):
         company=OuterRef('pk'),
     ).order_by('-date')
 
-    companies_to_be_archived = Company.objects.annotate(
+    candidate_companies_to_be_archived = list(Company.objects.annotate(
         latest_interaction_date=Subquery(
             latest_interaction.values('date')[:1],
         ),
@@ -38,13 +38,21 @@ def _automatic_company_archive(limit, simulate):
     ).filter(
         Q(latest_interaction_date__date__lt=_5y_ago) | Q(latest_interaction_date__isnull=True),
         archived=False,
-        duns_number__isnull=True,
         orders__isnull=True,
         investor_profiles__isnull=True,
         active_investment_projects__isnull=True,
         created_on__lt=_3m_ago,
         modified_on__lt=_3m_ago,
-    )[:limit]
+    ))
+
+    companies_to_be_archived = [
+        company
+        for company in candidate_companies_to_be_archived
+        if not any(
+            not (descendant.archived or descendant in candidate_companies_to_be_archived)
+            for descendant in company.descendants
+        )
+    ][:limit]
 
     for company in companies_to_be_archived:
         message = f'Automatically archived company: {company.id}'
@@ -63,7 +71,7 @@ def _automatic_company_archive(limit, simulate):
         )
         logger.info(message)
 
-    return companies_to_be_archived.count()
+    return len(companies_to_be_archived)
 
 
 def schedule_automatic_company_archive(limit=1000, simulate=True):
--- a/datahub/company/tasks/company.py
+++ b/datahub/company/tasks/company.py
@@ -49,8 +49,8 @@ def _automatic_company_archive(limit, simulate):
         company
         for company in candidate_companies_to_be_archived
         if not any(
-            not (descendant.archived or descendant in candidate_companies_to_be_archived)
-            for descendant in company.descendants
+            not (related_company.archived or related_company in candidate_companies_to_be_archived)
+            for related_company in company.related_companies
         )
     ][:limit]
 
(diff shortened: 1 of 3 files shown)


PippoRaimondiDIT (approved):

datahub/company/test/tasks/test_company.py
```diff
@@ -369,3 +369,55 @@ def test_investment_projects(
...
+        """
+        # Construct a tree of companies, but where some companies have an active
+        # descendant company, so we can assert they don't get archived.
+
+        gt_3m_ago = timezone.now() - relativedelta(months=3, days=1)
+        with freeze_time(gt_3m_ago):
+            company_root = CompanyFactory()
+            company_sub_1a = CompanyFactory(global_headquarters=company_root)
+            company_sub_1b = CompanyFactory(global_headquarters=company_root)
+            company_sub_2 = CompanyFactory(global_headquarters=company_sub_1a)
```
elcct, line 389: I think we don't have a case where subsidiary can be a global headquarters at the same time.
See: https://github.com/uktrade/data-hub-api/blob/main/datahub/company/serializers.py#L391

datahub/company/tasks/company.py
```diff
@@ -26,7 +26,7 @@ def _automatic_company_archive(limit, simulate):
         company=OuterRef('pk'),
     ).order_by('-date')
 
-    companies_to_be_archived = Company.objects.annotate(
+    candidate_companies_to_be_archived = list(Company.objects.annotate(
```
elcct, line 29: This may cause memory issues as it is loading all records matching the query.

datahub/company/test/tasks/test_company.py
michalc: Although it looks like this is enforced by data hub, it's not enforced in general in D&B data? For example, running:
```sql
SELECT
  c1.duns_number AS dn1,
  c2.duns_number AS dn2,
  c3.duns_number AS dn3,
  c4.duns_number AS dn4
FROM
  dun_and_bradstreet.global_companies c1
INNER JOIN
  dun_and_bradstreet.global_companies c2 ON c2.headquarter_or_parent_duns_number = c1.duns_number AND c2.duns_number != c1.duns_number
INNER JOIN
  dun_and_bradstreet.global_companies c3 ON c3.headquarter_or_parent_duns_number = c2.duns_number AND c3.duns_number != c2.duns_number
INNER JOIN
  dun_and_bradstreet.global_companies c4 ON c4.headquarter_or_parent_duns_number = c3.duns_number AND c4.duns_number != c3.duns_number
LIMIT
  5
```
in Data Workspace returns some rows, so I think it's possible in general? 

Although - am I misunderstanding what `global_headquarters` means in Data Hub terms? I _thought_ it just means "parent company" and is the same thing as `headquarter_or_parent_duns_number`. But does it more mean "the company right at the root of the tree of companies"?

datahub/company/tasks/company.py
michalc: This would at the absolute most be in the tens of thousands (and more usually < 100?) so I suspect would be fine? But - can change to a different pattern if you think we should?

datahub/company/test/tasks/test_company.py
michalc: Actually, stronger than that. Running this is Data Workspace on Data Hub's data:
```sql
SELECT
  c1.duns_number AS dn1,
  c2.duns_number AS dn2,
  c3.duns_number AS dn3
FROM
  dit.data_hub__companies c1
INNER JOIN
  dit.data_hub__companies c2 ON c2.global_headquarters_id = c1.id
INNER JOIN
  dit.data_hub__companies c3 ON c3.global_headquarters_id = c2.id
AND
  c1.duns_number IS NOT NULL AND
  c2.duns_number IS NOT NULL AND
  c3.duns_number IS NOT NULL
LIMIT
  5
```
Returns some rows. So it looks like even right now, Data Hub can have chains of of companies based on global_headquarters_id that are longer than just 2 - i.e. a subsidiary can have subsidiary?

elcct: > But does it more mean "the company right at the root of the tree of companies"?

Correct. The subsidiaries should connect directly to `global_headquarters` rather than through other subsidiaries. Maybe this changed and I am not aware of it.

> So it looks like even right now, Data Hub can have chains of of companies based on global_headquarters_id that are longer than just 2 - i.e. a subsidiary can have a subsidiary, so when archiving, to make sure we don't archive too many companies, we have to consider this case I think?

It may be good to figure out how this data entered Data Hub in such a way. Maybe it's better to correct it to ensure subsidiaries can't have subsidiaries and are attached to one common global headquarters, rather than archiving in incorrect state?

Or the rules around company hierarchy in Data Hub should be changed to more closely match D&B?

datahub/company/tasks/company.py
elcct: If the number of records will be small then that should be fine, but something to keep an eye on.

elcct (approved):

PR #4545, opened 2023-03-23
Feat/add 50 new countries

JosefSmith:
### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/metadata/migrations/0046_update_countries.yaml
```diff
@@ -0,0 +1,350 @@
...
+  fields:
+    disabled_on: null
+    name: Ascension
+    overseas_region: 8d4c4f31-06ce-4320-8e2f-1c13559e125f
+    iso_alpha2_code: SH-AC
+- model: metadata.country
+  pk: dc859cb8-23c4-4352-89cf-257fe9baaf34
+  fields:
+    disabled_on: null
+    name: Baker Island
```
cgsunkel, line 54: Do we really need to add [uninhabited islands](https://en.wikipedia.org/wiki/Baker_Island) to Data Hub? [There](https://en.wikipedia.org/wiki/Howland_Island) [are](https://en.wikipedia.org/wiki/Jarvis_Island) [more](https://en.wikipedia.org/wiki/Kingman_Reef) [of](https://en.wikipedia.org/wiki/Navassa_Island) [these](https://en.wikipedia.org/wiki/Ducie_Island) in the list of countries you want to add.

What is the value we would gain from adding these? Their inclusion in a reference data set does not really justify their inclusion here and it will just add further noise to an already large list.

michalc: @cgsunkel We've now had a brief exchange the Robert on Teams, @-mentioned you. I think the plan is to march forward and go for consistency. Apparently there is a type-ahead in most places, so negative UX impact should be minimal?

PippoRaimondiDIT (approved):

PR #4705, opened 2023-06-23
feature/CLS2-199-send-batch-requests-to-opensearch

chopkinsmade:
Because of the way the `get_search_by_entities_query` functions creates an opensearch query, which is to convert every duns number into a separate match filter, we need to batch the opensearch queries so only 1024 are sent at a time.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/dnb_api/utils.py
+++ b/datahub/dnb_api/utils.py
@@ -717,47 +717,28 @@ def append_datahub_details(family_tree_members: list):
                 break  # Stop once we've found the match
 
 
-def _batch_list(list, number_items):
-    """
-    Create a list of lists, with the maximum number of items in each list set to the number
-    provided in number_items
-
-    Args:
-        list (_type_): The list to create a batch of lists from
-        number_items (_type_): The maximum number of items
-
-    Returns:
-        A list of lists, with each inner list containing at most the number_items. The final inner
-        list may contain less then the number_items
-    """
-    list = iter(list)
-    return iter(lambda: tuple(islice(list, number_items)), ())
-
-
 def _load_datahub_details(family_tree_members_duns):
     """
     Load any known datahub details for the duns numbers provided
     """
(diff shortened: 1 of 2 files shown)


datahub/dnb_api/utils.py
```diff
@@ -715,27 +714,53 @@ def append_datahub_details(family_tree_members: list):
...
+                'uk_region',
+                'address',
+                'registered_address',
+                'sector',
+                'latest_interaction_date',
+                'archived',
+                'one_list_tier',
+                'number_of_employees',
+            ),
+        )[0:MAX_DUNS_NUMBERS_PER_REQUEST]
```
elcct, line 762: Is this necessary, since `batch_of_duns_numbers` should be within the limit already?

datahub/dnb_api/utils.py
```diff
@@ -715,27 +714,53 @@ def append_datahub_details(family_tree_members: list):
...
+        opensearch_results = execute_search_query(query)
+        results.extend(opensearch_results.hits)
+
+    return [x.to_dict() for x in results]
```
elcct, line 766: Rather than using `x`, it would be clearer to use `result` or `item`.

datahub/dnb_api/utils.py
chopkinsmade: The slice here is how the pagination is controlled with the `opensearch_dsl` library. I expected the pagination to be a prop on the search object, but this is how they've implemented it. The default return for opensearch is 10 so this is needed in addition to the batching, otherwise we'd send a maximum of 1024 duns numbers but only select 10 of them back.

datahub/dnb_api/utils.py
chopkinsmade: Good spot, updated

PR #4820, opened 2023-08-24
Cps 471 address county removal Colombia

bau123:
### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/metadata/migrations/0058_update_company_address.py
+++ b/datahub/metadata/migrations/0058_update_company_address.py
@@ -1,0 +1,41 @@
+from pathlib import PurePath
+
+import mptt
+from django.db import migrations
+
+from datahub.core.migration_utils import load_yaml_data_in_migration
+
+
+def load_services(apps, _):
+    load_yaml_data_in_migration(
+        apps,
+        PurePath(__file__).parent / '0057_update_services.yaml'
+    )
+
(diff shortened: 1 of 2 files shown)


datahub/metadata/migrations/0058_update_company_address.py
```diff
@@ -0,0 +1,41 @@
...
+from datahub.core.migration_utils import load_yaml_data_in_migration
+
+
+def load_services(apps, _):
+    load_yaml_data_in_migration(
+        apps,
+        PurePath(__file__).parent / '0057_update_services.yaml'
+    )
+
+def delete_address_colombia(apps, scheme_editor):
```
PippoRaimondiDIT, line 15: did you check in the DB there are no other columbia addresses referenced other than those 300 companies?

PippoRaimondiDIT: Not sure why we are deleting the address from metadata as well? was that the ask?

bau123: 0058_update_company_address.py is a old file that i made and i wasn't sure if this was the right way to do things instead of having it as a db command. This file could be ignored if it isn't the right way of doing things. update_company_address_county.py is another file i wrote which is should take a csv file but not sure if it's correct

PR #4985, opened 2023-11-06
bugfix/CLS2-556-fix-v3-field-errors

chopkinsmade:
### Description of change

Hide archived from the v3 opensearch queries, but available in the v4 search. A future ticket is going to add archived as a filterable column, however as the v3 search endpoints use a fuzzy search with a MultiMatch query, having archived in the list of queryable columns causes a text to bool parsing issue

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/search/query_builder.py
+++ b/datahub/search/query_builder.py
@@ -17,8 +17,12 @@ from datahub.search.apps import EXCLUDE_ALL, get_global_search_apps_as_mapping
 
 MAX_RESULTS = 10000
 V3_FIELD_EXCLUSIONS = ('archived',)
-# the V3 search endpoints use a MultiMatch query that cause issues with columns that do not have
-# an issue in the V4 endpoint.
+# The V3 search endpoint use a MultiMatch query, that combines every column defined in the
+# SEARCH_FIELDS property in each sub app of the main search app into a single query. When
+# adding a column to the list of fields that represent a boolean column in the search index,
+# for example 'archived', the MultiMatch query fails with a parsing error when a text value
+# is provided to the search endpoint. The error is due to opensearch attempting to parse a
+# text value into a boolean value of 'true' or 'false'
 
 
 class MatchNone(Query):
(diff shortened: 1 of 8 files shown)


marijnkampf:
Do you know what the roadmap is for deprecating the V3 search end points in Data Hub? Would there be a potential (quick) fix in the MultiMatch end point instead? If we are going with the current exclusion method, is there a test that fails when a field is not added to the exclusion fields but that should have?

chopkinsmade:
> Do you know what the roadmap is for deprecating the V3 search end points in Data Hub? Would there be a potential (quick) fix in the MultiMatch end point instead? If we are going with the current exclusion method, is there a test that fails when a field is not added to the exclusion fields but that should have?

1. Not sure on the deprecation of the v3 endpoints, I can see quite a few usages of these endpoints still in the front end
2. The MultiMatch isn't an endpoint it's a type of opensearch query that lets you query across multiple indexes which is how the V3 endpoints look to work. In the V4 search each type has it's own searchable index. I can't see a way of changing the error thrown by MultiMatch, as it's a valid error. We are grouping together every searchable column, ignoring the column data type, so when the MultiMatch query runs it fails with parsing text into a boolean for the archived column
3. Yep, without the exclusions being added, all the tests in this file fail datahub/search/company/test/test_basic_search_views.py, plus tests in a few other files. This test is used by the v4 endpoint, and has the archived present datahub/search/test/test_query_builder.py::test_get_search_by_entities_query

datahub/search/query_builder.py
```diff
@@ -16,6 +16,9 @@
 from datahub.search.apps import EXCLUDE_ALL, get_global_search_apps_as_mapping
 
 MAX_RESULTS = 10000
+V3_FIELD_EXCLUSIONS = ('archived',)
+# the V3 search endpoints use a MultiMatch query that cause issues with columns that do not have
```
elcct, line 20: Would be good to write a bit more about the issue.

elcct (approved):

PaulWheatcroft (approved):

PR #5062, opened 2023-12-07
Feature/cls2 587 task filter created by

marijnkampf:
### Description of change

On the Task search allow users to filter on `Not created by` in addition to the existing `created_by` filter:

> Given I have selected the ‘Created by' filter
> When the I select the checkbox labelled ‘Others’
> Then I should see the tasks created by others

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/search/task/views.py
+++ b/datahub/search/task/views.py
@@ -16,7 +16,7 @@ class SearchTaskAPIViewMixin:
     search_app = TaskSearchApp
     serializer_class = SearchTaskQuerySerializer
     es_sort_by_remappings = {}
-    fields_to_exclude = ()
+    fields_to_exclude = ('not_created_by',)
 
     FILTER_FIELDS = (
         'archived',
@@ -24,7 +24,6 @@ class SearchTaskAPIViewMixin:
         'title',
         'due_date',
         'created_by',
-        'not_created_by',
         'advisers',
         'company',
         'investment_project',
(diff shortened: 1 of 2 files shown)


samuele-mattiuzzo (approved):

datahub/search/task/views.py
```diff
@@ -34,9 +36,59 @@ class SearchTaskAPIViewMixin:
...
+    def get_base_query(self, request, validated_data):
+        """
+        Check for filters that filter for not_* parameters and extend the open search base query
+        with the must_not filters.
+        """
+        must_not = []
+        base_query = super().get_base_query(request, validated_data)
+
+        raw_query = base_query.to_dict()
+        filters = self.deep_get(raw_query, 'query|bool|filter')
```
chopkinsmade, line 63: The `filters` variable is only used if there is at least 1 not query. To avoid needing to do a deep get on every request, can the order of this function be adjusted so if the `request.data.get('not_created_by')` condition is false we return the base_query immediately?

datahub/search/task/views.py
```diff
@@ -34,9 +36,59 @@ class SearchTaskAPIViewMixin:
...
+        if not filters:
+            return base_query
+
+        if request.data.get('not_created_by'):
```
chopkinsmade, line 67: We will need to duplicate this for every not filter, we could look at changing this to have a variable that contains the not name and the corresponding match filter value.

chopkinsmade (approved):

PR #5064, opened 2023-12-07
 feature/rr-1257-function-to-get-all-fields-for-client-receipt-email

ince-dbt:
### Checklist

* [✅] Has this branch been rebased on top of the current `main` branch?

  <details>
  
  </details>

* [✅] Is the CircleCI build passing?

### General points

<details>

</details>


datahub/export_win/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+
+from datahub.export_win.models import CustomerResponse, CustomerResponseToken
+
+
+logger = getLogger(__name__)
+
+
+def get_all_fields_for_client_email_receipt(token, customer_response):
+    details = {
+        'customeremail': None,
```
chopkinsmade, line 13: Should these 2 values be `customer_email` and `customer_destination` for consistent naming? Or are these external fields that are named this way?

datahub/export_win/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+        'countrydestination': None,
+        'client_firstname': None,
+        'lead_officer_name': None,
+        'goods_services': None,
+        'url': None,
+    }
+    customer_response = CustomerResponse.objects.get(id=customer_response.id)
```
chopkinsmade, line 20: Are the CustomerResponse and CustomerResponseToken objects guaranteed to always have a matching `token.id` and `customer_response.id`? If not then could end up with some errors here without checking on the objects existing

chopkinsmade (approved):

datahub/export_win/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+    customer_response_token = CustomerResponseToken.objects.get(id=token.id)
+    win = customer_response.win
+    win_customer = customer_response_token.company_contact
+    details['customeremail'] = win_customer.email
```
elcct, line 24: You could do: ```
details = {
    'customer_email': win_customer.email,
    ... and so on
}
```
and then you don't have to initialise the dictionary as at the line 12.

datahub/export_win/tasks.py
ince-dbt: Thanks.. you are right.. we should keep consistent naming

datahub/export_win/tasks.py
ince-dbt: Pawel is right.. i dont think i need to fetch them again because it has been passed from up stream! So, I have amended the code

datahub/export_win/tasks.py
ince-dbt: Thanks! I have done as per suggestion

dredmonds (approved):

PR #5080, opened 2023-12-13
Update email notification URLs to include UTMs

samuele-mattiuzzo:
### Description of change

This task adds tracked URLs to the Task email reminders. The UTM_CAMPAIGN is generated by inspecting the class name and converting it to snake case

Jira ticket [here](https://uktrade.atlassian.net/browse/CLS2-645)

### Checklist

### DIFF
--- a/datahub/task/emails.py
+++ b/datahub/task/emails.py
@@ -5,6 +5,8 @@ from datahub.task.models import Task
 
 
 class EmailTemplate(ABC):
+    UTM_URL_BASE = "?utm_source=individual&utm_medium=email_notify&utm_campaign={}&utm_content=task"
+
     @abstractmethod
     def __init__(self, task: Task):
         self.task = task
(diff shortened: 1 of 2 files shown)


datahub/task/emails.py
```diff
@@ -47,6 +50,16 @@ def task_due_date(self):
             f'Date due: {self.task.due_date.strftime("%-d %B %Y")}' if self.task.due_date else None
         )
 
+    def _to_snake_case(self, str) -> str:
+        return ''.join(['_' + i.lower() if i.isupper()
+                       else i for i in str]).lstrip('_')
+
+    def get_utm_url(self) -> str:
```
chopkinsmade, line 57: Using an autogenerated utm based on the classname minus the word EmailTemplate directly links our classes to the GA params, and looking in the ticket our classnames don't match what has been configured. For example, `upcoming_task` is being sent from this code but it should be `task_due_date_approaching` in the email according to the ticket. There's also differences in the `task_assigned_to_others` and `task_amended_by_others`.

Each email has it's own implementation class based on EmailTemplate, to avoid linking our code with GA I think it would make more sense to define these utm values as separate props

chopkinsmade (approved):

PaulWheatcroft (approved):

PR #5240, opened 2024-03-05
Display win admin details groups labels

dredmonds:
### Description of change
Display win details including groups and labels from Django admin.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/export_win/models.py
```diff
@@ -342,6 +342,8 @@ class Win(BaseModel):
         blank=True,
         on_delete=models.PROTECT,
     )
+    updated = models.DateTimeField(auto_now_add=True, null=True, blank=True)
```
elcct, line 345: The model has `modified_on` and `modified_by` fields - inherited from `BaseModel`, so this would be redundant.

dredmonds: Done, please refer into my latest commit.

ince-dbt (approved):

datahub/export_win/migrations/0029_win_is_deleted_win_updated.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2 on 2024-03-05 13:56
```
elcct, line 1: It's better if you delete those two migrations 0029 and 0030 and then create them again. Otherwise we are adding the `updated` field and then removing it. If you create the migration again, that field will no longer be in the migration.

PR #5269, opened 2024-03-19
Admin setup for export win admin users

ince-dbt:
### Description of change

This PR relates to RR-1303, RR-1304, and RR-1319 which covers the work django admin dashboard for:

- deleting win (soft delete)
- reinstating win
- saving win form
- set ExportWinAdmin group user to view deleted win

ExportWinAdmin(2)

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/export_win/admin.py
```diff
@@ -1,17 +1,132 @@
...
+
+class WinAdminForm(ModelForm):
+    """Win admin form."""
+
+    class Meta:
+        model = Win
+        fields = '__all__'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
p3dr0migue1, line 99: You could have also done something like this:

```python
form_fields = [
    'cdms_reference',
    'customer_email_address',
    'customer_job_title',
    'line_manager_name',
    'lead_officer_email_address',
    'other_official_email_address'
]
if self.instance and self.instance.pk:
  for field in form_fields:
      if field in self.fields:
          self.fields[field].required = False
```

If you needed to add another field in the future, you would do it to the list. If you were to add another if statement it would be another test you would have to make.

datahub/export_win/admin.py
```diff
@@ -23,15 +138,126 @@ class WinAdmin(BaseModelAdminMixin, VersionAdmin):
...
+            'date',
+            'description',
+            'name_of_customer',
+            'goods_vs_services',
+            'name_of_export',
+            'sector',
+            'hvc',
+        )}),
+        ('Customer details', {'fields': (
+            'cdms_reference',  # Legacy field
```
p3dr0migue1, line 168: Does this mean these fields will be removed in the future?

p3dr0migue1:
The build is also failing atm, so we need to address these before approving the code

datahub/export_win/admin.py
ince-dbt: No, it will still be there but I believe from my understanding data wont be saved to this in the future as these fields are just like a place to store the data transferred from legacy win for those win transfered from legacy.

ince-dbt:
> The build is also failing atm, so we need to address these before approving the code

datahub/export_win/admin.py
ince-dbt: Thaks for the helpful suggestion. So, i think it is done as a list on lines 168 where you commented below and it needs to be arranged under DBT officer and Customer details.

datahub/export_win/test/factories.py
```diff
@@ -239,3 +241,36 @@ class LegacyExportWinsToDataHubCompanyFactory(factory.django.DjangoModelFactory)
 
     class Meta:
         model = 'export_win.LegacyExportWinsToDataHubCompany'
+
+
+class DeletedWinFactory(factory.django.DjangoModelFactory):
```
elcct, line 246: I don't think this is needed as `DeletedWin` is not a real model, just a proxy for `Win` so there can be a separate entry for deleted wins in Admin.
This also doesn't create a deleted win.
If you need a separate factory for deleted wins, you could inherit WinFactory and make is_deleted = True like this:
```
class DeletedWinFactory(WinFactory):
    is_deleted = True
```

datahub/export_win/test/test_admin.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.django_db
+def test_reinstate():
+    user = AdviserFactory()
+
+    deleted_win1 = DeletedWinFactory()
+    deleted_win2 = DeletedWinFactory()
+    deleted_win1.is_deleted = True
+    deleted_win2.is_deleted = True
```
elcct, line 46: Currently the `DeletedWinFactory` does not actually create a deleted win and later the `.is_deleted = True` is not actually saved to the database, so the win is not really soft deleted.

If you modify the `DeletedWinFactory` as suggested above, then instead of `deleted_win1.is_deleted = True` you could make an assertion that the created win is actually soft deleted like: 
```
    assert deleted_win1.is_deleted is True
    assert deleted_win2.is_deleted is True
```

datahub/export_win/test/test_admin.py
```diff
@@ -0,0 +1,60 @@
...
+
+    request_factory = RequestFactory()
+    request = request_factory.get('/')
+    request.user = user
+
+    admin = DeletedWinAdmin(model=DeletedWin, admin_site=None)
+
+    admin.undelete(request, queryset=DeletedWin.objects.all())
```
elcct, line 54: The `DeletedWin.objects.all()` wouldn't be returning soft deleted wins, because it is using default query from `BaseExportWinSoftDeleteManager`. Instead you could use `DeletedWin.objects.reinstate()` so that it will feed the soft deleted wins.

datahub/export_win/models.py
```diff
@@ -23,6 +23,24 @@
...
+    """Base class for Export win soft delete manager."""
+
+    def get_queryset(self):
+        return (
+            super()
+            .get_queryset()
+            .filter(is_deleted=False)
+        )
+
+    def reinstate(self, *args, **kwargs):
```
elcct, line 36: I would suggest renaming this method to `soft_deleted`, as `reinstate` can imply some action is being done to the objects, but that is not the case here.

The `Win.objects.soft_deleted()` would be clearer in my opinion.

datahub/export_win/test/test_admin.py
```diff
@@ -0,0 +1,60 @@
...
+    request.user = user
+    actions = admin.get_actions(request)
+    assert 'delete_selected' not in actions
+
+
+@pytest.mark.django_db
+def test_soft_delete():
+    """Return adviser as user with email."""
+    user = AdviserFactory()
+    win1 = DeletedWinFactory()
```
elcct, line 26: Here we should use `WinFactory` as we are trying to soft delete an active win.
Would be also good to assert that win is indeed not soft deleted yet by doing:
```
assert win1.is_deleted is False
```

datahub/export_win/test/test_admin.py
```diff
@@ -0,0 +1,60 @@
...
+    admin = WinAdmin(Win, admin_site)
+    request = RequestFactory().get('/')
+    request.user = user
+    actions = admin.get_actions(request)
+    assert 'delete_selected' not in actions
+
+
+@pytest.mark.django_db
+def test_soft_delete():
+    """Return adviser as user with email."""
```
elcct, line 24: Would be good to ensure the comments are describing the function correctly.

datahub/export_win/test/factories.py
ince-dbt: Thanks @elcct ! I didnt really want to create DeletedWinFactory and replicates all fields in Win in that sort of stupid way but few tests under orphaned test failed if I did not include it. However, I have now managed to fix and remove the DeletedWinFactory....

datahub/export_win/test/test_admin.py
ince-dbt: I made adjustment as per suggestion, but the assertion should be deleted_win1.is_deleted is False here because I am writing a test to undelete win or reinstating win. Hence, the result of is_deleted should be false to verify that the DeletedWin is successfully reinstated. Hope it makes sense, if not please read my new code. Happy to have further chat on this if it still does not make sense

datahub/export_win/test/test_admin.py
ince-dbt: Amended this as well!

datahub/export_win/models.py
ince-dbt: Renamed this as per suggestion

datahub/export_win/test/test_admin.py
ince-dbt: Done this

datahub/export_win/test/test_admin.py
ince-dbt: Done

paulgain (approved):

datahub/export_win/admin.py
```diff
@@ -1,17 +1,132 @@
...
+                'cdms_reference',
+                'customer_email_address',
+                'customer_job_title',
+                'line_manager_name',
+                'lead_officer_email_address',
+                'other_official_email_address',
+            ]
+            for field_name in fields_to_update:
+                field = self.fields.get(field_name)
+                if field:
```
elcct, line 111: I think this code has no effect - if the field doesn't exist, the `fields.get(field_name)` will throw `KeyError` exception before reaching this line. We don't need to do that check, so we can simply do this:
```
            for field_name in fields_to_update:
                self.fields[field_name].required = False
```

ince-dbt: As explianed previously, I am as much as keen as you to remove unnecessary extra line in my code but iIt gives sever error occured...

elcct (approved):

PR #5464, opened 2024-06-04
Add request retry for legacy export win migration.

elcct:
### Description of change

During migration someones the request times out or we get 500. This adds a retry functionality in case the request fails for any reason.

This also improves breakdowns and win advisers migration - rather than deleting all records associated with a win and then recreating them, it will update existing records or create missing ones.

I have not used built-in retry facility of requests, because that would require some refactoring of APIClient and I think that would not make sense for a one off like this.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/export_win/test/test_export_wins_api.py
+++ b/datahub/export_win/test/test_export_wins_api.py
@@ -1,13 +1,17 @@
+import time
+
 import pytest
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.test.utils import override_settings
+
 from requests.exceptions import (
     ConnectionError,
     ConnectTimeout,
     ReadTimeout,
     Timeout,
 )
+
 from rest_framework import status
 
 from datahub.company.export_wins_api import (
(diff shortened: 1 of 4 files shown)


datahub/export_win/test/test_export_wins_api.py
```diff
@@ -122,3 +135,85 @@ def test_get_call_invoked(
...
+        assert len(response) == 1
+        assert response[0] == {'test': 'response'}
+
+    def test_get_call_aborts(
+        self,
+        requests_mock,
+        mock_sleep,
+    ):
+        """
+        Check GET call will be retried
```
oliverjwroberts, line 181: Is this meant to be "Check GET call will be **aborted**"?

oliverjwroberts (approved):
PR adds retry functionality to GET requests as well as updates existing/creates missing records for breakdowns and win advisers, instead of deleting and recreating them. 

I agree with the custom retry function. I think for a one off solution, this is a sensible path of least resistance. Happy to approve!

datahub/export_win/test/test_export_wins_api.py
elcct: Nice! Just updated. Thank you.

dredmonds (approved):

DeanElliott96 (approved):

PR #5482, opened 2024-06-13
Rename capital only FDI type

oliverjwroberts:
### Description of change

To improve data quality, we have enforced a validation rule that states _Capital Only_ FDI investment projects must have an investment value >= £15 million. However, this validation only occurs when editing a project's value (which is done via a separate form that appears after the project has been created). 

So that users are aware of this rule earlier on in the journey, the team want to add the following text to the _Capital Only_ option within the FDI type dropdown:

image

After some deliberation, there seemed to be two approaches:
1. Rename the FDI type instance so the additional text appears across both the API and Frontend
2. Modify the data that is returned by the API and fed into the `FieldFDIType` component to include the additional text

Given the inconsistencies that may appear if going with option 2, it was decided to rename the instance across the system (option 1). 

>However, I am open to discussing this further.

In this PR, a migration has been created to change the name in environments where the instance already exists. Additionally, the fixture data has been updated so the name change appears in new/temporary environments where the instance will not exist during initialisation.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/metadata/migrations/0083_rename_capital_only_fdi_type.py
+++ b/datahub/metadata/migrations/0083_rename_capital_only_fdi_type.py
@@ -1,0 +1,35 @@
+from django.db import migrations
+
+
+CAPITAL_ONLY_FDI_TYPE_PK = '840f62c1-bbcb-44e4-b6d4-a258d2ffa07d'
+
+
+def change_fdi_type_name(apps, schema_editor):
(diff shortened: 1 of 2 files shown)


oliverjwroberts:
After further discussion, it was decided to go with option 2, instead. Closing this PR as it is has been superseded by https://github.com/uktrade/data-hub-frontend/pull/6927.

PR #5540, opened 2024-07-18
Fix storing email status.

elcct:
### Description of change

On rare occasion, when the `update_customer_response_for_lead_officer_notification_id` took longer time to execute it would overwrite the customer response object with its original state, essentially removing answers given by the customer.
This fix makes it so that the database changes are committed before the task executes, so that it will have fully updated object and also it only updates fields that are changed, so that even if between getting the object and updating it, there were changes to other fields, it wouldn't overwrite them.
I also updated other email status tasks, so they only update the changed fields.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/export_win/serializers.py
+++ b/datahub/export_win/serializers.py
@@ -169,6 +169,7 @@ class WinSerializer(ModelSerializer):
     complete = BooleanField(read_only=True)
     first_sent = DateTimeField(read_only=True)
     last_sent = DateTimeField(read_only=True)
+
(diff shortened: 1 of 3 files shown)


oliverjwroberts (approved):
PR ensures updates to the customer response are saved before triggering the `update_customer_response_for_lead_officer_notification_id` task. It also only saves the necessary fields, rather than the entire object.

Looks good to me!

paulgain (approved):

PR #5579, opened 2024-08-06
Only add administrative areas for country and area matches

DeanElliott96:
### Description of change

Some `AdministrativeArea`s do not have an abbreviated name associated with them, so we must match by country and name. For example, Tamil Nadu in India does not have an abbreviated name associated for us however from DNB it has an abbreviated name of "TN" which would have been matched to "Tennesse" in data-hub. My previously update would have not matched an `AdministrativeArea` here as it has no abbreviated name even though we have Tamil Nadu in our database. This update now checks by country and name instead of abbreviated name and name.

There are 3106 companies with the wrong `AdministrativeArea` associated with them, I'm currently writing a script to update them all.

### Checklist

### DIFF
--- a/datahub/dnb_api/utils.py
+++ b/datahub/dnb_api/utils.py
@@ -178,14 +178,15 @@ def extract_address_from_dnb_company(dnb_company, prefix, ignore_when_missing=()
         if dnb_company.get(f'{prefix}_country')
         else None
     )
-    area = (
-        AdministrativeArea.objects.filter(
-            area_code=dnb_company[f'{prefix}_area_abbrev_name'],
-            name=dnb_company[f'{prefix}_area_name'],
-        ).first()
-        if dnb_company.get(f'{prefix}_area_abbrev_name')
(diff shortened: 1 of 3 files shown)


p3dr0migue1 (approved):

datahub/dnb_api/utils.py
```diff
@@ -178,14 +178,15 @@ def extract_address_from_dnb_company(dnb_company, prefix, ignore_when_missing=()
...
-    )
+
+    area = None
+    if country and dnb_company.get(f'{prefix}_area_name'):
+        area = (
+            AdministrativeArea.objects.filter(
+                name=dnb_company[f'{prefix}_area_name'],
+                country_id=country.id,
+            ).first()
+        )
```
DeanElliott96, line 189: Not all `AdministrativeArea`s have area_codes, so filter by country and name instead.

cgsunkel (approved):

PR #5649, opened 2024-09-19
Feature/EYB Lead company matching

marijnkampf:
### Description of change

Feature branch of the following three PRs, which have been individually reviewed:

- Link EYB Lead to existing company https://github.com/uktrade/data-hub-api/pull/5647 or  
- Add new company from EYB Lead https://github.com/uktrade/data-hub-api/pull/5640
- Add contact record for EYB leads if there is no matching contact for the related company. https://github.com/uktrade/data-hub-api/pull/5643 

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/investment_lead/services.py
```diff
@@ -0,0 +1,123 @@
...
+        found (boolean): true/false based on success of search
+        company (object): a Company object or None
+    """
+    companies = Company.objects.filter(duns_number=duns_number)
+
+    if companies.count() == 1:
+        return True, companies[0]
+
+    # no match found
+    return False, None
```
DeanElliott96, line 34: Just checking that you want to ignore companies which share the same duns number (not sure if companies with same duns is even possible)? As the count check will only return a company if there is only a single match.

You could possibly do the following and remove the `found` logic. As you are checking for `found` in the other function, you could check for the company queryset instead. They have also already been reviewed so its up to you if you want to make the changes or not:
```
def match_by_duns_number(duns_number):
    companies = Company.objects.filter(duns_number=duns_number)

    if companies:
        return companies[0]

    return None
```

```
def process_eyb_lead(eyb_lead):
    if eyb_lead.duns_number is not None:
       company = match_by_duns_number(eyb_lead.duns_number)

        if company:
            eyb_lead.company = company
            eyb_lead.save()
            return company

    return add_new_company_from_eyb_lead(eyb_lead)
```

DeanElliott96 (approved):
Looks good, I like the small little functions doing one thing! 👍 

Small suggestion but since its already reviewed its up to you if you want to make it. Just a note as well on whether we want to ignore multiple companies with the same duns_number (if that's even possible).

datahub/investment_lead/services.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    if eyb_lead.duns_number is not None:
+        found, company = match_by_duns_number(eyb_lead.duns_number)
+
+        if found:
+            eyb_lead.company = company
+            eyb_lead.save()
+            return company
+
+    return add_new_company_from_eyb_lead(eyb_lead)
```
samuele-mattiuzzo, line 51: For any further reviewer: this logic is also present in the function below, `add_new_company_from_eyb_lead` and I was thinking about refactoring it before moving ahead with this PR.

I'll look at it tomorrow in the morning 👍🏻

datahub/investment_lead/services.py
samuele-mattiuzzo: @DeanElliott96 your edge case is something me and @marijnkampf discussed already. It won't happen as in DataHub `duns_number` has to be unique so it's either `found and one` or `none`.

```
def match_by_duns_number(duns_number):
    companies = Company.objects.filter(duns_number=duns_number)

    if companies:
        return companies[0]

    return None
```

while ☝🏻 looks a bit more elegant and clean it also means, unfortunately, that any integrity issue that somehow made it through our database would yield inconsistent results. We'd rather ignore results of that kind (companies sharing the same duns_number) than assign the first one, as it's better to consider them as invalid results, rather than pick the first one of the lot.

datahub/investment_lead/services.py
samuele-mattiuzzo: I've decided to refactor in a separate PR so to not hold up deployment.

samuele-mattiuzzo (approved):
🚀

datahub/investment_lead/services.py
marijnkampf: FYI the above is enforced on database level (duns_number field is unique) so the above isn't required.

PR #5665, opened 2024-09-26
Remove slow exclude on interactions and check in memory instead

DeanElliott96:
### Description of change

Do the exclude in memory instead of with the DB as the previous query was causing the CPU to max out.

### Checklist

### DIFF
--- a/datahub/company_activity/tasks.py
+++ b/datahub/company_activity/tasks.py
@@ -45,9 +45,7 @@ def relate_company_activity_to_interactions(batch_size=500):
         interaction__isnull=False,
     ).values_list('interaction_id', flat=True)
 
-    interactions = Interaction.objects.exclude(
-        id__in=activity_interactions,
-    ).filter(
+    interactions = Interaction.objects.filter(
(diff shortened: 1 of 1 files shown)


datahub/company_activity/tasks.py
```diff
@@ -45,9 +45,7 @@ def relate_company_activity_to_interactions(batch_size=500):
         interaction__isnull=False,
     ).values_list('interaction_id', flat=True)
 
-    interactions = Interaction.objects.exclude(
-        id__in=activity_interactions,
-    ).filter(
+    interactions = Interaction.objects.filter(
```
baarkerlounger, line 48: I think a set lookup would probably be faster for 700k records but let's see

baarkerlounger (approved):

datahub/company_activity/tasks.py
DeanElliott96: Thank you, good idea, added!

cgsunkel (approved):

PR #5689, opened 2024-10-09
Great URL can be null

baarkerlounger:
### Description of change

Fixes: `npsycopg2.errors.NotNullViolation: null value in column \"url\" of relation \"company_activity_great\" violates not-null constraint`

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/company_activity/models/great.py
```diff
@@ -27,7 +27,7 @@ class Great(models.Model):
     attributed_to_type = models.CharField(max_length=MAX_LENGTH)
     attributed_to_id = models.CharField(max_length=MAX_LENGTH)
 
-    url = models.CharField(max_length=MAX_LENGTH)
+    url = models.CharField(max_length=MAX_LENGTH, null=True)
```
DeanElliott96, line 30: Is there a reason to use `null=True` instead of going with the code-review-doctor suggestions? I don't think we have anywhere else which uses `null=True` on a `CharField`.
Django has some docs on it here as well:
https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null

> Avoid using [null](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null) on string-based fields such as [CharField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField) and [TextField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.TextField). If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a [CharField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField) has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

DeanElliott96 (changes requested):

datahub/company_activity/models/great.py
baarkerlounger: Mostly just that it more accurately reflects what we get in the dataset (there is an actual null value, rather than an empty string or missing key), but can default that to an empty string if you prefer.

DeanElliott96: Yeah that's a pretty valid reason! If we can though, I would follow the practices from Django as we are deciding how we want to store the data we get from Great.

DeanElliott96 (approved):

oliverjwroberts (approved):

PR #5696, opened 2024-10-11
Handle Great data fields longer than 255 chars

baarkerlounger:
### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


cgsunkel (approved):

datahub/company_activity/tasks/ingest_great_data.py
```diff
@@ -99,15 +99,15 @@ def json_to_model(self, jsn):
...
-            'data_company_name': data.get('company_name', ''),
-            'data_company_size': data.get('company_size', ''),
-            'data_phone_number': data.get('phone_number', ''),
-            'data_email_address': data.get('email_address', ''),
+            'data_full_name': data.get('full_name', '')[:255],
+            'data_website_url': data.get('website_url', '')[:255],
+            'data_company_name': data.get('company_name', '')[:255],
+            'data_company_size': data.get('company_size', '')[:255],
+            'data_phone_number': data.get('phone_number', '')[:255],
+            'data_email_address': data.get('email_address', '')[:255],
```
samuele-mattiuzzo, line 107: I have a question about this:

in the long-text below it's stated that `... we store as TextField if we need the full value, truncate to 255 if not ...`

In here, I think `website_url`, `email_address` and `phone_number` will need to be the full value to be useful, no? Why are they truncated and not converted as TextField? I see a `url` attribute being converted to TextField above, is it the same as `website_url`?

samuele-mattiuzzo (approved):
Approved regardless of my comment, as it shouldn't block the PR (I assume you've discussed this / know the answer, as opposed to this being a thing that was missed)

datahub/company_activity/tasks/ingest_great_data.py
baarkerlounger: The reason we're truncating those is that from the data we've got so far it looks like all valid answers to those should be within 255 chars i.e. where we see "phone numbers" over 255 chars it's because someone has put a comment in the phone number field or spam or similar, it's not likely to be a legitimate useful value. Ideally those sorts of values would be validated better upstream but in the meantime truncating seemed more reasonable here than changing the data type, and easier to spot and manually remediate than to drop the value entirely.

samuele-mattiuzzo: Yeah that's what I was wondering... A 255+ long phone or email address is most likely going to be unusable data anyway... a website url as well, I suppose.
I'll resolve

PR #5710, opened 2024-10-17
[2 of 2] Define and schedule job to ingest EYB triage and user data

oliverjwroberts:
### Description of change

This PR is 2 of 2 that ingests EYB data into Data Hub ([see 1 of 2 here](https://github.com/uktrade/data-hub-api/pull/5684)), and is going into a feature branch. The overall plan is to ingest EYB leads, stored as JSON in Data Hub's S3 bucket, into Data Hub via the company activity jobs/ingestion infrastructure.

This PR specifically defines and schedules the tasks to ingest the EYB triage and user data from the bucket.

Notably, rather than simultaneously schedule a task for the triage and user data respectively, we have asked the triage task to trigger the user task upon completion. This helps avoid duplicate instances of the same lead being created.

### Checklist

* [ ] Has this branch been rebased on top of the current `main` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [ ] Is the CircleCI build passing?

### General points

<details>
<summary><strong>Other things to check</strong></summary><p></p>

* Make sure `fixtures/test_data.yaml` is maintained when updating models
* Consider the admin site when making changes to models
* Use select-/prefetch-related field lists in views and search apps, and update them when fields are added
* Make sure the README is updated e.g. when adding new environment variables

</details>

See [docs/CONTRIBUTING.md](https://github.com/uktrade/data-hub-api/blob/main/docs/CONTRIBUTING.md) for more guidelines.

PR #5684, opened 2024-10-07
[1 of 2] Refactor EYB lead schema

oliverjwroberts:
### Description of change

This PR is 1 of 2 that works toward ingesting EYB data into Data Hub, and is going into a feature branch. The overall plan is to ingest EYB leads, stored as JSON in Data Hub's S3 bucket, into Data Hub via the company activity jobs/ingestion infrastructure.

Notable changes include:
- Removal of the EYB lead create view/endpoint as we are no longer going to ingest leads this way.
- Sector data will be received as four fields (`sectorID`, `sector`, `sectorSub`, `sectorSubSub`), but stored as a `sector` field that contains the related instance and `sector_segments` which contains the individual segment names; we cannot use `sectorID` to lookup a sector as this is the `SLXXX` ID from the reference Data Workspace dataset, which we do not store. Instead, we have to use the segment names.
- Additional company address fields will now be received, including `companyLocation` which contains a country's ISO Alpha 2 code (and is subsequently stored as `address_country` containing the specified country instance
- We also will get an optional `dunsNumber` to help us identify the specific company, when provided.

I have opted to use Serialisers for each _group_ of fields (triage and user) to provide validation of the incoming JSON and translate the data to Python types/an EYBLead instance (as a lot of the functionality comes straight out of the box). This is instead of going for a dictionary of mappings approach. 

At the moment, these Serialiser's aren't being used, but that will come in part 2 of the work where we shall define the job to actually ingest the JSON.

### camelCase fields in the Serialiser
As it currently stands, I have opted to use `camelCase` fields within the `CreateEYBLead` serialisers because the incoming fields will be camelCase. I think this makes the code more readable, but appreciate it goes against Python convention (and flake8 was not happy either). 

As a feasibility test, I have successfully removed the need for `camelCase` serialiser fields by adding a _camel-to-snake_ function in the `to_internal_value` method, among other changes:

```python
    def to_internal_value(self, data):
        """Convert unvalidated JSON data to validated data."""
        snake_case_data = {}   <------
        for key, value in data.items():    <------
            mapped_key = inflection.underscore(key)    <------
            snake_case_data[mapped_key] = value    <------
        internal_value = super().to_internal_value(snake_case_data)
        related_fields = self.get_related_fields_internal_value(data)
        internal_value.update(related_fields)
        return internal_value
```

[...]

PR #5719, opened 2024-10-22
Ingest EYB triage and user data

oliverjwroberts:
### Description of change

This PR adds functionality to ingest EYB triage and user records from Data Hub's S3 bucket.

Summary of changes:
- Removal of the `EYBLead` create view/endpoint (no longer needed)
- Various schema changes to the `EYBLead` model
- Created individual serialisers for the triage and user data
- Defined tasks to search for new EYB files and ingest their contents
- Scheduled the ingestion of user records after triage records to avoid duplicate `EYBLead` instances being created

Most of the changes in this feature branch have been approved in two sub-PRs:
- #5684 
- #5710 

> [!WARNING]
> Please note, a few small tweaks have been added since #5710 and are not pre-approved. These are the three most recent commits:
> - [ba9cbdfda](https://github.com/uktrade/data-hub-api/pull/5719/commits/ba9cbdfda149f49b6ec9019f0588fc2471b8e02a)
> - [5e5455d3f](https://github.com/uktrade/data-hub-api/pull/5719/commits/5e5455d3fc81e56cbf8bc8256f7beb16b83139da)
> - [9347643c2](https://github.com/uktrade/data-hub-api/pull/5719/commits/9347643c2c7bcaaf15999132a8420c3c5c574346)

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/investment_lead/serializers.py
```diff
@@ -86,137 +77,367 @@
...
 
     def validate_location(self, value):
-        if not UKRegion.objects.filter(name=value.title()).exists():
-            raise serializers.ValidationError(f'Location "{value}" does not exist.')
+        if isinstance(value, str):
+            transformed_value = value.replace('_', ' ').title()
+            if transformed_value == '':
+                return None
+            # Handle edge case where value is missing the word `The`
+            if transformed_value == 'Yorkshire And Humber':
```
baarkerlounger, line 217: I wonder if this is the only possible edge case of if it's worth pulling out into a method that checks a mapping dict.

samuele-mattiuzzo (approved):
Looks good, just `CodeCov` complaining about a couple of lines in the serializer file, worth testing those too

datahub/investment_lead/serializers.py
oliverjwroberts: That probably would be a better approach. Would you be happy for me to address this in the subsequent refactoring work? I'm conscious that we may find other snags/edge case that also need handling and can bundle them together; that way I can also do a more focused PR.

oliverjwroberts:
In `investment_lead/serializers.py`, CodeCov is flagging a few lines, but I think it's okay to leave them because:
- Lines 115 and 126 return an empty dict for the base class methods, which are intended to be overridden, but left in so if it's not overridden (by future code that inherits these classes), it doesn't fail. The classes that override these methods are tested completely.
- Lines 217 and 218 refer to the above regarding one edge case location value. On the EYB side, all new records are using a drop down with a fixed value. As a result, I'm still unsure if it's worth handling this edge case, rather than let it be skipped/not ingested.

baarkerlounger:
@oliverjwroberts I think it's still reasonable to:

a) Test what the base class does if used directly (how else do I know that removing that `{}` and returning `None` breaks things?
b) Test what inheriting code that doesn't override it would/should do - should it raise an error? Fail gracefully etc?

PR #5721, opened 2024-10-23
Bugfix/tet 606 fix company merge

DeanElliott96:
### Description of change

Updates the company merge tool to include the new models related to a `Company`.

There is a new tests which checks for any model related to a `Company`. If it is not part of the company merge, then the test fail.

For data you do not want to merge with the company merge you can:

- Add it into the `ALLOWED_RELATIONS_FOR_MERGING` but **NOT** the `MERGE_CONFIGURATION`.

For data you do want to merge with the company merge you you can:

- Add it into the `ALLOWED_RELATIONS_FOR_MERGING` **AND** the `MERGE_CONFIGURATION`.

### Checklist

### DIFF
--- a/datahub/company/merge_company.py
+++ b/datahub/company/merge_company.py
@@ -77,14 +77,6 @@ ALLOWED_RELATIONS_FOR_MERGING = {
     # the front end if required)
     CompanyExportCountry.company.field,
     CompanyExportCountryHistory.company.field,
-
-    # Not added to MERGE_CONFIGURATION as don't want to overwrite the existing companies global
-    # headquarters.
-    Company.global_headquarters.field,
-
-    # Not added to MERGE_CONFIGURATION as filled in as part of the company merge process.
-    Company.transferred_to.field,
-    Company.transferred_from.field,
 }
 
 
@@ -136,6 +128,9 @@ def merge_companies(source_company: Company, target_company: Company, user):
     """
     Merges the source company into the target company.
     MergeNotAllowedError will be raised if the merge is not allowed.
+
+    Companies with relations to another Company are not allowed to be merged. They would need to
+    be manually updated in the Admin before merging.
     """
     is_source_valid, invalid_obj = is_model_a_valid_merge_source(
         source_company,
--- a/datahub/company/merge_company.py
+++ b/datahub/company/merge_company.py
@@ -3,24 +3,35 @@ import logging
 import reversion
 
 from datahub.company.merge import (
-    _default_object_updater,
     is_model_a_valid_merge_source,
     is_model_a_valid_merge_target,
     MergeConfiguration,
     MergeNotAllowedError,
     update_objects,
 )
+from datahub.company.merge_utils.merge_relations import (
+    company_list_item_updater,
+    large_capital_opportunity_updater,
+    one_list_core_team_member_updater,
+    pipeline_item_updater,
+)
 from datahub.company.models import (
     Company,
+    CompanyExport,
     CompanyExportCountry,
     CompanyExportCountryHistory,
     Contact,
+    Objective,
+    OneListCoreTeamMember,
 )
 from datahub.company_activity.models import CompanyActivity
 from datahub.company_referral.models import CompanyReferral
 from datahub.dnb_api.utils import _get_rollback_version
 from datahub.interaction.models import Interaction
+from datahub.investment.investor_profile.models import LargeCapitalInvestorProfile
+from datahub.investment.opportunity.models import LargeCapitalOpportunity
 from datahub.investment.project.models import InvestmentProject
+from datahub.investment_lead.models import EYBLead
 from datahub.omis.order.models import Order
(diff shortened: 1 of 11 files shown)


datahub/company/test/test_merge_company.py
```diff
@@ -332,6 +302,7 @@ def test_merge_interactions_contacts_succeeds(
             result = merge_companies(source_company, target_company, user)
 
         assert result == {
+            **self.base_expected_result,
```
baarkerlounger, line 305: Why `self` here but not above?

datahub/company/test/test_merge_company.py
```diff
@@ -117,150 +134,103 @@ def company_with_investment_projects_factory():
 class TestDuplicateCompanyMerger:
     """Tests DuplicateCompanyMerger."""
 
+    # Default expect values for company changes to 0, changed per test.
+    # Usually need to add a new company relation in here for the tests, manually updating
```
baarkerlounger, line 138: I think this comment could be a bit clearer if it's needed?

datahub/company/test/test_merge_company_relations.py
```diff
@@ -0,0 +1,57 @@
...
+        Models for company merge also require the reversion decorator to undo the changes
+        if the merge fails:
+        ```
+            from datahub.core import reversion
+
+            @reversion.register_base_model()
+            class ModelRelatedToACompany(models.Model):
+        ```
+
+        To get the full error message, you may be better to run this test with
```
baarkerlounger, line 46: `To get the full error message run this test with the -s flag` although I'm not sure we need to be documenting pytest flags here

datahub/company/test/test_merge_company.py
DeanElliott96: `self` here as its inside the test function but above its in the pytest parametrize decorator and not needed/would error if you tried to use `self`.

datahub/company/models/company.py
```diff
@@ -680,6 +680,7 @@ def delete_export_country(self, country_id, adviser):
             export_country.delete()
 
 
+@reversion.register_base_model()
```
baarkerlounger, line 683: What's the intention behind this?

datahub/company/test/test_merge_company.py
DeanElliott96: What changes would you recommend to make it clearer?

datahub/company/test/test_merge_company_relations.py
DeanElliott96: True, just trying to make it clear what to do when another dev adds a company relation and wonders why this test is failing and doesn't see the full error.

I can remove it though.

datahub/company/test/test_merge_company.py
baarkerlounger: I'm not totally sure this is what it's getting at but maybe:

` New Company relations should be added here. Overriding Company relation values should be done in specific tests below`.?

baarkerlounger: Also is this `dict` derivable from the Company model so it doesn't need updating when new relations are added?

datahub/company/models/company.py
DeanElliott96: I've added it for the new relations added into the company merge as its required by the existing company merge code.

`reversion` allows you to rollback changes made to data in a model. I'm assuming its to allow you to revert the data changes back if the company merge tool fails.

baarkerlounger: I'm not totally understanding why the new relations require the model to be revertable but the old ones didn't though? What's special about the new relations?

datahub/company/test/test_merge_company.py
DeanElliott96: I wont be able to get it from `dict` as some fields are excluded from merging. But nice idea, I may be able to get it from `MERGE_CONFIGURATION` as that has all the relations used in the merge and their company field name.

datahub/company/models/company.py
DeanElliott96: All the old/existing relations in the company merge require and have the `reversion`. The new ones also have it now as well, its not something I've added only for the new ones.

datahub/company/test/test_merge_company.py
DeanElliott96: Now using the `MERGE_CONFIGURATION`

datahub/company/merge_company.py
```diff
@@ -66,40 +94,43 @@
...
-
-def _pipeline_item_updater(pipeline_item, field, target_company, source_company):
-    # If there is already a pipeline item for the adviser for the target company
-    # delete this item instead as the same company can't be added for the same adviser again
-    if PipelineItem.objects.filter(adviser=pipeline_item.adviser, company=target_company).exists():
-        pipeline_item.delete()
-    else:
-        _default_object_updater(pipeline_item, field, target_company, source_company)
-
-
```
DeanElliott96, line 86: Moved to own file.

baarkerlounger (approved):

PR #5723, opened 2024-10-24
feature/ingest-contact-consent-data

chopkinsmade:
### Description of change
Add new scheduled job to ingest consent data

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/company/tasks/contact.py
```diff
@@ -129,3 +143,120 @@ def automatic_contact_archive(limit=1000, simulate=False):
...
+            path,
+            'r',
+            transport_params={
+                'client': client,
+            },
+        ) as s3_file:
+            i = 0
+            for line in s3_file:
+                i = i + 1
+                if (i) % (10 ** min((max((math.floor(math.log10(i))), 2)), 6)) == 0:
```
baarkerlounger, line 210: What are we doing here?

datahub/company/tasks/contact.py
```diff
@@ -129,3 +143,120 @@ def automatic_contact_archive(limit=1000, simulate=False):
...
+                    logger.info('Processed %s rows from %s', i, path)
+                consent_row = json.loads(line)
+                if 'email' not in consent_row or 'consents' not in consent_row:
+                    continue
+
+                email = consent_row['email']
+                matching_contact = Contact.objects.filter(email=email).first()
```
baarkerlounger, line 217: If there's multiple contacts matching that email address (for example a duplicate record that hasn't been merged yet, or multiple people using a shared inbox), should this not update all found records rather than just the first?

.localstack/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM localstack/localstack:latest
```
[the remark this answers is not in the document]
chopkinsmade: My understanding of Moto is it is used for unit testing as it mocks the calls to AWS. Localstack is a full AWS emulator that runs locally, and can respond in the same way as AWS to SDK functions called from boto3. In this PR, I've been developing by uploading files to a localstack S3 bucket, that my code is then accessing in the same way it will when deployed to AWS.

datahub/company/tasks/contact.py
chopkinsmade: Good point about the possibility to duplicates. The S3 file will always contain a single email address, but I'll handle the scenario where there could be multiple in data hub

datahub/company/tasks/contact.py
chopkinsmade: It's writing to the logs in a way that handles both small and large files. It logs every 100 rows initially, going up to logging every 10,000 rows next then every 100,000 rows. I'll move it into an appropriately named helper function

.localstack/Dockerfile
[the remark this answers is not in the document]
baarkerlounger: I'm not sure there's much functional difference - in both cases your relying on 3rd party software to mock/emulate S3 correctly, in both cases you get an API that matches what you're really calling, and matches what's returned?

Only the Moto approach seems much more lightweight to me.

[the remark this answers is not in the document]
chopkinsmade: Localstack is running as part of the project docker-compose, so I can upload a file to a localstack S3 bucket and have my code running in the rq_long docker container read that file as if it's coming from S3. I'm not using it as part of any unit tests, I'm using it test my code is running as expected in the RQ worker. 
I see them as fulfilling 2 separate requirements - moto for unit tests to mock external dependancies, and localstack for acting as an AWS environment emulator so I can avoid reading / writing to a real S3 environment when developing locally.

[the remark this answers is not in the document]
baarkerlounger: Ah I see. I've been doing all my local dev via the unit tests so haven't really needed that but fair enough then.

datahub/company/tasks/contact.py
```diff
@@ -129,3 +143,134 @@ def automatic_contact_archive(limit=1000, simulate=False):
...
+                consent_row = json.loads(line)
+                if 'email' not in consent_row or 'consents' not in consent_row:
+                    continue
+
+                email = consent_row['email']
+                matching_contacts = Contact.objects.filter(email=email)
+                for matching_contact in matching_contacts:
+
+                    if not matching_contact:
+                        logger.debug('Email %s has no matching datahub contact', email)
```
baarkerlounger, line 231: Is there a test for this log message? I don't _think_ that condition will ever be met in the loop?

baarkerlounger (approved):

elcct (approved):

PR #5724, opened 2024-10-24
Fix EYB lead admin page crashing

oliverjwroberts:
### Description of change

In production, when you view an EYB lead in the admin, the server times out and crashes. This is because the EYB lead admin page uses the default fields for `EYBLead.company` and `EYBLead.investment_projects`, which subsequently incurs large overhead of having to select all the related instances to display in the drop-down. Spoiler alert: this is a lot in production.

This PR changes these fields to `raw_id_fields` which avoids this - see the Django docs [here](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields).

> For those curious, the default fields cause Django's backend to perform a `DECLARE "" NO SCROLL CURSOR WITH HOLD FOR SELECT` query on the related instance tables. Even for just 20,000 thousand related instances, this query was taking 30x longer than the rest.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/investment_lead/admin.py
+++ b/datahub/investment_lead/admin.py
@@ -4,6 +4,11 @@ from datahub.investment_lead.models import EYBLead
 
 
 class EYBLeadAdmin(admin.ModelAdmin):
+    search_fields = [
+        'triage_hashed_uuid',
+        'user_hashed_uuid',
+        'company_name',
+    ]
     list_display = [
         'id',
         'created_on',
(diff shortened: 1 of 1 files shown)


samuele-mattiuzzo (approved):

baarkerlounger (approved):
Would be interested to see some before and after comparison - maybe from Sentry's performance tab but sounds reasonable to me.

PR #5749, opened 2024-10-31
Handle OneToOneFields in opensearch

DeanElliott96:
### Description of change

For `OneToOneField`s in Django, getting the related object, e.g. `interaction.activity` returns the actual object itself and not a queryset. This causes an error with the Opensearch sync where it expects to use `values_list` and also expects to `filter` which works for `ForeignKey` fields as they return a queryset.

### Checklist

### DIFF
--- a/datahub/search/tasks.py
+++ b/datahub/search/tasks.py
@@ -1,6 +1,8 @@
 from logging import getLogger
 
 from django.apps import apps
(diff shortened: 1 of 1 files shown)


DeanElliott96:
Closed as decided to revert back to ForeignKey from OneToOne.

PR #5752, opened 2024-11-01
Check for already ingested Great files by prefix

baarkerlounger:
### Description of change

Since EYB now also use the IngestedFiles model we need to filter by prefix otherwise we'll pull the wrong latest ingest data.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company_activity/tasks/ingest_great_data.py
+++ b/datahub/company_activity/tasks/ingest_great_data.py
@@ -3,18 +3,15 @@ import logging
 
 from datetime import datetime
 
-import environ
-
 from smart_open import open
 
 from datahub.company.models.company import Company
 from datahub.company.models.contact import Contact
 from datahub.company_activity.models import GreatExportEnquiry, IngestedFile
+from datahub.company_activity.tasks.constants import GREAT_PREFIX
 from datahub.metadata.models import BusinessType, Country, EmployeeRange, Sector
 
 logger = logging.getLogger(__name__)
-env = environ.Env()
-REGION = env('AWS_DEFAULT_REGION', default='eu-west-2')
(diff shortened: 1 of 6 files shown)


cgsunkel (approved):

datahub/company_activity/tasks/ingest_great_data.py
```diff
@@ -46,7 +43,9 @@ def ingest(self, bucket, file):
 
     def _get_last_ingestion_datetime(self):
         try:
-            return IngestedFile.objects.latest('created_on').created_on.timestamp()
+            return IngestedFile.objects.filter(
+                filepath__icontains=GREAT_PREFIX,
+            ).latest('created_on').created_on.timestamp()
         except IngestedFile.DoesNotExist:
             return None
```
DeanElliott96, line 50: Should the `IngestedFile` have a new `source` field now its being used by other feeds (EYB, and more in future)?

Might possibly be flaky later where we are filtering on a filename or bucket path where we could instead filter on its source.

If we are adding any admin views for this and want to split it out into separate sources or ever filter on different sources, it may be easier and more performant to filter on a `source` field rather than a icontains lookup.

I might be underestimating the work involved in changing this now though.

samuele-mattiuzzo (approved):

DeanElliott96 (commented):
Just a comment to see if we should add a new source field onto the `IngestedFile` and  use that.

datahub/company_activity/tasks/ingest_great_data.py
baarkerlounger: Happy to consider that tech debt and add it when we next add a source.

PR #5790, opened 2024-11-14
feature/use-opensearch-for-consent-job-searching-instead-of-DB

chopkinsmade:
### Description of change
Using the django DB for checking if a contact in the S3 file existed was running slowly on production (200-300 milliseconds per query), with a file that contains 1.3 million rows.

To reduce impact on the DB, instead use the opensearch server to query if a contact has an email matching the row in the S3 file. This approach for checking contacts is much faster (single digit milliseconds), and a query is only made to the DB once validation checks have been ran on the last ingested date.

This has been tested on a local environment with 350,000 contacts, and took 30-35 minutes

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company/tasks/contact.py
+++ b/datahub/company/tasks/contact.py
@@ -23,7 +23,9 @@ from datahub.core.queues.errors import RetryError
 from datahub.core.queues.job_scheduler import job_scheduler
 from datahub.core.queues.scheduler import LONG_RUNNING_QUEUE
 from datahub.core.realtime_messaging import send_realtime_message
-
+from datahub.search.contact.models import Contact as SearchContact
+from datahub.search.execute_query import execute_search_query
+from datahub.search.query_builder import get_search_by_entities_query
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


datahub/company/tasks/contact.py
```diff
@@ -242,54 +279,51 @@ def sync_file_with_database(self, client, file_key):
...
-                        matching_contact.save()
 
-                        logger.info('Updated contact consent data for email %s', email)
-                    else:
+                    if not settings.ENABLE_CONTACT_CONSENT_INGEST:
                         logger.info(
                             'Email %s would have consent data updated, but setting is disabled',
                             email,
                         )
+                        continue
```
oliverjwroberts, line 308: Would it be more efficient to have this check before the for loop and exit before any processing has started if this setting is disabled? Or is it helpful to know individual emails that have qualified, but won't be ingested as a result?

datahub/company/test/tasks/test_contact_task.py
```diff
@@ -683,44 +722,218 @@ def test_sync_file_with_multiple_contacts_matching_email_does_update_contact(sel
...
+    ):
+        """
+        Test when a row has an email that matches a contact, but the contact has missing date,
+        returns True
+        """
+        task = ContactConsentIngestionTask()
+        contact = {
+            'id': uuid.uuid4(),
+            'email': 'a',
+            'consent_data_last_modified': '2023-07-20T12:00:00',
```
oliverjwroberts, line 859: Should the contact and row dates be swapped here. In this test, I think you wanted the contact to have the missing date?

datahub/company/test/tasks/test_contact_task.py
```diff
@@ -683,44 +722,218 @@ def test_sync_file_with_multiple_contacts_matching_email_does_update_contact(sel
...
+    ):
+        """
+        Test when a row has an email that matches a contact, but the file has missing date,
+        returns True
+        """
+        task = ContactConsentIngestionTask()
+        contact = {
+            'id': uuid.uuid4(),
+            'email': 'a',
+            'consent_data_last_modified': None,
```
oliverjwroberts, line 836: Should the contact and row dates be swapped here. In this test, I think you wanted the row to have the missing date?

oliverjwroberts (commented):
@chopkinsmade really nice approach using open-search instead of the ORM to lookup matching contacts; that time saving per query is huge. I know we've had some discussion already, but to reiterate, I think you've reasonably tested what you can (e.g. magnitude of data being processed) and covered most failure scenarios (e.g. contact data being out of sync and the opensearch index being wiped).

Just a couple of questions from me before I give it a green tick, but otherwise looks good. Nice work!

datahub/company/tasks/contact.py
chopkinsmade: Exactly the 2nd point, I wanted this to get as far as possible into the ingestion process without actually making a modification to the DB

datahub/company/test/tasks/test_contact_task.py
chopkinsmade: Good spot, I'll switch the test data in both of the tests to match the name of the test

datahub/company/tasks/contact.py
oliverjwroberts: I agree with trying to get as far as possible/filter out as many invalid updates before making a database hit. 

However, I was just thinking if this setting is disabled, you're going to end up with potentially hundreds of thousands of log messages saying essentially the same thing. Would it be preferable to have the check before the loop? The check itself doesn't use the database and irrespective of whether an email qualifies, nothing will be ingested if the setting is disabled.

PR #5791, opened 2024-11-14
Modify recorded ingestion datetime

oliverjwroberts:
### Description of change

During a recent incident where a stuck job blocked the long queue from processing any jobs for a few days, we found some EYB leads had not been ingested once the backlog cleared. The reason for this was the latest files contained "already ingested" records because their `modified` date was older than the last ingestion datetime, which was a couple of days later due to being stuck in the queue. 

This PR ensures the ingested datetime is set as the last modified datetime of the file being ingested (i.e. when it was created in the S3 bucket, not when it was actually ingested by the job). This will ensure delayed jobs will still ingest new records.

The `IngestedFile` model has also been added to the admin for increased visibility.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/company_activity/models/ingested_file.py
```diff
@@ -21,4 +22,4 @@ class IngestedFile(models.Model):
             'The S3 object path including prefix of the ingested file'
         ),
     )
-    created_on = models.DateTimeField(auto_now_add=True)
+    created_on = models.DateTimeField(default=timezone.now)
```
baarkerlounger, line 25: I wonder if this should be an additional field rather. You might still want to know when the `IngestedFile` record was created.

baarkerlounger: It also means you don't need to change the Great task at all

oliverjwroberts: Yes, that's a good shout. Will add now!

oliverjwroberts: Wouldn't it be more consistent for the Great task to use the same metric? Or is it less applicable to the Great data?

baarkerlounger: Great(ExportEnquiries) don't actually compare datetimes at all anymore since we assume our data can't be modified once submitted, we only check if the records (foreign) id exists in our database already. If not we create that record, if yes we skip it.

datahub/company_activity/tests/factories.py
```diff
@@ -116,31 +116,11 @@ class CompanyActivityIngestedFileFactory(factory.django.DjangoModelFactory):
     """
 
     filepath = 'data-flow/exports/ExportGreatContactFormData/20240920T000000.jsonl.gz'
-    created_on = now()
+    file_created = now()
 
     class Meta:
         model = 'company_activity.IngestedFile'
 
-    @classmethod
```
baarkerlounger, line 124: Do we not still want this now that it's a valid field again?

datahub/company_activity/models/ingested_file.py
```diff
@@ -2,6 +2,7 @@
 
 from django.conf import settings
 from django.db import models
+from django.utils import timezone
```
baarkerlounger, line 5: I wonder whether this whole model should move somewhere more generic now it's used across apps

datahub/company_activity/tests/factories.py
oliverjwroberts: It is no longer needed now we have the `file_created` field, which is overridable. I can see it potentially still being useful when trying to manipulate an instances _created_ date, but currently, no tests do this. If it's needed in the future, I suppose it can be re-added?

datahub/company_activity/models/ingested_file.py
oliverjwroberts: It probably should. We have an imminent PR that will add a task to ingest the marketing component of the EYB lead data. Once this is in, I will start the refactor. That way I can ensure I'm still covering all bases.

baarkerlounger (approved):

marijnkampf (approved):

PR #5807, opened 2024-11-19
Add one-off command to modify `referred` projects

oliverjwroberts:
### Description of change

This PR adds a one-off management command to modify `referred` investment projects, of which there are only a few hundred.

Projects that currently have `status == 'referred'` should, instead, have the [new](https://github.com/uktrade/data-hub-api/pull/5741) `Referred to EYB` specific programme added to their list of specific programmes and their status set as `'ongoing'`.

This command/code will be removed after the command has been run in production.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/investment/project/management/commands/modify_referred_projects.py
```diff
@@ -0,0 +1,53 @@
...
+                logger.info(f'Found {referred_projects.count()} referred projects. Modifying...')
+                # many-to-many field needs to be updated individually
+                for project in referred_projects:
+                    project.specific_programmes.add(referred_to_eyb_programme)
+                # bulk update status field
+                referred_projects.update(status=ONGOING_STATUS_VALUE)
+                logger.info('Finished modifying referred projects.')
+            else:
+                logger.warning('No referred projects found. Exiting...')
+        except Exception as e:
```
ince-dbt, line 52: Fantastic job with the error logging in the handle method! Just a little suggestion to make the exception handling even more robust and prevent unintentionally hiding unexpected issues. You might consider catching more specific exceptions instead of a general Exception. For example, handling SpecificProgramme.DoesNotExist separately could provide clearer error messages. Maybe something like this: 
```
except SpecificProgramme.DoesNotExist as e:
    logger.error(f'Referred to EYB specific programme not found: {e}')
```
This way, we can give more precise feedback when something goes wrong. Cheers! 😊

ince-dbt (approved):

datahub/investment/project/management/commands/modify_referred_projects.py
oliverjwroberts: Cheers for the feedback @ince-dbt. My thinking behind the generic Exception was that it allows me to catch a wide variety of issues that I'm not expecting. This avoids additional lines to catch and log several scenarios that I can think up, and shouldn't happen anyway, I felt this was an okay compromise; especially, as the plan is to remove the code/command after its been executed once. I hope that's acceptable?

ince-dbt: Understood, sounds good to me. Great work! 😊

oliverjwroberts:
> @oliverjwroberts is there another place (more visible than code) where we can track these two todos so that we don't forget to remove?

I'll add a comment to CLS2-1019 stating it cannot be moved to done until the command has been run. I'll also add a comment/link to CLS2-1020 so that this command gets removed as part of that work.

PR #5815, opened 2024-11-21
Tell browsers to use https instead of http for 1 hour

DeanElliott96:
### Description of change

This is a Django settings which instructs browsers to visit the site with HTTPS instead of HTTP. You can read me more about it in the [docs](https://docs.djangoproject.com/en/5.1/ref/middleware/#http-strict-transport-security).

As recommended by Django, I've set it to only do this for 1 hour at a time to ensure it doesn't break anything. Once its tested on UAT and been on PROD this can be updated to 1 year.

Later the preload option will also be added once it seems to be running fine on prod for a week or two.

### Checklist

### DIFF
--- a/config/settings/common.py
+++ b/config/settings/common.py
@@ -747,3 +747,7 @@ CSP_REPORT_ONLY = False
 
 S3_LOCAL_ENDPOINT_URL = env("S3_LOCAL_ENDPOINT_URL", default='')
 ENABLE_CONTACT_CONSENT_INGEST = env("ENABLE_CONTACT_CONSENT_INGEST", default=False)
+
+# Initially set to 1 hour, to increase once confirmed it's successful.
+SECURE_HSTS_SECONDS = 3600
+SECURE_HSTS_INCLUDE_SUBDOMAINS = True


config/settings/common.py
```diff
@@ -747,3 +747,7 @@ def _add_hawk_credentials(id_env_name, key_env_name, scopes):
 
 S3_LOCAL_ENDPOINT_URL = env("S3_LOCAL_ENDPOINT_URL", default='')
 ENABLE_CONTACT_CONSENT_INGEST = env("ENABLE_CONTACT_CONSENT_INGEST", default=False)
+
+# Initially set to 1 hour, to increase once confirmed it's successful.
+SECURE_HSTS_SECONDS = 3600
```
baarkerlounger, line 752: Why is this time limited at all?

DeanElliott96: The standard is to set this to 1year which tells the browser to allow a user to "try" a http request again after 1 year (which it wouldn't allow and issue another year).

Django recommends settings it to 1 hour and ensure it doesn't cause any issues first before setting it to the full year. If it blocks http for 1 year but someone expects to use the API via http, it would not allow that client access to the site for 1 year (without clearing their cookies I believe). At worst this would block them for 1 hour.

baarkerlounger (approved):

cgsunkel (approved):

PR #5817, opened 2024-11-21
Filter EYB leads by both `EYBLead.company_name` and `EYBLead.company.name`

oliverjwroberts:
### Description of change

When filtering EYB leads by company name, users were reporting unreliable results and that they were missing leads. 

This PR ensures the company name filter searches both the `EYBLead.company.name` and `EYBLead.company_name` fields. The latter contains the user entered name, the former is the name of the linked company record (matched on either duns number of created with the user-entered data).

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/investment_lead/test/test_views.py
+++ b/datahub/investment_lead/test/test_views.py
@@ -86,7 +86,18 @@ class TestEYBLeadListAPI(APITestMixin):
         assert len(response.data['results']) == pagination_limit
 
     def test_filter_by_company_name(self, test_user_with_view_permissions):
-        """Test filtering EYB leads by company name"""
+        """Test filtering EYB leads by the EYBLead.company_name field."""
+        company_name = 'Mars Exports Ltd'
+        EYBLeadFactory(company_name=company_name)
+        EYBLeadFactory()
+        api_client = self.create_api_client(user=test_user_with_view_permissions)
+        response = api_client.get(EYB_LEAD_COLLECTION_URL, data={'company': company_name})
+        assert response.status_code == status.HTTP_200_OK
+        assert response.data['count'] == 1
+        assert response.data['results'][0]['company_name'] == company_name
+
+    def test_filter_by_related_company_name(self, test_user_with_view_permissions):
+        """Test filtering EYB leads by the EYBLead.company.name field."""
         company_name = 'Mars Exports Ltd'
         company = CompanyFactory(name=company_name)
         EYBLeadFactory(company=company)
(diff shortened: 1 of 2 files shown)


samuele-mattiuzzo (approved):

baarkerlounger (approved):
In what scenarios would these be different?

samuele-mattiuzzo:
> In what scenarios would these be different?

`company.name` relies on the company being present in DataHub or being created brand new via the DNB service. Being that `company_name` is user inputted, then the two might not be the same, so we can use this safety net 👍🏻

baarkerlounger:
@samuele-mattiuzzo my assumption was that if a match is found the name would likely match (because it's been matched on user input name directly, or because the upstream data source has looked up the DNB/Company registration number based on user input), and if a match isn't found a company would be created with that name anyway. 

So in both those cases I'd have expected the names to match. Maybe not, if the match was done based on contact? Are there other scenarios I'm missing or is ☝️ wrong?

samuele-mattiuzzo:
> @samuele-mattiuzzo my assumption was that if a match is found the name would likely match (because it's been matched on user input name directly, or because the upstream data source has looked up the DNB/Company registration number based on user input), and if a match isn't found a company would be created with that name anyway.
> 
> So in both those cases I'd have expected the names to match. Maybe not, if the match was done based on contact? Are there other scenarios I'm missing or is ☝️ wrong?

I think I haven't explained myself properly, the issue can exist when we match with existing companies.
The match is done on `DNB` number and the creation is done based on the data that arrives via the `EYBLead` payload. When we create, then we use `eyb_lead.company_name` (which means, in turn, `company_name` and `company.name` will 100% match). When we match with an existing, we don't update `company_name`.

baarkerlounger:
That makes sense. In our case we only have a Company Registration Number if it was looked up by user input name upstream in the first place but it sounds like that's not the case for EYB? Do users just input a DNB number?

oliverjwroberts:
@baarkerlounger for EYB leads, users can search companies using a Dun and Bradstreet (D&B) lookup. However, if found, I'm not exactly sure what gets populated into the `EYBLead.company_name` field - whether its auto filled out from the D&B record or users are still able to modify it.

Your right, in most cases, these fields should match. However, when we look for a matching company record in Data Hub, we search for a D&B number and link the record without modification. In some cases, this has left us with slight mismatches in the `EYBLead.company_name` and `EYBLead.company.name` fields.

PR #5818, opened 2024-11-22
feature/sync-with-contacts-preloaded

chopkinsmade:
### Description of change
Using the django DB for individually checking if a contact in the S3 file existed was running slowly on production (200-300 milliseconds per query), with a file that contains 1.3 million rows.

To reduce the number of individual queries being executed against the DB, at the start of the job load all contacts from the DB into a python dictionary. This initial load of users takes around 4-5 minutes, however no further DB reads are required during the processing of the file.

This has been tested on a local environment with 600,000 contacts, and took 6-10 minutes to fully process the file with ENABLE_CONTACT_CONSENT_INGEST setting equal to `False`. Setting this value to `False` allows testing of the file processing logic, without making any changes to the contact in the DB.

When ENABLE_CONTACT_CONSENT_INGEST was set to `True`, the same 600,000 users takes 20-25 minutes to fully process the file and update the contact in the DB with the latest value for that contacts consent. As the only fields this job need to update are the `consent_data` and `consent_data_last_modified`, the `update` django function is used instead of `save` to avoid triggering the subscribed contact signals that submit additional sync jobs to the short queue 

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/cron-scheduler.py
+++ b/cron-scheduler.py
@@ -13,6 +13,7 @@ from django.conf import settings
 from datahub.company.tasks.adviser import schedule_automatic_adviser_deactivate
 from datahub.company.tasks.company import schedule_automatic_company_archive
 from datahub.company.tasks.contact import (
+    ingest_contact_consent_data,
     schedule_automatic_contact_archive,
 )
 from datahub.company.tasks.export_potential import update_company_export_potential_from_csv
@@ -246,6 +247,15 @@ def schedule_jobs():
     schedule_export_win_auto_resend_client_email()
     schedule_user_reminder_migration()
     schedule_update_company_export_potential_from_csv()
+    job_scheduler(
+        max_retries=3,
+        function=ingest_contact_consent_data,
+        cron=EVERY_HOUR,
+        queue_name=LONG_RUNNING_QUEUE,
+        retry_backoff=True,
+        description='Import contact consent data',
+        job_timeout=HALF_DAY_IN_SECONDS,
+    )
 
 
 def schedule_email_ingestion_tasks():
--- a/cron-scheduler.py
+++ b/cron-scheduler.py
@@ -34,6 +34,7 @@ from datahub.core.queues.constants import (
     # EVERY_THREE_AM_ON_TWENTY_EIGHTH_EACH_MONTH,
     EVERY_TWO_AM,
     HALF_DAY_IN_SECONDS,
+    ONE_HOUR_IN_SECONDS,
 )
 from datahub.core.queues.health_check import queue_health_check
 from datahub.core.queues.job_scheduler import job_scheduler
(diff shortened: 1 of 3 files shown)


cron-scheduler.py
```diff
@@ -246,6 +247,15 @@ def schedule_jobs():
...
     schedule_user_reminder_migration()
     schedule_update_company_export_potential_from_csv()
+    job_scheduler(
+        max_retries=3,
+        function=ingest_contact_consent_data,
+        cron=EVERY_HOUR,
+        queue_name=LONG_RUNNING_QUEUE,
+        retry_backoff=True,
+        description='Import contact consent data',
+        job_timeout=HALF_DAY_IN_SECONDS,
```
baarkerlounger, line 257: Does the timeout need to be this long if we're expecting it to run in ~25mins? Maybe 2 hours would be better?

datahub/company/tasks/contact.py
```diff
@@ -242,54 +269,50 @@ def sync_file_with_database(self, client, file_key):
...
+                if not email:
+                    logger.debug('Row %s has no email', i)
                     continue
 
-                for matching_contact in matching_contacts:
+                matching_contacts = contact_dict.get(email)
 
-                    last_modified = (
-                        consent_row['last_modified'] if 'last_modified' in consent_row else None
+                if not matching_contacts or len(matching_contacts) == 0:
```
baarkerlounger, line 279: Do you need the second check here? Wouldn't this always return `None` or 1+?

baarkerlounger (approved):

datahub/company/tasks/contact.py
```diff
@@ -242,54 +269,50 @@ def sync_file_with_database(self, client, file_key):
...
+                    contact.consent_data = consent_row['consents']
+                    contact.consent_data_last_modified = (
+                        consent_row['last_modified']
+                        if 'last_modified' in consent_row and consent_row['last_modified']
+                        else datetime.datetime.now()
+                    )
+
+                    # We don't need to trigger sync related data signals, use an update
+                    # instead of a save
+                    Contact.objects.filter(id=contact.id).update(
```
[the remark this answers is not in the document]
chopkinsmade: Batching is an option, which if there are any performance issues found we could look at. The benefit of doing it each time a contact needs to be updated - if this job fails after 20,000 rows and restarts, those 20,000 would all be skipped next time

cron-scheduler.py
chopkinsmade: Good point, I'll drop it down

datahub/company/tasks/contact.py
chopkinsmade: With the group by email logic earlier in the job, you're right if it's not None then it should have at least 1 contact in the list. There was a reason I added it, but that scenario might have been refactored out so will remove

oliverjwroberts (approved):

PR #5839, opened 2024-12-03
Feature CLS2-982 add missing EYB lead search fields

marijnkampf:
### Description of change

Add triage_created and is_high_value fields to open search for Investment Project EYB Leads.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/company_activity/tasks/sync.py
+++ b/datahub/company_activity/tasks/sync.py
@@ -184,8 +184,7 @@ def relate_company_activity_to_eyb_lead(batch_size=500):
     objs = [
         CompanyActivity(
             eyb_lead_id=eyb_lead['id'],
-            date=eyb_lead['triage_created'] if eyb_lead['triage_created']
-            else eyb_lead['created_on'],
+            date=eyb_lead['triage_created'],
             company_id=eyb_lead['company_id'],
             activity_source=CompanyActivity.ActivitySource.eyb_lead,
         )
--- a/datahub/company_activity/tasks/sync.py
+++ b/datahub/company_activity/tasks/sync.py
@@ -179,12 +179,13 @@ def relate_company_activity_to_eyb_lead(batch_size=500):
 
     eyb_leads = EYBLead.objects.filter(
         company__isnull=False,
-    ).values('id', 'created_on', 'company_id')
+    ).values('id', 'created_on', 'triage_created', 'company_id')
 
     objs = [
         CompanyActivity(
             eyb_lead_id=eyb_lead['id'],
-            date=eyb_lead['created_on'],
+            date=eyb_lead['triage_created'] if eyb_lead['triage_created']
+            else eyb_lead['created_on'],
             company_id=eyb_lead['company_id'],
             activity_source=CompanyActivity.ActivitySource.eyb_lead,
         )
--- a/datahub/company_activity/tasks/sync.py
+++ b/datahub/company_activity/tasks/sync.py
@@ -184,7 +184,8 @@ def relate_company_activity_to_eyb_lead(batch_size=500):
     objs = [
         CompanyActivity(
             eyb_lead_id=eyb_lead['id'],
-            date=eyb_lead['triage_created'],
+            date=eyb_lead['triage_created'] if eyb_lead['triage_created'] is not None
+            else eyb_lead['created_on'],
             company_id=eyb_lead['company_id'],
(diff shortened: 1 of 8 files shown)


datahub/company_activity/tasks/sync.py
```diff
@@ -179,12 +179,13 @@ def relate_company_activity_to_eyb_lead(batch_size=500):
...
         company__isnull=False,
-    ).values('id', 'created_on', 'company_id')
+    ).values('id', 'created_on', 'triage_created', 'company_id')
 
     objs = [
         CompanyActivity(
             eyb_lead_id=eyb_lead['id'],
-            date=eyb_lead['created_on'],
+            date=eyb_lead['triage_created'] if eyb_lead['triage_created']
+            else eyb_lead['created_on'],
```
DeanElliott96, line 188: Maybe a test for this logic?

DeanElliott96 (commented):
Looks good, if coverage doesn't pick it up do you think this should have a test?

datahub/company_activity/tasks/sync.py
marijnkampf: I've removed the check as it's superfluous. triage_created is required setting it to None results in a 
`django.db.utils.IntegrityError: null value in column "date" of relation "company_activity_companyactivity" violates not-null constraint`.

oliverjwroberts: @marijnkampf does this cover the scenario where it may attempt to sync leads that don't have the triage component (and subsequently no `triage_created` field)? In those cases, the field will be set to an empty string. Or do we filter out these 'incomplete' leads before syncing?

marijnkampf: Hopefully final go added check and fallback to created_on if there is no triage_created.

datahub/company_activity/tests/test_tasks/test_eyb_lead_tasks.py
```diff
@@ -32,7 +33,7 @@ def test_eyb_leads_are_copied_to_company_activity(self):
         assert CompanyActivity.objects.count() == len(eyb_leads)
 
         company_activity = CompanyActivity.objects.get(eyb_lead=eyb_leads[0])
-        assert company_activity.date == eyb_leads[0].created_on
+        assert company_activity.date == eyb_leads[0].triage_created.replace(tzinfo=pytz.UTC)
```
oliverjwroberts, line 36: I appreciate there are a couple of instances where `pytz.UTC` has been used in the export wins sub app, but for consistency would `datetime.timezone.utc` be more preferable?

datahub/investment_lead/models.py
```diff
@@ -234,5 +234,9 @@ def save(self, *args, **kwargs):
             CompanyActivity.objects.update_or_create(
                 eyb_lead_id=self.id,
                 activity_source=CompanyActivity.ActivitySource.eyb_lead,
-                defaults={'date': self.created_on, 'company_id': self.company_id},
+                defaults={
+                    'date': self.triage_created if self.triage_created is not None
```
oliverjwroberts, line 238: Same as above comment, this may need to cater for `triage_created` being an empty string, rather than `None`.

datahub/search/company_activity/fields.py
```diff
@@ -87,5 +87,7 @@ def activity_eyb_lead_field():
             'created_on': Date(),
             'company_name': Text(index=False),
             'duns_number': Text(index=False),
+            'triage_created': Date(),
```
oliverjwroberts, line 90: Is there a DateTime type? Or does this cover it?

oliverjwroberts (approved):
Looks good to me. A couple of queries, but nothing blocking the PR.

datahub/company_activity/tests/test_tasks/test_eyb_lead_tasks.py
marijnkampf: I've replaced pytz with datetime. It seems pytz is only used in one other test (/datahub/export_win/test/test_tasks.py). It however is included as it's required by pandas so changing this there may not make a huge difference. Happy for a tech debt ticket to be created for it though.

datahub/search/company_activity/fields.py
marijnkampf: There is no DateTime object the Date is stored in milliseconds https://opensearch.org/docs/latest/field-types/supported-field-types/index/

PR #5847, opened 2024-12-09
Extract common ingestion tasks

oliverjwroberts:
### Description of change

This PR creates a new _ingest_ sub-app that contains base functionality and utilities for ingesting objects from S3 buckets. The intention is the existing ingestion tasks can be refactored and plugged into this new common functionality in subsequent work.

Notably, it adds:
- A `get_s3_client()` function that returns a boto3 S3 client, either using the automatic endpoint URL or localstack's if the env variable `S3_LOCAL_ENDPOINT_URL` is set.
- An `S3ObjectProcessor` class for processing objects located at a specified prefix within an S3 bucket.
- Several constants for when using S3 buckets.
- A new `ingest.IngestedObject` model, that mirrors the functionality of `company_activity.IngestedFile`.
- A `BaseObjectIdentificationTask` class for identifying new objects in S3 and determining if they should be ingested.
- A `BaseObjectIngestionTask` class for ingesting a specified object from S3.

The _BaseObject...Task_ classes are intended to be inherited for each dataset that we wish to ingest. The class methods can be overridden to tailor the tasks to the specific dataset; the idea being that only minimal code changes are needed to do so.

In terms of testing, it also modifies/adds:
- An `s3_client` fixture
- An `s3_object_processor` fixture
- An `test_object_tuples` fixture which can be used to generate some object definitions for testing. This fixture also uses the utility function `compressed_json_faker` to serialize the contents and compress them into a `.jsonl.gz` object.
- An `upload_objects_to_s3` function that can be used to upload objects to a test S3 bucket.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


datahub/ingest/migrations/0001_initial.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 4.2.16 on 2024-11-21 17:14
+
+from django.db import migrations, models
+import django.utils.timezone
+import uuid
+
+
+class Migration(migrations.Migration):
```
baarkerlounger, line 8: What's your plan for moving the data?

datahub/ingest/tasks.py
```diff
@@ -0,0 +1,237 @@
...
+        return job.func_name == function_name and job.kwargs.get('object_key') == object_key
+
+    def is_job_queued(self, ingestion_task_function: callable, object_key: str) -> bool:
+        """Check if a job is queued or running."""
+        # Check queued jobs
+        if any(
+            self.match_job(job, ingestion_task_function, object_key) for job in self.queue.jobs
+        ):
+            return True
+        # Check running jobs
```
baarkerlounger, line 39: Could maybe extract this into a `is_job_running` method rather than adding a comment?

datahub/ingest/tasks.py
```diff
@@ -0,0 +1,237 @@
...
+                    record = self._get_record_from_line(deserialized_line)
+                    if self._should_process_record(record):
+                        self._process_record(record)
+                    else:
+                        self.skipped_counter += 1
+        except Exception as e:
+            logger.error(f'An error occurred trying to process {self.object_key}: {str(e)}')
+            raise e
+
+        # Record ingestion
```
baarkerlounger, line 148: Could also be a method perhaps

datahub/ingest/migrations/0001_initial.py
oliverjwroberts: My plan was to create a management command; probably at the same time I plug in the first ingestion tasks to this new common functionality. That way I can test it to ensure all instances are copied over correctly and it can be run multiple times while we refactor all ingestion tasks.

datahub/ingest/tasks.py
oliverjwroberts: Yes! I've made that change.

datahub/ingest/tasks.py
oliverjwroberts: Agreed. Same with the logging of various metrics (created, updated, failed, and skipped). I've made the changes in the new commit.

datahub/ingest/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    """Serializes and compresses records into a zipped JSON, encoded with UTF-8."""
+    json_lines = [
+        json.dumps({'object': record}, default=str)
+        for record in records
+    ]
+    compressed_content = gzip.compress('\n'.join(json_lines).encode('utf-8'))
+    return compressed_content
+
+
+def upload_objects_to_s3(
```
baarkerlounger, line 17: Should this be in `ingest/test/utils.py` or do we actually use it outside of tests?

datahub/ingest/tasks.py
```diff
@@ -0,0 +1,256 @@
...
+        """Record a successful ingestion by creating an IngestedObject instance."""
+        last_modified = self.s3_processor.get_object_last_modified_datetime(self.object_key)
+        IngestedObject.objects.create(object_key=self.object_key, object_created=last_modified)
+        logger.info(f'IngestObject instance created for {self.object_key}.')
+
+    def _log_ingestion_metrics(self):
+        """Log various metrics after a successful ingestion.
+
+        Metrics include:
+        - Number of and list of instance id's that have been created
```
baarkerlounger, line 238: No `'` on `ids`?

baarkerlounger (approved):
Looks good. I know preference is generally for smaller PRs but I'd say in cases like this it would be worth including at least one existing ingestion task to use this mechanism. Helps show it works, and makes it easier to review the new generic code side by side with the old being removed.

datahub/ingest/utils.py
oliverjwroberts: We don't at the moment; but it could be. It could be used to upload files to the localstack container during development, or buckets in the deployed environments.

baarkerlounger: Fair enough 👍

oliverjwroberts:
@baarkerlounger I was very conscious of how large this PR already is, but I agree, it probably would be easier to see a real example. I did attempt to describe an example in the docstring, but appreciate it's not the same. I was planning to plug the EYB ingestion tasks into this common functionality next; I wonder if it would be useful see that first before merging this in?

DeanElliott96 (changes requested):
Looks good, just a few suggestions/changes.

datahub/ingest/tasks.py
```diff
@@ -0,0 +1,256 @@
...
+        function_name = f'{ingestion_task_function.__module__}.{ingestion_task_function.__name__}'
+        return job.func_name == function_name and job.kwargs.get('object_key') == object_key
+
+    def is_job_queued(self, ingestion_task_function: callable, object_key: str) -> bool:
+        """Check if a job is queued."""
+        if any(
+            self.match_job(job, ingestion_task_function, object_key) for job in self.queue.jobs
+        ):
+            return True
+        return False
```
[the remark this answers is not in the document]
DeanElliott96: Nope, leftover from copy paste.

datahub/ingest/boto3.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            self.s3_client = s3_client
+
+    def list_objects(self) -> list[str]:
+        """Returns a list of all objects with specified prefix."""
+        response = self.s3_client.list_objects(
+            Bucket=self.bucket,
+            Prefix=self.prefix,
+        )
+        return [obj.get('Key') for obj in response.get('Contents', [])]
```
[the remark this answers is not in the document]
oliverjwroberts: Nice spot and your suggestion is a much better approach, thanks. Will fix now.

datahub/ingest/boto3.py
```diff
@@ -0,0 +1,85 @@
...
+    S3_BUCKET_NAME,
+)
+from datahub.ingest.models import IngestedObject
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_s3_client(region: str):
+    if settings.S3_LOCAL_ENDPOINT_URL:
```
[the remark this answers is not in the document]
oliverjwroberts: I do wonder if this should be in local or dev settings instead?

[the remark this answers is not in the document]
DeanElliott96: I done a search for it and never seen it but its been there for a month, my bad!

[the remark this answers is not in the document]
DeanElliott96: I guess since its an environment variable, common is fine?

DeanElliott96 (approved):
Looks great! Once merged I will try using it with Stova.

PR #5866, opened 2024-12-18
Feature/tet 812 ingest stova data

DeanElliott96:
### Description of change

Ingest Stova event data into DataHub `StovaEvent` model which copies it into a DataHub `Event`.

This uses the new generic ingest app from here: https://github.com/uktrade/data-hub-api/pull/5847.

Events require a Service which is not provided by Stova, after speaking with Tom Aldridge and we are going to create a new service "Stova Event Service" which will be applied to all ingested Stova Events.

Each row (stova event) will not be updated but the ingested file will be updated with new events so there is no requirement to update the already ingested Stova Events.

Depending on: https://github.com/uktrade/data-hub-frontend/pull/7422 as the service requires being added to the frontend as well.

### Checklist

### DIFF
--- a/cron-scheduler.py
+++ b/cron-scheduler.py
@@ -18,7 +18,7 @@ from datahub.company.tasks.contact import (
 )
 from datahub.company.tasks.export_potential import update_company_export_potential_from_csv
 from datahub.company_activity.tasks.ingest_company_activity import ingest_activity_data
-from datahub.company_activity.tasks.ingest_stova_events import stova_identification_and_ingest_task
+from datahub.company_activity.tasks.ingest_stova_events import ingest_stova_event_data
 from datahub.core.queues.constants import (
     EVERY_EIGHT_AM,
     EVERY_EIGHT_THIRTY_AM_ON_FIRST_EACH_MONTH,
@@ -141,7 +141,7 @@ def schedule_jobs():
         description='Check S3 for new EYB triage data files and schedule ingestion',
     )
     job_scheduler(
-        function=stova_identification_and_ingest_task,
+        function=ingest_stova_event_data,
         cron=EVERY_HOUR,
         description='Check S3 for new Stova Event files and schedule ingestion',
     )
--- a/cron-scheduler.py
+++ b/cron-scheduler.py
@@ -18,7 +18,7 @@ from datahub.company.tasks.contact import (
 )
 from datahub.company.tasks.export_potential import update_company_export_potential_from_csv
 from datahub.company_activity.tasks.ingest_company_activity import ingest_activity_data
-from datahub.company_activity.tasks.ingest_stova_events import stova_identification_task
+from datahub.company_activity.tasks.ingest_stova_events import stova_identification_and_ingest_task
 from datahub.core.queues.constants import (
     EVERY_EIGHT_AM,
     EVERY_EIGHT_THIRTY_AM_ON_FIRST_EACH_MONTH,
@@ -141,7 +141,7 @@ def schedule_jobs():
         description='Check S3 for new EYB triage data files and schedule ingestion',
     )
     job_scheduler(
-        function=stova_identification_task,
+        function=stova_identification_and_ingest_task,
         cron=EVERY_HOUR,
         description='Check S3 for new Stova Event files and schedule ingestion',
     )
--- a/cron-scheduler.py
+++ b/cron-scheduler.py
@@ -18,6 +18,7 @@ from datahub.company.tasks.contact import (
 )
 from datahub.company.tasks.export_potential import update_company_export_potential_from_csv
 from datahub.company_activity.tasks.ingest_company_activity import ingest_activity_data
+from datahub.company_activity.tasks.ingest_stova_events import stova_identification_task
 from datahub.core.queues.constants import (
     EVERY_EIGHT_AM,
     EVERY_EIGHT_THIRTY_AM_ON_FIRST_EACH_MONTH,
@@ -138,6 +139,11 @@ def schedule_jobs():
         function=ingest_eyb_triage_file,
         cron=EVERY_HOUR,
         description='Check S3 for new EYB triage data files and schedule ingestion',
+    )
+    job_scheduler(
+        function=stova_identification_task,
+        cron=EVERY_HOUR,
+        description='Check S3 for new Stova Event files and schedule ingestion',
     )
 
     if settings.ENABLE_ESTIMATED_LAND_DATE_REMINDERS:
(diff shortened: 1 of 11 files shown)


DeanElliott96:
Forgot to add it to the cron.

DeanElliott96:
Paused for now, the frontend expects a "service" and crashes without one.

cron-scheduler.py
```diff
@@ -139,6 +140,11 @@ def schedule_jobs():
         cron=EVERY_HOUR,
         description='Check S3 for new EYB triage data files and schedule ingestion',
     )
+    job_scheduler(
+        function=stova_identification_task,
```
baarkerlounger, line 144: Is `identification` different from ingestion? If not might be worth sticking to the naming pattern of the other jobs for consistency?

datahub/company_activity/tasks/ingest_stova_events.py
```diff
@@ -1,78 +1,99 @@
...
+            'location_address1': record.get('location_address1', ''),
+            'location_address2': record.get('location_address2', ''),
+            'location_address3': record.get('location_address3', ''),
+            'location_postcode': record.get('location_postcode', ''),
+            'standard_currency': record.get('standard_currency', ''),
         }
 
-        StovaEvent.objects.create(**values)
+        try:
+            StovaEvent.objects.create(**values)
```
baarkerlounger, line 88: Can these events be updated or is it just create?

cron-scheduler.py
DeanElliott96: The identification function identifies the next latest key (file). The identification code takes an ingestion function as a parameter and once a file (key) is found it schedules the given ingestion function.

A little confusing and not clear here, I can add some comments and look at renaming them.

datahub/company_activity/tasks/ingest_stova_events.py
DeanElliott96: Good thought! I've not considered this, I'll raise on stand up tomorrow and see if it will be updated or who to contact and ask.

cron-scheduler.py
baarkerlounger: I think I'd say that's just part of the overall "ingestion system" rather than a thing in it's own right but I see what you mean.

datahub/company_activity/tasks/ingest_stova_events.py
DeanElliott96: @chopkinsmade has said the fields will not change, only additional new rows (events) will be in the file.

cron-scheduler.py
DeanElliott96: I've renamed the function and added some docstrings.

datahub/company_activity/models/stova_event.py
```diff
@@ -82,11 +82,23 @@ def create_or_update_datahub_event(self) -> Event:
...
                 'address_postcode': self.location_postcode,
                 'address_country': get_country_by_country_name(self.country, 'GB'),
+                'service_id': self.get_stova_event_service_id(),
                 'notes': self.description,
             },
         )
         return event
 
+    @staticmethod
+    def get_stova_event_service_id() -> str:
```
baarkerlounger, line 92: Does this even need a method? Why not just a constant?

cron-scheduler.py
```diff
@@ -139,6 +140,11 @@ def schedule_jobs():
         cron=EVERY_HOUR,
         description='Check S3 for new EYB triage data files and schedule ingestion',
     )
+    job_scheduler(
+        function=stova_identification_and_ingest_task,
```
baarkerlounger, line 144: `ingest_stova_event_data` ? Would match the others more closely but just a personal preference/nit really.

datahub/company_activity/models/stova_event.py
baarkerlounger: Actually where does this `id` come from? Should there be a real lookup here?

datahub/company_activity/tasks/ingest_stova_events.py
```diff
@@ -1,78 +1,101 @@
...
+
+
+class StovaEventIngestionTask(BaseObjectIngestionTask):
+
+    existing_ids = []
+
+    def _process_record(self, record: dict) -> None:
+        """Saves an event from Stova from the S3 bucket into a `StovaEvent`"""
+        if not self.existing_ids:
+            self.existing_ids = list(StovaEvent.objects.values_list('stova_event_id', flat=True))
```
baarkerlounger, line 46: How big is this list going to be? Do you want a set here rather for const time lookup?

datahub/company_activity/models/stova_event.py
DeanElliott96: > Does this even need a method? Why not just a constant?

I put it in a function as the comment was a few lines long so though it would be better suited on its own with a docstring. It's also handling a model relation similar to the event_type and country which both have their own functions.

> Actually where does this id come from? Should there be a real lookup here?

Services are added in the metadata app which are created with migrations. These are run prior to the app starting. Since we already specify its uuid in the migration, we know what it is so this saves a database lookup for every new stova event we ingest.

The file: https://github.com/uktrade/data-hub-api/blob/47cef81b5648fe238d27f480631c5fc1efcbb3ed/datahub/metadata/migrations/0089_add_stova_serivce.yaml

datahub/company_activity/tasks/ingest_stova_events.py
DeanElliott96: "Probably" not too big for this but I agree a set would be better. I'll update to use a set.

datahub/metadata/migrations/0089_add_stova_serivce.yaml
```diff
@@ -0,0 +1,12 @@
+- model: metadata.service
+  pk: f6671176-6493-43ba-a92d-899281efcb55
```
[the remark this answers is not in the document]
DeanElliott96: No this is randomly generated but once migrated will be the same for local, dev, staging and prod. These are just to create rows in the database. This is how all the models in the metadata app are used.

This is also the id used in the ingestion for the `service_id` to save a lookup.

cron-scheduler.py
DeanElliott96: I missed that, thought the name was not as descriptive with just the "identify".

I'll update, agree with making consistent with the others.

datahub/metadata/migrations/0089_add_stova_serivce.yaml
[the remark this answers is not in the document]
baarkerlounger: Ah ok. Seems a bit odd to hardcode something like that but I assume the chances of a clash are near enough to 0 to not matter.

baarkerlounger (approved):

santoshdasa12345 (approved):

PR #5852, opened 2024-12-11
Feature CLS2 -1085 Accept any valid string for Eyblead

santoshdasa12345:
### Description of change

Similar to #5849, EYB leads were being rejected by the serialiser at ingestion time due to _invalid_ choices for `spend`, when in reality we hadn't updated our list of choices to match new ones implemented on EYB's side.

To avoid this issue occurring again, it was decided to remove the choice restriction on this field, as well as `intent` and `landing_timeframe`. 

### Checklist

* [x] Has this branch been rebased on top of the current `main` branch?

  <details>
  <summary>Explanation</summary>
  
  The branch should not be stale or have conflicts at the time reviews are requested.
  
  </details>

* [x] Is the CircleCI build passing?

### General points

<details>
<summary><strong>Other things to check</strong></summary><p></p>

* Make sure `fixtures/test_data.yaml` is maintained when updating models
* Consider the admin site when making changes to models
* Use select-/prefetch-related field lists in views and search apps, and update them when fields are added
* Make sure the README is updated e.g. when adding new environment variables

</details>

See [docs/CONTRIBUTING.md](https://github.com/uktrade/data-hub-api/blob/main/docs/CONTRIBUTING.md) for more guidelines.

PR #5890, opened 2025-01-08
User friendly numbers for EYB Leads

marijnkampf:
### Description of change

Return user friendly currency and number ranges for EYB Lead data (spend and hiring).

Guidance is taken from the [GOV UK style guide](https://www.gov.uk/guidance/style-guide/a-to-z)

For any ranges, use ‘to’ rather than '-'
E.g. 1-5 should be 1 to 5
For numbers over 999 a comma should be used e.g. 1,000
For numbers over 1,000,000, they should be displayed as 1 million for better readability
Currency symbols should be used
E.g. £0 to £499,999

Some possible options include text as UPPER_SNAKE_CASE, where as a fallback this is formatted as Sentence case instead. 

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/core/test/test_utils.py
+++ b/datahub/core/test/test_utils.py
@@ -203,8 +203,8 @@ def test_format_currency_range_string_separator(string, expected):
         ('0-9999', False, False, '£0 to £9,999'),
         ('0-10000', False, False, '£0 to £10,000'),
         ('0-1000000', False, False, '£0 to £1 million'),
+        ('10000001+', False, False, '£10 million+'),
         # Return string as Sentence case for invalid numbers
-        ('10000001+', False, False, '10000001+'),
         ('SPECIFIC_AMOUNT', False, False, 'Specific amount'),
     ),
 )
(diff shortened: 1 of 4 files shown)


baarkerlounger (commented):
Given this is purely about presentation(?) and the backend doesn't do any presentation(?) shouldn't this happen on the react side really?

p3dr0migue1 (approved):

oliverjwroberts:
> Given this is purely about presentation(?) and the backend doesn't do any presentation(?) shouldn't this happen on the react side really?

You do make a good point. For additional context, similar functionality was previously implemented, but removed in #5852. A potential argument for keeping this in the API is then the Data Workspace datasets would also be formatted this way...

I'd be interested to hear others' thoughts on this. I know Santosh currently has a [PR](https://github.com/uktrade/data-hub-frontend/pull/7452) for similar functionality in the Frontend.

marijnkampf:
@baarkerlounger it's basically picking your design principle to stick to; either keeping it DRY or arguable having formatting/presentation in the API. 

As @oliverjwroberts mentioned this formatting will need to be done for the Data Workspace API regardless. One thing we could consider is formatting this when ingesting the data, but the downside of that would be we would loose the format of the original data. I think having this in the API here is an acceptable solution as it avoids having to do the calculations in both the API and the front end. Which would use different technologies and the changes of different non consistent results would increase.

baarkerlounger (approved):

PR #5911, opened 2025-01-21
Refactor EYB ingestion tasks

oliverjwroberts:
### Description of change

This PR refactors the EYB ingestion tasks to use the new common functionality provided by the `datahub.ingest` sub app. The tasks should functionally be the same.

The PR also consolidates how test objects are defined and uploaded in EYB tests, as well as, renames identification tasks to be consistent in the `cron-scheduler.py` and `datahub.company_activity` sub app.

To summarise the business rules behind the EYB tasks:
- EYB leads are made up of 3 components: triage, user, and marketing
- Each component is ingested separately
- EYB tasks are chained as follows: triage (identification and ingestion) --> user (identification and ingestion) -> linking leads to company and contact records -> marketing (identification and ingestion)
- Components must be matched on their respective `hashed_uuid` fields, so when looking for an existing record, we search the `triage_`, `user_`, and `marketing_hashed_uuid` fields.

#### Triage
- Triage records are nested under an `'object'` key; e.g. `{'object': {'key': value, ...}}`
- Triage components of a lead can be updated
- Triage records have the primary key `hashedUuid`
- Triage ingestion task should trigger user identification task

#### User
- User records are also nested under an `'object'` key; e.g. `{'object': {'key': value, ...}}`
- User components of a lead can be updated
- User records have the primary key `hashedUuid`
- User ingestion task should trigger company and contact matching, and marketing identification task

#### Marketing
- Marketing records are **not** nested under an `'object'` key; e.g. `{'key': value, ...}`
- Marketing components of a lead should **not** be updated
- Marketing records have the primary key `hashed_uuid` (snake case versus the camelCase that the triage and user components have)

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


cron-scheduler.py
```diff
@@ -137,9 +137,9 @@ def schedule_jobs():
         description='Check S3 for new Company Activity data files and schedule ingestion',
     )
     job_scheduler(
-        function=ingest_eyb_triage_file,
```
baarkerlounger, line 140: Would prefer to keep the existing here, or update the other tasks below as well - I think it's useful for them to be consistently named and described so it's clear they're branches of the same thing.

datahub/ingest/utils.py
```diff
@@ -4,10 +4,11 @@
 from datahub.ingest.boto3 import S3ObjectProcessor
 
 
-def compressed_json_faker(records: list[dict] = None) -> bytes:
+def compressed_json_faker(records: list[dict] = None, nested: bool = True) -> bytes:
```
baarkerlounger, line 7: What distinction is `nested` making here? (Why is that handled differently?)

oliverjwroberts: I've just updated the PR description, but the reason this is here is because the triage and user components are nested under an `'object'` key, whereas, the marketing component is not. I wanted this helper function to build the records for both. However, you could argue the fakers should nest the record instead.

cron-scheduler.py
```diff
@@ -132,24 +136,24 @@ def schedule_jobs():
         description='Update companies from dnb service',
     )
     job_scheduler(
-        function=ingest_activity_data,
+        function=company_activity_identification_task,
         cron=EVERY_HOUR,
-        description='Check S3 for new Company Activity data files and schedule ingestion',
+        description='Identify new company activity objects',
```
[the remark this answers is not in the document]
baarkerlounger: That's why I prefered ingest initially (to me the fact it does some identification first is almost an implementation detail - the primary purpose is file ingestion). Don't feel strongly though - I think this works fine too.

datahub/ingest/utils.py
baarkerlounger: Might be more reusable then to pass in what the record is being nested under rather than a boolean, so that if we have a future dataset where it's nested under say `'data': 'object':` it'll work for that too.

baarkerlounger (approved):

cron-scheduler.py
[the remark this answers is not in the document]
oliverjwroberts: I've added "*and schedule their ingestion*" to the task descriptions.

datahub/ingest/utils.py
oliverjwroberts: I've modified the argument to accept a string that can be used as the nested key name.

baarkerlounger (approved):

DeanElliott96 (approved):

PR #5916, opened 2025-01-22
Use existing _should_process_record function to decide whether to ingest record

DeanElliott96:
### Description of change

The base ingestion class we use to ingest data already has a function `_should_process_record` to determine whether to ingest an individual record or not. By default this function checks for a `modified` key in the JSON to compare its datetime to decide it is should be ingested. The key `modified` does not exist for Stova and the error is handled however the base function uses `logger.error` which throws a sentry alert.

This PR overrides this function for Stova and moves the logic for checking whether a record should be ingested or not into the overriden `_should_process_record` function. This will prevent the checks for the `modified` field which does not exist for Stova.

Fixes sentry alert: https://sentry.ci.uktrade.digital/organizations/dit/issues/149837/?environment=staging&project=235&query=is%3Aunresolved+is%3Aneeds_review&statsPeriod=24h

### Checklist

### DIFF
--- a/datahub/company_activity/tests/test_tasks/test_stova_ingestion_task.py
+++ b/datahub/company_activity/tests/test_tasks/test_stova_ingestion_task.py
@@ -205,12 +205,14 @@ class TestStovaIngestionTasks:
         data = test_base_stova_event
         data['id'] = existing_stova_event.stova_event_id
 
+        # Check that the record should not be processed if it already exists.
         with caplog.at_level(logging.INFO):
-            task._process_record(data)
+            assert task._should_process_record(data) is False
             assert (
                 f'Record already exists for stova_event_id: {existing_stova_event.stova_event_id}'
             ) in caplog.text
 
+        # Check that if there are duplicate IDs in the JSON but not our DB, this is handled.
         data['id'] = 999999
         task._process_record(data)
         with caplog.at_level(logging.ERROR):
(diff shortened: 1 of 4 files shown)


datahub/company_activity/tests/test_tasks/test_stova_ingestion_task.py
```diff
@@ -205,12 +205,14 @@ def test_stova_event_fields_with_duplicate_event_ids(self, caplog, test_base_sto
...
 
+        # Check that the record should not be processed if it already exists.
         with caplog.at_level(logging.INFO):
-            task._process_record(data)
+            assert task._should_process_record(data) is False
             assert (
                 f'Record already exists for stova_event_id: {existing_stova_event.stova_event_id}'
             ) in caplog.text
 
+        # Check that if there are duplicate IDs in the JSON but not our DB, this is handled.
```
baarkerlounger, line 215: Can these comments go into the doc string for the test?

baarkerlounger (approved):

datahub/company_activity/tests/test_tasks/test_stova_ingestion_task.py
DeanElliott96: They are in the docstring as well.
I've put a comment above each one so you know which scenario is being tested.

oliverjwroberts: Or maybe a separate test for the two cases?

DeanElliott96: Yeah I'm testing two different things here. I'll move into their own tests! Cheers @baarkerlounger @oliverjwroberts

baarkerlounger (approved):

PR #5924, opened 2025-01-26
Create Interaction for Stova Attendee with default fake advisor

DeanElliott96:
### Description of change

To show as an Attendee of a DataHub `Event`, you must be a `Contact` in an `Interaction` for that `Event`.
This PR creates interactions from StovaAttendees. It also creates a new fake default advisor for assigning these interactions as the Stova Data does not provide an accurate way to give us Advisor emails.

### Checklist

### DIFF
--- a/datahub/company_activity/tests/test_tasks/test_stova_attendee_ingestion_task.py
+++ b/datahub/company_activity/tests/test_tasks/test_stova_attendee_ingestion_task.py
@@ -1,22 +1,17 @@
-import gzip
-import json
 import logging
 
 from datetime import datetime
 from unittest import mock
 
-import boto3
 import pytest
 
 from django.test import override_settings
 from moto import mock_aws
-from sentry_sdk import init
-from sentry_sdk.transport import Transport
 
-from datahub.company.models import Company
-from datahub.company.test.factories import AdviserFactory, CompanyFactory, ContactFactory
+from datahub.company.models import Advisor as Adviser, Company
+from datahub.company.test.factories import CompanyFactory, ContactFactory
 from datahub.company_activity.models import StovaAttendee
-from datahub.company_activity.tasks.constants import BUCKET, REGION, STOVA_ATTENDEE_PREFIX
+from datahub.company_activity.tasks.constants import STOVA_ATTENDEE_PREFIX
 from datahub.company_activity.tasks.ingest_stova_attendees import (
     stova_attendee_identification_task,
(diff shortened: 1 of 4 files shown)


baarkerlounger (approved):
Nice

DeanElliott96:
Running some tests on the frontend against Interactions and contacts created via the Stova Ingestion and there are a few fields the frontend is expecting that is missing.

datahub/company_activity/tests/test_tasks/test_stova_attendee_ingestion_task.py
```diff
@@ -420,3 +372,156 @@ def test_stova_attendee_not_created_if_event_does_not_exist(
...
+        object_definition = (test_file_path, compressed_json_faker([data]))
+        upload_objects_to_s3(s3_object_processor, [object_definition])
+
+        ingestion_task = StovaAttendeeIngestionTask(test_file_path, s3_object_processor)
+        assert Adviser.objects.filter(
+            email='stova_default@businessandtrade.gov.uk',
+            first_name='Stova Default',
+            last_name='Adviser',
+            is_active=False,
+        ).first() is None
```
oliverjwroberts, line 391: [.exists()](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#exists) could also be used here

oliverjwroberts (approved):
Looks good to me, nice work @DeanElliott96!

datahub/company_activity/tests/test_tasks/test_stova_attendee_ingestion_task.py
```diff
@@ -460,6 +460,8 @@ def test_stova_attendee_ingestion_creates_interaction(
         ).first()
         assert interaction is not None
         assert interaction.company == contact.company
+        assert interaction.companies.all().count() == 1
```
baarkerlounger, line 463: Don't think you need the `.all()` here do you?

DeanElliott96: No, it's a bad habit and I keep forgetting!

datahub/company_activity/tests/test_tasks/test_stova_attendee_ingestion_task.py
DeanElliott96: Yeah it would make more sense to use it here, added now!

baarkerlounger (approved):

PR #5929, opened 2025-01-28
Sync UK company and project site address fields on applicable projects

santoshdasa12345:
### Description of change

We are [replacing the old site decided question with a new one](https://github.com/uktrade/data-hub-frontend/pull/7511): is the [investment project] site address the same as the [uk recipient] company address? If users select "yes", the investment project's site address should be set to the uk company address; functionality that will be handled by the frontend before sending the PATCH request.

However, if the UK company address is subsequently updated, we wanted to ensure that applicable investment projects had their site address fields updated too; hence this PR.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/investment/project/signals.py
+++ b/datahub/investment/project/signals.py
@@ -1,3 +1,5 @@
+import reversion
+
 from django.db.models import Q
 from django.db.models.signals import m2m_changed, post_save, pre_save
 from django.dispatch import receiver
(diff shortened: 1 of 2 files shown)


datahub/investment/project/signals.py
```diff
@@ -121,3 +121,29 @@ def update_country_investment_originates_from(sender, **kwargs):
...
+        )
+
+        for investment_project in investment_projects:
+            if investment_project.site_address_is_company_address:
+                investment_project.address_1 = instance.address_1
+                investment_project.address_2 = instance.address_2
+                investment_project.address_town = instance.address_town
+                investment_project.address_postcode = instance.address_postcode
+
+            investment_project.save(
```
elcct, line 144: This should only save investment project if it is updated.

datahub/investment/project/test/test_signals.py
```diff
@@ -136,3 +137,49 @@ def test_update_investor_company_doesnt_update_country_of_origin_for_won_project
...
+        uk_company.address_1 = '10 Downing Street'
+        uk_company.address_2 = 'Whitehall'
+        uk_company.address_town = 'Manchester'
+        uk_company.address_postcode = 'M1 1AA'
+        uk_company.save()
+
+        for project in projects:
+            project.refresh_from_db()
+
+            assert project.site_address_is_company_address
```
elcct, line 185: This should check if address is update.

The test should also check if reversion version is created, so that history of changes to investment project is maintained.

oliverjwroberts:
> Maybe it's worth exploring whether this is a good approach to keep a copy of company address in the investment project. Maybe downstream service could look up the company address if `site_address_is_company_address` is true instead or this could be "baked in" the serialiser.

@elcct I agree, it's not best practice to copy and store duplicate data. However, may I provide a little more context behind the intention of this change and reasons why I think this is still a "good enough" approach?

We are [replacing the old site decided question with a new one](https://github.com/uktrade/data-hub-frontend/pull/7511): is the [investment project] site address the same as the [uk recipient] company address? If users select "yes", the investment project's site address should be set to the uk company address; functionality that will be handled by the frontend before sending the PATCH request. 

However, if the UK company address is subsequently updated, we wanted to ensure that applicable investment projects had their site address fields updated too; something we thought best handled by signals and hence this PR.

Due to the fact the project site address can be either the company address, or another manually entered address, I think we still need to populate these fields either way, and make sure they stay up to date when the company address changes. Ensuring these fields are populated also allows us to piggy-back off existing functionality such as stage validation and minimise changes to analytical reporting, with fewer lines of code modified. 

However, I would be happy to discuss further if you would like? In the meantime, I'll work with @santoshdasa12345 to make amendments as per your comments. Thanks.

datahub/investment/project/signals.py
santoshdasa12345: resolved

datahub/investment/project/test/test_signals.py
santoshdasa12345: @elcct Updated with changes could you please review

oliverjwroberts (approved):

elcct (approved):

PR #5938, opened 2025-02-03
Add api for stova events

bau123:
### Description of change

- Adds API endpoint for StovaEvent for the frontend to see a detail view of a stova event.
- Adds the stova_event_id to the Event Opensearch endpoint to click through to the stova event detail on the frontend.
- Adds the stova_event_id to the interaction detail view for clicking through to a stova event when viewing an interaction with an event

The stova_event_id has been added to the other endpoints to give the frontend the id to link to a stova event instead of a datahub event. Clicking this button on the frontend would take you to the stova event detail page which would use the new API view created here.

### Checklist

### DIFF
--- a/datahub/company_activity/serializers/stova.py
+++ b/datahub/company_activity/serializers/stova.py
@@ -8,6 +8,7 @@ class StovaEventSerializer(serializers.ModelSerializer):
     class Meta:
         model = StovaEvent
         fields = [
+            # Reverse FK to data hub `Event`
             'datahub_event',
             'id',
             'stova_event_id',
--- a/datahub/company_activity/serializers/stova.py
+++ b/datahub/company_activity/serializers/stova.py
@@ -1,0 +1,46 @@
+from rest_framework import serializers
+
+from datahub.company_activity.models.stova_event import StovaEvent
+
+
+class StovaEventSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = StovaEvent
+        fields = [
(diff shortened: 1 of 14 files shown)


datahub/company_activity/serializers/stova.py
```diff
@@ -0,0 +1,46 @@
...
+
+from datahub.company_activity.models.stova_event import StovaEvent
+
+
+class StovaEventSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = StovaEvent
+        fields = [
+            'datahub_event',
```
baarkerlounger, line 11: Think it'd be worth a comment here to explain this is the extra field and where it comes from

datahub/company_activity/urls/stova_urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from .. import views
```
baarkerlounger, line 3: Absolute import here?

baarkerlounger (approved):

datahub/company_activity/serializers/stova.py
DeanElliott96: Added comment now

datahub/company_activity/urls/stova_urls.py
DeanElliott96: Updated to absolute now.

oliverjwroberts (approved):

PR #5941, opened 2025-02-04
Update python version in containers.

elcct:
### Description of change

As in the title.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>


baarkerlounger (approved):

datahub/interaction/email_processors/parsers.py
```diff
@@ -111,10 +111,9 @@ def _extract_and_validate_contacts(self, all_recipients):
         return contacts
 
     def _extract_and_validate_sender_adviser(self):
-        try:
-            sender_email = self.message.from_[0][1]
-        except IndexError:
+        if not any(from_[0] or from_[1] for from_ in self.message.from_):
```
oliverjwroberts, line 114: I appreciate I don't have context of what `from_` contains exactly, but instead of `... or from_[1] ...`, do you want to assert `from_[0][1]` here (i.e. the 2nd element of `from_[0]`), similar to what was being tried before?

elcct: Perhaps I could add a comment why this is that way?

elcct: I added a comment.

oliverjwroberts: Ahh okay, I'm with you now. Maybe the following would be more clear in terms of what you're checking for?
```python
if not self.message.from_ or self.message.from_ == [('', '')]:
```
Either way, I'm happy for this to go in as is! Nice work on the Python version bump @elcct 🚀

oliverjwroberts (approved):

datahub/interaction/email_processors/parsers.py
elcct: I updated it to your version as it is clearer.

PR #5998, opened 2025-03-12
Add management command to re-ingest latest EYB objects

oliverjwroberts:
### Description of change

This PR adds a management command that schedules the re-ingestion of the latest EYB objects. 

The creation of this command was prompted by the fact some leads appear to missing and we want to re-ingest all records to ensure none have been missed; as this isn't the first time, and probably won't be the last, it seemed sensible a management command was created.

Rather than change the behaviour of the ingestion tasks (specifically, setting `_should_process_record` to always return true, among other changes), which would've required a lot more code and tests, I took an easier approach: remove the history of ingested objects/records, forcing the ingestion tasks to presume all records contain unseen changes. It then schedules the same identification task that is scheduled in `cron-scheduler.py`.

### Checklist

  <details>
  
  </details>

### General points

<details>

</details>

### DIFF
--- a/datahub/investment_lead/management/commands/schedule_reingestion_of_eyb_leads.py
+++ b/datahub/investment_lead/management/commands/schedule_reingestion_of_eyb_leads.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
-    """Command to schedule the re-ingestion all records from the latest EYB objects.
+    """Command to schedule the re-ingestion of all records from the latest EYB objects.
 
     It does this by removing all EYB IngestedObject instances and then triggering the existing
     eyb_triage_identification_task that usually runs every hour. By removing the record of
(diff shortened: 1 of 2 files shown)


datahub/investment_lead/management/commands/schedule_reingestion_of_eyb_leads.py
```diff
@@ -0,0 +1,48 @@
...
+    TRIAGE_PREFIX,
+)
+from datahub.investment_lead.tasks.ingest_eyb_user import USER_PREFIX
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """Command to schedule the re-ingestion all records from the latest EYB objects.
```
baarkerlounger, line 20: re-ingestion *of* all records?

baarkerlounger (approved):
May have misunderstood the context but if Leads were missed on the original ingestion is there a reason to think they'll be included when re-ingesting?

PR #6023, opened 2025-03-26
Bugfix/tet 998 remove stova relations

DeanElliott96:
### Description of change

Temporary management command which will be removed once run.
This management commands deletes all the interactions, contacts and companies created from Stova. There are around 35000 companies to delete, 70150 contacts and around 67000 interactions 😨.

Some safety nets in place:

- Simulation to try a fake run first
- Can specify a batch size, so will try small batches before committing to larger batches
- Reversion to undo any changes
- Additional temporary model to track id of objects deleted to make it easy to undo the deletion with reversion
- Lots of logging

### Checklist


DeanElliott96:
Should probably add some sort of batching to do this in stages. Or even to specify a batch_size and could specify a tiny batch (around 5) to confirm the command before committing to larger deletions.

datahub/company_activity/models/stova_attendee.py
```diff
@@ -46,15 +46,15 @@ class StovaAttendee(models.Model):
     created_on = models.DateTimeField(auto_now_add=True)
     company = models.ForeignKey(
         Company,
-        on_delete=models.PROTECT,
+        on_delete=models.SET_NULL,
```
baarkerlounger, line 49: Will this be reverted back after this has been run or is this a permanent change?

DeanElliott96: These fields will be removed once this is run in a new PR.

baarkerlounger (approved):

ince-dbt (approved):

PR #6026, opened 2025-03-27
Add signal to delete company from opensearch when company is deleted

DeanElliott96:
### Description of change

Currently if a company gets deleted on Data Hub, it does not trigger OpenSearch to also remove the company. This means that when a company is removed from the database, the company will still appear for users on the frontend but clicking on the company will show a 404 (as it no longer exists). See error video below.

You can't rerun the sync as it won't pick this change up, currently you would need to delete all the indices and recreate them which would take hours on prod.

This PR fixes this by also deleting the company from opensearch if it has been deleted from the database.

Related PRs - other OpenSearch indices are not being updated when their database object is deleted.
Contacts: https://github.com/uktrade/data-hub-api/pull/6027
Event: https://github.com/uktrade/data-hub-api/pull/6028
Adviser: https://github.com/uktrade/data-hub-api/pull/6029
Export Country History: https://github.com/uktrade/data-hub-api/pull/6030

https://github.com/user-attachments/assets/c2164010-1538-49ef-a433-a65d7788e5ef

### Checklist

### DIFF
--- a/datahub/search/company/signals.py
+++ b/datahub/search/company/signals.py
@@ -1,9 +1,13 @@
 from django.db import transaction
-from django.db.models.signals import post_save
+from django.db.models.signals import post_delete, post_save
 
 from datahub.company.models import Company as DBCompany
 from datahub.interaction.models import Interaction as DBInteraction
 from datahub.search.company import CompanySearchApp
+from datahub.search.company.models import (
+    Company as SearchCompany,
+)
+from datahub.search.deletion import delete_document
 from datahub.search.signals import SignalReceiver
 from datahub.search.sync_object import sync_object_async, sync_related_objects_async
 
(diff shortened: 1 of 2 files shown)


baarkerlounger (commented):
Should this also trigger removing contacts or anything else from OpenSearch?

DeanElliott96:
> > Should this also trigger removing contacts or anything else from OpenSearch?
> 
> I checked interactions and they were deleted but just checked with Contacts and they are not deleted either as its missing a post_delete signal as well. Removing a contact from the DB would still show a Contact on the Frontend.
> 
> I'll add a fix into this PR for it. Will also take a look at the other OpenSearch indices and see if they have delete signals.

Missing delete signals:

- advisers
- contact
- event
- export_country_history

baarkerlounger (approved):

ince-dbt (approved):

PR #6074, opened 2025-04-23
feat: hcsat backend

joshwong-cs:
#### TET-988
new hcsat app to collect customer satisfaction feedback throughout datahub

* added `CustomerSatisfactionToolFeedback`
  * excluded user tracking (`created_by`, `modified_by`) for anonymity
* added `post /v4/hcsat`: creates an initial feedback record
* added `patch /v4/hcsat/<uuid:pk>`: updates initial feedback to populate reasons for dissatisfaction
* added serializer for data validation and repr:
  * enforce read-only fields on updates
  * prevent dissatisfaction feedback submission during initial post or on 'useful' records during patches
  * added tests

#### TET-989
hcsat `/dataset` endpoint for data workspace

* bog standard dataset view
* bog standard dataset tests 

### Checklist

  <details>
  
  </details>


datahub/hcsat/admin.py
```diff
@@ -0,0 +1,2 @@
+
+# Register your models here.
```
oliverjwroberts, line 2: Did you want to include an admin page at this stage? If not, maybe best to remove?

datahub/hcsat/test/test_serializers.py
```diff
@@ -0,0 +1,245 @@
...
+            'was_useful': feedback.was_useful,
+            'did_not_find_what_i_wanted': feedback.did_not_find_what_i_wanted,
+            'difficult_navigation': feedback.difficult_navigation,
+            'lacks_feature': feedback.lacks_feature,
+            'unable_to_load': feedback.unable_to_load,
+            'inaccurate_information': feedback.inaccurate_information,
+            'other_issues': feedback.other_issues,
+            'other_issues_detail': feedback.other_issues_detail,
+            'improvement_suggestion': feedback.improvement_suggestion,
+            'created_on': feedback.created_on.isoformat().replace('+00:00', 'Z'),
```
oliverjwroberts, line 38: Is it possible/would it be easier to use the `format_date_or_datetime` function here?

datahub/hcsat/views.py
```diff
@@ -0,0 +1,18 @@
...
+from datahub.hcsat.models import CustomerSatisfactionToolFeedback
+from datahub.hcsat.serializers import CustomerSatisfactionToolFeedbackSerializer
+
+
+class CustomerSatisfactionToolFeedbackViewSet(CoreViewSet):
+    """Views for creating anonymous feedback (POST) and adding context (PATCH)."""
+
+    serializer_class = CustomerSatisfactionToolFeedbackSerializer
+
+    permission_classes = (AllowAny,) # TODO IS THIS CORRECT???
```
oliverjwroberts, line 13: Assuming there's no special permission requirements for feedback, I think the default is to use `from rest_framework.permissions import IsAuthenticated`.

oliverjwroberts (commented):
I really like the factory implementation with the conditional fields and serialiser validation.

A couple of minor comments, and if it was me, I'd maybe consider splitting/moving some of the serialiser tests to a separate `test_views.py` script as it feels theres some overlap. 

Otherwise, it looks great.  Nice work Josh!

datahub/hcsat/serializers.py
```diff
@@ -0,0 +1,125 @@
...
+        for field in detailed_fields_to_remove:
+            validated_data.pop(field, None)
+
+        return super().create(validated_data)
+
+    def update(self, instance, validated_data):
+        """Explicitly remove fields that should be read-only during update."""
+        # ensure these fields are *never* passed to the superclass update method
+        validated_data.pop('url', None)
+        validated_data.pop('was_useful', None)
```
baarkerlounger, line 122: Is a user not able to edit this? (e.g. hit the back button and change their answer?)

baarkerlounger (approved):

oliverjwroberts (approved):

joshwong-cs:
do not merge yet, waiting for confirmation on official questions, and possibly adding test_views.py for api-level consistency

baarkerlounger:
Checked the questions doc and what you have here is correct 👍

datahub/hcsat/views.py
joshwong-cs: oh, actually, would isAuthenticated de-anonymize the response?

baarkerlounger: I don't think so. You're authenticating that the request is coming from the frontend, not that it's coming from a specific user. @oliverjwroberts agree?

oliverjwroberts: I don't think so, or at least, it wouldn't change the current serializer behaviour of not saving `created_by` and `modified_by` fields. I believe it uses the `request.user` object to check permissions. However, this information might show up in logs.

baarkerlounger: Right but who is the user here, the SSO identity or just the frontend app? Locally `request.user` just returns ``(no name)` for me so I think it's the later

baarkerlounger: Hmm that might be wrong actually - we pass a session token here: https://github.com/uktrade/data-hub-frontend/blob/8f716e72c6f46c4c6f88af18199b2ab4fa16429f/src/middleware/api-proxy.js#L126.

Not entirely sure how easy it would be to work from that back to a user but we shouldn't be logging that anywhere anyway, so I think this is sufficiently anonymous for HCSAT.

baarkerlounger: I'm inclined to think as long as we're not storing a user identifier against the response this is still sufficiently anonymous for HCSAT.

PR #6094, opened 2025-05-06
devops: replace dockerize with healthchecks

joshwong-cs:
### Description of change

* all services (except rq_sched, which has no dependents) now implement docker healthchecks
* replace dockerize with docker services dependency
* e.g., docker inspect --format='{{json .State.Health}}' data-hub-postgres-1

### Checklist

  <details>
  
  </details>


setup-uat.sh
```diff
@@ -3,12 +3,12 @@
...
 # https://github.com/uktrade/data-hub-frontend#continuous-integration
 
-es_amp=""
-## If ES_APM_ENABLED is True wait for it to start
-if [ "$ES_APM_ENABLED" = 'True' ] ; then
-   es_amp="-wait tcp://es-apm:8200"
-fi
-dockerize -wait  tcp://postgres:5432 -wait tcp://opensearch:9200  -wait tcp://redis:6379 $es_amp
+# es_amp=""
+# ## If ES_APM_ENABLED is True wait for it to start
```
baarkerlounger, line 7: Why is this being commented out?

oliverjwroberts: Is this script even needed any more? I've not looked into, but my gut feeling is since the migration, this is no longer required.

baarkerlounger: Might be worth looking at whether we definitely _need_ to wait for external services at the container level or whether we could adjust our application logic to do that?

baarkerlounger (approved):

setup-uat.sh
joshwong-cs: will add pg, redis, opensearch, es_apm service checks to django CoreConfig.ready()

baarkerlounger:
@joshwong-cs you still intending to merge this?

joshwong-cs:
> @joshwong-cs you still intending to merge this?

@baarkerlounger is it confirmed that [setup-uat.sh](https://github.com/uktrade/data-hub-api/pull/6094/files#diff-b9b9e426453232202a248eec5966195347af1e441bd0df9c11a1149048e2a356) is unused legacy code? if so then we can merge and leave the pg, redis, opensearch, es_apm django internal service checks to later?

oliverjwroberts:
@baarkerlounger @joshwong-cs yes, it looks like both the Frontend and OMIS use it in their `docker-compose.backend.yml` files, which are called in CI when running tests.

PR #6121, opened 2025-05-20
feat: tet 1028 prompt payments backend

joshwong-cs:
#### TET-1028
* adds prompt payments model/serializer/ to company activity app
* ingestion stores full prompt payments dataset, regardless of matched status
* exposes (matched) company payments data via `/company/<uuid:pk>/prompt-payments`
* tests for base model, ingest, company integration
* TODO company matching can happen anytime inside data hub via matching ch number
* TODO contact matching can happen anytime inside data hub via matching ch number and email

### Checklist


datahub/company_activity/models/prompt_payments.py
```diff
@@ -0,0 +1,75 @@
...
+from django.db import models
+
+from datahub.company.models.company import Company
+from datahub.company.models.contact import Contact
+from datahub.core import reversion
+from datahub.core.models import BaseModel
+
+
+@reversion.register_base_model()
+class PromptPayments(BaseModel):
```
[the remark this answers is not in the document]
joshwong-cs: no

datahub/company_activity/tasks/ingest_prompt_payments.py
```diff
@@ -0,0 +1,212 @@
...
+
+        if company_house_number:
+            archived_companies_by_chn = Company.objects.filter(
+                company_number=company_house_number,
+                archived=True,
+            )
+            if archived_companies_by_chn.count() == 1:
+                company = archived_companies_by_chn.first()
+                logger.info(
+                    f'Found archived company by CHN {company_house_number} for source_id {source_id}.',
```
baarkerlounger, line 98: Worth adding tests for the log messages throughout I think

datahub/company/models/company.py
```diff
@@ -470,7 +470,10 @@ class Meta:
                 CompanyPermission.change_one_list_core_team_member.value,
                 'Can change one list core team member associated with company',
             ),
-            (CompanyPermission.view_export_win.value, 'Can view company export win'),
+            (
```
baarkerlounger, line 473: This is just formatting right? Might be worth removing from this PR.

datahub/company_activity/models/prompt_payments.py
```diff
@@ -0,0 +1,75 @@
...
+        on_delete=models.SET_NULL,
+        related_name='prompt_payments',
+    )
+
+    approved_by = models.TextField()
+    qualifying_contracts_in_period = models.BooleanField()
+    average_paid_days = models.IntegerField()
+    paid_within_30_days_pct = models.IntegerField()
+    paid_31_to_60_days_pct = models.IntegerField()
+    paid_after_61_days_pct = models.IntegerField()
```
baarkerlounger, line 48: This seems like an odd thing to store as a field? I assume it would be updated at worst every day - is it not possible to store the raw data and calculate this where needed?

datahub/company_activity/models/prompt_payments.py
```diff
@@ -0,0 +1,75 @@
...
+    company_house_number = models.TextField(db_index=True)
+    company = models.ForeignKey(
+        Company,
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='prompt_payments',
+    )
+
+    # stores the raw email address from the s3 data file
```
baarkerlounger, line 33: When you say "raw", as opposed to what? Do we process email addresses/CH numbers before storing elsewhere?

datahub/company/test/test_signals.py
```diff
@@ -194,3 +198,185 @@ def test_delete_company_export_country_no_signal(self):
...
+        pp_no_email.refresh_from_db()
+        pp_to_be_linked.refresh_from_db()
+
+        assert pp_already_linked.contact != contact_in_company
+        assert pp_no_email.contact is None
+        assert pp_to_be_linked.contact == contact_in_company
+
+    def test_contact_linking_handles_multiple_contacts_with_same_email_in_company(self, caplog):
+        """If multiple contacts in the same company have the same email,
+        a warning should be logged and no contact should be linked to avoid ambiguity.
```
baarkerlounger, line 340: What's the impact of not linking a contact? (I know we do this elsewhere but there are legitimate cases e.g. `payroll@company.co.uk` type emails)

datahub/company_activity/tasks/ingest_prompt_payments.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+
+        source_id = record.get('id')
+        if source_id in self.existing_ids:
+            logger.info(f'Record already exists for prompt payments source_id: {source_id}')
+            return False
+        return True
+
+    def _get_company(self, company_house_number, company_name, source_id):  # noqa
+        """Attempts to find a company, prioritizing non-archived records.
```
baarkerlounger, line 53: Could this also try to match on `contact`? Is so could the logic be extracted since then I think it's the same as the Great ingestion task?

datahub/company/signals.py
```diff
@@ -85,3 +87,57 @@ def _record_export_country_history(export_country, action, adviser):
         country=export_country.country,
         status=export_country.status,
     )
+
+
+@receiver(post_save, sender=Company, dispatch_uid='link_prompt_payments_on_company_save')
+def link_prompt_payments_on_company_save(sender, instance, created, **kwargs):
```
DeanElliott96, line 93: I wonder if we could avoid the use of signals?

They can be an anti pattern in Django.
https://www.django-antipatterns.com/antipattern/signals.html

Could this be part of the company save method?

datahub/company_activity/models/prompt_payments.py
joshwong-cs: as opposed to directly storing contacts and inheritting email from there, since we want to be in a position of over-storing to enable ex-post matching

(but i agree raw is ambiguous, will change)

datahub/company_activity/models/prompt_payments.py
joshwong-cs: these are the "raw" data we get from prompt payments 💀🤢🤢

